PR #2316, opened 2024-09-04
feat(zero-cache): avoid rewriting the lmids query

darkgnotic:
The query for last modification ids used to contain an `OR` of all of the `clientIDs`, but now that it no longer does (as of https://github.com/rocicorp/mono/commit/160c4b2ad9d912ab351e83d9589210b247834ad3#diff-bb6552d08e698958f0505e6bfcab7828b0e1df6d8f467a3903383c315380157b), it need not be re-put on every new client. Just write it the first time.


tantaman (commented):
lgtm

tantaman:
lol bencher resurrecting stuff from 2024 🫠

PR #2339, opened 2024-09-09
refactor(zero-cache): Rename CancelableAsyncIterable to "Source"

darkgnotic:
Also add a "Sink" counterpart, both of which are implemented to Subscription.


arv (commented):
LGTM

PR #2357, opened 2024-09-12
chore(zero): Miscellaneous fixes to get RN building.

aboodman:
Originally from https://github.com/rocicorp/mono/pull/2333/files but simplified and cleaned up.


arv (commented):
LGTM

PR #2359, opened 2024-09-12
fix(zero-cache): cancel pending transactions if change stream is aborted

darkgnotic:
Avoid any dangling transactions (that can prevent WAL cleanup) if a change stream is aborted midway through a transaction.


arv (commented):
LGTM

PR #2373, opened 2024-09-14
fix(zero-cache): kill child processes on exit

grgbkr:
Fixes #2372


darkgnotic (approved):
Thank you @grgbkr !!!

PR #2374, opened 2024-09-14
!feat(zero-cache): new change streamer schema and protocol

darkgnotic:
**BREAKING CHANGE**

⚠️  wipe your replica.db
⚠️  delete the "cdc" schema in your Postgres CHANGE_DB

New scheme for watermarking and ordering Changes in the ChangeDB such that they preserve the order in which they were received from Postgres, regardless of their LSN (which may not represent the correct order).

The scheme is documented in `change-streamer-service.ts`

Also update the ChangeStreamer API such that watermarks are only exported to subscribers in the final `commit` message, so that there is no ambiguity as to which watermark to store and use for subscription resumption. Non-commit messages are only used internally within the change-streamer implementation.

Fixes #2361


arv (commented):
LGTM

PR #2387, opened 2024-09-16
feat(zqlite): Use UPDATE if possible TableSource

arv:
In TableSource if we have an 'edit' SouceChange we can use the UPDATE operation instead of DELETE and INSERT.


tantaman (commented):
lgtm

PR #2388, opened 2024-09-16
feat(zero-react): move react to a peer dep

tantaman:
React needs to be a peer dependency for when we ship the package to users.


arv (approved):
LGTM

PR #2389, opened 2024-09-16
feat(zero-cache): Deal with edit changes in zero-cache 

arv:
This now allows for the edit changes to be handled in zero-cache.

To allow this I had to add a type property to the RowChange type so that
we can distinguish between the different types of changes.


arv (commented):
Review with ignore whitespace (the inline snapshot got indented differently for some reason)

darkgnotic (approved):
This is exciting. 

Modulo the ref counting issue, this looks good to me, but I would appreciate it @grgbkr could also give this a look as he knows the tricky CVR ref-counting logic very well.  🙏

PR #2390, opened 2024-09-16
feat(zero-react): query hook that auto-tracks deps

tantaman:
This was @arv's original idea. We don't need to use a dependency array and can instead track if the query changed.


arv (approved):
LGTM

PR #3053, opened 2024-11-19
wip(zqlite): Fix LIKE operator

arv:
I thought that this would work now that we have an ICU build. But it doesn't. I'm going to have to do some more work on this.


arv:
This is still broken 😢

PR #3307, opened 2024-12-13
feat(zbugs): useWatchQuery

arv:
This is WIP but I want some feedback.

This creates a new hook that takes a callback that gets called with the changes that come out at the end of the pipeline.

This is based on the ideas I had last week (or the week before that?) when I first worked on adding the emoji change callbacks.

Usage:

```ts
useWatchQuery(z.query.comment.where('issueID', id), change => {
  if (change.type === 'add') {
    console.log(change.row, 'was added');
  } else if (change.type === 'remove') {
    console.log(change.row, 'was removed');
  }
});
```

Things that are not obvious/clean.
- Name of the hook?
- The implementation waits for the transaction end but this information is lost in the callback. We could call the callback with an array of all the changes instead.
- What to do about child changes since at this level there is no memory of the previous children?
- Should the hook have an option to also be called for non complete results or maybe add that info to the change objects?

### DIFF
--- a/apps/zbugs/src/hooks/use-watch-query.tsx
+++ b/apps/zbugs/src/hooks/use-watch-query.tsx
@@ -1,0 +1,151 @@
+import type {Query, QueryType, Row, TableSchema} from '@rocicorp/zero';
+import {useEffect} from 'react';
+import {unreachable} from 'shared/src/asserts.js';
+import type {Change as IVMChange} from 'zql/src/ivm/change.js';
+import type {Input, Output} from 'zql/src/ivm/operator.js';
+import type {Format} from 'zql/src/ivm/view.js';
+import type {AdvancedQuery} from 'zql/src/query/query-internal.js';
+
+export function useWatchQuery<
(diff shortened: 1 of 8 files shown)


apps/zbugs/src/hooks/use-watch-query.tsx
```diff
@@ -0,0 +1,151 @@
...
+    for (const change of this.#pendingChanges) {
+      for (const cb of this.#subscribers) {
+        cb(toChange(change));
+      }
+    }
+    this.#pendingChanges = [];
+  }
+}
+
+function changeViewFactory<
```
grgbkr, line 79: Hmmm I wonder if we can reasonably push this down a layer so that it can be used in non-React environments like Cloudflare-DO, NodeJS, vanilla JS.

aboodman:
Your perlow contact in the channel had an amazing idea: to propagate
mutation name and args to client and drive effects off that. There was a
long thread.

a (phone)

PR #3311, opened 2024-12-13
feat(zbugs): load 100 and display older link

grgbkr:


apps/zbugs/src/zero-setup.ts
```diff
@@ -62,14 +63,24 @@ export function preload(z: Zero<Schema>) {
...
     cleanup();
     baseIssueQuery
-      .related('comments', q => q.related('emoji').limit(10))
+      .related('comments', comments =>
+        comments
+          .related('creator', creator => creator.one())
+          .related('emoji', emoji =>
+            emoji.related('creator', creator => creator.one()),
+          )
+          .limit(INITIAL_COMMENT_LIMIT)
```
aboodman, line 81: Why do we preload this many comments @grgbkr ?

aboodman: I think 10 was the right number before at least from a ux pov. Was there some consistency thing?

aboodman: Oh I guess it has something to do with the 'load more' effect you were creating? Is it required? This seems like too aggressive of a preload.

PR #3325, opened 2024-12-14
Add some safety limits

aboodman:


arv:
It is not clear to me why we are limiting the size of strings at this level? Shouldn't that be done at the application layer and not in the crud mutators?

aboodman:
Yeah I ended up doing this a different way. Closing this.

PR #3390, opened 2024-12-17
feat(zero): simple covering algorithm

tantaman:
Putting up for any feedback & help (see TODOs) since we're down to the wire. Still needs more tests, esp around the query manager.


tantaman:
I'm closing this in favor of implementing it on the server.

General idea:
1. All materialized/preloaded queries get registered with the server
2. The server can decide to create a pipeline or not depending on the presence of covering queries for the client group
3. Covering is determined _before_ permission transformations since those transformations are a constant mapping of the query.

PR #3409, opened 2024-12-18
fix(zero-react): added zero client overrides for SSR

0xcadams:
Fixes Zero throwing an error in SSR where `WebSocket` is not available.

### DIFF
--- a/packages/zero-client/src/client/server-option.ts
+++ b/packages/zero-client/src/client/server-option.ts
@@ -56,6 +56,10 @@ function validateServerParam(paramName: string, server: string): HTTPString {
 export function getServer(
   server: string | undefined | null,
 ): HTTPString | null {
+  if (typeof window === 'undefined') {
+    console.debug('Zero started in a non-browser environment, no data will be synced.');
+    return null;
+  }
   if (server === undefined || server === null) {
     console.warn(
       'Zero starting up with no server URL. This is supported for unit testing ' +
--- a/packages/zero-client/src/client/server-option.ts
+++ b/packages/zero-client/src/client/server-option.ts
@@ -1,3 +1,4 @@
+import {getBrowserGlobal} from '../../../shared/src/browser-env.js';
 import type {HTTPString} from './http-string.js';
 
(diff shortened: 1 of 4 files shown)


tantaman:
seems reasonable to me as a stop gap to prevent blowing up in ssr environments. @aboodman

packages/zero-client/src/client/server-option.ts
```diff
@@ -56,6 +57,13 @@ function validateServerParam(paramName: string, server: string): HTTPString {
 export function getServer(
   server: string | undefined | null,
 ): HTTPString | null {
+  const WS = getBrowserGlobal('WebSocket');
+  if (!WS) {
+    console.debug(
+      'Zero started in an unsupported environment, no data will be synced.',
```
aboodman, line 63: I would rather have this be warn. Will add a follow-up commit.

packages/zero-react/src/use-query.tsx
```diff
@@ -30,10 +30,14 @@ export function useQuery<
   const view = viewStore.getView(
     z.clientID,
     q as AdvancedQuery<TSchema, TReturn>,
-    enable,
+    enable && z.server !== null,
```
aboodman, line 33: We already allow the server to be specified as null, so it seems like this should have already worked?

aboodman:
Moved to https://github.com/rocicorp/mono/pull/3445

0xcadams:
Great, thanks!

packages/zero-react/src/use-query.tsx
aboodman: @0xcadams - why was this change needed. It ended up breaking reactivity when a server isn't specified which is something we want to work, e.g., for tests. Can I undo this line?

aboodman: Also can you tell me what environment you experienced this in? I tried Next.js and it already has `WebSocket` polyfilled so it can't be that.

0xcadams: Yeah we ended up moving away from Remix because it was more trouble than it was worth

PR #3422, opened 2024-12-19
feat(zero-vue): Vue bindings for Zero

markusgeert:
This PR adds a `useQuery` composable in order to use Zero in the context of Vue. The bindings are heavily based on the existing Solid ones, but so far they seem to work correctly.

A sample project showcasing the client in a Vue project, based on hello-zero-solid is available [here](https://github.com/markusgeert/hello-zero-vue).


arv:
Very cool.

I think this looks good but we haven't updated the Solid bindings to match the React bindings with the `resultType` part. Maybe it makes sense to do this in this PR or in a followup?

markusgeert:
> I think this looks good but we haven't updated the Solid bindings to match the React bindings with the `resultType` part. Maybe it makes sense to do this in this PR or in a followup?

I brought the return of `useQuery` in line with the react one, probably nice to have that immediately.

One minor thing is that the `QueryResultDetails` type is now duplicated between zero-vue and zero-react, since I didn't really know where to put it otherwise.

arv:
Thank you.

Once things settle a bit after the holidays I would like to merge this.

packages/zero-vue/src/use-query.ts
```diff
@@ -0,0 +1,48 @@
...
+  AdvancedQuery,
+  QueryType,
+  Smash,
+  TableSchema,
+} from '../../zero-advanced/src/mod.js';
+import {readonly, ref, watch, type DeepReadonly, type Ref} from 'vue';
+import {vueViewFactory} from './vue-view.js';
+import type {ResultType} from '../../zql/src/query/typed-view.js';
+
+export type QueryResultDetails = {
```
arv, line 12: This should be Readonly. 

https://github.com/rocicorp/mono/pull/3460/files#diff-bd9e0d16f2377f03eadaae972a1023f3657c6ef25df2886cb1086d5a6ea3939dR16

packages/zero-vue/src/use-query.ts
```diff
@@ -0,0 +1,48 @@
...
+  type: ResultType;
+};
+
+export type QueryResult<TReturn extends QueryType> = [
+  Ref<Smash<TReturn>>,
```
arv, line 17: And this too?

arv (approved):
LGTM

Thanks. This is great.

packages/zero-vue/src/use-query.ts
arv: This cannot be readonly as is because we are mutating the `type` property.

Maybe:

```ts
  readonly #details: QueryResultDetails;
```

should be 

```ts
  readonly #details: Ref<QueryResultDetails>;
```

and then we use a shallowRef for the details?

packages/zero-vue/src/use-query.ts
```diff
@@ -0,0 +1,48 @@
...
+>(
+  queryGetter: () => Query<TSchema, TReturn>,
+): DeepReadonly<QueryResult<TReturn>> {
+  const queryResult: Ref<Smash<TReturn> | undefined> = ref();
+  const details = ref<QueryResultDetails>({type: 'unknown'});
+
+  watch(
+    queryGetter,
+    (query, _, onCleanup) => {
+      const view = (query as AdvancedQuery<TSchema, TReturn>).materialize(
```
arv, line 33: I'm not sure I understand how this works any more :-)

The `view` contains a reactive object. However, the `watch` is only watching the query?

packages/zero-vue/src/use-query.ts
markusgeert: What I've done now is wrap the details in the return of `useQuery` in a `readonly` primitive. The returntype already made it readonly (with `DeepReadonly`), and wrapping it in a `readonly` will also enforce this at runtime.

packages/zero-vue/src/use-query.ts
markusgeert: I made the type readonly. Is this something that should be enforced/checked at runtime?

packages/zero-vue/src/use-query.ts
markusgeert: Hehe, I had to refresh as well.

The data property of the view is a reactive object. This is mutated whenever `push` is called on the view. When the `queryGetter` is mutated, the watch callback is run, which just creates a new view. 

Effectively this _should_ mean that the value of `queryResult` changes when either:

1. The queryGetter is triggered (by creating a new view)
2. OR when push is called on the view (by mutating the view's data)

packages/zero-vue/src/use-query.ts
arv: No need to check these things at runtime.

We use a lot of readonly types to make it clear that changing the properties might change an internal cached value.

packages/zero-vue/src/use-query.ts
arv: Ah, maybe Vue does something with the getter?

arv:
I added a basic useQuery test.

I also want to add this test: https://github.com/rocicorp/mono/pull/3460/files#diff-7756058ea220d207e955c23c130886c54e727c6c5737ae3f895669d52a4d745cR77

Let me know if you want me to take over that?

Other than adding that test and merging/rebasing with main I think this is ready to go.

arv:
@markusgeert There is also this: https://github.com/danielroe/zero-vue/blob/main/src/query.ts. There are some small differences. It is not clear to me how these differences matter?

@danielroe Could you take a look at this PR?

markusgeert:
> @markusgeert There is also this: https://github.com/danielroe/zero-vue/blob/main/src/query.ts. There are some small differences. It is not clear to me how these differences matter?

For one @danielroe is _way_ more qualified than me when it comes to implementing these bindings :). Other than that, I believe the differences are minor. Their implementation allows for a more flexible input type (`MaybeRefOrGetter`) which is great. Cleanup is also handled a bit differently (at start of watch + `onUnmounted` vs `onCleanup`). Although @danielroe might be able to weigh in on what these differences actually entail?

danielroe:
Also, to reiterate what I said to Aaron, more than happy to donate the `zero-vue` package name if you want to keep the bindings here - I would be _delighted_ if so.

arv:
We discussed this yesterday and since we are not Vue experts we want to continue to keep it outside of the mono repo.

At this point we would recommend using @danielroe's implementation.

Hopefully this will all settle after a while and the churn in the advanced API will allow some stability for the vue bindings.

danielroe:
@markumark you would be very welcome to join in maintaining zero-vue, and @arv, if you have any suggestions for it and its implementation/continued development, please let me know. I would love it to be a first class experience for Vue users.

markusgeert:
@arv understandable, thank you for taking a look.

@danielroe I would be happy to. I'll create some issues to start.

PR #3423, opened 2024-12-19
Expose `getDefaultPusher`

masylum:
Current workaround is:

```js
function getDefaultPusher() {
	const r = new Replicache({ name: "foo", licenseKey: "bar" })
	return r.pusher
}
```


aboodman:
Seems reasonable to me.

PR #3425, opened 2024-12-20
chore(zero-cache): log more information when a client is not found

darkgnotic:
Logs show an instance of an assert that should not be possible given the CVR schema constraints (i.e. `clientID` is a foreign key reference):

Screenshot 2024-12-19 at 16 58 42

Indeed, the data in the DB looks well formed:

Screenshot 2024-12-19 at 17 00 41

So this incidence is a bit of a mystery. Add more logging to debug if/when it happens again.


arv (commented):
LGTM

PR #3426, opened 2024-12-20
fix(zero-cache): fix graceful shutdown by awaiting process manager before exiting

darkgnotic:
Reverts #3424 as that did not fix the issue.

Instead, have each process managing worker (`multi/main.ts` and `server/main.ts`) await its process manager before exiting. 

This avoids sending an additional `SIGTERM` to child processes when they are still draining.


darkgnotic:
Graceful shutdowns now work properly with the new multi-tenant process hierarchy:

Screenshot 2024-12-19 at 20 02 47

PR #3432, opened 2024-12-20
fix(sandbox): disable timeouts for requests to change-streamer

darkgnotic:
Another configuration that we need to carry over from the previous incarnation of sandbox. 

Screenshot 2024-12-20 at 15 31 50

Requests to the `change-streamer` are long-lived websockets and thus should not be subject to timeouts.


darkgnotic:
I'll make the same change to prod once sandbox succeeds.

PR #3438, opened 2024-12-21
chore(zero): Switch xxhash to pure js version.

aboodman:
I did some quick benchmarking and though this is much slower than the wasm version, it's in the neighborhood of 4-5x not 10x, so that's nice. Also hashing is just so fast it's not likely to be the biggest of our problems for ast hashing or inserting optimistic data. Let's unblock people from using Zero in env's like RN first and optimize this if it profiles show it is a problem later.

CleanShot 2024-12-20 at 21 20 31@2x


aboodman:
@natew do you have a way to test this locally w/o me doing a build to ensure it fixes things for you.

aboodman:
@darkgnotic do I need to do anything other than bump PROTOCOL_VERSION?

aboodman:
@darkgnotic FYI I don't think `invalidation.ts` was being used, so removed.

tantaman:
we don't even use `xxhash` to hash anything besides the AST.

`export function rowIDHash(id: RowID): string {` is dead code. Check it out: https://github.com/rocicorp/mono/pull/3441

grgbkr:
> `export function rowIDHash(id: RowID): string {` is also [dead code](https://github.com/rocicorp/mono/pull/344). So we don't use xxhash in that many places. Just `toPrimaryKeyString`? which appears to only be used in crud mutations.

It's also used when inserting entities into replicache when processing pokes https://github.com/rocicorp/mono/blob/ab69c6cc8f0cca9d568bdc2372bf555319cd9a33/packages/zero-client/src/client/zero-poke-handler.ts#L317
this is probably the most performance sensitive usage.

darkgnotic:
> @darkgnotic do I need to do anything other than bump PROTOCOL_VERSION?

That should be it. 

Good catch on needing to bump the PROTOCOL_VERSION. Even though the structure doesn't change, all of the query's hashes will.  👍

packages/zero-client/src/client/keys.test.ts
```diff
@@ -29,21 +29,27 @@ test('toPrimaryKeyString', () => {
       issueID: 'issue1',
       labelID: 'label1',
     }),
-  ).toMatchInlineSnapshot(`"e/issue_label/dqncq7kzfa6fpmhj26azwoy9t"`);
+  ).toMatchInlineSnapshot(
+    `"e/issue_label/15328927053344014787296837153592575662"`,
```
arv, line 33: What was the motivation to not use `toString(36)` any more?

arv (commented):
LGTM

packages/zero-client/src/client/keys.test.ts
aboodman: Whoops, it wasn't a conscious choice. Now I get to redo all these tests 😂.

PR #3440, opened 2024-12-21
fix(zero-cache): fix changedb race in rep-manager handoff

darkgnotic:
Delay accessing the change db until the replication stream has been successfully subscribed to.

This provides the single writer guarantee that the new change-streamer does not miss any changes written by the old change-streamer.


darkgnotic:
FYI, I think ultimately it would be clearer and safer to use some kind of locking or test-and-set mechanism on the change DB itself. I'll think about this more.

PR #3446, opened 2024-12-22
feat(zero-cache): split the row key out of put-row patches

darkgnotic:
Bumps the `zero-protocol` version to 4, in which a `put` row patch contains not a full row `value`, but separate `id` and `rest` fields, the first comprising the primary key (as it does for `del` and `update` patches), and the latter comprising the remainder of the row.

All fields are made optional for backwards compatibility. For v3 and under, the legacy `value` field is set, and for v4 and up, the `id` and `rest` fields are set. The client code handles both cases.

In a later release, we will bump the version to 5, remove the legacy `value` field, and make `id` and `rest` required.


packages/zero-cache/src/services/view-syncer/client-handler.ts
```diff
@@ -289,6 +308,16 @@ function makeRowPatch(patch: RowPatch): RowPatchOp {
   }
 }
 
+export function rest(id: RowKey, value: Row) {
+  const rest: Record<string, Value> = {};
+  for (const key in value) {
```
tantaman, line 313: Use `for (const key of Object.keys(value))` so it only checks the current object and not also the prototype.

packages/zero-cache/src/services/view-syncer/client-handler.ts
```diff
@@ -263,19 +273,28 @@ const lmidRowSchema = v.object({
...
     op,
     id: {table: tableName, rowKey: id},
   } = patch;
 
   switch (op) {
-    case 'put':
+    case 'put': {
+      const value = v.parse(ensureSafeJSON(patch.contents), rowSchema);
+      const contents =
+        protocolVersion >= 4
```
tantaman, line 286: this protocolVersion is a property of the connection and set by the client, correct?

packages/zero-client/src/client/zero-poke-handler.ts
```diff
@@ -320,16 +320,18 @@ function rowsPatchOpToReplicachePatchOp(
           op.id,
         ),
       };
-    case 'put':
+    case 'put': {
+      const value = op.value ? op.value : {...op.id, ...op.rest};
       return {
         op: 'put',
         key: toPrimaryKeyString(
```
tantaman, line 327: would you update `toPrimaryKeyString` in v5 to not rely on the schema and only take in `op.id`?

tantaman: Did you and @grgbkr  already sketch out a plan on how to handle when the primary key format changes on the client? We don't even handle this today. I'd imagine everything gets bricked (or behaves strangely) today if we change the columns involved in a primary key in `schema.ts`.

darkgnotic:
Thank you @tantaman. As discussed, abandoning this approach as it is not necessary.

PR #3450, opened 2024-12-23
feat(zero-cache): fix paths to enable running under windows

tantaman:


tantaman:
closed in favor of https://github.com/rocicorp/mono/pull/3453

PR #3459, opened 2024-12-28
feat(zql): fix `IS` to take null, fix `Row` to take `Query`

tantaman:
This came up from a user here: https://discord.com/channels/830183651022471199/1322274277453594705/1322276660627898470

The question was how they could derive the types to use for `props` in a component from a query.

### DIFF
--- a/apps/zbugs/src/pages/issue/comment.tsx
+++ b/apps/zbugs/src/pages/issue/comment.tsx
@@ -1,7 +1,6 @@
 import type {Row} from '@rocicorp/zero';
 import classNames from 'classnames';
 import {memo, useState} from 'react';
-import type {Schema} from '../../../schema.js';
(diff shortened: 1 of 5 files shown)


apps/zbugs/src/pages/issue/comment.tsx
```diff
@@ -10,21 +9,17 @@ import {EmojiPanel} from '../../components/emoji-panel.js';
...
 
 type Props = {
   id: string;
   issueID: string;
-  comment: Row<Schema['tables']['comment']> & {
-    readonly creator: Row<Schema['tables']['user']> | undefined;
-  } & {
-    readonly emoji: readonly Emoji[];
-  };
+  comment: Row<ReturnType<typeof commentQuery>>;
```
arv, line 22: Maybe export the type instead?

arv (commented):
LGTM

PR #3460, opened 2024-12-28
feat(zero-solid): Add resultType support to zero-solid

grgbkr:
Bring zero-solid up to par with React features added in https://github.com/rocicorp/mono/pull/3229/files

### DIFF
--- a/packages/zero-solid/src/use-query.ts
+++ b/packages/zero-solid/src/use-query.ts
@@ -7,11 +7,21 @@ import type {
   TableSchema,
 } from '../../zero-advanced/src/mod.js';
 import {solidViewFactory} from './solid-view.js';
+import type {ResultType} from '../../zql/src/query/typed-view.js';
+
+export type QueryResultDetails = Readonly<{
+  type: ResultType;
+}>;
+
+export type QueryResult<TReturn extends QueryType> = readonly [
+  Smash<TReturn>,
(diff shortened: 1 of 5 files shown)


arv (approved):
LGTM

arv (commented):
LGTM

packages/zero-solid/src/use-query.ts
```diff
@@ -7,11 +7,21 @@ import type {
...
+export type QueryResult<TReturn extends QueryType> = readonly [
+  Smash<TReturn>,
+  QueryResultDetails,
+];
 
 export function useQuery<
   TSchema extends TableSchema,
   TReturn extends QueryType,
->(querySignal: () => Query<TSchema, TReturn>): Accessor<Smash<TReturn>> {
+>(querySignal: () => Query<TSchema, TReturn>): Accessor<QueryResult<TReturn>> {
```
arv, line 24: After using this in hello-zero-solid it is clear to me that the return type of this should be:

```ts
[
  Accessor<Smash<TReturn>>,
  Accessor<QueryResultDetails>,
]
```

The way it is done now, one cannot use destructuring which makes the code a bit inconvenient.

arv: https://github.com/rocicorp/hello-zero-solid/pull/9

PR #3461, opened 2024-12-29
fix(zero-cache): fix replication of TRUNCATE statements

darkgnotic:
Truncates were initially processed by converting the TRUNCATE event into a sequence of `del` patches that were to be pushed through IVM pipelines. This never worked, as the generation of this sequence opened an iterator over the `prev` database handle, and that database needs to be concurrently advanced by the IVM push.

Instead, truncates are now processed in the same manner as that of schema changes; the advancement is aborted and pipelines are rehydrated from the `curr` database handle.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/snapshotter.ts
+++ b/packages/zero-cache/src/services/view-syncer/snapshotter.ts
@@ -222,8 +222,8 @@ export interface SnapshotDiff extends Iterable<Change> {
  * change or truncate is encountered, which result in aborting the
  * advancement and resetting / rehydrating the pipelines.
  */
-export class ResetPipelineSignal extends Error {
-  readonly name = 'ResetPipelineSignal';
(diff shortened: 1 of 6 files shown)


packages/zero-cache/src/services/view-syncer/snapshotter.ts
```diff
@@ -219,13 +219,14 @@ export interface SnapshotDiff extends Iterable<Change> {
...
-export class SchemaChangeError extends Error {
-  readonly name = 'SchemaChangeError';
+export class ResetPipelinesSignal extends Error {
+  readonly name = 'ResetPipelinesSignal';
 
-  constructor(table: string) {
-    super(`schema for table ${table} has changed`);
+  constructor(msg: string) {
+    super(msg);
   }
```
arv, line 230: you can remove the constructor now

arv (commented):
LGTM

PR #3472, opened 2025-01-03
feat(zql): support nested property access in JWTs

tantaman:


packages/zero-cache/src/auth/read-authorizer.query.test.ts
```diff
@@ -312,6 +316,13 @@ const permissions = must(
       authData: AuthData,
       {cmpLit}: ExpressionBuilder<TableSchema>,
     ) => cmpLit(authData.role, '=', 'admin');
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    type TODO = any;
+    const isAdminThroughNestedData = (
+      authData: AuthData,
+      {cmpLit}: ExpressionBuilder<TableSchema>,
+      // TODO: proxy should return parameter references instead....
+    ) => cmpLit(authData.properties?.role as TODO, 'IS', 'admin');
```
aboodman, line 325: Do arrays work too?

packages/zero-protocol/src/ast.ts
```diff
@@ -320,7 +321,7 @@ type StaticParameter = {
   // Read and write authorization will both send the
   // current authentication data (authData).
   readonly anchor: 'authData' | 'preMutationRow';
-  readonly field: string;
+  readonly field: string | string[];
```
arv, line 324: Does AST changes require a version change?

If the client is at commit n+1 and the server at n there will be problems

arv (commented):
LGTM

Very nice.

Like @aboodman said, maybe add a test with an array.

PR #3473, opened 2025-01-05
feat(zero-react): do not render with empty data on first pass

tantaman:
First render pass was returning an empty array then the actual contents.


grgbkr:
LGTM

PR #3474, opened 2025-01-06
feat(schema): added strong optional typing

0xcadams:
This adds strong typing to the optional property passed into e.g. `column.string(false)`, instead of `optional: boolean`

```
const _demo = string(false)
//     ^ { type: 'string', optional: false, customType: string }

const _demo2 = string()
//     ^ { type: 'string', optional: false, customType: string }

const _demo3 = string(true)
//     ^ { type: 'string', optional: true, customType: string }
```


arv (approved):
LGTM

Thank you. Much nicer

arv:
FYI, these things can be "unit" tested but no need at this point.

PR #3476, opened 2025-01-06
fix(zero-react): useSyncExternalStore shim

arv:
https://bugs.rocicorp.dev/issue/3388

This is to allow zero-react to work prior to the release of react 18.


arv:
Thanks @jorroll for the bug report.

PR #3478, opened 2025-01-06
chore(zero): Replace `lodash` with alternatives

beeequeue:
This PR replaces usages of `lodash` with smaller alternatives as part of the [e18e](https://e18e.dev) initiative.

---

I replaced `lodash.kebabcase` and `lodash.snakecase` with [`kasi`](https://github.com/fabiospampinato/kasi) as it was recommended in the latest [blog post](https://e18e.dev/blog/november-contributions-showcase.html#kasi). This should result in about a 40x size reduction for this code (10kb -> 250b)

---

I replaced `lodash.merge` with [`defu`](https://github.com/unjs/defu) from unjs, which should should result in a 13x size reduction (10.6kb -> 800b) and a perf improvement:
image


packages/shared/src/options.ts
```diff
@@ -163,6 +162,16 @@ export type Config<O extends Options> = {
     : never;
 };
 
+/**
+ * Creates a defu instance that overrides arrays instead of merging them.
+ */
+const defu = createDefu((obj, key, value) => {
```
arv, line 168: @darkgnotic I'm not familiar with intent here. Is this the right merging behavior?

arv (approved):
LGTM

packages/shared/src/options.ts
beeequeue: It's what the tests are expecting, so it's what I made it do 😄

darkgnotic: Good question! Currently we never set any default arrays other than the empty array, and we haven't yet encountered a use case for a non-empty default, but my hunch is that we will want flags to override the default rather than add to it. So this change is good. If we change our minds, we know where to look.  😄

darkgnotic (approved):
Thank you @beeequeue!

PR #3479, opened 2025-01-06
chore(zero,zero-cache): Replace strip-ansi with native impl

beeequeue:
This PR replaces usages of `strip-ansi` with the [native node implementation](https://nodejs.org/api/util.html#utilstripvtcontrolcharactersstr) as part of the [e18e](https://e18e.dev) initiative.

`stripVTControlCharacter` is available from node v16.11.0 which should be fine as `@rocicorp/zero` requires v20

### DIFF
--- a/packages/shared/src/options.ts
+++ b/packages/shared/src/options.ts
@@ -6,7 +6,7 @@ import commandLineUsage from 'command-line-usage';
 import kebabcase from 'lodash.kebabcase';
 import merge from 'lodash.merge';
 import snakeCase from 'lodash.snakecase';
-import {stripVTControlCharacters} from 'node:util';
+import {stripVTControlCharacters as stripAnsi} from 'node:util';
 import {assert} from './asserts.js';
 import {must} from './must.js';
 import * as v from './valita.js';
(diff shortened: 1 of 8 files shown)


packages/shared/src/options.ts
```diff
@@ -6,7 +6,7 @@ import commandLineUsage from 'command-line-usage';
 import kebabcase from 'lodash.kebabcase';
 import merge from 'lodash.merge';
 import snakeCase from 'lodash.snakecase';
-import stripAnsi from 'strip-ansi';
+import {stripVTControlCharacters as stripAnsi} from 'node:util';
```
arv, line 9: @darkgnotic Is it Ok to depend on `node:util` here?

I also wonder about CF workers and other environments for this "library".

beeequeue: I didn't actually look up what runtimes you guys support before opening this, but it's supported in node, deno, and [CF workers (with nodejs_compat enabled)](https://workers-nodejs-compat-matrix.pages.dev/), but [not bun](https://bun.sh/docs/runtime/nodejs-apis#node-util).

darkgnotic: Yes, we only support running `zero-cache` on node.

darkgnotic (approved):

beeequeue:
(the tests process never exited after finishing for whatever reason)

darkgnotic:
> (the tests process never exited after finishing for whatever reason)

We haven't figured out why this occasionally happens. Can you try re-running after updating with the base branch?

arv (approved):

PR #3480, opened 2025-01-07
fix(zero-solid)!: useQuery return tuple of Accessor

arv:
BREAKING CHANGE!

Previously it returned an accessor of a tuple which is not ergonomic.

It forced you to do:

```tsx
const rows = () = useQuery(q)()[0]; // yes you need the ()[0]
</List items={rows()}/>

// or
const res = useQuery(q);
</List items={rows()[0]}/>
```

With this change you can do:

```tsx
const [rows] = useQuery(q);
</List items={rows()}/>
```

### DIFF
--- a/packages/zero-solid/src/use-query.ts
+++ b/packages/zero-solid/src/use-query.ts
@@ -1,37 +1,35 @@
-import {type Accessor, createMemo, onCleanup} from 'solid-js';
+import {type Accessor, onCleanup} from 'solid-js';
 import type {
-  Query,
   AdvancedQuery,
+  Query,
   QueryType,
   Smash,
   TableSchema,
 } from '../../zero-advanced/src/mod.js';
-import {solidViewFactory} from './solid-view.js';
 import type {ResultType} from '../../zql/src/query/typed-view.js';
+import {solidViewFactory} from './solid-view.js';
 
 export type QueryResultDetails = Readonly<{
   type: ResultType;
 }>;
 
 export type QueryResult<TReturn extends QueryType> = readonly [
-  Smash<TReturn>,
-  QueryResultDetails,
+  Accessor<Smash<TReturn>>,
+  Accessor<QueryResultDetails>,
 ];
 
 export function useQuery<
   TSchema extends TableSchema,
   TReturn extends QueryType,
->(querySignal: () => Query<TSchema, TReturn>): Accessor<QueryResult<TReturn>> {
-  return createMemo(() => {
-    const query = querySignal();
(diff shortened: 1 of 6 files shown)


grgbkr:
Agree this is a significant improvement in DX. 

Have you tested this in hello-zero-solid sample app? (something I should have done with the original).

packages/zero-solid/src/use-query.ts
```diff
@@ -1,37 +1,35 @@
...
-    const query = querySignal();
-    const view = (query as AdvancedQuery<TSchema, TReturn>).materialize(
-      solidViewFactory,
-    );
-
-    onCleanup(() => {
-      view.destroy();
-    });
+>(querySignal: () => Query<TSchema, TReturn>): QueryResult<TReturn> {
+  const query = querySignal();
```
grgbkr, line 25: I'm worried without the use of createMemo that this function body can be run multiple times without querySignal's value changing. 

https://docs.solidjs.com/concepts/derived-values/memos

grgbkr: or maybe even better
```
export function useQuery<
  TSchema extends TableSchema,
  TReturn extends QueryType,
>(
  querySignal: () => Query<TSchema, TReturn>,
): [Accessor<Smash<TReturn>>, Accessor<QueryResultDetails>] {
  const memoed = createMemo(() => {
    const query = querySignal();
    const view = (query as AdvancedQuery<TSchema, TReturn>).materialize(
      solidViewFactory,
    );

    onCleanup(() => {
      view.destroy();
    });

    return [view.data, {type: view.resultType}] as const;
  });

  return [() => memoed()[0], () => memoed()[1]];
}
```

grgbkr: Looking at this I think I see a bug with the previous code which may be related to the iloop.  `view.data` was accessed in the `createMemo` and `view.data` is a property of a solid Store, thus a reactive value.  

So I think anytime the value of `view.data` changed (as determined using ===), the `createMemo` body was rerun, materializing a new view.

packages/zero-solid/src/use-query.ts
```diff
@@ -31,7 +27,8 @@ export function useQuery<
     onCleanup(() => {
       view.destroy();
     });
-
-    return [view.data, {type: view.resultType}] as const;
+    return view;
   });
+
+  return [() => view().data, () => view().resultDetails];
```
arv, line 33: I could not come up with anything simpler than this.

arv (commented):
PTAL

arv:
Tomorrow I'll checkout the old version and test your theory related to the store changing inside the createMemo.

PR #3481, opened 2025-01-07
feat(zero-schema): Add runtime assertion for all relationship tables in createSchema

jer-k:
zbugs: [Potential for a better error message when a table is referenced in a relationship but not included in createSchema](https://bugs.rocicorp.dev/issue/3370)

Thank you @arv for the work on https://github.com/rocicorp/mono/pull/3477. It pointed me in the direction of where to look code wise. I wanted to expand the work that was done to resolve the issue I ran into where defining a `destSchema` for a table that isn't referenced in `schema.tables` causes an error.

I added test cases for two potential ways to have a missing table. If anyone can think of other scenarios I can gladly write more tests.

My only question about adding this, while I think it improves the DX and will help people to ensure their schema is valid, is there any concern about runtime cost and having `createSchema` doing this additional work?

### DIFF
--- a/packages/zero-schema/src/normalize-table-schema.ts
+++ b/packages/zero-schema/src/normalize-table-schema.ts
@@ -33,7 +33,7 @@ export class NormalizedTableSchema implements TableSchema {
   constructor(
     tableSchema: TableSchema,
     tableSchemaCache: TableSchemaCache,
-    assertDest: AssertDestFunc,
+    assertFieldRelation: AssertFieldRelation,
   ) {
     this.tableName = tableSchema.tableName;
     const primaryKey = normalizePrimaryKey(tableSchema.primaryKey);
@@ -44,7 +44,7 @@ export class NormalizedTableSchema implements TableSchema {
       this.tableName,
       tableSchema.relationships,
       tableSchemaCache,
-      assertDest,
+      assertFieldRelation,
     );
   }
 }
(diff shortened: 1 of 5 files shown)


tantaman:
Thanks. Also feels like I should resurrect the conversation around the builder API for the schema as these can be statically checked in that case.

https://github.com/rocicorp/mono/pull/3137

arv:
Sorry. I just realized that we already have some of this code in a different place.

https://github.com/rocicorp/mono/blob/b4615feaa51aa7035d3411d8c8bf8bf9e4f1f99c/packages/zero-schema/src/normalized-schema.ts#L17

In the spirit of reducing code duplication I think `createSchema` should be:

```ts
export function createSchema<const S extends Schema>(schema: S): S {
  // Ignore return value of normalizeSchema. At this level we want to return the
  // same passed in object for less surprises.
  normalizeSchema(schema);
  return schema as S;
}
```

then update normalizeSchema to also do the validation you are adding in this PR.

As a bonus we could attach the NormalizedSchema to the schema using a WeakMap so we do not have to redo this work later.

WDYT?

jer-k:
> Sorry. I just realized that we already have some of this code in a different place.
> 
> https://github.com/rocicorp/mono/blob/b4615feaa51aa7035d3411d8c8bf8bf9e4f1f99c/packages/zero-schema/src/normalized-schema.ts#L17
> 
> In the spirit of reducing code duplication I think `createSchema` should be:
> 
> ```ts
> export function createSchema<const S extends Schema>(schema: S): S {
>   // Ignore return value of normalizeSchema. At this level we want to return the
>   // same passed in object for less surprises.
>   normalizeSchema(schema);
>   return schema as S;
> }
> ```
> 
> then update normalizeSchema to also do the validation you are adding in this PR.
> 
> As a bonus we could attach the NormalizedSchema to the schema using a WeakMap so we do not have to redo this work later.
> 
> WDYT?

I think that sounds like a good approach. Might be a little over my head in terms of implementation, but having the all the validation be co-located seems like the right approach.

arv:
Ok. I'll take over this PR.

PR #3482, opened 2025-01-07
feat(zbugs): add bundle analyzer so we can start paring down size

tantaman:
```ts
npm run analyze
```
t
CleanShot 2025-01-07 at 15 33 17

One quick win: we could replace `b+tree` with the `treap` we were using a ways back.


aboodman:
Does it know about dead code? I'm always wary about these things because so
much changes at bundle time.

PR #3484, opened 2025-01-08
fix(zql): prevent take's early return assert from swallowing errors

grgbkr:
If during hydrate, fetching from take's input throws an error this error was masked by take's finally throwing an error with the message 'Unexpected early return prevented full hydration'.

Update the code to only throw 'Unexpected early return prevented full hydration' only in the case of an early return, not in the case of an exception.


RobertoSnap:
Fixed 👍

PR #3485, opened 2025-01-08
chore(zero-cache): api to run unit tests on custom pg instance

darkgnotic:
Adds a `CUSTOM_PG=postgresql?... npx run tests` option for running the `zero-cache` tests on a custom PG instance, such as an Aurora instance.

Fix a tests to clean up global state, so that it can be rerun on the same pg instance multiple times.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/pg/schema/shard.pg-test.ts
+++ b/packages/zero-cache/src/services/change-streamer/pg/schema/shard.pg-test.ts
@@ -19,6 +19,7 @@ describe('change-source/pg', () => {
 
   afterEach(async () => {
     await testDBs.drop(db);
+    await testDBs.sql`RESET ROLE; DROP ROLE IF EXISTS supaneon`.simple();
   });
 
   function publications() {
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/services/change-streamer/pg/schema/shard.pg-test.ts
```diff
@@ -19,6 +19,7 @@ describe('change-source/pg', () => {
 
   afterEach(async () => {
     await testDBs.drop(db);
+    await testDBs.sql`RESET ROLE; DROP ROLE IF EXISTS supaneon`.simple();
```
arv, line 22: why was this needed?

darkgnotic: Good question. Roles are not scoped to a database, so they don't get cleaned up when we drop the test db. This doesn't matter when we're running against a brand new Postgres cluster from testcontainers, but if for example, we reuse an Aurora instance and run tests again, the role is still around and the test fails to create the role.

So using long-lived database instances of testing is a bit tricky because cleanup is especially important. For now I'm only using it manually to vet Aurora compatibility.

PR #3489, opened 2025-01-08
chore(zero-protocol): bump to v4, start renaming delPatch.id to delPatch.value

darkgnotic:
Start the migration to send the full row in `del` patches to make the protocol agnostic to the primary key.

* A new (optional) `value` field is introduced to hold the full row value
* `id` is made optional
* server must set one or the other
* client accepts either; the logic for extracting the key from the full row is already in place

TODO: Remove obsolete `update` patch.

### DIFF
--- a/packages/zero-protocol/src/row-patch.ts
+++ b/packages/zero-protocol/src/row-patch.ts
@@ -23,7 +23,7 @@ const delOpSchema = v.object({
   // Either `id` or `value` must be set.
   // Migration plan:
   // - Start setting `value` on the server instead of `id` and
-  //   set MIN_SUPPORTED_PROTOCOL_VERSION = 4.
+  //   set MIN_SERVER_SUPPORTED_PROTOCOL_VERSION = 4.
   // - Remove `id` and make `value` required to make
   //   MIN_SERVER_SUPPORTED_PROTOCOL_VERSION = 5.
   id: primaryKeyValueRecordSchema.optional(),
--- a/packages/zero-protocol/src/row-patch.ts
+++ b/packages/zero-protocol/src/row-patch.ts
@@ -20,7 +20,14 @@ const updateOpSchema = v.object({
 const delOpSchema = v.object({
   op: v.literal('del'),
   tableName: v.string(),
-  id: primaryKeyValueRecordSchema,
+  // Either `id` or `value` must be set.
+  // Migration plan:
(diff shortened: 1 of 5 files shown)


packages/zero-protocol/src/row-patch.ts
```diff
@@ -20,7 +20,14 @@ const updateOpSchema = v.object({
 const delOpSchema = v.object({
   op: v.literal('del'),
   tableName: v.string(),
-  id: primaryKeyValueRecordSchema,
+  // Either `id` or `value` must be set.
+  // Migration plan:
+  // - Start setting `value` on the server instead of `id` and
+  //   set MIN_SUPPORTED_PROTOCOL_VERSION = 4.
```
grgbkr, line 26: s/MIN_SUPPORTED_PROTOCOL_VERSION/MIN_SERVER_SUPPORTED_PROTOCOL_VERSION

packages/zero-protocol/src/row-patch.ts
```diff
@@ -20,7 +20,14 @@ const updateOpSchema = v.object({
 const delOpSchema = v.object({
   op: v.literal('del'),
   tableName: v.string(),
-  id: primaryKeyValueRecordSchema,
+  // Either `id` or `value` must be set.
```
grgbkr, line 23: Right now id always need to be set until MIN_SUPPORTED_PROTOCOL_VERSION is bumped to 4 right?

grgbkr, line 23: Would this perhaps simpler migration plan work:
1. PROTOCOL_VERSION = 4 id is optional value is required, server sets both, client reads value (no fallback to id)
2. PROTOCOL_VERSION = 5 id is removed

grgbkr:
LGTM, though maybe the migration can be simplified.

packages/zero-protocol/src/row-patch.ts
darkgnotic: Thank you, fixed!

packages/zero-protocol/src/row-patch.ts
darkgnotic: This will work too. So many choices.  😁 

The one initially outlined avoids sending more data, so I have a slight preference for that. But I think either is fine; I don't think we need to think too deeply about it since it's all internal.

darkgnotic:
Thanks @grgbkr !

PR #3491, opened 2025-01-09
feat(change-protocol): reorganize types and export a change source protocol

darkgnotic:
🚧  Under construction. Not yet ready for fully supported developer use. 🚧 

1. code reorganization: moved `change-source` related packages out of the `change-streamer` package to make it easier to understand the separation of reponsibilities
2. exported a `@rocicorp/zero/change-protocol` package for that developers to reference when building custom `change-source` implementations
3. added two new messages in the change-protocol that are not yet used, but will be soon: 
  * an "ack-requested" downstream message from the `change-source` implementation, and 
  * an "ack" upstream message that the `change-streamer` send in response.

The current postgres `change-source` implementation will be migrated to use the new messages in [3] in a subsequent step.


darkgnotic:
@arv, would appreciate if you verified that I got the exporting of the module right from the `zero` npm.  🙏 

@cbnsndwch feedback welcome!

darkgnotic:
> @arv, would appreciate if you verified that I got the exporting of the module right from the `zero` npm. 🙏
> 

For example, I don't see a `mod.d.ts` file in the `zero/out/zero-cache/src/services/change-source/protocol/` directory, so I'm quite sure I'm missing something.  🤔

packages/zero-cache/src/services/change-source/protocol/mod.ts
```diff
@@ -0,0 +1,4 @@
+// Types are externally exported under stable versions. Code within
```
arv, line 1: We can enforce this with eslint. See /eslint-config.json

packages/zero/tool/build.js
```diff
@@ -100,6 +100,7 @@ async function buildZeroClient() {
     react: basePath('src/react.ts'),
     solid: basePath('src/solid.ts'),
     advanced: basePath('src/advanced.ts'),
+    ['change-protocol']: basePath('src/change-protocol.ts'),
```
arv, line 103: This is not quite right.

These are only for the client side bundle (with code splitting)

Look at the `/server` part in `package.json` instead.

arv:
> For example, I don't see a `mod.d.ts` file in the `zero/out/zero-cache/src/services/change-source/protocol/` directory, so I'm quite sure I'm missing something. 🤔

You need to add `src/change-protocol.ts` to  packages/zero/tsconfig.server.json

packages/zero/tool/build.js
darkgnotic: I would like to export the `change-source/protocol/...` valita and type definitions as something like `@rocicorp/zero/change-protocol`, so that developers can reference it when building their custom change sources. 

I was hoping I could still keep the code in that deeply nested zero-cache package, but I can move it out if I need to.

I have this in `package.json` but it doesn't seem to work on its own:

```json
    "./server": {
      "types": "./out/zero-cache/src/server/main.d.ts",
      "default": "./out/zero-cache/src/server/main.js"
    },
    "./advanced": {
      "types": "./out/zero/src/advanced.d.ts",
      "default": "./out/advanced.js"
    },
    "./change-protocol": {
      "types": "./out/zero-cache/src/services/change-source/protocol/mod.d.ts",
      "default": "./out/change-protocol.js"
    }
```

packages/zero-cache/src/services/change-source/protocol/mod.ts
darkgnotic: I'll sync up with you on this. Not sure I understand.

cbnsndwch:
Oh wow, I missed this! Super excited for this!

PR #3496, opened 2025-01-10
feat(zero-cache): run/manage litestream from within zero-cache

darkgnotic:
Run `litestream restore` and `litestream replicate` as subprocesses of `zero-cache`.

This inverts the previous setup in which `litestream restore` was run in the `start-dual.sh` script and `litestream replicate` executed the `zero-cache` as a subprocess.

This new structure allows the following:
* Waiting for and retrying `litestream restore` in the `view-syncer` configuration if the backup is not yet available
* Delaying `litestream replicate` until initial sync has happened, avoiding database lock timeout situations like https://bugs.rocicorp.dev/issue/3209
* Allowing `--auto-reset` configured `zero-cache`s (e.g. running on supabase or neon) to work with `litestream`, since the auto reset would now happen before `litestream replicate` is called
* Supporting `litestream` restore / replicate in a multi-tenant environment, with the backup location ferried in the per-tenant `ZERO_LITESTREAM_BACKUP_LOCATION` variable


cesara (approved):
lgtm!

PR #3500, opened 2025-01-10
chore(replicache): Remove MetaType.IndexChangeSDD

arv:
This is one of many to remove SDD from Replicache


arv:
Motivation to see the zero size bundle checker get some results.

PR #3511, opened 2025-01-11
chore(zero-cache): faster CVR row flushes with json_to_recordset()

darkgnotic:
Replace the multi-`INSERT` array statement with a `json_to_recordset()` based strategy, as recommended by the author of the postgres client: https://github.com/porsager/postgres/issues/997#issuecomment-2552133528

This method:
* Avoids maxing out on the number of parameters per statement, as it becomes a single parameter, obviating the need to split the commit into batches.
* Avoids arbitrary numbers of prepared statements, obviating the prepared / unprepared workaround used to reduce the number of prepared statements. (Instead, only a single prepared statement is used.)

Local testing shows that this results in up to 25% improvement in flush times, presumably due to the elimination of unprepared statements for arrays of length < 128.


darkgnotic:
FYI, I confirmed an even bigger performance improvement with a loadtest in production.

Before:

Screenshot 2025-01-15 at 16 36 57

After:

Screenshot 2025-01-15 at 16 37 11

And the icing on the cake:

Screenshot 2025-01-15 at 15 57 33

PR #3513, opened 2025-01-13
chore(zero-cache): add the commit watermark to "begin" messages

darkgnotic:
The first step in migrating to replica schema v2 (https://bugs.rocicorp.dev/issue/3404).

The ChangeSource and ChangeStreamer now use the `commitLSN` for the `commitWatermark` of `begin` messages and the `watermark` of `commit` messages. The former are not used yet, but will be in a subsequent step.


arv (commented):
LGTM

PR #3515, opened 2025-01-13
chore(zero-cache): finish migration of change protocol to include commit watermarks

darkgnotic:
This is a followup to #3513 to complete the migration of the change protocol to include commit watermarks in the `["begin", ... ]` message.

They are not used yet; this is just a protocol change, which had to be done in two steps because it is sent over the wire between `view-syncer`s and the `replication-manager`.


arv (commented):
LGTM

This must have been fun :')

PR #3517, opened 2025-01-13
fix(zero): Do not "bundle" change-protocol

arv:
The server part of zero is not bundled using esbuild. It is just compiled using tsc and the code structure maps the source code structure.

Followup to https://github.com/rocicorp/mono/pull/3491

### DIFF
--- a/packages/zero/package.json
+++ b/packages/zero/package.json
@@ -93,7 +93,7 @@
     },
     "./change-protocol": {
       "types": "./out/zero-cache/src/services/change-source/protocol/mod.d.ts",
-      "default": "./out/change-protocol.js"
+      "default": "./out/zero-cache/src/services/change-source/protocol/mod.js"
     }
   },
   "bin": {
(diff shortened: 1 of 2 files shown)


darkgnotic (commented):
The change-protocol is intended to be an interface between zero-cache and custom upstream providers (like the one @cbnsndwch is building for Mongo). So the intention is allow developers to reference these classes in their implementations, like:

Screenshot 2025-01-09 at 11 15 10

What's the right way to do that?

arv:
The code in your screenshot works with this PR.

I think we should change it to:

```js
import {ChangeSourceDownstream} from '@rocicorp/zero/change-protocol/v0';
```

darkgnotic:
Ah, got it. So the "exports" in the `package.json` is all that's needed. Thanks @arv!

> The code in your screenshot works with this PR.
> 
> I think we should change it to:
> 
> ```js
> import {ChangeSourceDownstream} from '@rocicorp/zero/change-protocol/v0';
> ```

PR #3520, opened 2025-01-13
feat(zero-cache): remove initial-watermark logic from change source protocol

darkgnotic:
Remove the complex, fragile logic around determining the "initialWatermark" of a change stream. As part of the migration to Replica schema v2 (https://bugs.rocicorp.dev/issue/3404), we can now simply rely on the commit watermarks declared in "begin" messages (#3513 #3515), avoiding the need to compute a watermark one-after the last commit. 

This also simplifies the task of implementing a change-source for a custom upstream.


arv (commented):
LGTM

PR #3524, opened 2025-01-13
fix(zbugs): Playing how many times can we fix the same tiny bug.

aboodman:
A few things were broken:

- j/k were busted because I didn't remember that data- attributes have to be strings.
- the logo no longer soft-nav'd because it was an image nested inside a link.


arv (commented):
LGTM

PR #3529, opened 2025-01-14
feat(zero-client): Add `storageKey` constructor param.

aboodman:
I couldn't figure out how to test this. I added a test like this to `zero.test.ts` and it behaves as if the storage isn't shared while I can plainly see it is in the real browser. Also couldn't figure out how to open a debugger to test:

```ts
test('storage partioning', async () => {
  const schema = {
    version: 1,
    tables: {
      foo: {
        columns: {
          id: {type: 'string'},
          val: {type: 'number'},
        },
        primaryKey: ['id'],
        tableName: 'foo',
      },
    },
  } as const;

  const z1 = zeroForTest({
    schema,
    userID: 'u1',
    kvStore: 'idb',
  });

  await z1.mutate.foo.insert({id: 'a', val: 1});
  await z1.close();

  const z2 = zeroForTest({
    schema,
    userID: 'u1',
    kvStore: 'idb',
  });

  const {promise, resolve} = resolver();
  const view = z2.query.foo.where('id', 'a').one().materialize();
  view.addListener((row, type) => {
    console.log(row, type);
    if (type === 'complete') {
      expect(row).deep.equal({id: 'a', val: 1});
      resolve();
    }
  });

  await promise;
});

```

### DIFF
--- a/packages/zero-client/src/client/options.ts
+++ b/packages/zero-client/src/client/options.ts
@@ -38,14 +38,21 @@ export interface ZeroOptions<S extends Schema> {
   /**
    * A unique identifier for the user. Must be non-empty.
    *
-   * For efficiency, a new Zero instance will initialize its state from
-   * the persisted state of an existing Zero instance with the same
-   * `userID`, domain and browser profile.
+   * Each userID gets its own client-side storage so that the app can switch
+   * between users without losing state.
    *
    * This must match the user identified by the `auth` token if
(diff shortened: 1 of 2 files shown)


arv:
To get a debugger in a real browser:

```
npx vitest --browser.headless=false
```

You sample code should work.

arv (commented):
LGM

PR #3530, opened 2025-01-14
chore(zero): Replace b+tree with treap

arv:
This Treap implementation is based on the one @tantaman wrote for [Materialite](https://github.com/vlcn-io/materialite/tree/main).

I stripped it of things we do not need and rewrote the iterator functions since they could be simplified by using generators and `yield*`.


aboodman:
Whoa, this seems like a huge change. Do we have any perf tests, even rough
ones to have a sense of the impact? I think the b tree is central to perf.

Further we’re going to soon need forking of the memory sources. Will this
support it?

a (phone)

tantaman:
There is a persistent version that will fork which we used in zero’s in ivm
1.

tantaman:
The perf diff between persistent treap from ivm1 and btree wasn't that huge and it was on my list to switch back. We should measure for ivm 2 though.

arv:
Yup. Perf tests should be able to tell us if it is worth 10kb of code...

Even if we end up keeping a BTree we should be able to strip down "btree" to a bare minimum. It has a lot of things we do not need.

aboodman:
I'm very sus because the treap –– IIUC is basically a binary tree. So it
should destroy the locality that makes b-trees work so well:

https://www.cs.cornell.edu/courses/cs312/2004sp/lectures/lec24.html#:~:text=B%2Dtrees%20are%20a%20way,more%20crucial%20than%20with%20memory
.

... but the proof is in the pudding, so let's just test it to be sure.

arv:
I tested it and iteration is 15x slower.

I've been mulling on this overnight. I feel like we could maybe store the binary tree in an array but it does add an amortized grow/shrink step. However, iteration should be even better than with a btree because now we are only iterating an array, with holes but we can keep the ratio at 2.

I'm still having a hard time letting this go so I might experiment a bit more.

The other option is to strip things out of btree. I'm pretty certain I can get that from 15k to 5k in an hour or so.

arv:
I'm going to upload the bench.

arv:
```
 ✓ src/ordered-set.bench.ts (8) 5485ms
   ✓ iteration (2) 1792ms
     name        hz      min      max     mean      p75      p99     p995     p999     rme  samples
   · btree   482.36   1.9000   2.5000   2.0731   2.1000   2.3000   2.3000   2.5000  ±0.59%      242   fastest
   · treap  14.5624  66.4000  71.8000  68.6700  69.7000  71.8000  71.8000  71.8000  ±1.84%       10
   ✓ insertion (2) 1210ms
     name         hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · btree  1,693.66  0.3000  3.1000  0.5904  0.7000  1.0000  1.1000  3.1000  ±1.93%      847   fastest
   · treap    816.77  0.7000  3.3000  1.2243  1.3000  2.3000  2.5000  3.3000  ±2.44%      411
   ✓ get (2) 1211ms
     name          hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · btree  11,658.00  0.0000  0.4000  0.0858  0.1000  0.2000  0.2000  0.3000  ±1.78%     5829
   · treap  29,424.00  0.0000  0.3000  0.0340  0.1000  0.1000  0.2000  0.2000  ±2.33%    14712   fastest
   ✓ delete (2) 1270ms
     name           hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · btree  144,015.20  0.0000  0.2000  0.0069  0.0000  0.1000  0.1000  0.1000  ±2.69%    72022
   · treap  240,000.00  0.0000  0.3000  0.0042  0.0000  0.1000  0.1000  0.1000  ±2.74%   120000   fastest

 BENCH  Summary

  btree - src/ordered-set.bench.ts > iteration
    33.12x faster than treap

  btree - src/ordered-set.bench.ts > insertion
    2.07x faster than treap

  treap - src/ordered-set.bench.ts > get
    2.52x faster than btree

  treap - src/ordered-set.bench.ts > delete
    1.67x faster than btree
```

tantaman:
Which version of the treap? From materialite or latest that was checked into mono? I don't recall iteration being more than a 2-3x difference.

arv:
> Which version of the treap? From materialite or latest that was checked into mono? I don't recall iteration being more than a 2-3x difference.

This was the treap from your repo

arv:
Superseded by https://github.com/rocicorp/mono/pull/3544

PR #3531, opened 2025-01-14
fix(zero): Change export name for change-source

arv:
```js
import {Change} from '@rocicorp/zero/change-protocol/v0';
```


arv:
I've verified that this works.

@darkgnotic If you are still having issues let me know and we can look at this in "person".

darkgnotic:
> I've verified that this works.
> 
> @darkgnotic If you are still having issues let me know and we can look at this in "person".

This works for me too. Thank you @arv!

PR #3533, opened 2025-01-14
chore(zql): add noop output to ivm operators

arv:
Instead of having to keep checking if output is set we initialize it to a noop Output object.

### DIFF
--- a/packages/zql/src/ivm/operator.ts
+++ b/packages/zql/src/ivm/operator.ts
@@ -74,6 +74,16 @@ export interface Output {
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/zql/src/ivm/operator.ts
```diff
@@ -73,6 +73,16 @@ export interface Output {
   push(change: Change): void;
 }
 
+/**
+ * An implementation of Output that throws if pushed to. It is used as the
+ * initial value for for an operator's output before it is set.
+ */
+export const throwOutput: Output = {
```
aboodman, line 80: cute

PR #3534, opened 2025-01-14
repro(zql): not exists bug

tantaman:
This reproduces a user report with queries that use `NOT EXISTS`.

Thread: https://discord.com/channels/830183651022471199/1326515508534579240/1326515834763612241
Summary of what they saw --

1. Given the following query:

```ts
const query = newQuery(queryDelegate, schema.tables.TYL_trackable)
   .where(({not, exists}) =>
     not(exists('trackableGroup', q => q.where('group', '=', 'archived'))),
   )
   .related('trackableGroup');
```

2. A single `trackable` row `{id: '001', name: 'trackable 1'}`.
3. 0 trackableGroup rows.

After adding a `trackableGroup` row with `group` set to `archived`, the `trackable` row returned
by the query is removed. As expected.

When deleting the `trackableGroup` row, the `trackable` row is not returned by the query. This is unexpected. The row should be returned since it is no longer archived.

This repro shows that `zql` (not-exists.test.ts) does the correct thing but the `pipeline-driver` (pipeline-driver.repro.test.ts) is sending the wrong number of `add` and `remove` changes.

After the `trackableGroup` row is deleted from the server, the client is still left with 1 `trackableGroup`row on their device.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/pipeline-driver.repro.test.ts
+++ b/packages/zero-cache/src/services/view-syncer/pipeline-driver.repro.test.ts
@@ -1,0 +1,273 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import {afterEach, beforeEach, expect, test} from 'vitest';
+import {createSilentLogContext} from '../../../../shared/src/logging-test-utils.js';
+import {DbFile} from '../../test/lite.js';
+import {Database} from '../../../../zqlite/src/db.js';
+import {PipelineDriver} from './pipeline-driver.js';
+import {CREATE_STORAGE_TABLE, DatabaseStorage} from './database-storage.js';
+import {Snapshotter} from './snapshotter.js';
+import {initReplicationState} from '../replicator/schema/replication-state.js';
+import {initChangeLog} from '../replicator/schema/change-log.js';
+import {
+  fakeReplicator,
+  ReplicationMessages,
+  type FakeReplicator,
+} from '../replicator/test-utils.js';
+import type {LogContext} from '@rocicorp/logger';
+import type {AST} from '../../../../zero-protocol/src/ast.js';
+
+let dbFile: DbFile;
+let db: Database;
+let lc: LogContext;
+let pipelines: PipelineDriver;
+let replicator: FakeReplicator;
+
+const TRACKABLE_AST: AST = {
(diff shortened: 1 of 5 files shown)


grgbkr:
@tantaman  Wow.  Great job reproducing! 

Any theories on where the bug is?

tantaman:
> Any theories on where the bug is?

Not yet.

A potentially related issue has to do with trying to get the client to keep rows so it can compute `exists` and `not exists` locally.

This is broken for `NOT EXISTS` when a row is added.

`SELECT * FROM issue WHERE NOT EXISTS comment`

A comment is added to issue 1, becoming the first comment for that issue. Issue 1 is then removed from the query results due to `NOT EXISTS`. Since issue 1 is a parent of the comment, the comment row will be removed too since the pipeline driver will remove all rows in relationships hanging off of the parent.

tantaman:
On second reading, the output from `pipelineDriver` looks ok with respect to the original problem.

tantaman:
> A potentially related issue has to do with trying to get the client to keep rows so it can compute exists and not exists locally.

This does present a problem although maybe not the exact problem the user described. If we are not sending down the rows to compute `NOT EXISTS` then the client will compute incorrect query results.

tantaman:
We'll be disabling `not exists` on the client for the time being.

Some ideas for fixes:
1. Custom queries where the user specifies what to sync in their server side query via `related` and we do not ever sync `exists` clauses.
2. Creating a synthetic query where the CVR is the source and the `whereNotExists` clause is appended.

Example of (1):

```ts
// client query:
z.issue.whereExists('assignee')

// server query
z.issue.whereExists('assignee').related('assignee', q => q.limit(2));
// ^- use related to sync what rows you'd like to use for the client to satisfy exists clauses

// todo: work out how this looks for `not exists`
```

Example of (2):

`not exists` means we need to fetch a row for everything that does not match. We can do this by looking at the data on device and getting related rows to define the "negative space" of not exists.

If we have `issue.whereNotExists('assignee')` then we need to fetch an `assignee` row for every `issue` row we have on device. This way the client can figure out which issues on device actually do not have an assignee.

To compute this for every row on device using the CVR synthetic source:
```ts
cvr.issue.related('assignee')
```

TODO: how does this work with many nested `notExists`?

PR #3540, opened 2025-01-15
!feat(zero-cache): multi-node dispatch for multi-tenant zero-cache

darkgnotic:
⚠️   **BREAKING CHANGE**

* The tenant `id` fields in `--tenants-json` must now be unique and url-friendly (alphanumeric, hyphens, underscores)

### Feature

Adds support for the multi-node configuration to multi-tenant `zero-cache`

### Topology

* Review: For client sync connections, multi-tenant dispatch consults the `host` and `path` fields in the `--tenants-json` config to determine which tenant to route an incoming sync connection to.
* 🆕 The tenant `id` is then internally used to distinguish / dispatch change subscriptions made from the `view-syncer` to the `replication-manager`
* Consequently, tenant `id`s are required to be unique and url safe.

> Note: The `host` and `path` fields, on the other hand, are optional, changeable, and much less suitable for internal dispatch, which must be stable in order to avoid downtime or tenant-to-tenant mismatches.

`view-syncer`:

Screenshot 2025-01-15 at 09 19 18

`replication-manager`:

Screenshot 2025-01-15 at 09 20 14

### Implementation

* The `view-syncer` appends the tenant `id` to the path of `ZERO_CHANGE_STREAMER_URI` values.
* The `replication-manager`:
  * Adjusts the behavior of its TenantDispatcher to use the tenant `id` (only) as the path distinguisher
  * Adds websocket handoff logic to the ChangeStreamerHttpService to accept not only direct connections to its listening port, but sockets handed off from the `main` process to the `dispatch` process to the `change-streamer` process. (This is analogous to how dispatch is done to `syncer` processes in the `view-syncer`)


arv (commented):
LGTM

PR #3541, opened 2025-01-15
feat(zero-cache): define a "status" message for the change-protocol

darkgnotic:
Adds a `['status', ..., {watermark: string}]` message to the `change-protocol` that encapsulates the communication of log positions between a ChangeSource and the `zero-cache`.

Move the ack logic previously confined to the Postgres implementation down into the ChangeStreamerService implementation so that it is applicable to all ChangeSource implementations.

@cbnsndwch


arv (commented):
LGTM

PR #3542, opened 2025-01-16
chore(zero-cache): set windowsHide for litestream subprocesses

darkgnotic:
This doesn't fix the popups for all of the other Workers (which are created via `fork()`), but we might as well do this for `litestream`, which is invoked via `spawn()`.

https://discord.com/channels/830183651022471199/1288232858795769917/1329188972743622676


tantaman:
lgtm

PR #3547, opened 2025-01-16
feat(zero-cache): allow TableSource to work on tables without a `PRIMARY KEY`

darkgnotic:
Remove the requirement of an explicit `PRIMARY KEY` from TableSource.

Instead, rely on the table's `UNIQUE` indexes.

* TableSource is still configured with a "PrimaryKey" for the purposes of uniquely identifying rows. However, this key can be any uniquely indexed set of columns, and need not be an explicit `PRIMARY KEY` (which, in SQLite, mean the same thing, as the [rowid](https://www.sqlite.org/withoutrowid.html) is the only real "primary key").
* Similarly, the `Ordering` check is relaxed to allow an ordering that includes the columns of any unique index, and not necessarily those of the primary key.

https://bugs.rocicorp.dev/issue/3186
https://bugs.rocicorp.dev/issue/3198

### DIFF
--- a/packages/zqlite/src/table-source.ts
+++ b/packages/zqlite/src/table-source.ts
@@ -1,6 +1,7 @@
 import type {SQLQuery} from '@databases/sql';
 import {assert, unreachable} from '../../shared/src/asserts.js';
 import {must} from '../../shared/src/must.js';
+import {difference} from '../../shared/src/set-utils.js';
 import type {Writable} from '../../shared/src/writable.js';
 import type {
   Condition,
@@ -14,7 +15,6 @@ import type {
   SchemaValue,
   ValueType,
 } from '../../zero-schema/src/table-schema.js';
-import {assertOrderingIncludesPK} from '../../zql/src/builder/builder.js';
 import {
   createPredicate,
   transformFilters,
@@ -90,6 +90,8 @@ export class TableSource implements Source {
   readonly #connections: Connection[] = [];
   readonly #table: string;
   readonly #columns: Record<string, SchemaValue>;
+  // Maps sorted columns JSON string (e.g. '["a","b"]) to Set of columns.
+  readonly #uniqueIndexes: Map<string, Set<string>>;
   readonly #primaryKey: PrimaryKey;
   readonly #clientGroupID: string;
   #stmts: Statements;
@@ -105,8 +107,14 @@ export class TableSource implements Source {
     this.#clientGroupID = clientGroupID;
     this.#table = tableName;
     this.#columns = columns;
+    this.#uniqueIndexes = getUniqueIndexes(db, tableName);
     this.#primaryKey = primaryKey;
     this.#stmts = this.#getStatementsFor(db);
+
+    assert(
+      this.#uniqueIndexes.has(JSON.stringify([...primaryKey].sort())),
+      `primary key ${primaryKey} does not have a UNIQUE index`,
+    );
   }
 
   get table() {
@@ -126,7 +134,6 @@ export class TableSource implements Source {
     if (cached) {
       return cached;
     }
-    assertPrimaryKeyMatch(db, this.#table, this.#primaryKey);
 
     const stmts = {
       cache: new StatementCache(db),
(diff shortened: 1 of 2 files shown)


packages/zqlite/src/table-source.ts
```diff
@@ -105,8 +107,14 @@ export class TableSource implements Source {
     this.#clientGroupID = clientGroupID;
     this.#table = tableName;
     this.#columns = columns;
+    this.#uniqueIndexes = getUniqueIndexes(db, tableName);
     this.#primaryKey = primaryKey;
     this.#stmts = this.#getStatementsFor(db);
+
+    assert(
+      this.#uniqueIndexes.has(JSON.stringify([...primaryKey].sort())),
```
tantaman, line 115: nit: why do you need to spread the primary key here?

tantaman: We'll need to document for the user that the `primaryKey` field in the schema is either:

1. the actual primary key
2. the columns of a unique index

Maybe even go so far as to change the `primaryKey` field of the schema builder to be called `uniqueKey` instead of `primaryKey`

cc @aboodman

tantaman: I'd wait till https://github.com/rocicorp/mono/pull/3545 lands before changing the `primaryKey` field name in TableSchema though. #3545 will let us change the user facing name without having to change anything internally.

packages/zqlite/src/table-source.ts
```diff
@@ -685,24 +706,31 @@ function gatherStartConstraints(
...
+    sql`
+    SELECT idx.name, json_group_array(col.name) as columnsJSON
+      FROM sqlite_master as idx
+      JOIN pragma_index_list(idx.tbl_name) AS info ON info.name = idx.name
+      JOIN pragma_index_info(idx.name) as col
+      WHERE idx.tbl_name = ${tableName} AND
+            idx.type = 'index' AND 
+            info."unique" != 0
+      GROUP BY idx.name
+      ORDER BY idx.name`,
```
tantaman, line 723: is this also supposed to return the primary key column?

It doesn't seem to:

```
sqlite> create table foo (a integer primary key, b);
sqlite> SELECT idx.name, json_group_array(col.name) as columnsJSON
   ...>       FROM sqlite_master as idx
   ...>       JOIN pragma_index_list(idx.tbl_name) AS info ON info.name = idx.name
   ...>       JOIN pragma_index_info(idx.name) as col
   ...>       WHERE idx.tbl_name = "foo" AND
   ...>             idx.type = 'index' AND 
   ...>             info."unique" != 0
   ...>       GROUP BY idx.name
   ...>       ORDER BY idx.name;
sqlite>
```

packages/zqlite/src/table-source.ts
darkgnotic: Agree on adjusting the name to reflect the wider semantics.

> nit: why do you need to spread the primary key here?

In order to call `Array.sort()`. I guess tuples don't get the array goodies.

packages/zqlite/src/table-source.ts
darkgnotic: It does. Just watch out for the tricky `INTEGER PRIMARY KEY` special case, which doesn't result in creating an index in SQLite:

https://www.sqlite.org/lang_createtable.html#rowid

(I only know this because I had to figure out why some unit tests were failing)

darkgnotic (commented):
Thanks @tantaman !

packages/zqlite/src/table-source.ts
tantaman: Right, so don't you need to handle the `integer primary key` case with this query?

tantaman:
> In order to call Array.sort(). I guess tuples don't get the array goodies

right, it is because sort sorts in place in JavaScript even though it also returns a result. Calling `sort` on a tuple would destroy its identity.

packages/zqlite/src/table-source.ts
darkgnotic: Ah, I understand the question now. Starting from https://github.com/rocicorp/mono/pull/3555, we don't need to handle `primary key` columns declared in the replica; we instead depend on the unique indexes replicated from upstream (which includes the primary keys declared there).

PR #3552, opened 2025-01-17
fix(zql): whereExists before where

grgbkr:


packages/zqlite/src/table-source.ts
```diff
@@ -512,26 +519,30 @@ export class TableSource implements Source {
...
+export function filtersToSQL(filters: NoSubqueryCondition): SQLQuery {
   switch (filters.type) {
     case 'simple':
       return simpleConditionToSQL(filters);
     case 'and':
-      return sql`(${sql.join(
-        filters.conditions.map(condition => optionalFiltersToSQL(condition)),
-        sql` AND `,
-      )})`;
+      return filters.conditions.length > 0
```
grgbkr, line 533: The need to handle the empty cases was revealed by dropping where conditions when they were fully applied by the source.

packages/zql/src/builder/builder.ts
```diff
@@ -200,68 +200,57 @@ function buildPipelineInternal(
...
-    return otherConditions.length > 0
-      ? new Filter(
-          input,
-          appliedFilters ? 'push-only' : 'all',
-          createPredicate({
-            type: 'or',
-            conditions: otherConditions,
-          }),
-        )
-      : new FanIn(new FanOut(input), []);
```
grgbkr, line 253: this was producing a pipeline that was not fully connected.  not sure if this was intentional or not

aboodman:
oooh because the `where` was downstream of `whereExists`. Subtle!

aboodman:
What are the remaining cases where filters are not handled by sources?

grgbkr:
> What are the remaining cases where filters are not handled by sources?

When the condition contains an exist condition.  In this case the sources may partially filter for example if you have

`a = b and exists(c)` the sources will filter by `a = b`, but if you have 
`a = b or exists(c)` the sources cannot do any filtering.

arv (commented):
LGTM

Very nice

packages/zql/src/builder/builder.ts
grgbkr: Got it, the filtering at the source ensures empty or is always empty now.

tantaman:
I think we should still update the builder to move all `where` filters above `exists` filters. This'll help in the `or` case where the source cannot apply filters but there is filtering that we can do specific to each branch.

e.g.,

`(a = b AND exists(c)) OR (d = f AND exists(g))`

PR #3553, opened 2025-01-17
feat(zqlite): select an appriopriate "primary key" for a given sort

darkgnotic:
When connecting an Input, select an appropriate `primaryKey` (amongst the table's unique indexes) to for the specified `sort` Ordering. Take will assert that the Ordering contains all of the columns in the primaryKey.

This will also be tested at a higher level (pipeline-driver) in a subsequent PR.

### DIFF
--- a/packages/zqlite/src/table-source.ts
+++ b/packages/zqlite/src/table-source.ts
@@ -210,7 +210,7 @@ export class TableSource implements Source {
     return {
       tableName: this.#table,
       columns: this.#columns,
-      primaryKey: this.#primaryKey,
+      primaryKey: this.#selectPrimaryKeyFor(connection.sort),
       sort: connection.sort,
       relationships: {},
       isHidden: false,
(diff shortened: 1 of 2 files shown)


packages/zqlite/src/table-source.ts
```diff
@@ -210,7 +210,7 @@ export class TableSource implements Source {
     return {
       tableName: this.#table,
       columns: this.#columns,
-      primaryKey: this.#primaryKey,
+      primaryKey: this.#selectPrimaryKeyFor(connection.sort),
```
tantaman, line 213: I think choosing the primary key from the sort is safe. It gives me some pause since it creates an opportunity for the primary key to diverge. That is, one query gets one primary key and another query gets a different one.


What you have here seems to be ok since:
1. A sort is unique per connection.
2. You're calculating the "primary key" from the sort
3. You're setting it as the `primaryKey` at the source of the connection
4. All downstream operators will use this as their `primaryKey` as they all get their schema from their input(s).

It does seem like a place where we can introduce subtle bugs, however. Say if an operator ever starts getting primary key information from something that is not an upstream input to it.


---

`join` and `exists` refer to the primary key of a node. Join uses it to index its operator storage. `exists` uses it to determine if caching `exists` results would be effective or not.

tantaman: I think I'd be happier of the "primary key" information could be static and not subject to different computations at runtime.

cc @grgbkr

PR #3555, opened 2025-01-18
feat(zero-cache): index and primary key-handling overhaul

darkgnotic:
### Bugs fixed:

* Correctly ignore [indexes with expression or partial indexes](https://bugs.rocicorp.dev/issue/3402)
* [Handle primary key changes](https://bugs.rocicorp.dev/issue/3186)

With setup for adding [support for Prisma join tables](https://bugs.rocicorp.dev/issue/3198) (i.e. not implemented yet but this is a step towards it)

### Features

* Support changes to the PRIMARY KEY
* Partial support for tables without PRIMARY KEYs (i.e. removing a roadblock when setting up a project)

### DX

Initial sync / replication no longer fails for tables without a PRIMARY KEY. Instead, a warning is displayed:

Screenshot 2025-01-17 at 12 39 14

but the data is otherwise replicated to the replica.

However, this data is not synced to the clients, because primary keys are necessary for downstream logic (e.g. IVM, Replicache). 

> Note: From Postgres' perspective, primary keys are also necessary to replicate `UPDATE` and `DELETE` commands, and while the developer is publishing a table with no primary key, Postgres will refuse to execute those commands:

Screenshot 2025-01-17 at 16 27 39

Once the user has added a primary key to the table, the table becomes visible to zql and downstream clients. 

Resetting of the replica is not necessary. It is handled just as other schema changes are handled.

### Implementation

* PRIMARY KEYs are no longer written to the SQLite replica.
* Instead, we now replicate primary key indexes from Postgres (in addition to non-primary key indexes that we already replicate)
* In zqlite, we choose an appropriate unique index for IVM operations according to the sort order requested. This was added in #3547 and #3553

### TODO

Some of the code in this PR is setup for upcoming features:

* The `unionKey` is the union of all uniquely indexed columns. These will be used as CVR row keys in order to [facilitate clients migrating between primary keys on client-side schemas](https://rocicorp.slack.com/archives/C013XFG80JC/p1736376499195429?thread_ts=1736355005.417389&cid=C013XFG80JC)
* The `isReplicaIdentity` field looked up for Postgres indexes will be used to support tables with unique indexes but no primary key, such as prisma join tables

### DIFF
--- a/packages/zero-cache/bench/benchmark.ts
+++ b/packages/zero-cache/bench/benchmark.ts
@@ -7,7 +7,7 @@ import type {Source} from '../../zql/src/ivm/source.js';
 import {newQuery, type QueryDelegate} from '../../zql/src/query/query-impl.js';
 import {Database} from '../../zqlite/src/db.js';
 import {TableSource} from '../../zqlite/src/table-source.js';
-import {listTables} from '../src/db/lite-tables.js';
+import {computeZqlSpecs} from '../src/db/lite-tables.js';
 import {mapLiteDataTypeToZqlSchemaValue} from '../src/types/lite.js';
 import {schema} from './schema.js';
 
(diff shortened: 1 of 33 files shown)


darkgnotic:
Sorry this PR is so yuge. I intended to compose it in smaller pieces, but everything is kind of tied together.

I _was_ able to delay some pieces to later PRs though (as explained in the TODO section) .  😅

PR #3561, opened 2025-01-18
fix(zero-cache): fix client catchup

darkgnotic:
For client catchup. The CVR rows keys may differ from the TableSource row keys, so the latter must be able to look up rows by any row key.


tantaman:
lgtm

PR #3563, opened 2025-01-19
chore(zqlite): remove debug console logging

darkgnotic:
From #3552, presumably put in just for debugging.


grgbkr:
Thanks.  Sorry about that.  Going to see if there is something i can add to our lint config.

PR #3565, opened 2025-01-20
fix(zero-cache): do not consider nullable columns in keys

darkgnotic:
When choosing a row key from UNIQUE indexes, excluded indexes with nullable columns.

SQLite considers `NULL` values distinct, making nullable columns unsuitable to be part of a row key. Postgres, for example, requires that indexes used for `REPLICA IDENTITY USING INDEX` [contain only non null columns](https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-REPLICA-IDENTITY).

Thus, indexes with nullable (upstream) columns must be excluded when determining the row key.

Because the `NOT NULL` constraint on upstream columns is not actually applied to the replicated columns, it is instead encode in the SQLite data type, similar to how the `TEXT_ENUM` attribute of enum columns is encoded.


packages/zero-cache/src/db/lite-tables.test.ts
```diff
@@ -292,7 +295,7 @@ describe('computeZqlSpec', () => {
     expect(
       t(`
     CREATE TABLE nopk(a INT, b INT, c INT, d INT);
-    CREATE TABLE foo(a INT, b INT, c INT, d INT);
+    CREATE TABLE foo(a INT, b "INT|NOT_NULL", c INT, d INT);
```
tantaman, line 298: why encode `not null` into the data type rather than setting the column to be `not null`?

I think? the reason is write auth which creates an optimistic copy of the write before replication which may be missing values. Do you agree that we should get rid of this weird encoding here once we move everything to custom mutators?

darkgnotic: Yes, that was my original thought, to wait for custom mutators so that we can start replicating NOT NULL again and use that instead.

But, I also realized that we have a precedence for encoding upstream information in a way that does not affect SQLite behavior, namely the TEXT_ENUM property. So it seems reasonable to do the same for NOT NULL-ness, where we care about properties on upstream but don't necessarily need / want to replicate the behavior on the replica.

I agree it looks kind of weird, but I think it's useful to have a way to encode upstream information into the sqlite schema without having to maintain a separate table.

PR #3567, opened 2025-01-20
chore(zql): Smaller syntax

arv:
Just testing the CI


arv:
Hah. The non br version is smaller but the br version is larger.

PR #3568, opened 2025-01-20
chore(zero-client): Run performance benchmark

arv:


arv:
This needs some more work but I need something running on main at this point so merging.

PR #3571, opened 2025-01-20
chore: update @rocicorp/eslint-config to 0.7.0

grgbkr:
In particular this pull in "no-console".

### DIFF
--- a/packages/zero-cache/src/server/life-cycle.ts
+++ b/packages/zero-cache/src/server/life-cycle.ts
@@ -244,9 +244,11 @@ export async function runUntilKilled(
 export async function exitAfter(run: () => Promise<void>) {
   try {
     await run();
+    // eslint-disable-next-line no-console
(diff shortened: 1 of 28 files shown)


packages/zero-cache/src/server/life-cycle.ts
```diff
@@ -244,9 +244,11 @@ export async function runUntilKilled(
 export async function exitAfter(run: () => Promise<void>) {
   try {
     await run();
+    // eslint-disable-next-line no-console
```
grgbkr, line 247: Should these be using the lc: LogContext rather than console?

darkgnotic: I couldn't immediately think of an elegant way to do that. These are top-level wrappers that call the methods that create the LogContext ...

darkgnotic (commented):
Thanks @grgbkr !

PR #3572, opened 2025-01-21
fix(zero-cache): stop piping stdio between parent and child

darkgnotic:
Remove the stdio piping between parent/child processes, as it causes event listener warnings. This was a workaround for Windows that should not longer be necessary now that the child processes are no longer `detached`.


tantaman:
works for me on windows 10.

PR #3573, opened 2025-01-21
feat(zql)!: `tables` and `relationships` are now arrays not objects

tantaman:
Uses:

```ts
{
  [K in TTables[number]['schema']['name']]: ...
}
```

to pull unique names from array members to use as object keys.

And then

```ts
Extract<
      TTables[number]['schema'],
      {name: K}
    >
```

to pull the corresponding value type with the given name from the array.

### DIFF
--- a/apps/zbugs/schema.ts
+++ b/apps/zbugs/schema.ts
@@ -189,26 +189,16 @@ type AuthData = {
   role: 'crew' | 'user';
 };
 
-export const schema = createSchema(
-  5,
-  {
-    user,
-    issue,
-    comment,
-    label,
-    issueLabel,
-    viewState,
-    emoji,
-    userPref,
-  },
-  {
+export const schema = createSchema(5, {
+  tables: [user, issue, comment, label, issueLabel, viewState, emoji, userPref],
+  relationships: [
     userRelationships,
     issueRelationships,
     commentRelationships,
     issueLabelRelationships,
     emojiRelationships,
-  },
-);
+  ],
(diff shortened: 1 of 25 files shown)


apps/zbugs/schema.ts
```diff
@@ -189,26 +189,16 @@ type AuthData = {
...
-    comment,
-    label,
-    issueLabel,
-    viewState,
-    emoji,
-    userPref,
-  },
-  {
+export const schema = createSchema(5, {
+  tables: [user, issue, comment, label, issueLabel, viewState, emoji, userPref],
```
aboodman, line 193: What happens if there are duplicate table names in here? Is there a runtime error?

apps/zbugs/schema.ts
```diff
@@ -189,26 +189,16 @@ type AuthData = {
...
+  relationships: [
```
aboodman, line 194: Same question here. what happens if two of these relationships objects have the same source? What if the same relationship name is declared twice?

aboodman (approved):
API changes approved ✅. Some minor questions about validation.

apps/zbugs/schema.ts
tantaman: There's validation but it got added to the second commit rather than the first.  I miss mercurial's absorb command that automatically amends working changes to the correct commits :(

tantaman:
My only worry with this change is how `Extract` will perform in very large codebases.

PR #3576, opened 2025-01-21
fix(zql): `null` should not be allowed in `where` except with `IS`.

aboodman:
This originally worked but got broken with some restructuring in query.ts. Adds some tests to make sure it doesn't break again.

Fixes: https://bugs.rocicorp.dev/issue/3426i

BREAKING CHANGE: This narrows the allowed types for where both in all cases. For IS/IS NOT, null is only allowed if the column type is optional. For other operators, null is now never allowed.


aboodman:
@tantaman can you please think a bit about the api change and make sure this narrowing is correct.

arv:
> @arv is there a way to test the typescript errors more specifically? I don't like these generic ts-expect-error style tests because any ts error will make the test "pass", potentially hiding real problems.

As far as I know there is no way to make sure that you got the expected error.

packages/zql/src/query/query.ts
```diff
@@ -32,18 +32,20 @@ export type Operator =
...
-  ? Exclude<
-      SchemaValueToTSType<TSchema['columns'][TColumn]>,
-      null | undefined
-    >[]
-  : TOperator extends 'IS' | 'IS NOT'
-  ? Exclude<SchemaValueToTSType<TSchema['columns'][TColumn]>, undefined> | null
-  : Exclude<SchemaValueToTSType<TSchema['columns'][TColumn]>, undefined>;
+> = TOperator extends 'IS' | 'IS NOT'
+  ? // SchemaValueToTSType adds null if the type is optional
+    SchemaValueToTSType<TSchema['columns'][TColumn]>
```
tantaman, line 42: I'm having trouble deciding if required columns should be allowed to be compared to null.

My immediate reaction was no and that it is nonsensical to compare a required column to null. On the other hand, it does make some things easier on the user.

Maybe `IssueDetail` loads in `parentIssue` information lazily rather than preloading.

```ts
function IssueDetail({issue}: {issue: Issue) {
  const parentInfo = useQuery(z.query.issue.where('id', 'IS', issue.parentId);
  return <>...</>
}
```

If `IS` cannot compare to null for required fields that gets a bit more awkward and error prone.

```ts
function IssueDetail({issue}: {issue: Issue) {
  const parentInfo = useQuery(z.query.issue.where('id', 'IS', issue.parentId ?? '', issue.parentId !== null);
  return <>...</>
}
```

1. They have to use this weird `empty string` sentinel
2. They have to remember to disable the query

tantaman: tldr: I think `IS` should be able to take `null` even for required columns.

aboodman: will make this change.

KaiSpencer:
> > @arv is there a way to test the typescript errors more specifically? I don't like these generic ts-expect-error style tests because any ts error will make the test "pass", potentially hiding real problems.
> 
> As far as I know there is no way to make sure that you got the expected error.

Just jumping in here @arv @aboodman. Since you use Vitest for running your tests you could use the [type testing feature](https://vitest.dev/guide/testing-types)

This allows you to make specific assertions on the shape of types. Would be happy to make a more general PR bringing this into the codebase if it was something you guys wanted implementing.

aboodman:
On Wed, Jan 22, 2025 at 11:49 PM Kai Spencer ***@***.***>
wrote:

> @arv <https://github.com/arv> is there a way to test the typescript
> errors more specifically? I don't like these generic ts-expect-error style
> tests because any ts error will make the test "pass", potentially hiding
> real problems.
>
> As far as I know there is no way to make sure that you got the expected
> error.
>
> Just jumping in here @arv <https://github.com/arv> @aboodman
> <https://github.com/aboodman>. Since you use Vitest for running your
> tests you could use the type testing feature
> <https://vitest.dev/guide/testing-types>
>
> This allows you to make specific assertions on the shape of types. Would
> be happy to make a more general PR bringing this into the codebase if it
> was something you guys wanted implementing.
>

Hi Kai, thanks for the note. We actually already use this Vitest feature,
but it didn't occur to me I could use it to test error messages. But maybe
I can using compile-time string matching.

Will try. thanks for the idea

PR #3578, opened 2025-01-21
fix(zero): Add missing type exports

arv:
https://bugs.rocicorp.dev/issue/3459

When a library imports `@rocicorp/zero` and uses `definePermission`, and the library has a `tsconfig.json` that includes `declaration: true`, then tsc complains that it cannot infer the type:

```
The inferred type of 'permissions' cannot be named without a reference
to '../node_modules/@rocicorp/zero/out/zero-protocol/src/ast'. This is
likely not portable. A type annotation is necessary.
```

The reason is that `definePermission` has a dependency on the `Condition` type from `ast.ts` from `zero-protocol`, but that file is never loaded by tsc so it does not know about this type.

The solution is to export a type dependency from `ast.ts`. That way `tsc` is happy.

However, to be less magic we now export all the **transitive** dependencies of `definePermission`. This leads to us eventually exporting `AST` etc from `ast.ts`. This is not ideal. I didn't want us to have to commit to the shape of the AST but we are already making it accessible to the outside world via `definePermission` (even though it was not very straight forward to get to it before).

### DIFF
--- a/packages/zero-client/src/mod.ts
+++ b/packages/zero-client/src/mod.ts
@@ -57,25 +57,60 @@ export type {
   VersionNotSupportedResponse,
   WriteTransaction,
 } from '../../replicache/src/mod.js';
+export type {
+  AST,
+  Bound,
+  ColumnReference,
+  CompoundKey,
+  Condition,
+  Conjunction,
+  CorrelatedSubquery,
+  CorrelatedSubqueryCondition,
+  CorrelatedSubqueryConditionOperator,
+  Disjunction,
+  EqualityOps,
+  InOps,
+  LikeOps,
+  LiteralReference,
+  LiteralValue,
+  Ordering,
+  OrderOps,
+  OrderPart,
+  Parameter,
+  SimpleCondition,
+  SimpleOperator,
+  ValuePosition,
+} from '../../zero-protocol/src/ast.js';
+export {relationships} from '../../zero-schema/src/builder/relationship-builder.js';
+export {
+  createSchema,
+  type Schema,
+} from '../../zero-schema/src/builder/schema-builder.js';
+export {
+  boolean,
+  enumeration,
+  json,
+  number,
+  string,
+  table,
+} from '../../zero-schema/src/builder/table-builder.js';
+export type {
+  AssetPermissions as CompiledAssetPermissions,
+  PermissionsConfig as CompiledPermissionsConfig,
+  Policy as CompiledPermissionsPolicy,
+  Rule as CompiledPermissionsRule,
+} from '../../zero-schema/src/compiled-permissions.js';
 export {
   ANYONE_CAN,
   definePermissions,
   NOBODY_CAN,
 } from '../../zero-schema/src/permissions.js';
-export {
-  createSchema,
-  type Schema,
-} from '../../zero-schema/src/builder/schema-builder.js';
+export type {
+  AssetPermissions,
+  PermissionRule,
+  PermissionsConfig,
+} from '../../zero-schema/src/permissions.js';
 export {type TableSchema} from '../../zero-schema/src/table-schema.js';
-export {
-  table,
-  string,
-  number,
-  boolean,
-  enumeration,
-  json,
-} from '../../zero-schema/src/builder/table-builder.js';
-export {relationships} from '../../zero-schema/src/builder/relationship-builder.js';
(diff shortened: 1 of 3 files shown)


aboodman:
A few things:

- Is it reasonable that `definePermission` has a dependency on `Condition`? Should we fix that instead?
- Can you add `declaration: true` to zbugs so that we catch regressions in the future?

cbnsndwch:
Follow-up comments:

1. there are a few other types that cause this for me:

```typescript
import type { Condition } from '@rocicorp/zero-protocol/src/ast.js';
import type { PermissionRule, SchemaValue } from '@rocicorp/zero/schema';
```

2. In apps with mixed ESM/CommonJS sources it is not possible to directly reference `@rocicorp/zero` because it doesn't provide CJS exports (e.g.: any app that creates/loads zero schemas in a NestJS server).

I resolved this in https://github.com/cbnsndwch/zero-sources by adding a library that wraps zero in a `tsup` project with CJS output and re-exports the exports I needed, but it's an ugly and brittle solution that I'm using as a temporary workaround until we could have a conversation about npm packages and exports.

arv:
> Is it reasonable that definePermission has a dependency on Condition? Should we fix that instead?

An alternative is to have definePermission return a opaque type. I was considering this but it felt too much work to hide the AST types.

arv:
@cbnsndwch I don't think we want to support cjs at this point. The world is moving away from it so it is generally better to go the other direction... fix the cjs dependencies for something more modern.

tantaman:
> However, to be less magic we now export all the transitive dependencies of definePermission. This leads to us eventually exporting AST etc from ast.ts. This is not ideal. I didn't want us to have to commit to the shape of the AST but we are already making it accessible to the outside world via definePermission (even though it was not very straight forward to get to it before).

An unfortunate situation but I'm ok with treating these as an internal interface (even though they're exported) and breaking people who decide to depend on these types.

If we add a new type that `Condition` depends on, how will we know that we need to export it?

wdyt of exporting these with a prefix of `internal_`?

cbnsndwch:
> @cbnsndwch I don't think we want to support cjs at this point. The world is moving away from it so it is generally better to go the other direction... fix the cjs dependencies for something more modern.

that's totally reasonable. And for the record, I agree with you (see my ongoing work in https://github.com/mjackson/remix-the-web/issues/18 to add Web APIs compatibility to `node:http`). It's really unfortunate that the maintainers of Nest don't seem to get the memo, and the framework is just too good to move to something else at this point. 

I'm of course not expecting or asking you to directly support CJS, i can take care of bridging the gap. The blocker for me is that `@rocicorp/zero` is just one npm package is very selective with what it exports, hence the workaround. 

This message from @arv seems like it's meant to address that? https://discord.com/channels/830183651022471199/1331312171212210207/1331917456515666012

> I find these errors very confusing. I think, what we need to do is to ensure that we export all the types that an exported symbol references.

cbnsndwch:
> @cbnsndwch I'm not sure what your doing here?
> 
> ```ts
> import type { Condition } from '@rocicorp/zero-protocol/src/ast.js';
> ```

I needed access to the types in order to generate table schemas from entity classes and their decorators. See <https://github.com/cbnsndwch/zero-sources/blob/main/libs/zero-nest-mongoose/src/tables.mts>

cbnsndwch:
> > However, to be less magic we now export all the transitive dependencies of definePermission. This leads to us eventually exporting AST etc from ast.ts. This is not ideal. I didn't want us to have to commit to the shape of the AST but we are already making it accessible to the outside world via definePermission (even though it was not very straight forward to get to it before).
> 
> An unfortunate situation but I'm ok with treating these as an internal interface (even though they're exported) and breaking people who decide to depend on these types.
> 
> If we add a new type that `Condition` depends on, how will we know that we need to export it?
> 
> wdyt of exporting these with a prefix of `internal_`?

this may be a good approach in some cases, though one of the goals I'm after is automatically generating schemas from existing in-app data model. Those generated schema would eventually need to be consumed by a front-end app so the different export names may get tricky. Worth exploring though

cbnsndwch:
@arv @tantaman I figured out how to make Nest consume ESM directly 

in Node 22.12+ synchronous `require(ESM)` is no longer behind an experimental flag. I'll chime in on the types I'm seeing are missing in ZBugs, but wanted to close the loop here as well.

PR #3579, opened 2025-01-21
feat(zero-react): delay unsubscribing queries

tantaman:
strict mode will mount, unmount, remount a component causing `changeDesiredQueries` to remove the query then re-add it. This creates flickering for users who do not have a preload or covering query higher up in the component hierarchy.

It also creates unnecessary query churn in dev mode.

https://discord.com/channels/830183651022471199/1322102519987572747

### DIFF
--- a/packages/zero-react/package.json
+++ b/packages/zero-react/package.json
@@ -6,7 +6,8 @@
   "module": "true",
   "scripts": {
     "check-types": "tsc",
-    "check-types:watch": "tsc --watch"
+    "check-types:watch": "tsc --watch",
+    "test": "vitest run"
   },
   "peerDependencies": {
     "react": "^18.3.1"
--- a/packages/zero-react/package.json
+++ b/packages/zero-react/package.json
@@ -7,7 +7,8 @@
   "scripts": {
     "check-types": "tsc",
     "check-types:watch": "tsc --watch",
-    "test": "vitest run"
+    "test": "vitest run",
+    "test:watch": "vitest watch"
   },
   "peerDependencies": {
     "react": "^18.3.1"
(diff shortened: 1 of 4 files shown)


packages/zero-react/package.json
```diff
@@ -6,7 +6,8 @@
   "module": "true",
   "scripts": {
     "check-types": "tsc",
-    "check-types:watch": "tsc --watch"
+    "check-types:watch": "tsc --watch",
+    "test": "vitest run"
```
arv, line 10: Also add test:watch for consistency with all our other packages.

packages/zero-react/src/use-query.test.ts
```diff
@@ -0,0 +1,151 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
arv, line 1: Can we not do this? You can just add it on the actual line that you want to allow any.

packages/zero-react/src/use-query.tsx
```diff
@@ -57,6 +57,8 @@ function getDefaultSnapshot<TReturn>(singular: boolean): QueryResult<TReturn> {
   ) as QueryResult<TReturn>;
 }
 
+export const getAllViews = Symbol();
```
arv, line 60: Using a WeakMap pattern with a TESTING guard allows the code to be removed.

```ts
const allViews = new WeakMap();

declare const TESTING: boolean

export class ViewStore {
  constructor() {
    if (TESTING) {
      const view = {};
      allViews.set(this, view);
    }
  }
}

export function getAllViews(store) {
  return allViews.get(store);
}
```

packages/zero-react/src/use-query.tsx
```diff
@@ -236,10 +242,24 @@ class ViewWrapper<
     this.#materializeIfNeeded();
     return () => {
       this.#reactInternals.delete(internals);
+
+      // only schedule a cleanup task if we have no listeners left
       if (this.#reactInternals.size === 0) {
-        this.#view?.destroy();
-        this.#view = undefined;
-        this.#onDematerialized();
+        setTimeout(() => {
```
arv, line 248: Is there no case where we want to cancel the timeout?

arv (approved):
LGTM

Thanks. This was on my todo list for today. Another job well done!

PR #3582, opened 2025-01-22
fix(zero-cache): correctly handle row key changes in CVR updates

darkgnotic:
Make the CVR update logic detect row key changes and recognize when two different keys reference the same row.

Algorithm:
* The `KeyColumns` class determines the (potentially old) key columns referenced in the CVR.
  * It does this purely by examining the current CVR rows.
  * This relies on an invariant that a CVR's row keys are self-consistent, explained in the js doc.
* When processing rows produced by the IVM pipeline, check if the row exists in the CVR by either the new key (from the pipeline) or the old key (used by the CVR). This is possible at this point because the full row is available, and thus the old key columns can be looked up.
  * Edge case: If the old key columns no longer exist in the replica, we throw `ClientNotFound` and drop the CVR. This would only happen if the developer didn't properly update their schema versions; the CVR should have ideally been rejected earlier on.
* In the case of the row being referenced by the old key, the old row is deleted (actually deleted from the DB) because it is replaced by the new row. 
* However, the row contents are not sent in the poke if the row's patch version has not changed, so extraneous `put` and `del` patches are avoided.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/cvr-store.ts
+++ b/packages/zero-cache/src/services/view-syncer/cvr-store.ts
@@ -635,6 +635,28 @@ export class CVRStore {
     this.#pendingRowRecordPuts.set(row.id, row);
   }
 
+  /**
+   * Note: Removing a row from the CVR should be represented by a
+   *       {@link putRowRecord()} with `refCounts: null` in order to properly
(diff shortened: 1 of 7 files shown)


darkgnotic:
I think this doesn't update the row record cache correctly. Followup fix coming soon.

PR #3586, opened 2025-01-22
fix(zero): Add more exports

arv:
https://bugs.rocicorp.dev/issue/3459

And set `declarations: true` for zbugs so we can catch these errors in the future.


arv:
@cbnsndwch FYI

PR #3588, opened 2025-01-22
fix(zero-cache): fix changelog purging

darkgnotic:
The sanity check for purging changelog records is not compatible with the new commit watermark scheme introduced in #3520. In other words, now that all message for a transaction have the same watermark, the sanity check doesn't make sense.

Remove the sanity check and just purge transactions before the transactions in the watermark. We no longer purge all the way up to the final "commit" message, but that's fine.


arv (commented):
LGTM

PR #3593, opened 2025-01-23
chore(zero-cache): improve error message when the `--replica-file` cannot be opened

darkgnotic:
Another #papercut that we see a lot, particularly among Windows users that copy the quickstart config verbatim.

The error message now includes "ZERO_REPLICA_FILE" and the (invalid) path. 

Screenshot 2025-01-22 at 19 04 08


tantaman (commented):
lgtm

PR #3594, opened 2025-01-23
fix(zero-react): Export transitive types

arv:
https://discord.com/channels/830183651022471199/1288232858795769917/1331312171212210207

This is another issue where `declations: true` is causing issues because we are not exporting all the types we are referencing.


tantaman:
Why don't we run into these problems in things like `hello-zero`?

I think this is basically what the user was doing with their `useFolders` hook but inside of `hello-zero`:

CleanShot 2025-01-23 at 09 40 26

arv:
> Why don't we run into these problems in things like `hello-zero`?

We are not triggering that bug yet in hello-zero.

Adding 

```ts
// use-bug.ts
import { useQuery, useZero } from "@rocicorp/zero/react";
import { Schema } from "./schema";

export function useBug() {
  const z = useZero<Schema>();
  const query = z.query.medium;
  return useQuery(query);
}
```

triggers the same issue.

PR #3598, opened 2025-01-23
fix(zql): make valita the source of truth for the `Parameter` type

tantaman:
the valita schema and typescript type for `Parameter` drifted.  There's no need to have two types, just infer the `Parameter` type from the valita schema.

### DIFF
--- a/packages/zero-protocol/src/ast.test.ts
+++ b/packages/zero-protocol/src/ast.test.ts
@@ -243,6 +243,6 @@ test('protocol version', () => {
   // If this test fails because the AST schema has changed such that
   // old code will not understand the new schema, bump the
   // PROTOCOL_VERSION and update the expected values.
-  expect(hash).toEqual('1g60qx4dfwety');
-  expect(PROTOCOL_VERSION).toEqual(3);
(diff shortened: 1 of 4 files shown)


packages/zero-protocol/src/ast.test.ts
```diff
@@ -243,6 +243,6 @@ test('protocol version', () => {
   // If this test fails because the AST schema has changed such that
   // old code will not understand the new schema, bump the
   // PROTOCOL_VERSION and update the expected values.
-  expect(hash).toEqual('1g60qx4dfwety');
+  expect(hash).toEqual('2zzy9s2lcdcms');
```
tantaman, line 246: @darkgnotic - the protocol did change but only by growing wider. `field` is now `string | string[]` instead of `string`.

Does this require a protocol version bump since it will forever be backwards compatible? There will not be a narrowing of the field later.

darkgnotic: Also, if we're storing AST's with this new repeated field (e.g. in the CVR or idb), old code will fail when attempting to deserialize it.

PR #3601, opened 2025-01-23
work

grgbkr:


packages/zql/src/ivm/change.ts
```diff
@@ -28,7 +28,7 @@ export type RemoveChange = {
  */
 export type ChildChange = {
   type: 'child';
-  row: Row;
+  rows: Row[];
```
grgbkr, line 31: eventually this could be a Stream<Row>

PR #3606, opened 2025-01-23
feat(zero-schema): allow remapping column and table names

tantaman:
This is just the API layer. We'll have to update the query-builder to use these names when creating the AST next (https://rocicorp.slack.com/archives/C013XFG80JC/p1737662271150499?thread_ts=1737653133.055509&cid=C013XFG80JC).

API looks like:

```ts
const user = table('user')
  .upstreamName('users')
  .columns({
    id: string().upstreamName('user_id'),
    name: string().upstreamName('user_name'),
    recruiterId: number().upstreamName('user_recruiter_id'),
  })
  .primaryKey('id');
```

### DIFF
--- a/packages/zero-schema/src/builder/schema-builder.test.ts
+++ b/packages/zero-schema/src/builder/schema-builder.test.ts
@@ -2,7 +2,7 @@
  * Relationships as secondary.
  */
 
-import {expectTypeOf, test} from 'vitest';
+import {expect, expectTypeOf, test} from 'vitest';
 import {table, number, string} from './table-builder.js';
 import {relationships} from './relationship-builder.js';
 import type {Query} from '../../../zql/src/query/query.js';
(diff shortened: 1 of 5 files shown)


aboodman:
Where should the remapping actually occur? At SQLite or JS layer? Are there any tradeoffs?

aboodman:
API wise I find `upstreamName` awkward and long if it's going to be in every column.

It's tempting to use `as()` since that word has known meaning from SQL. But that would mean that the identifiers in the columns object would be snake_case or whatever which would make linters unhappy.

So what about `s/upstreamName/from/g` ? or `s/upstreamName/source/g` ?

darkgnotic:
> API wise I find `upstreamName` awkward and long if it's going to be in every column.
> 
> It's tempting to use `as()` since that word has known meaning from SQL. But that would mean that the identifiers in the columns object would be snake_case or whatever which would make linters unhappy.
> 
> So what about `s/upstreamName/from/g` ? or `s/upstreamName/source/g` ?

I think `from(...)` reads quite nicely. Not sure if it provides enough context but I assume that people who need this will read the documentation for the "from()" method and thus understand?

@tantaman FYI I'm take the baton for this PR.

darkgnotic:
> So what about s/upstreamName/from/g ? or s/upstreamName/source/g ?

Forking off into https://github.com/rocicorp/mono/pull/3681

That's what it looks like with `.from()`. 

I'm working on wiring this up but you can see what the API looks like in the meantime.

PR #3608, opened 2025-01-24
fix(zero-solid): Only call setState once per transaction

arv:


grgbkr:
LGTM.  This looks like a great change, and is the right thing to do.

I still dont understand how this just became a problem for Dax, as this aspect of the solid-view has not changed since solid-view was created.

grgbkr (approved):

PR #3612, opened 2025-01-25
fix(zbugs): README link typo

devdumpling:
Fixes md link to the app so it doesn't try to relatively redirect in github.

> Caught this while trying to find the link


arv (approved):
Thanks

PR #3620, opened 2025-01-27
fix(zero-schema): Fix an issue with TS2589

arv:
TS2589: Type instantiation is excessively deep and possibly infinite

The hint to solve this was here:

https://github.com/microsoft/TypeScript/issues/30188#issuecomment-478938437

The solution is to compute the types that cause problems step by step. By using `T extends infer U`, we can force the computation of `T` instead of deferring it. This way the instantiation does not get deeper.

 instantiation is excessively deep and possibly infinite

### DIFF
--- a/packages/zero-schema/src/builder/schema-builder.test.ts
+++ b/packages/zero-schema/src/builder/schema-builder.test.ts
@@ -1,12 +1,8 @@
-/**
- * Relationships as secondary.
- */
-
 import {expectTypeOf, test} from 'vitest';
-import {table, number, string} from './table-builder.js';
-import {relationships} from './relationship-builder.js';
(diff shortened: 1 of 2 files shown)


arv:
https://bugs.rocicorp.dev/issue/3477

tantaman:
🧙‍♂️

Using the [TS example](https://github.com/microsoft/TypeScript/issues/30188#issuecomment-478938437):
`Reverse<Reverse<T1> extends infer R ? Cast<R, any[]> : never, T2>`

So this causes the compiler to figure out the type of `Reverse<T1>` before passing it to `Cast`? Whereas the opposite formulation: `Reverse<Cast<Reverse<T1>, any[]>, T2>` would create a tree of expressions of arbitrary depth to be evaluated?

PR #3627, opened 2025-01-28
chore(zero-cache): move MessageProcessor out of IncrementalSyncer

darkgnotic:
(Pure refactoring)

Moves the `MessageProcessor` (and TransactionProcessor helper) class out of `incremental-sync.ts` and into its own `change-processor.ts` file, renamed appropriately to `ChangeProcessor`.

This will facilitate using the same Change replication logic to perform initial sync for custom change sources. 

@cbnsndwch


packages/zero-cache/src/services/replicator/change-processor.ts
```diff
@@ -0,0 +1,561 @@
...
+export type TransactionMode = 'IMMEDIATE' | 'CONCURRENT';
+
+/**
+ * The ChangeProcessor partitions the stream of messages into transactions
+ * by creating a {@link TransactionProcessor} when a transaction begins, and dispatching
+ * messages to it until the commit is received.
+ *
+ * From https://www.postgresql.org/docs/current/protocol-logical-replication.html#PROTOCOL-LOGICAL-MESSAGES-FLOW :
+ *
+ * "The logical replication protocol sends individual transactions one by one.
```
cbnsndwch, line 61: In Mongo there are not transaction commit markers, only transaction identification.

Changes however only show up in the oplog after they've committed, which is delayed until the issuing application's requested write concern is matched. This allows applications to dynamically control durability on a more granular level than what is common in RDBMS's. E.g.: for a write concern of 'majority' in a three node cluster transactions would only appear in the oplog after two nodes have acknowledged them.

The tradeoff is that changes belonging to different transactions may be interleaved in the change stream.

I've considered two approaches to handle this:

1. Not streaming transactions to zero at all, stream individual changes as standalone
  - Pro: simpler handling, matches eventual consistency semantics from Mongo
  - Con: may lead to inconsistent view state for large change sets while replication from zero cache down to client converges to the state at the source

2. Buffer changes belonging to transactions on the change source server and stream them to zero only after a timeout has elapsed from the latest change in the transaction was seen in the oplog
  - Pro: unit of work semantics
  - Con: latency
  - Con: higher memory use in change source server
  - Con: insufficiently large timeout may lead to changes from a transaction showing up AFTER a commit message has been streamed to zero. Could be addressed by treating changes as a net-new transaction but this would negates the consistency Pro

I'm unsure how likely the edge cases above will be for most apps but given the emphasis on de-normalization and eventual consistency typical in mongo I'm leaning towards option 1

cbnsndwch (commented):
Added a comment on handling lack of commit markers in the MongoDB change source

cbnsndwch:
https://github.com/rocicorp/mono/pull/3627#discussion_r1931504386

packages/zero-cache/src/services/replicator/change-processor.ts
darkgnotic: Great question.

Interestingly, Postgres has a similar problem (although not quite for the same reason). Namely, changes from different transactions can be interleaved in the log, and thus cannot be streamed immediately to subscribers. Postgres does what you describe in [2]: buffer changes and stream them only after the final commit. Here is a good description of this:

https://www.dolthub.com/blog/2024-03-08-postgres-logical-replication/#wal-data-and-replication-example

I would lean towards going with this approach to begin with, based on the assumption that transactional semantics are generally important to Mongo users as they are for traditional RDBMS users (an assumption which could be wrong!).

If the cons become a big enough problem for our users, there _is_ a Postgres protocol option called "streaming" mode in which large transactions can be streamed while in progress. This addresses the cons that you mentioned for the buffering approach, as detailed here:  https://blog.peerdb.io/exploring-versions-of-the-postgres-logical-replication-protocol

It's a lot of work to implement support for streaming in-progress transactions; each of the downstream layers needs to be augmented to track multiple in-progress transactions, possibly roll them back, and retroactively set the versions of the rows when the transaction ultimately commits. However, it's something we would certainly consider if necessary, and we would add support at the change protocol level so that it would work for both postgres and custom change sources.

arv (commented):
LGTM

packages/zero-cache/src/services/replicator/change-processor.ts
cbnsndwch: > based on the assumption that transactional semantics are generally important to Mongo users as they are for traditional RDBMS users

In some cases, yes.

In my apps, for example, we've modeled the domain so that DB records are essentially full aggregate roots, and writes that cross aggregate boundaries are implemented in such a way that they're very rare. We handle consistency at the application layer.

ACID compliance will definitely be important for users/apps that specify a non-default write concern and use transactions. The issue will be the lack of commit markers in the oplog. I'll work on option 2 with a timeout. Once the initial release of the changelog is ready and I can recruit some other mongo users I'll be able to poll them for their insights.

PR #3628, opened 2025-01-28
chore(zql): log if `next` in `table-source` takes too long

tantaman:
Exists and join will query both sides of a join.

Example:

```ts
z.query.issue.whereExists('comment')
```

Will query:
1. comment.issueID
2. issue.id

because something on the left or the right can change.

A user was missing an index on the opposite side of their join, causing fetches of parent rows to be very slow.

Pulling a single row from SQLite should not take more than 1ms. This adds some logging if/when it does.

https://rocicorp.slack.com/archives/C013XFG80JC/p1738069279890379?thread_ts=1737997408.519909&cid=C013XFG80JC

### DIFF
--- a/packages/zero-cache/src/server/multi/config.test.ts
+++ b/packages/zero-cache/src/server/multi/config.test.ts
@@ -69,7 +69,9 @@ test('parse options', () => {
         },
         "log": {
           "format": "text",
+          "ivmSampling": 0,
           "level": "info",
+          "slowRowThreshold": 3,
         },
         "perUserMutationLimit": {
           "windowMs": 60000,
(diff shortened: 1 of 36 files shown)


packages/zero-cache/src/server/multi/config.test.ts
```diff
@@ -320,6 +324,14 @@ test('zero-cache --help', () => {
                                                    The URL of the trace collector to which to send trace data. Traces are sent over http.            
                                                    Port defaults to 4318 for most collectors.                                                        
                                                                                                                                                      
+     --log-slow-row-threshold number               default: 2
```
tantaman, line 327: note: the expected fetch time is 0.1ms.

PR #3631, opened 2025-01-28
chore(zero-cache): re-enable single process mode

tantaman:
It is easier to profile CPU usage in single process mode. `inspect-brk` seems to sometimes lose track of the workers otherwise if you start the program before enabling cpu profiling.


darkgnotic (approved):

packages/zero-cache/src/server/multi/main.ts
```diff
@@ -120,4 +120,6 @@ export default async function runWorker(
 
 if (!singleProcessMode()) {
   void exitAfter(() => runWorker(parentWorker, process.env));
+} else {
+  await runWorker(null, process.env);
```
darkgnotic, line 124: Actually, I think this might break the integration tests.   🤔

darkgnotic: Yeah, it breaks it.

Maybe the way to do this is to have another entry point, like `src/debug/single.ts`, which sets SINGLE_PROCESS to 1 and calls `loadWorker('../server/multi/main.ts')`.

PR #3634, opened 2025-01-28
feat(zero-cache): back-pressure aware handling of websockets

darkgnotic:
Transforms the incoming and outgoing streams of a WebSocket into Subscription objects with proper back-pressure signaled backoff. 

This will be used for communicating with the custom Change Source endpoint. (@cbnsndwch)

In the future, we should also change our pre-existing transformation logic (i.e. `streamIn()` and `streamOut()` used for internal websocket communication) to use this under the covers in order to properly handle back pressure there as well.

### DIFF
--- a/packages/zero-cache/src/types/streams.ts
+++ b/packages/zero-cache/src/types/streams.ts
@@ -89,9 +89,9 @@ export function stream<In extends JSONValue, Out extends JSONValue>(
 
   // Outgoing transform.
   async function sendOut() {
-    for await (const up of outstream) {
+    for await (const msg of outstream) {
       // Upstream backpressure is signaled by the tcp socket buffer.
-      if (!duplex.write(BigIntJSON.stringify(up))) {
+      if (!duplex.write(BigIntJSON.stringify(msg))) {
         const {promise: canWriteMore, resolve: drained} = resolver();
 
         const start = Date.now();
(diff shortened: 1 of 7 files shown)


arv:
I know this is not stable and widely available but we should keep this in mind:

https://developer.mozilla.org/en-US/docs/Web/API/WebSocketStream

packages/zero-cache/src/types/streams.ts
```diff
@@ -32,6 +40,120 @@ export type Sink<T> = {
...
+  inOptions: Options<In> = {},
+  streamOptions: DuplexOptions = {},
+): {outstream: Sink<Out>; instream: Source<In>} {
+  const endpoint = ws.url ?? 'client';
+  function close(err?: unknown) {
+    if (ws.readyState !== ws.CLOSED && ws.readyState !== ws.CLOSING) {
+      if (err) {
+        closeWithError(lc, ws, err);
+      } else {
+        lc.info?.(`closing conneciton to ${endpoint}`);
```
arv, line 64: typo

packages/zero-cache/src/types/ws.ts
```diff
@@ -1,14 +1,22 @@
 import type {LogContext} from '@rocicorp/logger';
 import type {WebSocket} from 'ws';
 
-export function closeWithProtocolError(
+// https://github.com/Luka967/websocket-close-codes
```
arv, line 4: Or https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4.1

arv (commented):
LGTM

PR #3637, opened 2025-01-29
feat(zero-cache): run `pragma optimize` on connection close

tantaman:
The docs suggest this and:

> This pragma is usually a no-op or nearly so and is very fast. On the occasions where it does need to run ANALYZE on one or more tables, it sets a temporary analysis limit, valid for the duration of this pragma only, that prevents the ANALYZE invocations from running for too long.

https://www.sqlite.org/pragma.html#pragma_optimize


packages/zqlite/src/db.ts
```diff
@@ -72,6 +72,7 @@ export class Database {
   }
 
   close(): void {
+    this.#db.pragma('optimize');
```
darkgnotic, line 75: One worry I had was whether this could ever fail with a "database locked" error since this would be our first case of actually writing to the db from the non-replicator thread.

I guess we'll see. If we see this we can put a try/catch around this and just ignore I guess.

darkgnotic: Actually, looks like it behaves the way we want. `ANALYZE` will fail of the database is locked, but `PRAGMA OPTIMIZE` seems to be best effort.  👍 

Screenshot 2025-01-28 at 21 57 52

tantaman: You bring up a good point though. Isn't the DB almost always locked by the replicator? My prior understanding was that the replicator always holds a write transaction open.

darkgnotic: It only holds one when it's processing a transaction.

I'll put the clause in a try/catch and perhaps re-examine our busy timeouts. Rereading the docs, `PRAGMA optimize` is _often_ a no-op, which is probably why it didn't show any lock errors in my simple test, but my bet is that when it does run `ANALYZE` it will try to grab the lock.

PR #3644, opened 2025-01-29
chore(zero-client)!: Reduce exports

arv:
BREAKING CHANGE!

We do not need to expose these Replicache types...


arv:
@aboodman Technically a breaking change.

aboodman (approved):

PR #3645, opened 2025-01-29
feat(zero): add type functions for custom mutators

tantaman:
Mainly lifting from https://github.com/rocicorp/mono/pull/3493/files#diff-eb7e31f6f0fe54b4a8955bb87dbf0f33aaa890361470233ad2875cc2e6f4ce7a

tweaks:
- multiple args to mutators
- add a default type parameter for `MD` so it need not be specified

### DIFF
--- a/packages/zero-client/src/client/custom.ts
+++ b/packages/zero-client/src/client/custom.ts
@@ -1,0 +1,79 @@
+import type {
+  DeleteID,
+  InsertValue,
+  Schema,
+  TableSchema,
+  UpdateValue,
+  UpsertValue,
+} from '../mod.ts';
+import type {ClientID} from '../types/client-state.ts';
+
+/**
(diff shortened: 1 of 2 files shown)


packages/zero-client/src/client/custom.ts
```diff
@@ -0,0 +1,79 @@
...
+  tx: Transaction<S>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ...args: any[]
+) => void;
+
+export type TransactionReason = 'optimistic' | 'rebase';
+
+/**
+ * WriteTransactions are used with *mutators* which are registered using
+ * {@link ReplicacheOptions.mutators} and allows read and write operations on the
```
arv, line 28: fix link

and description is also not correct

arv (commented):
LGTM

PR #3646, opened 2025-01-29
feat(zero-cache): deprecate unnecessary fields from change-protocol

darkgnotic:
Make a couple of unnecessary fields optional in the change-protocol, to be removed in a later release.

@cbnsndwch


cbnsndwch (commented):
lgtm

PR #3648, opened 2025-01-29
feat(zero): namespace custom mutators to the table they impact

tantaman:
Rather than having all mutators at the top level, they'll need to be namespaced.

Example:

```ts
const mutators = {
    issue: {
      setTitle: (tx, id: string, title: string) => {
        tx.mutate.issue.update({id, title});
      },
    },
   notATableName: { doThing: () => {} }
};
```

The namespace is not required to match a table name but tables names are provided to intellisense at this callsite.

CleanShot 2025-01-29 at 14 15 05

### DIFF
--- a/packages/zero-client/src/client/custom.test.ts
+++ b/packages/zero-client/src/client/custom.test.ts
@@ -1,0 +1,50 @@
+import {expectTypeOf, test} from 'vitest';
+
+import {schema} from '../../../zql/src/query/test/test-schemas.ts';
+import type {CustomMutatorDefs, MakeCustomMutatorInterfaces} from './custom.ts';
+type Schema = typeof schema;
+
+test('argument types are preserved on the generated mutator interface', () => {
+  const mutators = {
+    issue: {
+      setTitle: (tx, id: string, title: string) => {
+        tx.mutate.issue.update({id, title});
+      },
+      setProps: (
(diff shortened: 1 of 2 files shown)


aboodman:
Not sure how I feel about prompting the table names. I'll try it out.

aboodman:
(because people might misunderstand that they can only use those names)

tantaman:
> (because people might misunderstand that they can only use those names)

good point.

aboodman:
I don't know how hard it is but I thought people might enjoy creating
whatever hierarchy they want there? Like having more than two levels? You
know cuz I know how much you love TS wrangling.

tantaman:
> You know cuz I know how much you love TS wrangling.

I'm honestly getting tired of it 😅. I was more than happy when arv took on the "type heirarchy too deep" bug

> I don't know how hard it is but I thought people might enjoy creating whatever hierarchy they want there? 

Mind if I circle back to this after we have custom mutators w/o reads working end to end?

arv (commented):
LGTM

arv:
Can you add a test where there isn't a "namespace" too?

```ts
const mutators = {
   setTitle: (tx, id: string, title: string) => {
     tx.mutate.issue.update({id, title});
   },  
};
```

PR #3650, opened 2025-01-29
feat(zero-client): attach custom mutator implementations to `zero` in…

tantaman:
…stance

Additionally:

- mutators are async! Update their signatures
- ~~support varargs to custom mutators~~ edit: going to roll this change back and do it separately.

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -113,6 +113,12 @@ import {
 import {getServer} from './server-option.ts';
 import {version} from './version.ts';
 import {PokeHandler} from './zero-poke-handler.ts';
+import {
+  makeReplicacheMutator,
+  type CustomMutatorDefs,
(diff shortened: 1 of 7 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -463,7 +486,24 @@ export class Zero<const S extends Schema> {
...
+      let existing = mutate[namespace];
+      if (existing === undefined) {
+        existing = {};
+        mutate[namespace] = existing;
+      }
+
+      for (const name of Object.keys(
+        mutatorsForNamespace as Record<string, CustomMutatorImpl<Schema>>,
+      )) {
+        existing[name] = must(rep.mutate[customMutatorKey(namespace, name)]);
```
arv, line 504: TODO: Support "namespacing", ie zero.mutate.x.y.z

arv (commented):
LGTM

PR #3652, opened 2025-01-30
fix(zero-cache): correctly handle multiple-column schema changes

darkgnotic:
The previous schema change logic relied on the assumption that for an `ALTER TABLE` statement, only one column change can happen at a time. This is incorrect, as multiple column actions can be present in the statement. 

The corrected logic leverages the tables' `oid` and associated columns' `attnum` values to determine the diffs that need to be applied to the replica. As such, it no longer needs to know what the DDL command was (e.g. `ALTER TABLE` vs `ALTER PUBLICATION`).

https://discord.com/channels/830183651022471199/1333484736814776360/1333497466363973642

Tangential: Also fix the degraded-mode schema change detection logic (e.g. supabase, neon) to detect changes in the columns' NOT NULL attribute.


arv (commented):
LGTM

PR #3655, opened 2025-01-30
fix(zero-cache): handle subscribers ahead of the change-streamer change-log

darkgnotic:
Because the `change-streamer` forwards messages to view-syncers immediately, and concurrently stores and acks upstream, it is possible for a replica to be ahead of the change log if a server is killed before the change can be stored. 

Handle this case by considering the subscriber "caught up" if it is ahead of the change log. The existing forwarder logic already handles filtering out messages that have already been seen.

https://discord.com/channels/830183651022471199/1288232858795769917/1334328586265563260

https://discord.com/channels/830183651022471199/1332721644527157319


arv (commented):
LGTM

PR #3661, opened 2025-01-31
chore(infra): sandbox sst deploy

cesara:


prod/sst/sst.config.ts
```diff
@@ -0,0 +1,203 @@
...
+  app(input) {
+    return {
+      name: process.env.APP_NAME || "zero",
+      removal: input?.stage === "production" ? "retain" : "remove",
+      home: "aws",
+      region: process.env.AWS_REGION || "us-east-1",
+    };
+  },
+  async run() {
+
```
darkgnotic, line 17: uber nit: remove one blank line here?

prod/sst/sst.config.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    const loadSchemaJson = () => {
+      if (process.env.ZERO_SCHEMA_JSON) {
+        return process.env.ZERO_SCHEMA_JSON;
+      }
+
+      try {
+        const schema = readFileSync("zero-schema.json", "utf8");
```
darkgnotic, line 25: what's the intention here? is the idea that this deploy script would be run with a zero-schema.json in the same directory?

prod/sst/sst.config.ts
```diff
@@ -0,0 +1,203 @@
...
+        command: ["CMD-SHELL", "curl -f http://localhost:4848/ || exit 1"],
+        interval: "5 seconds",
+        retries: 3,
+        startPeriod: "300 seconds",
+      },
+      environment: {
+        ...commonEnv,
+        ZERO_CHANGE_STREAMER_URI: replicationManager.url,
+        ZERO_UPSTREAM_MAX_CONNS: "15",
+        ZERO_CVR_MAX_CONNS: "160",
```
darkgnotic, line 145: At some point I'd like to make these numbers automatically calculate based on higher level configuration that we pass into a library. Like `maxTotalConnections` that gets divvied up according to the task scaling options. This is where it's really nice to have a turing complete language for the configuration (as opposed to Yaml).

Excited about evolving this.

prod/sst/sst.config.ts
```diff
@@ -0,0 +1,203 @@
...
+            healthyThreshold: 2,
+            timeout: 3,
+          },
+        },
+      },
+    });
+
+    // View Syncer Service
+    cluster.addService(`view-syncer`, {
+      cpu: "2 vCPU",
```
darkgnotic, line 132: Should we bump this to our target prod resource usage? I think its "8 vCPU" and "16 GB" of ram. Or were you thinking of making this parameterizable in a library first?

darkgnotic (approved):
🚀

prod/sst/sst.config.ts
cesara: ya the idea is that the sst deploy would read the zero-schema.json. in the same directory you are deploying from and use that value for the env variable. now that i think of it i should probably allow this to be an env variable also.

.prettierignore
```diff
@@ -9,8 +9,8 @@ packages/replicache/doc/docs/api
 packages/replicache/doc/.docusaurus/
 **/.next
 **/.vercel
+*.tsbuildinfo
 renderer
 vendor
 !/renderer/src/constants.ts
-**/bot-recordings.ts
```
cesara, line 15: did not mean to remove, re-adding

PR #3671, opened 2025-01-31
chore(zql-to-sql): harness for testing against pg

tantaman:
the `zql-to-sql` package will also need all the PG test utilities. Since `zql-to-sql` is more primitive, move the test utils there and have `zero-cache` import them.


darkgnotic (commented):
Defining a hierarchy based on some notion of "primitiveness" seems a bit artificial. (If there were or are going to be actual dependencies that might make a little sense ...)

Regardless, it might be better to put the testing harness stuff in a shared directory that both packages reference. Then we avoid any implications of code dependencies between unrelated packages.

I don't know of the perfect name, but throwing some out there:

* packages/pg-shared
* packages/pg-test
* something else ...

wdyt?

tantaman:
> Defining a hierarchy based on some notion of "primitiveness" seems a bit artificial

Specifically, `zero-cache` depends on `zql-to-sql` rather than the other way around.

I like `packages/pg-test`

darkgnotic:
> > Defining a hierarchy based on some notion of "primitiveness" seems a bit artificial
> 
> @darkgnotic - specifically, `zero-cache` depends on `zql-to-sql` rather than the other way around. So `zql-to-sql` is objectively more primitive than `zero-cache`.

I didn't realize that. When will `zero-cache` compile to sql? (My assumption that zql-to-sql would be in the mutation-server, after we push mutations to it).

> 
> I like `packages/pg-test`

👍

darkgnotic:
> > (My assumption that zql-to-sql would be in the mutation-server, after we push mutations to it).
> 
> @darkgnotic - since all the services (replicator, view syncer, mutagen1) are in zero-cache right now I was going to put the mutation server code there too. Maybe that doesn't make sense though. Hadn't thought too much about it till just now.

That totally makes sense. I was just under the impression that we would only be pushing mutations to some external mutation server, and not handling the mutations ourselves. But don't worry, I'll catch up with the plan at some point.  😅

darkgnotic:
> > (My assumption that zql-to-sql would be in the mutation-server, after we push mutations to it).
> 
> @darkgnotic - since all the services (replicator, view syncer, mutagen1) are in zero-cache right now I was going to put the mutation server code there too. Maybe that doesn't make sense though. Hadn't thought too much about it till just now.

Regardless, let's do the `packages/pg-test` thing so that the setup files are more "shared" than "inherited".

tantaman:
I actually don’t know what I was thinking . Only the user’s api server will
use z2s, not zero cache. Need sleep 😴

darkgnotic:
> I actually don’t know what I was thinking . Only the user’s api server will use z2s, not zero cache. Need sleep 😴

🐑 🐑 🐑 
    😴

PR #3672, opened 2025-01-31
feat(shared) re-add `clone` to `btree-set`

tantaman:
This'll be needed to support reads in custom mutators on the client.

Design: https://www.notion.so/replicache/Custom-Mutators-1353bed8954580f09875f1484e7ce7d4#18b3bed8954580f38321e897032b0fdf


arv (commented):
LGTM

PR #3677, opened 2025-01-31
fix(zero-cache): make replication-manager handoff watertight

darkgnotic:
The `replication-manager` reads the `lastWatermark` from the change DB in order to take over the postgres replication slot from the right starting point.

Because the old replication manager is not killed until the replication slot is taken over, there is a potential race condition in which the old replication manager, before shutting down, commits a new change to the change log, moving the watermark forward. This would result in an INSERT conflict when the new replication manager attempts to persist that same transaction.

Although ignoring conflicts is a potential workaround, it has the drawback of masking other types of subtle bugs.

The solution implemented here uses a new `replicationState` table with a single row used to coordinate handoffs. The row contains:
* an `owner` column for claiming ownership (similar to CVR ownership)
* a `lastWatermark` column that is updated whenever a new transaction is committed

When the `replication-manager` starts, it writes itself as the `owner` and reads the `lastWatermark`. This initial write prevents any subsequent writes from the old `replication-manager`, which always reads the `owner` in each transaction before writing the new `lastWatermark`. A change in ownership is detected either:
* via the read, in the case where the ownership change was written before the read, or
* during the commit, which would manifest as a SERIALIZATION_FAILURE because a row that was read during the transaction has changed

### DIFF
--- a/packages/zero-cache/src/server/change-streamer.ts
+++ b/packages/zero-cache/src/server/change-streamer.ts
@@ -23,16 +23,24 @@ export default async function runWorker(
   env: NodeJS.ProcessEnv,
 ): Promise<void> {
   const config = getZeroConfig(env);
-  const port = config.changeStreamerPort ?? config.port + 1;
+  const {
+    taskID,
+    changeStreamerPort: port = config.port + 1,
+    upstream,
+    change,
+    shard,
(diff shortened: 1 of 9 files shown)


arv (commented):
LGTM

PR #3681, opened 2025-02-01
feat(zero-schema): allow remapping column and table names

darkgnotic:
API looks like:

```ts
const user = table('user')
  .from('users')
  .columns({
    id: string().from('user_id'),
    name: string().from('user_name'),
    recruiterId: number().from('user_recruiter_id'),
  })
  .primaryKey('id');
```


arv:
I haven't looked at the details yet.

The main question/concern I have is that we should do this transformation once, preferably on the server and have the client be totally oblivious to this.

I think it needs some design.

packages/zero-client/src/client/context.test.ts
```diff
@@ -66,26 +71,30 @@ test('getSource', () => {
   expect(context.getSource('nonexistent')).toBeUndefined();
 
   // Should work for other table too.
-  const source2 = context.getSource('userStates');
+  const source2 = context.getSource('user_states');
```
tantaman, line 74: Looking at https://github.com/rocicorp/mono/pull/3684/files#diff-5add7e0d6b7b33e13c3a57846ae287c60e56e40352b80ad5ad26f64c870c2e26R338 it looks like the client still gets to refer to everything by client names but I don't see this test updated in that pr to go back to using `userStates`.

packages/zero-client/src/client/context.test.ts
```diff
@@ -66,26 +71,30 @@ test('getSource', () => {
...
       "columns": {
         "stateCode": {
           "customType": null,
+          "name": "stateCode",
           "optional": false,
+          "sourceName": "stateCode",
           "type": "string",
         },
-        "userID": {
+        "user_id": {
```
tantaman, line 85: ditto.

packages/zero-client/src/client/context.test.ts
```diff
@@ -66,26 +71,30 @@ test('getSource', () => {
...
           "customType": null,
+          "name": "userID",
           "optional": false,
+          "sourceName": "user_id",
           "type": "string",
         },
       },
       "primaryKey": [
-        "userID",
+        "user_id",
```
tantaman, line 94: ditto.

tantaman:
> The main question/concern I have is that we should do this transformation once, preferably on the server and have the client be totally oblivious to this.

In https://github.com/rocicorp/mono/pull/3684 it looks like:

1. The client transforms the AST to a "server AST" when sending it to the server (https://github.com/rocicorp/mono/pull/3684/files#diff-acf758b9ee0921927ea5abfbcae3f6f82d658f4e38284abcbcdaee1bce43ed8cR140)
2. The server pokes server results which the client re-maps to client names before writing to storage (https://github.com/rocicorp/mono/pull/3684/files#diff-23a32e8a18fc970895939c0cef1412fbb0b730cf1f19053151f69ae62fd0712cR352)

Since the transforms happen at the two choke points (push and poke), the client gets to refer to everything by client names and server by server names.


This looks reasonable to me but I'm still thinking through schema changes. Having the client do the mapping does seem more correct than having the server do it since a client could be on a different schema version than the server.

darkgnotic:
Very sorry. This PR is an old, obsolete strategy. Will send out the new one.

PR #3682, opened 2025-02-01
perf(zero-solid): use reconcile

grgbkr:
Setting solid store state many times is very slow, even if it is within a Solid batch call.  In the hello-zero-solid app, if there are 3000 rows, processing the pushes during initial load takes 30 seconds. Nearly all of the time is in an internal solid function called unwrap.  
image
image

The previous approach to solving this performance issue, https://github.com/rocicorp/mono/pull/3608, is incorrect cause we cannot defer processing pushes, pushes have to be processed as received, because the relationship streams will be incorrect if processing is deferred.  This was surfaced by issue https://bugs.rocicorp.dev/issue/3488
For comparison here is the performance of that approach:
image

This new approach maintains a copy of the view as a plain JS object which is updated on each push as the push is received, and then uses solid's `reconcile` to update the store at the end of the transaction.
image
The downside is we are ~2xing memory (actually this depends on how much of the data is interned string, its probably actually much less than 2x), and we are diffing the views (which doesnt seem to match the solid way of precise reactivity, we are translating a stream of precise updates into a big diff).

The upside is its 
1. correct
2. event faster than the previous optimization 120 ms vs 970 ms.  Probably because in this test case the diff is trivial.

### DIFF
--- a/packages/zero-solid/src/solid-view.ts
+++ b/packages/zero-solid/src/solid-view.ts
@@ -34,17 +34,13 @@ export class SolidView<V> implements Output {
   #state: Store<State>;
   #setState: SetStoreFunction<State>;
 
-  #pendingChanges: Change[] = [];
-
   constructor(
     input: Input,
-    onTransactionCommit: (cb: () => void) => void,
-    format: Format,
-    onDestroy: () => void,
-    queryComplete: true | Promise<true>,
+    format: Format = {singular: false, relationships: {}},
+    onDestroy: () => void = () => {},
+    queryComplete: true | Promise<true> = true,
   ) {
     this.#input = input;
-    onTransactionCommit(this.#onTransactionCommit);
     this.#format = format;
     this.#onDestroy = onDestroy;
     [this.#state, this.#setState] = createStore<State>([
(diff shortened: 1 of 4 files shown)


arv:
I have a branch that treats the data as immutable and does the bare updates. My initial goal was to use that for react because in react we currently deep copy (since the react external store needs a new object as a signal that something changed)..

If we are using reconcile we should use that.

I'll upload my branch for your looking...

packages/zero-solid/src/solid-view.ts
```diff
@@ -47,60 +49,55 @@ export class SolidView<V> implements Output {
...
-      );
-    } finally {
-      this.#pendingChanges = [];
-    }
-  }
-
   push(change: Change): void {
-    // Delay setting the state until the transaction commit.
-    this.#pendingChanges.push(change);
+    applyChange(this.#root, change, this.#input.getSchema(), '', this.#format);
```
arv, line 100: Is this correct? We are now changing the root without triggering updates. I'm concerned that there might be some way to see this intermediary state?

Maybe we should have a #newRoot and swap the root to that using reconcile in commit?

arv (commented):
LGTM but I'm afraid this might have other problems.

arv:
https://github.com/rocicorp/mono/pull/3688

PR #3683, opened 2025-02-02
fix(zero): fix change-protocol/v0 export paths in package.json

cbnsndwch:
The `change-protocol/v0` export paths in `@rocicorp/zero@0.13.2025013101` point to a `mod.ts` file that doesn't exist in the `out` directory, which breaks imports for custom change sources.

This PR fixes this by:

- adding a `change-protocol/v0.ts` file to `zero/src`
- adding the `change-protocol` directory to the `include` setting in `packages/zero/tsconfig.server.json`
- updating the `./change-protocol/v0` export in `packages/zero/package.json` to point to the new file


darkgnotic:
Thanks @cbnsndwch. I will defer to build-master @arv  🙏

arv (approved):
LGTM

Thanks

PR #3684, opened 2025-02-02
feat(zero-schema): allow mapping of server->client table and column names

darkgnotic:


packages/zero-client/src/client/query-manager.ts
```diff
@@ -129,17 +137,16 @@ export class QueryManager {
     let entry = this.#queries.get(astHash);
     this.#recentQueries.delete(astHash);
     if (!entry) {
+      const serverAST = makeServerAST(normalized, this.#tables);
```
darkgnotic, line 140: @grgbkr this all works locally end-to-end but I wanted to confirm that this is okay:

* The IVM pipeline (and the query manager here) compute the hash of the AST in the client namespace.
* The query manager than converts the AST to the server namespace (using the same hash) when sending the AST to the server
* The server returns the same (server-namespaced) AST's in its query patches, which are stored in replicache

The last point is what I'm not 100% confident about. Is it okay that the client is storing server-namespaced AST's in IDB? I don't know how they are used from there ...

grgbkr: The stored ASTs are currently not used.  The client currently only checks for equality of queries, which it does by comparing hashes. 

The eventual use case is 'consistency', where we will want to know does 'got query' X cover 'desired query' Y (more sophisticated comparison than just equality).  Here we will need namespaces to be consistent. 

My instinct is we should push the mapping to the very edge of the api (query builder and view), and have everything else work on the server namespace.

darkgnotic: My first iteration (#3681) was along these lines. It was slightly messy, but it can work.

The ArrayView would have to use a different comparator than the rest of the IVM pipeline. I was also unsure whether the ArrayView is suitable as the boundary (i.e. whether other parts of the pipeline might get exposed to the application layer).

> The eventual use case is 'consistency', where we will want to know does 'got query' X cover 'desired query' Y (more sophisticated comparison than just equality). Here we will need namespaces to be consistent.

If we stick to the wire-protocol being the conversion boundary, we can convert the server AST's back to the client namespace. In the poke handler.

I think @tantaman was leaning towards the conversion boundaries in this PR, but I could go either way.

darkgnotic: @grgbkr and I decided to continue with this approach. I'll add server-to-client name mapping to the AST's before storing them in IDB.

darkgnotic: Okay, ast's from poked queries are now mapped back to the client namespace.

The test looks much nicer with the queries and rows (in the Replicache poke) consistent with each other.

arv:
LGTM

The code size increase is a bit unfortunate but this is adding a major feature and I do not see a way to get this to compile much smaller.

PR #3685, opened 2025-02-02
fix(react): Allows useQuery to work again when server is null.

aboodman:
Fixes https://bugs.rocicorp.dev/issue/3497

This was broken with:

https://github.com/rocicorp/mono/pull/3409#discussion_r1894702111

I don't think the server check was actually needed to accomplish the goal of that PR, but will check w/ chase. Want to land this now to unblock user with canary.


arv (approved):
LGTM

PR #3686, opened 2025-02-02
feat(react): Allow passing a server snapshot to useQuery.

aboodman:
This enables basic SSR/handoff.

### DIFF
--- a/packages/zero-react/src/use-query.tsx
+++ b/packages/zero-react/src/use-query.tsx
@@ -23,19 +23,38 @@ export function useQuery<
   TReturn,
 >(
   q: Query<TSchema, TTable, TReturn>,
-  enable: boolean = true,
+  enableOrOptions:
+    | boolean
+    | {
+        enable?: boolean | undefined;
+        serverSnapshot: TReturn | undefined;
+      }
+    | undefined = true,
 ): QueryResult<TReturn> {
   const z = useZero();
+  const enable =
+    typeof enableOrOptions === 'boolean'
+      ? enableOrOptions
+      : enableOrOptions.enable ?? false;
+  const options =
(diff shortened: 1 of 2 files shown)


packages/zero-react/src/use-query.tsx
```diff
@@ -164,6 +183,7 @@ export class ViewStore {
     clientID: string,
     query: AdvancedQuery<TSchema, TTable, TReturn>,
     enabled: boolean,
+    requireComplete: boolean,
```
tantaman, line 186: switch to an `options` object now that we're at 2 boolean args?

packages/zero-react/src/use-query.tsx
```diff
@@ -23,19 +23,38 @@ export function useQuery<
...
     q as AdvancedQuery<TSchema, TTable, TReturn>,
     enable && z.server !== null,
+    serverSnapshot !== undefined,
   );
   // https://react.dev/reference/react/useSyncExternalStore
   return useSyncExternalStore(
     view.subscribeReactInternals,
     view.getSnapshot,
-    view.getSnapshot,
+    serverSnapshot
```
tantaman, line 52: Do you have an example of how someone uses zero on the server to populate this?

Something like this?

```ts
const q = z.query.issue.related('assignee');
export async function getServerSideProps() {
  return { issues: await q.run() }
}

function MyComponent({issues}) {
  const q = useQuery(q, {serverSnapshot: issues});
  ...
}
```

Unclear to me how you hydrate a tree of components with SSR without one giant query at the root.

aboodman:
Sorry this wasn’t ready for review yet. Put into a draft pr to share early work with @jjenzz

aboodman:
When I tested this, it was not working for some reason – as in, we were still rendering an incomplete result in between the SSR result and the complete client result for some reason.

PR #3687, opened 2025-02-03
chore(infra): zbugs dockerfile to match zero

cesara:


prod/zbugs/Dockerfile
```diff
@@ -1,19 +1,27 @@
...
-RUN tar -C /usr/local/bin -xzf /tmp/litestream.tar.gz
-RUN chmod +x /usr/local/bin/litestream
-RUN echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > .npmrc && \
-    npm run build-ci && \
-    rm -f .npmrc
-RUN apk add --update curl
-WORKDIR /opt/app/packages/zero-cache
+
+# Copy the tarball from the build context
+COPY ./prod/zbugs/rocicorp-zero-*.tgz .
```
darkgnotic, line 12: So close!

I wonder if we can modify the main Dockerfile to do either or? Like, in pseudo code it would be something like:

```
if $ZERO_VERSION ends with ".tgz"
  COPY ./${ZERO_VERSION} .
  RUN npm install -g ./${ZERO_VERSION}
else
  npm install -g @rocicorp@${ZERO_VERSION}
```

PR #3691, opened 2025-02-03
feat(z2s): test against pg

tantaman:
Reveals that the compiler will need to:
1. support `hidden` tables for junction edges
2. support `one`

Will provide the `format` as an arg to the `compiler` in a follow up PR so we can take `one` and `junction` edges into account.


darkgnotic (commented):
LGTM

PR #3693, opened 2025-02-03
feat(zero-client): support zql reads in optimistic mutations

tantaman:


packages/zero-client/src/client/custom.test.ts
```diff
@@ -124,3 +124,41 @@ test('custom mutators write to the local store', async () => {
   issues = z.query.issue.run();
   expect(issues.length).toEqual(0);
 });
+
+test('custom mutators can query the local store during an optimistic mutation', async () => {
+  const z = zeroForTest({
+    schema,
+    mutators: {
+      issue: {
```
arv, line 132: for testing, we do not need the extra fancy mutate.x.y.z namespacing.

packages/zero-client/src/client/custom.test.ts
```diff
@@ -124,3 +124,41 @@ test('custom mutators write to the local store', async () => {
...
+                tx.mutate.issue.update({id: issue.id, closed: true}),
+              ),
+          );
+        },
+      },
+    },
+  });
+
+  await Promise.all(
+    Array.from({length: 10}).map(async (_, i) => {
```
arv, line 147:
```suggestion
    Array.from({length: 10}, async (_, i) => {
```

packages/zero-client/src/client/custom.test.ts
arv: Maybe I misunderstood the intended use of the extra levels?

@aboodman

arv (approved):
LGTM

PR #3700, opened 2025-02-04
feat(zero-cache): support syncing tables from non-public schemas

darkgnotic:
### Feature

Tables from non-public schemas can be synced to the client.

### Instructions

This requires [creating a publication](https://www.postgresql.org/docs/current/sql-createpublication.html) on upstream (as the default will only sync tables from the "public" schema):

```sql
CREATE PUBLICATION zero_data FOR TABLE foo, TABLE bar, TABLE my_schema.baz, ...;
```

and specifying that publication in the `ZERO_SHARD_PUBLICATIONS` server option:

```env
# .env

ZERO_SHARD_PUBLICATIONS=zero_data
```

Finally, on the client, the schema must be configured to reference the table correctly:

```ts
// schema.ts

const baz = table('baz')
    .from('my_schema.baz')
    .columns({...})
    .build();
```

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/schema/validation.ts
+++ b/packages/zero-cache/src/services/change-source/pg/schema/validation.ts
@@ -6,22 +6,10 @@ import {
 import {Default} from '../../../../db/postgres-replica-identity-enum.ts';
 import type {PublishedTableSpec} from '../../../../db/specs.ts';
 import {ZERO_VERSION_COLUMN_NAME} from '../../../replicator/schema/replication-state.ts';
-import {unescapedSchema} from './shard.ts';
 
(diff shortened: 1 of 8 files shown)


packages/zero-cache/src/services/change-source/pg/schema/validation.ts
```diff
@@ -6,22 +6,10 @@ import {
...
 
 const ALLOWED_IDENTIFIER_CHARS = /^[A-Za-z_]+[A-Za-z0-9_-]*$/;
 
-export function validate(
-  lc: LogContext,
-  shardID: string,
-  table: PublishedTableSpec,
-) {
-  const shardSchema = unescapedSchema(shardID);
-  if (!['public', 'zero', shardSchema].includes(table.schema)) {
```
darkgnotic, line 19: The removal of this check is the only behavioral code change.

Everything else in this PR is tests.  😀

PR #3701, opened 2025-02-04
perf,fix(zero-solid): back to produce but pre-reading relationships and optimizing initial view build

grgbkr:
The `reconcile` optimization approach (https://github.com/rocicorp/mono/pull/3682) was incorrect (I misunderstood how reconcile works), as was revealed by the tests added here: https://github.com/rocicorp/mono/pull/3692.  

This approach goes back to using `produce`, but addresses the correctness issue by using an idea @tantaman came up with of reading the relationships of the changes at the time the push is received, but still queueing them to be applied to the solid store state on transaction completion.   We are careful to only read relationships in the changes actually used by `applyChanges`.

This solved the correctness problem, but while much faster than the 30 seconds prior to us trying to optimize this, the 3000 row initial load case in hello-zero-solid was still spending ~742ms in  SolidView#applyChanges.  
Pasted Graphic 2

This is because the proxy used by solid `produce` is fairly slow.   To address this, I added an optimization for the case where the store's view is currently empty.  If the store is currently empty build up the view on a new plain old JS object root, which is returned for the new state.  This avoids building up large views from scratch using solid `produce`.  This brings the time for the 3000 row initial load case in hello-zero-solid to ~133ms.
Pasted Graphic 1

### DIFF
--- a/packages/zero-solid/src/solid-view.ts
+++ b/packages/zero-solid/src/solid-view.ts
@@ -1,6 +1,6 @@
 import {
   createStore,
-  reconcile,
+  produce,
   type SetStoreFunction,
   type Store,
 } from 'solid-js/store';
@@ -14,14 +14,17 @@ import {
   type Output,
   type Query,
   type ViewFactory,
-} from '../../zero/src/advanced.ts';
-import type {ResultType, Schema} from '../../zero/src/zero.ts';
-import {deepClone} from '../../shared/src/deep-clone.ts';
-import {createSignal} from 'solid-js';
+} from '../../zero-advanced/src/mod.js';
+import type {Schema} from '../../zero-schema/src/mod.js';
+import type {ResultType} from '../../zql/src/query/typed-view.js';
+import type {Row} from '../../zero-protocol/src/data.ts';
+import type {Node} from '../../zql/src/ivm/data.ts';
 
 export type QueryResultDetails = {
   readonly type: ResultType;
 };
+
+type State = [Entry, QueryResultDetails];
 
 const complete = {type: 'complete'} as const;
 const unknown = {type: 'unknown'} as const;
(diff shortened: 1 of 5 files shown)


packages/zql/src/ivm/view-apply-change.ts
```diff
@@ -7,14 +7,51 @@ import {
...
+import {drainStreams, type Comparator, type Node} from './data.ts';
 import type {SourceSchema} from './schema.ts';
 import type {Entry, EntryList, Format} from './view.ts';
 
+/**
+ * `applyChange` does not consume the `relationships` of `ChildChange#node`,
+ * `EditChange#node` and `EditChange#oldNode`.  The `ViewChange` type
+ * documents and enforces this via the type system.
+ */
+export type ViewChange =
```
grgbkr, line 19: I don't love having these new types exposed in `@rocicorp/zero/advanced`, but it seemed better than just dropping the relationships from ChildChange and EditChange.

packages/zero-solid/src/solid-view.ts
```diff
@@ -49,56 +50,130 @@ export class SolidView<V> implements Output {
...
+        node: {row: change.node.row},
+        oldNode: {row: change.oldNode.row},
+      };
+  }
+}
+
+function materializeNodesRelationships(node: Node): Node {
+  return {
+    row: node.row,
+    relationships: Object.fromEntries(
```
arv, line 170: Do we really need all these arrays?

I'm worried about the strain we are putting on the GC here

arv (commented):
LGTM

I still don't feel good about this. Why is it so slow?

aboodman:
Remember that the test case is pushing like 9k rows. So it's 1.5ms/row.
Which ... still surprisingly slow, but tolerable.

aboodman:
Sorry I mean 0.01ms/row. Maybe within range of reason.

PR #3702, opened 2025-02-04
perf(zero-solid): avoid temp arrays when building from empty

grgbkr:
This further optimizes the case of building from empty by avoiding allocating temp arrays.

Now we only have to allocate temp arrays for storing relationships between push and transaction commit when updating a non empty view, which is often smaller updates.

This didn't have an appreciable impact on the test case we've been using of 3000 rows (with 2, 1 child, relationships per row) in hello-zero-solid, but those rows all have very small non-nested relationships, so thats understandable.  I still think this is a worthwhile optimization despite it adding a bit of complexity.
image

### DIFF
--- a/packages/zero-solid/src/solid-view.ts
+++ b/packages/zero-solid/src/solid-view.ts
@@ -34,6 +34,15 @@ export class SolidView<V> implements Output {
   readonly #format: Format;
   readonly #onDestroy: () => void;
 
+  // Optimization: if the store is currently empty we build up
+  // the view on a plain old JS object stored at #builderRoot, and return
+  // that for the new state on transaction commit.  This avoids building up
+  // large views from scratch via solid produce.  The proxy object used by
+  // solid produce is slow and in this case we don't care about solid tracking
+  // the fine grained changes (everything has changed, it's all new).  For a
+  // test case with a view with 3000 rows, each row having 2 children, this
+  // optimization reduced #applyChanges time from 743ms to 133ms.
+  #builderRoot: Entry | undefined;
   #state: Store<State>;
   #setState: SetStoreFunction<State>;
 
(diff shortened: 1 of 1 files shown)


arv:
In general I think we should be more careful about allocations. They all add up in the end.

packages/zero-solid/src/solid-view.ts
```diff
@@ -164,18 +180,23 @@ function materializeRelationships(change: Change): ViewChange {
...
 
-function materializeNodesRelationships(node: Node): Node {
+function materializeNodeRelationships(node: Node): Node {
   return {
     row: node.row,
     relationships: Object.fromEntries(
       Object.entries(node.relationships).map(([relationship, stream]) => {
-        const drained = [...stream()].map(materializeNodesRelationships);
-        return [relationship, () => drained];
+        const materialized = [...stream()].map(materializeNodeRelationships);
```
arv, line 188: This function was the one that stood out to me.

If you write in a non functional style there are no extra arrays created.

```ts
export function materializeNodeRelationships(node: Node): Node {
  const relationships: Record<string, () => Stream<Node>> = {};
  for (const relationship in node.relationships) {
    const materialized: Node[] = [];
    for (const n of node.relationships[relationship]()) {
      materialized.push(materializeNodeRelationships(n));
    }
    relationships[relationship] = () => materialized;
  }
  return {
    row: node.row,
    relationships,
  };
}
```

packages/zero-solid/src/solid-view.ts
```diff
@@ -50,13 +59,21 @@ export class SolidView<V> implements Output {
...
     [this.#state, this.#setState] = createStore<State>([
-      this.#initialEmptyEntry(),
+      initialRoot,
       queryComplete === true ? complete : unknown,
     ]);
-    input.setOutput(this);
-
-    this.#applyChanges(input.fetch({}), node => ({type: 'add', node}));
+    if (isEmptyRoot(initialRoot)) {
+      this.#builderRoot = this.#createEmptyRoot();
```
arv, line 75: why not use initialRoot here?

arv (approved):
LGTM

packages/zero-solid/src/solid-view.ts
grgbkr: doing a quick follow up for this.

packages/zero-solid/src/solid-view.ts
grgbkr: To avoid directly modifying the object that we set as the stores state.

packages/zero-solid/src/solid-view.ts
grgbkr: https://github.com/rocicorp/mono/pull/3703

PR #3705, opened 2025-02-04
feat(zero-cache): map CRUD operations to server names

darkgnotic:
Maps CRUD operations to their server names if aliased in the schema.

Also consolidates the name mapping logic.

### DIFF
--- a/packages/zero-protocol/src/push.ts
+++ b/packages/zero-protocol/src/push.ts
@@ -1,8 +1,13 @@
 import {jsonSchema} from '../../shared/src/json-schema.ts';
 import * as v from '../../shared/src/valita.ts';
+import type {NameMapper} from '../../zero-schema/src/name-mapper.ts';
 import {rowSchema} from './data.ts';
 import * as MutationType from './mutation-type-enum.ts';
-import {primaryKeySchema, primaryKeyValueRecordSchema} from './primary-key.ts';
+import {
+  primaryKeySchema,
+  primaryKeyValueRecordSchema,
+  type PrimaryKey,
+} from './primary-key.ts';
 
 export const CRUD_MUTATION_NAME = '_zero_crud';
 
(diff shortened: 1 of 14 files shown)


packages/zero-protocol/src/push.ts
```diff
@@ -105,3 +110,37 @@ export type CustomMutation = v.Infer<typeof customMutationSchema>;
...
+
+export function mapCRUD(
+  arg: CRUDMutationArg,
+  map: NameMapper,
+): CRUDMutationArg {
+  return {
+    ops: arg.ops.map(({op, tableName, primaryKey, value}) => {
+      switch (op) {
+        case 'delete': {
+          // 'delete' is separated to appease ts since its `value` is of a different type.
```
arv, line 122: Is the code identical? If so static unsafe cat would be reasonable

packages/zero-schema/src/name-mapper.ts
```diff
@@ -0,0 +1,118 @@
...
+      throw new Error(
+        `unknown column "${src}" of "${table}" table ${
+          !ctx ? '' : `in ${JSON.stringify(ctx)}`
+        }`,
+      );
+    }
+    return dst;
+  }
+
+  row<R extends Row>(table: string, row: R): R {
```
arv, line 92: This is not really correct, is it? The input row and the output row are not always the same type

darkgnotic: They should be. The only thing that can differ are the object keys. The values will be the same, so I think that means it's the same type? Unless I'm misunderstanding something.

For clarity, I did this in order to handle both `Record<string, JSONValue>` and `Record<string, JSONValue | undefined>`.

packages/zero-protocol/src/push.ts
darkgnotic: Yeah, it's identical. I might have had trouble figuring out what to cast.   🥵

packages/zero-schema/src/name-mapper.ts
arv:
```ts
row(table: string, row: Row): Row {
```

would be correct but not what you have. Let me show a TS playground.

https://www.typescriptlang.org/play/?#code/FAFwngDgpgBASgewO4wLzygYwQJwCYA8AziDgJYB2A5gDQwUCuAtgEZQ4B8A3MMAGYMKmEGQQUYfAnBhQAHiCgU8ReMg4AKHAC54ASh3SA3sBgxsFEjBwA3A8jQxDAXxMTcMdecsBrKGBiUVrqOrqY2ANq+YAB0IAgAqhDQOADCAIZEUOq6ALoOOJF+OTymLmFQIAw44jYwGfA8LqCQsHAAjA6GaTptXDAsOgBMTjxeIFZtBh3oXT10AzDDPM3Q8IOdAIJzMABCQy5jVoMG6+h8mm26PEA

What you really need is:

```ts
row<In extends Row, Out extends Row>(table: string, row: In): Out {
```

But a better much more advanced solution would be to use the schema to figure out the correct type

```ts
row<S extends Schema, R extends Row>(table: string, row: R): MapRowNames<S, R> {
```

darkgnotic: Hows this approach?  #3712

PR #3708, opened 2025-02-05
chore(zero-protocol): combine identical code with unsafe cast

darkgnotic:


darkgnotic:
@arv are you able to ts-fu a way around this?

PR #3709, opened 2025-02-05
feat(zero-client): Send deleteClients to server

arv:
This is work in progress to support cleaning up clients on the server. At this point the server doesn't do anything with this message.

Here is how it is intended to work:
- Replicache tells zero client when a replicache client is deleted.
- Zero client sends a `deleteClients` message to the server.
- It also stores the IDs of the deleted clients persistently (using the perdag). It stores these in a new head. The deleted clients is a shared state between all client groups.
- When the server receives a `deleteClients` message it deletes the clients (not yet implemented), and...
- Sends a deleteClients message back to the client with the IDs of the clients that were actually deleted.
- The client then removes the IDs from the list of deleted clients.
- Zero client also sends the list of deleted clients to the server when it connects, so the server can clean up clients that were deleted while the client was offline. This is done after a short timeout to avoid competing with more important work.


arv:
@grgbkr 

```
// TODO: Maybe should send this as part of initConnection? It would be good to
// delete the clients before we send down the results of the queries in case
// some of those queries should be deleted.
```

I think this is the right thing to do. WDYT?

packages/replicache/src/persist/collect-idb-databases.ts
```diff
@@ -305,6 +305,7 @@ function canDatabaseBeCollectedAndGetDeletedClientIDs(
 ): Promise<
   [canCollect: false] | [canCollect: true, deletedClientIDs: ClientID[]]
 > {
+  // TODO: Also see if there is a deleted-clients head and include that information.
```
arv, line 308: When doing that I think it would make sense to have Replicache be responsible of this head instead of zero dealing with it.

grgbkr:
> @grgbkr
> 
> ```
> // TODO: Maybe should send this as part of initConnection? It would be good to
> // delete the clients before we send down the results of the queries in case
> // some of those queries should be deleted.
> ```
> 
> I think this is the right thing to do. WDYT?

Agree, it would be good to communicate this as part of initConnection.

packages/zero-client/src/client/delete-clients-manager.ts
```diff
@@ -0,0 +1,148 @@
...
+  #listenToConnect(): void {
+    this.#getConnectPromise()
+      .then(() => this.#onConnect())
+      .catch(() => {});
+  }
+
+  #onConnect(): void {
+    this.#lc.debug?.('DeleteClientsManager: onConnect');
+    this.#sendTimerID = setTimeout(() => {
+      this.sendDeletedClientsToServer().catch(() => {});
```
grgbkr, line 61: Should we log something on error?

packages/zero-client/src/client/zero.ts
```diff
@@ -524,6 +529,13 @@ export class Zero<
     );
     this.#clientToServer = clientToServer(schema.tables);
 
+    this.#deleteClientsManager = new DeleteClientsManager(
+      () => this.#connectResolver.promise,
+      msg => this.#send(msg),
+      rep.memdag,
```
grgbkr, line 535: I think this should be the perdag.  

When working with the CLIENT_GROUPS_HEAD_NAME and CLIENTS_HEAD_NAME heads, replicache works directly with the perdag.

If we don't write this to the perdag, nothing currently will persist it to the perdag.  Persist only operates on the Commit chain, it doesn't persist all memdag heads to the perdag.

packages/replicache/src/persist/collect-idb-databases.ts
grgbkr: This does seem to fit more naturally in the Replicache layer.  It also would be a nice addition to Replicache (though we may never prioritize shipping it), as Replicache currently has a problem where the server can never correctly delete client records (https://www.notion.so/replicache/Deleting-Client-entries-from-backend-database-bdc4592792f94e7489507bba693c8359?pvs=4#bf968d1ca19b4f2e8d909d3350e353c5)

packages/zero-client/src/client/delete-clients-manager.ts
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * Zero calls this after it connects to ensure that the server knows about all
+   * the clients that might have been deleted locally since the last connection.
+   */
+  async sendDeletedClientsToServer(): Promise<void> {
+    const deletedClients = await withRead(this.#dagStore, dagRead =>
+      getDeletedClients(dagRead),
+    );
+    if (deletedClients.length > 0) {
+      this.#send(['deleteClients', {clientIDs: deletedClients}]);
```
grgbkr, line 99: Perhaps we should also send message informing the server of deleted client groups, this would allow it to cleanup CVR data scoped to client groups (though if all the clients in a client group are deleted, the remaining client group specific CVR data is quite minimal).  cc @darkgnotic

darkgnotic: We'll have to implement it before we do that.  😛 

https://github.com/rocicorp/mono/blob/7d142474fb1596833c293032bbfda48ffabcfd99/packages/zero-cache/src/workers/connection.ts#L219

Is now the time?

darkgnotic: yeah, the data should be minimal assuming the queries, and thus the rows are removed. i'd be fine with leaving it to garbage collection (which I do plan to add).

darkgnotic: Although the other thought is, if a client group becomes empty, we can just delete the full client group (i guess when the last client deletes itself?).

arv: I'll add that since we have that info

packages/zero-client/src/client/delete-clients-manager.ts
```diff
@@ -0,0 +1,91 @@
...
+    this.#lc = lc;
+  }
+
+  /**
+   * This gets called by Replicache when it deletes clients from the persistent
+   * storage.
+   */
+  async onClientsDeleted(clientIDs: ClientID[]): Promise<void> {
+    this.#lc.debug?.('onClientsDeleted:', clientIDs);
+    const deletedClients = await withWrite(this.#dagStore, async dagWrite => {
```
grgbkr, line 44: Maybe this part of updating the deleted clients in the perdag should be auto handled in the replicache layer.

packages/zero-client/src/client/delete-clients-manager.ts
```diff
@@ -0,0 +1,91 @@
...
+    this.#lc.debug?.('DeletedClientsManager, send:', deletedClients);
+    this.#send(['deleteClients', {clientIDs: deletedClients}]);
+  }
+
+  /**
+   * Zero calls this after it connects to ensure that the server knows about all
+   * the clients that might have been deleted locally since the last connection.
+   */
+  async sendDeletedClientsToServer(): Promise<void> {
+    const deletedClients = await withRead(this.#dagStore, dagRead =>
```
grgbkr, line 62: this.getDeletedClients();

packages/zero-client/src/client/delete-clients-manager.ts
grgbkr, line 44: Maybe Replicache should be adding the deleted clients to the perdag deleted clients automatically at its layer, and then calling the onClientDeleted callback to inform layers above that things have been added.

grgbkr (commented):
LGTM, very nice.

packages/zero-client/src/client/delete-clients-manager.ts
grgbkr: In the case that the server deletes it when the last client is deleted from the client group, it is possible a client will later connect with this server-deleted client group id (as it may not be gc'd client side yes, so new clients can be assigned to it).  So we'd have to be careful to handle this correctly.  

We don't necessarily want to GC client groups as soon as there are no clients in them (this just means all tabs have been closed).  A new tab opening should be able to boot strap from a recent client group.

darkgnotic: My understanding is that if there are no clients, then there are no desires and no queries and no rows. In that case, I think we should be able to delete the client group from the CVR.

> So we'd have to be careful to handle this correctly.

Yes, good point. I think the current view-syncer code will fail if a client comes in with a non-null base cookie but there's no CVR associated with the client (i.e. if it was deleted). So we'd have to handle that case better. Currently the browser will get an InvalidBaseCookie error, resulting in a browser refresh; not ideal, but at least not stuck.

arv: > In the case that the server deletes it when the last client is deleted from the client group, it is possible a client will later connect with this server-deleted client group id (as it may not be gc'd client side yes, so new clients can be assigned to it). So we'd have to be careful to handle this correctly.
> 
> We don't necessarily want to GC client groups as soon as there are no clients in them (this just means all tabs have been closed). A new tab opening should be able to boot strap from a recent client group.

OK. That settles that quuestion. I will add a clientGroupIDs array to the message as well.

PR #3710, opened 2025-02-05
fix(replicache): Collecting IDB with pending mutations...

arv:
Allow collecting IDB databases with pending mutations if mutation recovery is disabled.

This is important because in Zero client we have disabled mutation recovery and if we do not collect old IDB databases with pending mutations, we will never collect them.

### DIFF
--- a/packages/replicache/src/persist/collect-idb-databases.ts
+++ b/packages/replicache/src/persist/collect-idb-databases.ts
@@ -37,6 +37,7 @@ export function initCollectIDBDatabases(
   collectInterval: number,
   initialCollectDelay: number,
   maxAge: number,
+  enableMutationRecovery: boolean,
   onClientsDeleted: OnClientsDeleted,
   lc: LogContext,
   signal: AbortSignal,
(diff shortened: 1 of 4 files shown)


packages/replicache/src/persist/collect-idb-databases.ts
```diff
@@ -286,14 +299,21 @@ export function deleteAllReplicacheData(
  * `[false]`. Otherwise we return `true` and an array of the clientIDs to
  * remove.
  */
-function gatherPendingMutationsInClientGroups(
+function canDatabaseBeCollectedAndGetDeletedClientIDs(
```
grgbkr, line 302: there is similar logic in client-group-gc that we should update as well.  https://github.com/rocicorp/mono/blob/3399ecc55c86dfbd8a63e247b20c2da608f1c221/packages/replicache/src/persist/client-group-gc.ts#L52

arv: #3719

PR #3711, opened 2025-02-05
chore(zero-client): Fix flaky test

arv:
By retrying the assertion a few times.

### DIFF
--- a/packages/zero-client/src/client/zero.test.ts
+++ b/packages/zero-client/src/client/zero.test.ts
@@ -2220,12 +2220,12 @@ suite('Invalid Downstream message', () => {
 });
 
 async function tryNTimesAsync(n: number, fn: () => Promise<void>) {
-  for (let i = 0; i < n; i++) {
+  for (let i = 0; ; ) {
     try {
       await fn();
       return;
     } catch (e) {
-      if (i === n - 1) {
+      if (++i === n) {
         throw e;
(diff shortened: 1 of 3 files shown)


packages/zero-client/src/client/zero.test.ts
```diff
@@ -2253,10 +2266,11 @@ test('kvStore option', async () => {
 
     // Firefox is flaky... it takes longer time than Chromium and WebKit.
     // We therefore give it a few times to pass the expectation.
-
-    await tickAFewTimes(clock);
-    await new Promise(resolve => realSetTimeout(resolve, 100));
-    expect(allDataView.data).deep.equal(expectedValue);
+    await tryNTimesAsync(10, async () => {
```
tantaman, line 2269: Why doesn't `await r.mutate.e.insert({id: 'a', value: 1});` wait until the insertion has been persisted to IDB?  (line 2274)

arv: Because w3e write to the LazyStore... We actually have to wait for the persist process for it to become persistent. In Replicache tests we expose persist... That reminds me, we can get the ReplicacheImpl from testing and we can call persist on that.

tantaman: How much work is that? That would be ideal.

PR #3712, opened 2025-02-05
fix(zero-cache): more correct type parameters

darkgnotic:


packages/zero-schema/src/name-mapper.ts
```diff
@@ -89,18 +89,21 @@ export class NameMapper {
     return dst;
   }
 
-  row<R extends Row>(table: string, row: R): R {
+  row<V extends Value>(
```
tantaman, line 92: Not ideal (the return type becomes a union of all possible value types for each key) but I guess it doesn't matter that much since you don't care much about the types at this level.

CleanShot 2025-02-05 at 18 49 49@2x

https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABAJzgdwDwDVEFMAeUuYAJgM6JggC2ARrsogD6JlTIxgDmLtccAG1wBDJCyoCBAPgAU+AFyIASiJIIBATwwqIcZCQxsO3ADSIsUqQEpFK4WrCbtuXfsPtOXMxamIA3gBQiCi4UCDISPgA3AEAvkA

---

Given this layer doesn't seem to mind dropping the names of the keys --

```ts
row(table: string, row: Row): Row
```

would be better as @arv suggested.

darkgnotic: It was originally just `Row` in, `Row` out, but I was trying to handle this case:

https://github.com/rocicorp/mono/blob/7d142474fb1596833c293032bbfda48ffabcfd99/packages/zero-protocol/src/push.ts#L50

Where the `value` (Row) type is more constrained for 'delete' ops than for other ops, and I wanted ts to know that the converted `value` adheres to the (more constrained) inputted type. 

Then again, I casted away this particular case because I couldn't get typescript to understand this anyway (but I think for a different reason).

https://github.com/rocicorp/mono/blob/main/packages/zero-protocol/src/push.ts#L124

😮‍💨

arv (commented):
LGTM

packages/zero-schema/src/name-mapper.ts
arv: Yeah, this is better. It is not lying any more :-)

PR #3713, opened 2025-02-05
feat(zero-client): implement rebasing of custom mutators

tantaman:
I did not update crud mutators to write to the IVM source  during a rebase as the intent is to remove support for CRUD mutators when custom mutators are ready.

### DIFF
--- a/packages/zero-client/src/client/custom.test.ts
+++ b/packages/zero-client/src/client/custom.test.ts
@@ -1,8 +1,17 @@
-import {expect, expectTypeOf, test} from 'vitest';
+import {beforeEach, describe, expect, expectTypeOf, test} from 'vitest';
 import {schema} from '../../../zql/src/query/test/test-schemas.ts';
-import type {CustomMutatorDefs, MakeCustomMutatorInterfaces} from './custom.ts';
+import {
+  TransactionImpl,
+  type CustomMutatorDefs,
+  type MakeCustomMutatorInterfaces,
+  type Transaction,
+} from './custom.ts';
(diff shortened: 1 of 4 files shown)


packages/zero-client/src/client/custom.test.ts
```diff
@@ -1,8 +1,17 @@
...
-import type {CustomMutatorDefs, MakeCustomMutatorInterfaces} from './custom.ts';
+import {
+  TransactionImpl,
+  type CustomMutatorDefs,
+  type MakeCustomMutatorInterfaces,
+  type Transaction,
+} from './custom.ts';
 import {zeroForTest} from './test-utils.ts';
 import {nanoid} from '../util/nanoid.ts';
+import {createDb} from './ivm-source-repo.test.ts';
```
arv, line 11: Do not import from *.test.ts files because it will cause those beforeEach, test, describe etc to run when this test file is run.

Can you update our eslint config to disallow this?

Instead move test util functions to test-util.ts.

packages/zero-client/src/client/ivm-source-repo.test.ts
```diff
@@ -109,7 +109,7 @@ test('fork', () => {
 
 const lc = createSilentLogContext();
 let timestamp = 42;
-async function createDb(
+export async function createDb(
```
arv, line 112: Do not import from .test.ts

packages/zero-client/src/client/custom.test.ts
tantaman: https://github.com/rocicorp/mono/pull/3720

PR #3714, opened 2025-02-05
feat(zero-cache): allow multiple zero-cache shards to use the same CVR db

darkgnotic:
For dev instances, "preview" instances, and sharding, zero-caches with different `--shard-id` values can point to the same UPSTREAM_DB 

They can now also point to the same CVR_DB, as they will each operate in their own `cvr_{shard-id}` schema.

Data from the existing global `cvr` schema will be migrated the first time.

(Similar change for the CHANGE_DB is forthcoming)

https://bugs.rocicorp.dev/issue/3502

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/cvr-store.pg-test.ts
+++ b/packages/zero-cache/src/services/view-syncer/cvr-store.pg-test.ts
@@ -23,6 +23,8 @@ import {
 import {setupCVRTables, type RowsRow} from './schema/cvr.ts';
 import type {CVRVersion} from './schema/types.ts';
 
+const SHARD_ID = 'sdf';
+
 describe('view-syncer/cvr-store', () => {
   const lc = createSilentLogContext();
   let db: PostgresDB;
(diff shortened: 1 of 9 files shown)


packages/zero-cache/src/services/view-syncer/cvr-store.pg-test.ts
```diff
@@ -23,6 +23,8 @@ import {
 import {setupCVRTables, type RowsRow} from './schema/cvr.ts';
 import type {CVRVersion} from './schema/types.ts';
 
+const SHARD_ID = 'sdf';
```
arv, line 26: sdf?

arv (commented):
LGTM

packages/zero-cache/src/services/view-syncer/cvr-store.pg-test.ts
darkgnotic: qwerty  😛

PR #3718, opened 2025-02-06
wip: Aa/snake default

aboodman:


packages/zero-schema/src/builder/schema-builder.ts
```diff
@@ -56,12 +58,14 @@ export function createSchema<
     >['relationships'];
   };
 } {
+  const nameMapper = getNameMapper(options.nameMapping ?? 'snake');
```
darkgnotic, line 61: probably want the default to be 'none', so that we're not doing anything the user didn't ask for?

or do you think most users would expect us to map names to snake case? feels a bit magicky.

packages/zero-schema/src/name-mapper.ts
```diff
@@ -1,118 +1,22 @@
...
-      ? cols
-      : (cols.map(col => columns[col] ?? col) as unknown as Columns);
+export function getNameMapper(nameMapping: NameMapping): NameMapper {
+  switch (nameMapping) {
+    case 'snake':
+      return snakeCase;
+    case 'none':
+      return identity;
+    default:
+      return identity;
```
darkgnotic, line 20: I think this can be:

```ts
  default:
    unreachable(nameMapping);
```

darkgnotic (approved):
This is awesome!

PR #3719, opened 2025-02-06
fix(replicache): Add GC for client group without mutation recovery

arv:
When enableMutationRecovery is false we gc client groups that are not referenced by any client even if they have pending mutations.


packages/replicache/src/persist/client-group-gc.ts
```diff
@@ -37,19 +38,21 @@ export function initClientGroupGC(
 /**
  * This removes client groups that have no clients and no pending mutations.
```
grgbkr, line 39: Please update comment.

PR #3722, opened 2025-02-06
fix(zero-cache): work around bug re: generated columns in PG15

darkgnotic:
* Confirms that generated columns (`GENERATE ALWAYS AS (expr)`) are not replicated by PG.
* Confirms that identity columns (`GENERATED ... AS IDENTITY`) _are_ replicated by PG.
* Works around a bug in PG-15 whereby it incorrectly includes the former in pb_publication_tables.attnames. It won't replicate those columns, so they should not be included in attnames. (PG-16 and PG-17 do the right thing)


arv (commented):
LGTM

PR #3723, opened 2025-02-06
feat(zero-cache): push url configuration option

tantaman:
split this out into its own PR since there may be feedback on how we want to authenticate zero-cache to the user's API server.


packages/zero-cache/src/config/zero-config.ts
```diff
@@ -137,6 +137,21 @@ export const zeroOptions = {
...
   },
 
+  push: {
+    url: {
+      type: v.string().optional(), // optional until we remove CRUD mutations
+      desc: [
+        `The URL of the API server to which zero-cache will push mutations.`,
+      ],
+    },
+    apiKey: {
```
tantaman, line 147: a JWT/auth token will also be passed along for each `push` if a user is logged in. The api key is for added security in the case of logged out users.

darkgnotic: Is it passed in the `Authorization` header? I'm sure you'll document it somewhere, but it might be nice to put it here as well.

darkgnotic (commented):
LGTM

PR #3726, opened 2025-02-06
chore: Fix lint

arv:
We no longer allow importing from .test.ts files. This commit fixes all the lint errors.


tantaman (commented):
lgtm

PR #3727, opened 2025-02-06
chore(infra): 1gb zbugs docker 

cesara:
-  ./get-data.sh 
-  `docker compose down -v`
-  `docker compose -f ./docker-compose-1gb.yml up`
- probably want to limit issues to 100 

image


apps/zbugs/docker/1gb_upstream/init.sql
```diff
@@ -0,0 +1,379 @@
+DROP TABLE IF EXISTS "user",
```
tantaman, line 1: is it possible to share the 1gb upstream with zbugs upstream? I would have expected zbugs and rocinante to have the same schema but maybe I'm mistaken?

tantaman (approved):
lgtm

apps/zbugs/docker/1gb_upstream/init.sql
cesara: I DRYd it so we can share the same base.sql

PR #3728, opened 2025-02-07
feat(zero-cache): take 2: time-slice IVM for fairness and responsiveness

darkgnotic:
When running an IVM pipeline, yield the process every 500ms, preventing a long-running query from hogging the CPU. This should improve fairness as well as eliminate connection timeouts (from missed pongs).

The original implementation in #3651 did not work because the magic stopwatch was used incorrectly. 

The new implementation just uses Date.now().

Verified functionality thanks to @cesara's gigabugs dataset.

Two ViewSyncers (`clientGroupID`) are taking turns processing chunks of their pipelines in the same process (`pid`), and ping-pong is responsive all the while:

Screenshot 2025-02-06 at 16 03 02

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/view-syncer.ts
+++ b/packages/zero-cache/src/services/view-syncer/view-syncer.ts
@@ -3,7 +3,6 @@ import {Lock} from '@rocicorp/lock';
 import type {LogContext} from '@rocicorp/logger';
 import {resolver} from '@rocicorp/resolver';
 import type {JWTPayload} from 'jose';
-import {Stopwatch} from 'magic-stopwatch';
 import type {Row} from 'postgres';
 import {
   manualSpan,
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -961,8 +961,9 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
         span.setAttribute('totalRows', total);
       });
 
-      stopwatch.stop(RECORD_LAP);
-      return stopwatch.laps.reduce((total, lap) => total + lap.elapsed, 0);
```
tantaman, line 965: did the laps grow without out previously?

arv:
Yay! Less dependencies is always a plus!

packages/zero-cache/src/services/view-syncer/view-syncer.ts
darkgnotic: Previously, the stopwatch didn't even start.  😝 

So there was no yielding, as elapsed was always 0.

But, even after I fixed the code to start the stopwatch, the semantics of `lap.elapsed` is not what I thought it was. It's always "elapsed since the stopwatch started", and no "elapsed since the lap started", so this summation would have been wrong.

The magic-stopwatch has lost its magic. (Or maybe it was never there to begin with.  🤣 )

PR #3729, opened 2025-02-07
chore(zero-cache): Refactor putDesiredQueries

arv:
So that we can add ttl to it.


packages/zero-cache/src/services/view-syncer/cvr.ts
```diff
@@ -215,11 +216,11 @@ export class CVRConfigDrivenUpdater extends CVRUpdater {
 
   putDesiredQueries(
     clientID: string,
-    queries: {[id: string]: AST},
+    queries: Readonly<{hash: string; ast: AST}>[],
```
arv, line 219: In practice we always call this with a single query. The test calls this with 0, 1 and 2 queries. We could simplify to only support 0 or 1.

PR #3730, opened 2025-02-07
chore(zero-cache): use existing queue in `PusherService`

tantaman:
addresses https://github.com/rocicorp/mono/pull/3724#discussion_r1946270166 to use the existing queue class.


darkgnotic (commented):
LGTM

PR #3731, opened 2025-02-07
wip(zero): Add Time To Live -- TTL

arv:
- This adds a ttl option to useQuery. This then gets added to the `QueryImpl`.
- When set, it gets included in changeDesiredQueries messages.
- The `ttl`  and the `expiresAt`) are stored in the CVR.

### DIFF
--- a/packages/zero-client/src/client/query-manager.ts
+++ b/packages/zero-client/src/client/query-manager.ts
@@ -21,6 +21,13 @@ import type {ReadTransaction} from './replicache-types.ts';
 
 type QueryHash = string;
 
+type QueryEntry = {
+  normalized: AST;
+  count: number;
+  gotCallbacks: GotCallback[];
+  ttl: number | undefined;
+};
+
 /**
  * Tracks what queries the client is currently subscribed to on the server.
  * Sends `changeDesiredQueries` message to server when this changes.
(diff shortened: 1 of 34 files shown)


packages/zero-client/src/client/query-manager.ts
```diff
@@ -109,9 +114,18 @@ export class QueryManager {
...
-    for (const [hash, {normalized}] of this.#queries) {
+    for (const [hash, {normalized, expiresAt}] of this.#queries) {
       if (!existingQueryHashes.has(hash)) {
-        patch.set(hash, {op: 'put', hash, ast: normalized});
+        let ttl: number | undefined;
+        if (expiresAt !== undefined) {
+          ttl = expiresAt - now;
+        }
+        patch.set(hash, {
+          op: 'put',
```
tantaman, line 124: I find this a bit confusing.

I'd expect the TTL to be "forever" until the query is unmounted and then it gets set to a value. The reason being that we don't ever want to expire a query that is actively in use.

arv: The intention is that active queries will never be removed. When they are "removed", the ttl will kick in.

The intended (future) algorithm is something like this:
- First send all active queries, keep track of number of rows sent. 
- Then sort the remaining queries based on ttl. No ttl means as long as there is space so that comes before a query with ttl.
- then send queries as long as there is more space. 
- once the space limit is reached, finish sending the current query. 
- delete the rest of the queries.

Maybe this is too confusing? I guess we could mean that the ttl kicks in when deactivating?

PR #3735, opened 2025-02-07
feat(zero-protocol): client-side of revised-cookie poke protocol

darkgnotic:
Adds an optional `cookie` field to `pokeEnd` which overrides the value that was specified in `pokeStart`.

(In the future, we should considering moving the field to only be in `pokeEnd`.)

The motivation for this change is to allow the `view-syncer` to avoid CVR updates for IVM pushes that don't change the CVR (thereby avoiding the quadratic writes to the CVR db). 

When this happens, the associated pokes can either be:
1. cancelled if nothing was sent, or
2. completed (e.g. if catchup rows were sent) with a revised cookie reflecting what is stored in the CVR

### DIFF
--- a/packages/zero-client/src/client/zero-poke-handler.test.ts
+++ b/packages/zero-client/src/client/zero-poke-handler.test.ts
@@ -1311,6 +1311,9 @@ test('mergePokes with all optionals defined', () => {
             ],
           },
         ],
+        pokeEnd: {
+          pokeID: 'poke1',
+        },
       },
       {
         pokeStart: {
@@ -1346,6 +1349,9 @@ test('mergePokes with all optionals defined', () => {
             ],
           },
         ],
+        pokeEnd: {
+          pokeID: 'poke2',
+        },
       },
     ],
     schema,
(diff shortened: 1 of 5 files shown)


packages/zero-client/src/client/zero-poke-handler.test.ts
```diff
@@ -1601,6 +1613,64 @@ test('mergePokes sparse', () => {
...
+          pokeStart: {
+            pokeID: 'poke1',
+            baseCookie: '3',
+            cookie: '5',
+            schemaVersions: {minSupportedVersion: 1, maxSupportedVersion: 1},
+          },
+          parts: [
+            {
+              pokeID: 'poke1',
+              lastMutationIDChanges: {c1: 1, c2: 2},
```
grgbkr, line 1630: This doesn't seem like a valid poke.  How can there be state changes but the cookie not change?  Why would these lastMutationIDChanges be sent for poke1 starting from baseCookie 3 but not for poke2 also starting from baseCookie 3?

darkgnotic: Ah good point. I should change the test to return `cookie: '4'` or something. This would essentially mean: "please advance your base cookie but to 4 instead of 5".

I think the protocol will also be easier to understand if we only send the `cookie` at the pokeEnd (and remove it from pokeStart). I plan to migrate to that protocol unless you see an issue with it.

PR #3736, opened 2025-02-08
feat(zero-cache): eliminate quadratic CVR writes

darkgnotic:
System scalability improvement which eliminates CVR db updates and pokes for replica changes that do not affect the client group. 

This mitigates the quadratic nature of replication processing by pruning processing branches before writing to the CVR postgres DB and downstream clients, limiting the blast radius to IVM push processing. 

* The CVR pruning avoids a flush if there are no pending writes. Note that this means that we will update the `lastActive` field of a CVR less often, but that only affects garbage collection.
* The poke pruning involves buffering the initial `pokeStart` until a subsequent message is sent, and avoiding sending anything at all if the `finalVersion` passed to `pokeEnd` is equal to the client's `baseCookie`

Note that both features rely on the client correctly understanding the revised poke protocol in #3735, and thus the features are only enabled if all of the connected clients are up to `PROTOCOL_VERSION = 5`. If older clients are connected, the pruning does not happen and the previous behavior is invoked. 

In a few releases we can decommission the old versions and make the pruning behavior unconditional.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/client-handler.ts
+++ b/packages/zero-cache/src/services/view-syncer/client-handler.ts
@@ -64,7 +64,7 @@ export type PatchToVersion = {
 export interface PokeHandler {
   addPatch(patch: PatchToVersion): void;
   cancel(): void;
-  end(): void;
+  end(finalVersion: CVRVersion): void;
 }
 
 const NOOP: PokeHandler = {
@@ -78,7 +78,7 @@ const NOOP: PokeHandler = {
 const PART_COUNT_FLUSH_THRESHOLD = 100;
 
 /**
- * Handles a single {@link ViewSyncer.sync()} connection.
+ * Handles a single `ViewSyncer` connection.
  */
 export class ClientHandler {
   readonly #clientGroupID: string;
@@ -88,6 +88,7 @@ export class ClientHandler {
   readonly #lc: LogContext;
   readonly #pokes: Subscription<Downstream>;
   #baseVersion: NullableCVRVersion;
+  readonly #protocolVersion: number;
   readonly #schemaVersion: number;
 
   constructor(
@@ -97,6 +98,7 @@ export class ClientHandler {
     wsID: string,
     shardID: string,
     baseCookie: string | null,
+    protocolVersion: number,
     schemaVersion: number,
     pokes: Subscription<Downstream>,
   ) {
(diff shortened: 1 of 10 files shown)


aboodman:
Did this come up from a customer issue, or just something that was
bothering you?

darkgnotic:
> Did this come up from a customer issue, or just something that was bothering you?
> 

Something that has bothered me for a long time. I had put it off because it doesn't make much of a difference in an app like zbugs where everyone kind of sees the same thing, but I anticipate that it will be an issue with the types of apps our users are building (email clients, customer dashboards, etc.).

arv (commented):
LGTM

packages/zero-cache/src/services/view-syncer/client-handler.ts
```diff
@@ -222,12 +236,24 @@ export class ClientHandler {
...
-        this.#pokes.push(['pokeEnd', {pokeID, cancel: true}]);
+        if (pokeStarted) {
+          this.#pokes.push(['pokeEnd', {pokeID, cancel: true}]);
+        }
       },
 
-      end: () => {
+      end: (finalVersion: CVRVersion) => {
+        const cookie = versionToCookie(finalVersion);
+        if (!pokeStarted) {
```
grgbkr, line 246: It is possible for 
`pokeStarted && cmpVersions(this.#baseVersion, finalVersion) === 0`?

If so should we send a cancel in this case?

darkgnotic: 🤔 

I think it should be impossible.

`addPatch()` will ignore any patches that are before or equal to the `#baseVersion`:

```ts
    const addPatch = (patchToVersion: PatchToVersion) => {
      const {patch, toVersion} = patchToVersion;
      if (cmpVersions(toVersion, this.#baseVersion) <= 0) {
        return;
      }
      const body = ensureBody();
```

and so `ensureBody()` would have never been called, and `pokeStarted` would always be false.

If it _does_ happen, it means we somehow sent a patch that is greater than the `#baseVersion`, and then set the `finalVersion` to something earlier than the patch, which would be a bug in our logic.

It might be worth adding an `assert()` to this effect though, as a sanity check. wdyt?

grgbkr: Assert sgtm.

grgbkr:
LGTM

PR #3737, opened 2025-02-08
feat(zero-cache): log the server version at startup

darkgnotic:
Logs the server version at startup (when using the Docker image) to identify what's running in production. This should facilitate debugging client/server protocol mismatches such as https://discord.com/channels/830183651022471199/1337466220894617650/1337866243730051113

Screenshot 2025-02-08 at 13 37 45


darkgnotic:
@cesara this doesn't quite do what we want for our Github bulids:

Screenshot 2025-02-08 at 13 54 02

Would it be straightforward to pass the exact filename into the CD script so that it shows up in the logs?

Not a big deal. Just nice to have on the zbugs side too.

PR #3739, opened 2025-02-10
fix(zql,z2s): place `hidden` flag on the correct relationship

tantaman:
`query-impl` put the `hidden` flag at the final table in a junction edge, which was incorrect. ZQL did the right thing because `join` moved the hidden flag up a level incorrectly. Two errors made a right in this case.

Concretely:
- query impl would hide the `label` table
- join would move the `hidden` flag back up to the `issue-label` table.

The z2s compiler, however, couldn't work properly with the `hidden` flag being placed in the wrong level by `query-impl`


grgbkr:
LGTM.  Good find.

PR #3740, opened 2025-02-10
chore(zero, replicache): Let replicache manage the deleted clients head

arv:
Now, replicache writes the deleted clients head as clients are deleted.

This was previously managed by zero since replicache does not really need this information.


grgbkr:
LGTM.  nice.

PR #3742, opened 2025-02-10
feat(zero-cache): treat `numeric` as `number` 

tantaman:
ZQL returns `numeric` as a number.
PG returns it as a `string`.

Change `pg` to return it as a number if it fits JS precision.

If we go this route, `table-source` should be update to convert to `string` on an overflow as well.

### DIFF
--- a/packages/z2s/src/test/chinook/chinook.pg-test.ts
+++ b/packages/z2s/src/test/chinook/chinook.pg-test.ts
@@ -39,13 +39,12 @@ let artistQuery: Query<Schema, 'artist'>;
 let customerQuery: Query<Schema, 'customer'>;
 let employeeQuery: Query<Schema, 'employee'>;
 let genreQuery: Query<Schema, 'genre'>;
+let invoiceQuery: Query<Schema, 'invoice'>;
+let invoiceLineQuery: Query<Schema, 'invoice_line'>;
 let mediaTypeQuery: Query<Schema, 'media_type'>;
 let playlistQuery: Query<Schema, 'playlist'>;
 let playlistTrackQuery: Query<Schema, 'playlist_track'>;
-// TODO: buggy tables
-// let invoiceQuery: Query<Schema, 'invoice'>;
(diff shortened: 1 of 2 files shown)


packages/z2s/src/test/chinook/chinook.pg-test.ts
```diff
@@ -82,12 +81,12 @@ describe('basic select', () => {
...
-    // ['invoice', () => invoiceQuery], --> total is showing up as a string
-    // ['invoice_line', () => invoiceLineQuery], --> numeric columns are showing up as strings
+    ['invoice', () => invoiceQuery],
+    ['invoice_line', () => invoiceLineQuery],
     ['media_type', () => mediaTypeQuery],
     ['playlist', () => playlistQuery],
-    // ['playlist_track', () => playlistTrackQuery], --> this is not sorting correctly between zql and pg
-    // ['track', () => trackQuery], --> numeric columns are showing up as strings
+    ['playlist_track', () => playlistTrackQuery],
+    // ['track', () => trackQuery], // some tracks are coming back in the wrong position!
```
tantaman, line 89: track is broken for some other reason now.

darkgnotic:
Eek. I don't know what the right answer is. I think returning different js types for a single data type is weird. I would recommend one of the following semantics: 
* Return number, and error when overflowing max-safe-integer (like we do with ints)
* Sticking with string, since the type is intended for "arbitrary precision"

This is really a dx decision. Perhaps @aboodman has thoughts?

aboodman:
I also really dislike using different js types for same sql type dependent
on size.

I think the most dx friendly options for these types is to default to js
number since that is commonly going to work and be easiest to use and most
efficient, but give used option (in zero schema) to override and map to
some other type.

Not sure about api design for mapping. There are some choices in there:
should it error if narrowing type? Should users be able to provide their
own converters? Etc

For a short term thing I prefer number always, and just lose precision, and
pg should match zql.

a (phone)

PR #3744, opened 2025-02-10
chore(zero-cache): derive the server-side schema from the replica

darkgnotic:
For the purpose of constructing queries from ASTs (or fragments thereof), derive a `Schema` object from the replica itself rather than depending on the developer-written client-side schema.

This currently happens in:
* the WriteAuthorizer (going away with custom mutators)
* the `run-query` script (used for debugging queries)

This also obsoletes the method that converted client-side schemas to server names.


packages/zero-cache/src/auth/load-schema.ts
```diff
@@ -32,13 +27,25 @@ export function getSchema(config: ZeroConfig): Promise<{
...
+        tableSpec: {name, primaryKey},
+        zqlSpec: columns,
+      } = table;
+      return [name, {name, columns, primaryKey} satisfies TableSchema];
+    }),
+  );
+  return {
+    version: 1, // only used on the client-side
+    tables,
+    relationships: {}, // relationships are already denormalized in ASTs
```
tantaman, line 49: We should create a new type as a downstream user of this type may not understand all the caveats around it.

"Yes, it is the schema except when loaded on the server then relationships are missing and version is incorrect"

```ts
type TableSchemas = {
  tables: {readonly [table: string]: TableSchema}
};
```

darkgnotic: I like that. Will do!

PR #3745, opened 2025-02-10
chore(zero-cache): add sanity check, fix test

darkgnotic:
Adds a sanity check that the finalVersion always advances if pokes were sent.

Fix unit test to advance the cookie since a poke was sent (otherwise it is exercising a scenario that should not be possible)


grgbkr:
LGTM!

PR #3746, opened 2025-02-10
fix(zql): join should add relationship to child change pushes

grgbkr:
This impacted queries with multiple whereExist conditions.  The missing relationships on child changes would cause an assertion error in the Exist operator [here](https://github.com/rocicorp/mono/blob/23b57636e7bf9bc51c933e87d3dcfff63cf73843/packages/zql/src/ivm/exists.ts#L277).  This bug was introduced in https://github.com/rocicorp/mono/pull/3629

Issue reported on Discord https://discord.com/channels/830183651022471199/1288232858795769917/1338368796825620531

### DIFF
--- a/packages/zql/src/query/query-impl.query.test.ts
+++ b/packages/zql/src/query/query-impl.query.test.ts
@@ -1199,6 +1199,85 @@ test('where exists', () => {
   expect(materialized.data).toEqual([]);
 });
 
+test('duplicative where exists', () => {
+  const queryDelegate = new QueryDelegateImpl();
+  const issueSource = must(queryDelegate.getSource('issue'));
+  const labelSource = must(queryDelegate.getSource('label'));
+  const issueLabelSource = must(queryDelegate.getSource('issueLabel'));
+  issueSource.push({
+    type: 'add',
+    row: {
+      id: '0001',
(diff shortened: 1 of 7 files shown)


packages/zql/src/query/query-impl.query.test.ts
```diff
@@ -1199,6 +1199,85 @@ test('where exists', () => {
...
+    row: {
+      id: '0001',
+      name: 'bug',
+    },
+  });
+
+  const materialized = newQuery(queryDelegate, schema, 'issue')
+    .where('closed', true)
+    .whereExists('labels', q => q.where('name', 'bug'))
+    .whereExists('labels', q => q.where('name', 'bug'))
```
tantaman, line 1238: huh. Don't these end up with the same relationship name? I'm surprised there are not other bugs with this.

grgbkr: This logic in builder assures they end up with unique aliases, https://github.com/rocicorp/mono/blob/77b2edf0f3e69a7b16483c783f37e813c656ee94/packages/zql/src/builder/builder.ts#L389

Though, I'm not sure what happens if you duplicate related...

PR #3750, opened 2025-02-10
feat(zero-cache): handle numeric in the replication stream

darkgnotic:
Followup to #3742: handle the `numeric` type in the replication stream (which is the domain of the `pg` client) in the same way that it is handled during initial-sync (which is the domain of the `postgres.js` client).


tantaman (commented):
thanks

PR #3751, opened 2025-02-10
chore(zero-cache): move debugging options into debugging scripts

darkgnotic:
Move the options relevant to each script into the script itself. This should make it more sustainable to add other scripts (like an up-and-coming `deploy-permissions` script) without making the main server options a nexus.


tantaman (commented):
lgtm

PR #3752, opened 2025-02-10
chore(zero, replicache): include client group ids in deleted clients

arv:
Now we store `{clientIDs, clientGroupIDs}` in the deleted clients head and update these as needed.

When sending `deletedClients` (as part of connection, initConnection etc) we send both client IDs and client group IDS.

### DIFF
--- a/packages/zero-protocol/src/connect.ts
+++ b/packages/zero-protocol/src/connect.ts
@@ -1,4 +1,5 @@
 import * as v from '../../shared/src/valita.ts';
+import {deleteClientsBodySchema} from './delete-clients.ts';
 import {queriesPatchSchema} from './queries-patch.ts';
 
 /**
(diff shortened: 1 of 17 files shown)


packages/zero-protocol/src/connect.ts
```diff
@@ -21,7 +22,7 @@ export const connectedMessageSchema = v.tuple([
 
 const initConnectionBodySchema = v.object({
   desiredQueriesPatch: queriesPatchSchema,
-  deletedClients: v.array(v.string()).optional(),
+  deleted: deleteClientsBodySchema.optional(),
```
darkgnotic, line 25: FYI this needed a PROTOCOL_VERSION bump. 

https://discord.com/channels/830183651022471199/1346879106616135845/1346915202398027858

I will add a test to notifies us of this when the wire protocol changes (currently we're only testing AST changes).

arv: Sorry

arv: Actually, should this not have worked since they are optional?

darkgnotic: We use strict parsing, so if the server sees an unknown field (in this case, `deletedClients`), it will reject it (because it does not know how to interpret it).

I think it's safer this way ... with "passthrough", the server may not necessarily understand how to correctly handle new fields.

PR #3755, opened 2025-02-11
feat(zero-cache): upstream zero.permissions table and deploy tool

darkgnotic:
Adds an upstream-global (i.e. unsharded) `zero.permissions` table to hold the permissions JSON object, and an `npx zero-deploy-permissions` tool to deploy it to an upstream postgres.

The table is not yet consulted; that is coming soon, along with integration with `npx zero-cache-dev`.

### Deploy DX

`--help`:

Screenshot 2025-02-10 at 16 49 39

In the directory with `.env`:

Screenshot 2025-02-10 at 16 51 24

With an `--output-file`:

Screenshot 2025-02-10 at 16 51 55

### DIFF
--- a/packages/zero-cache/src/scripts/deploy-permissions.ts
+++ b/packages/zero-cache/src/scripts/deploy-permissions.ts
@@ -1,0 +1,158 @@
+import {consoleLogSink, LogContext} from '@rocicorp/logger';
+import 'dotenv/config';
+import {writeFile} from 'node:fs/promises';
+import {basename, dirname, join, relative, resolve} from 'node:path';
+import {fileURLToPath} from 'node:url';
+import {tsImport} from 'tsx/esm/api';
+import {parseOptions} from '../../../shared/src/options.ts';
+import * as v from '../../../shared/src/valita.ts';
+import {
+  permissionsConfigSchema,
+  type PermissionsConfig,
+} from '../../../zero-schema/src/compiled-permissions.ts';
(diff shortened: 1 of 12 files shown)


packages/zero-cache/src/services/change-source/custom/change-source.ts
```diff
@@ -205,6 +205,10 @@ function getRequiredTables(
       lastMutationID: {type: 'number'},
       userID: {type: 'string'},
     },
+    [`zero.permissions`]: {
```
darkgnotic, line 208: @cbnsndwch FYI, there's a new upstream table requirement.

Going forward I would love to have your thoughts on how we can/should handle these types of upstream schema changes.

packages/zero-cache/src/scripts/deploy-permissions.ts
```diff
@@ -0,0 +1,158 @@
...
+if (config.output.file) {
+  await writeFile(
+    config.output.file,
+    JSON.stringify(permissions, null, config.output.pretty ? 2 : 0),
+  );
+  lc.info?.(`Wrote permissions JSON to ${config.output.file}`);
+} else if (config.upstream.db) {
+  await deployPermissions(lc, config.upstream.db, permissions);
+} else {
+  lc.error?.(`No --output-file or --upstream-db specified`);
```
tantaman, line 155: nit: throw if both options are specified too

PR #3757, opened 2025-02-11
fix(zero-client): Allow overlapped mutators.

aboodman:
Fixes https://bugs.rocicorp.dev/issue/3529


arv:
I would like to see/test what the push body looks like. All the calls to mutate inside a batch should be batched. I think this is where the problems show up. The CRUD mutators you get through the mutateBatch argument all get grouped into a single CRUD mutation. If you use z.mutate in there you will get multiple mutations and the order is probably not what you expect.

PR #3758, opened 2025-02-11
fix(shared): Immutable type too deep

arv:
We were getting:

```
Type instantiation is excessively deep and possibly infinite.ts(2589)
```

When a ReadonlyArray was passed into `Immutable`. The reason was that we were using `extends Array<infer U>`, but ReadonlyArray does not extend Array.


tantaman:
lgtm

PR #3763, opened 2025-02-11
feat(zero-cache): update `dispatcher` to be able to hand-off to `mutator`

tantaman:
Follows the same pattern as `syncer` to set up a `mutator` worker that will receive websocket connections from the dispatcher.

### DIFF
--- a/packages/zero-cache/src/server/main.ts
+++ b/packages/zero-cache/src/server/main.ts
@@ -147,6 +147,10 @@ export default async function runWorker(
     }
     syncers.forEach(syncer => handleSubscriptionsFrom(lc, syncer, notifier));
   }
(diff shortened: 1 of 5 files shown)


packages/zero-cache/src/server/main.ts
```diff
@@ -147,6 +147,10 @@ export default async function runWorker(
     }
     syncers.forEach(syncer => handleSubscriptionsFrom(lc, syncer, notifier));
   }
+  let mutator: Worker | undefined;
+  if (config.push.url !== undefined) {
```
tantaman, line 151: custom mutators are optional until we are ready to release them and deprecate crud mutators.

PR #3764, opened 2025-02-11
feat(zero-cache): publish zero.permissions table, use a trigger for hash

darkgnotic:
Add the `zero.permissions` table to the shard's publication, and use a trigger to set the `hash` value instead of a generated columns. (Postgres does not replicated generated columns)


packages/zero-cache/src/integration/integration.pg-test.ts
```diff
@@ -284,6 +284,18 @@ const INITIAL_CUSTOM_SETUP: ChangeStreamMessage[] = [
...
+  [
+    'data',
+    {
+      tag: 'insert',
+      relation: {
+        schema: 'zero',
+        name: 'permissions',
+        keyColumns: ['lock'],
+      },
+      new: {lock: true, permissions: null, hash: null},
```
darkgnotic, line 296: FYI, @cbnsndwch, this is one more requirement of upstreams, that there exist exactly one row in the permissions table.

At some point we'll have to document all of this. Thanks for living on the bleeding edge!

PR #3768, opened 2025-02-11
feat(zero): Flip default permissions to deny

aboodman:


packages/zero-schema/src/permissions.ts
```diff
@@ -16,8 +16,13 @@ import type {
 } from './compiled-permissions.ts';
 import {clientToServer, NameMapper} from './name-mapper.ts';
 
-export const ANYONE_CAN = undefined;
+export const ANYONE_CAN = [
+  (authData: {sub: string}, eb: ExpressionBuilder<Schema, never>) =>
+    // TODO: Could just add a true() predicate?
+    eb.cmpLit(authData.sub, 'NOT IN', []),
```
tantaman, line 22: nit: `eb.cmpLit(true, true)` will work

```
npm run run-query -- --debug-query 'z.query.issue.where(({cmpLit}) => cmpLit(false, "=", true))'
> zbugs@0.0.0 run-query
> npm run zero-build-schema && tsx ../../packages/zero-cache/src/scripts/run-query.ts --debug-query z.query.issue.where(({cmpLit}) => cmpLit(false, "=", true))


> zbugs@0.0.0 zero-build-schema
> tsx ../../packages/zero-schema/src/build-schema.ts

=== Query Stats: ===

issue vended: []
total rows considered: 0
time: 1.45ms ms


=== Query Plans: ===
```

packages/zero-cache/src/auth/write-authorizer.ts
```diff
@@ -436,13 +428,17 @@ export class WriteAuthorizerImpl implements WriteAuthorizer {
     return true;
   }
 
+  /**
+   * Defaults to *false* if the policy is empty. At least one rule has to pass
+   * for the policy to pass.
+   */
   #passesPolicy(
     policy: Policy | undefined,
```
tantaman, line 436: cell policies also come in through here which have the opposite default.

packages/zero-schema/src/permissions.ts
arv: would `eb.and()` work?

I tried running npm run run-query to check but it does not work for me:

image

tantaman: I think that would work too.

That's really strange that `where` is undefined for you. Maybe it has to do with the updates that remove `schema.json`? Since the schema is required for `run-query` to create the query interface.

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -609,16 +609,13 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
       }
       const {query: transformedAst, hash: newTransformationHash} =
         transformAndHashQuery(
+          lc,
           ast,
-          must(this.#permissions).permissions,
+          // TODO(darick): confirm this change is OK
```
darkgnotic, line 614: `this.#permissions` is a `LoadedPermissions` object, and its state of being null or defined represents whether we've loaded permissions. So we want to `must()` only that, essentially asserting that the calling code has already loaded permissions.

The `permissions` field _inside_ of `LoadedPermissions`, otoh, can be `null`, which means that the user has not yet deployed any permissions at all. (i.e. we loaded the permissions, and they were `null`).

How you want to deal with that is up to you. You can do something like:

```ts
must(this.#permissions).permissions ?? DEFAULT_PERMISSIONS,
```

Or have the ReadAuthorizer accept `null` (again) as the PermissionsConfig and default it in there.

packages/zero-cache/src/services/mutagen/mutagen.ts
```diff
@@ -263,6 +263,8 @@ async function processMutationWithTx(
     }
   }
 
+  authorizer.reloadPermissions();
```
darkgnotic, line 266: Thank you for fixing this.  🙏

packages/zero-cache/src/auth/write-authorizer.ts
aboodman: Yeah but the caller handles this by only calling `passesPolicy` in the cases where a cell policy actually exists.

packages/zero-schema/src/permissions.ts
aboodman: Done! Super elegant, I love it. And I used `or([])` for `false`.

PR #3769, opened 2025-02-11
chore(zero-cache): use the schema load logic in deploy-zero-permissions

darkgnotic:
Have the `transform-query` script use the schema loading logic from `deploy-permissions`, to remove a dependency on the soon-to-go-away schema.json logic.


tantaman (approved):
lgtm

PR #3775, opened 2025-02-12
fix(zero-cache-dev): better error message during redeploy failures

darkgnotic:
An error can happen during a redeploy. This doesn't stop an already running server.


grgbkr:
LGTM

PR #3783, opened 2025-02-12
feat(zero-deploy-permissions): validate rules before deploying to upstream

darkgnotic:
Screenshot 2025-02-12 at 11 58 12


tantaman (commented):
Nice. 

An alternative would be to validate the schema itself rather than the permission rules.

- The permission rules cannot refer to things not in the schema so validating the schema validates the rules.
- If the schema refers to missing tables and columns, a user would want to know that too as the queries in their app will break at runtime

PR #3784, opened 2025-02-12
feat(zero-cache): avoid applying application permissions to internal queries

darkgnotic:
This never mattered before, but will matter when permissions are flipped to disallow by default.


tantaman (commented):
lgtm

PR #3792, opened 2025-02-13
chore(zero): Remove syncing of connected clients

arv:
We used to sync the clientIDs of the connected clients. The idea was that we would use this information to determine which clients we cannot delete. However, this information is not used anywhere in the codebase and is not necessary. This commit removes the syncing of the clientIDs.

This means that we can remove the clientPatches that we send.

Since we are not sending client patches we can also remove this state from the CRV.


packages/zero-cache/src/services/view-syncer/schema/init.ts
```diff
@@ -63,14 +63,23 @@ export async function initViewSyncerSchema(
...
       await tx`
-      ALTER TABLE ${tx(schema)}."rowsVersion" 
+      ALTER TABLE ${tx(schema)}."rowsVersion"
         DROP CONSTRAINT fk_rows_version_client_group`;
     },
   };
 
+  const migrateV6ToV7: Migration = {
+    migrateSchema: async (_, tx) => {
+      await tx`DROP INDEX IF EXISTS client_patch_version`;
```
arv, line 76: I'm not sure why I need `IF EXISTS` here?

packages/zero-cache/src/services/view-syncer/schema/init.ts
```diff
@@ -63,14 +63,23 @@ export async function initViewSyncerSchema(
...
+      await tx`ALTER TABLE ${tx(schema)}.clients DROP COLUMN deleted`;
+      await tx`ALTER TABLE ${tx(schema)}.clients DROP COLUMN "patchVersion"`;
+    },
+    minSafeVersion: 7,
```
arv, line 80: @darkgnotic Is this migration code correct? Once the columns have been deleted v6 will not work.

grgbkr:
LGTM, I'll let @darkgnotic review the migration code.

packages/zero-cache/src/services/view-syncer/schema/init.ts
darkgnotic: True, not needed. I agree that removing it is safer in that it would flag a typo, for instance.

packages/zero-cache/src/services/view-syncer/schema/init.ts
darkgnotic: Yes, nicely done!

darkgnotic (approved):

arv:
I was getting errors when not including `IF EXISTS`. I will try again and see if I can track down how I got into that state.

packages/zero-cache/src/services/view-syncer/schema/init.ts
darkgnotic: Actually, thinking about this a bit more, I think we need to do this in two steps.

The CVR is read by view-syncers, for which there can be many instances, and some people (like Izak) will canary view-syncers with a new version at the same time as running old version (or go back and forth).

So bumping from 6 to 7 without allowing 6 to run will break them. This was similar to a mistake I made here:

https://discord.com/channels/830183651022471199/1339066868245467157/1339074101079441460

So I think the way to roll something like this out is to first not use the columns, but allow older versions to use them, (version=7, minVersion=6), and then in a subsequent release roll out the change that removes the columns (version=8, minVersion=7).

darkgnotic: This means, I think in 7, the schema change is to do:

```sql
ALTER TABLE ${tx(schema)}.clients ALTER COLUMN "patchVersion" DROP NOT NULL;
```

And stop reading / writing it in the code.

Once everyone's on 7, we can then drop the columns and indexes when bumping to 8 and moving minSafeVersion to 7.

arv: Looking at this.

It seems better to keep the schema and data as is.

If I make the patchVersion nullable old code that is running may get a null patchVersion when that isn't expected.

Instead, I'm removing all the read code paths and once we feel certain that there are no more reads out there we can remove the write paths too.

PR #3797, opened 2025-02-13
feat(zero-permissions): add "type":"module" recommendation when schema load fails

darkgnotic:
Also make any exceptions more readable by throwing them to the console (rather than logging them as objects).


grgbkr:
LGTM

grgbkr:
Thanks!

PR #3804, opened 2025-02-13
feat(zero-pg): implement `createPushHandler` to process mutations on the user's API server

tantaman:
For a higher level overview see: https://www.notion.so/replicache/User-s-Push-Endpoint-1983bed895458069a83ee2ab73d54b0f

The basic idea is that the `PushHandler` will be wired into the user's express/fastify/hono/cloudflare/next/whatever server.

```ts
app.post('/api/zero-push', createPushHandler({
  schema,
  mutators,
  shardID,
  dbConnectionProvider
}));
```

The one surprise I had while implementing this is the need to pass a `shardID` to the push handler. How would a user determine what to set this value to? cc @darkgnotic


packages/zero-cache/src/services/change-source/pg/schema/shard.ts
```diff
@@ -87,7 +87,17 @@ export async function ensureGlobalTables(db: PostgresDB) {
   await db.unsafe(GLOBAL_SETUP);
 }
 
-function shardSetup(shardID: string, publications: string[]): string {
+export function getClientsTableDefinition(schema: string) {
```
tantaman, line 90: exported so I can use this in `web.pg-test.ts`

packages/zero-pg/src/custom.ts
```diff
@@ -27,14 +31,66 @@ export type CustomMutatorImpl<S extends Schema, TDBTransaction> = (
...
 
+type Options<
+  S extends Schema,
+  TDBTransaction,
+  MD extends CustomMutatorDefs<S, TDBTransaction>,
+> = {
+  schema: S;
+  dbConnectionProvider: ConnectionProvider<TDBTransaction>;
+  mutators: MD;
+  shardID?: string;
```
tantaman, line 42: 🤔

packages/zero-pg/src/custom.ts
```diff
@@ -27,14 +31,66 @@ export type CustomMutatorImpl<S extends Schema, TDBTransaction> = (
...
+    this.dbTransaction = dbTransaction;
+    this.token = token;
+    this.clientID = clientID;
+    this.mutationID = mutationID;
+
+    // TODO: implement both of these.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.mutate = {} as any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.query = {} as any;
```
tantaman, line 92: to be done in a follow up

PR #3809, opened 2025-02-13
feat(prod): fully migrate to SST for permissions deployment

darkgnotic:
Use SST for deploying permissions, removing the redundant GitHub actions.

Also, set `Service.wait` to `false` by default to keep deployments fast in the common case.

We only need to set them to `true` for the occasional protocol / AST version upgrade, which we can do manually (in the same PR as the protocol change).


cesara (commented):
lgtm

PR #3813, opened 2025-02-14
feat(zero-cache): Implement server side of deleteClients

arv:
When the server receives a `DeleteClients` message, it will delete the client (and client group) from the CRV database. This will send `'desiredQueriesPatches'` patches to connected
clients that are affected by the deletions.

The server also confirms the deletion by sending a `'deleteClients'` back to the connected clients that requested the deletion.

Towards https://bugs.rocicorp.dev/issue/3476


packages/zero-cache/src/services/view-syncer/view-syncer.pg-test.ts
```diff
@@ -882,6 +1466,147 @@ describe('view-syncer/service', () => {
...
+          "schema": "",
+          "table": "users",
+        },
+      ]
+    `,
+    );
+
+    expect(client1.drain()).toMatchInlineSnapshot(`[]`);
+
+    // TODO: How do I get ViewSyncer to "close" client2?
```
arv, line 1514: @darkgnotic ?

darkgnotic: To close a connection, you need to call `stream.cancel()` on the `stream` returned by `viewSyncer.initConnection()` (way up on line 415).

Currently the code fires off a background task that just reads the stream and drops messages into the queue, so you'd have to plumb the stream out of that somehow. Maybe have `connect()` return `[Queue, Source]` or something.

packages/zero-protocol/src/poke.ts
```diff
@@ -53,9 +52,6 @@ export const pokePartBodySchema = v.object({
   pokeID: v.string(),
   // Changes to last mutation id by client id.
   lastMutationIDChanges: v.record(v.number()).optional(),
-  // Patches to the set of "alive" clients (according to server) belonging to
-  // this client group.
-  clientsPatch: clientsPatchSchema.optional(),
```
darkgnotic, line 58: The nice thing here is that removing an optional field doesn't require a protocol version bump.   👍

PR #3816, opened 2025-02-14
feat(zero-pg): implement `makeSchemaCRUD` for the server

tantaman:
In the body of a custom mutator on the server, CRUD operators are provided to update the underlying tables.
This matches what also exists on the client, allowing users to share their mutator code between client and server.

### DIFF
--- a/packages/zero-pg/package.json
+++ b/packages/zero-pg/package.json
@@ -18,6 +18,8 @@
   },
   "prettier": "@rocicorp/prettier-config",
   "dependencies": {
+    "@databases/escape-identifier": "^1.0.3",
+    "@databases/sql": "^3.3.0",
     "@rocicorp/logger": "^5.3.0",
     "vitest": "^2.1.5"
   }
--- a/packages/zero-pg/src/custom.ts
+++ b/packages/zero-pg/src/custom.ts
@@ -1,12 +1,15 @@
 import type {ReadonlyJSONValue} from '../../shared/src/json.ts';
 import type {Schema} from '../../zero-schema/src/builder/schema-builder.ts';
+import type {TableSchema} from '../../zero-schema/src/table-schema.ts';
 import type {
   SchemaCRUD,
   SchemaQuery,
+  TableCRUD,
   TransactionBase,
 } from '../../zql/src/mutate/custom.ts';
 import type {ConnectionProvider, DBTransaction} from './db.ts';
 import {PushProcessor, type PushHandler} from './web.ts';
+import {formatPg, sql} from '../../z2s/src/sql.ts';
 
 export interface Transaction<S extends Schema, TWrappedTransaction>
   extends TransactionBase<S> {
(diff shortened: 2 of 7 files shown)


packages/zero-pg/src/custom.ts
```diff
@@ -75,22 +78,173 @@ export class TransactionImpl<S extends Schema, TWrappedTransaction>
...
+  )}`;
+}
+
+function origAndServerNamesFor(
+  originalNames: readonly string[],
+  schema: TableSchema,
+): [origName: string, serverName: string][] {
+  return originalNames.map(name => {
+    const col = schema.columns[name];
+    return [name, col.serverName ?? name] as const;
```
tantaman, line 248: we should make `serverName` non-optional in the underlying structure to remove these checks. It can remain optional in the builder interface but later filled in by `createSchema`.

packages/zero-pg/src/custom.ts
```diff
@@ -1,12 +1,15 @@
...
+import type {TableSchema} from '../../zero-schema/src/table-schema.ts';
 import type {
   SchemaCRUD,
   SchemaQuery,
+  TableCRUD,
   TransactionBase,
 } from '../../zql/src/mutate/custom.ts';
 import type {ConnectionProvider, DBTransaction} from './db.ts';
 import {PushProcessor, type PushHandler} from './web.ts';
+import {formatPg, sql} from '../../z2s/src/sql.ts';
```
tantaman, line 12: we build our sql with the `z2s` query builder rather than pulling in a full-fledged postgres library. The reason is that the user can bring their own postgres library. We're not opinionated about this.

packages/zero-pg/src/custom.ts
tantaman: we should also validate the schema.ts that exists on the server matches the user's pg db when `createHandler` is invoked.

Related: https://github.com/rocicorp/mono/pull/3783

packages/zero-pg/package.json
```diff
@@ -18,6 +18,8 @@
   },
   "prettier": "@rocicorp/prettier-config",
   "dependencies": {
+    "@databases/escape-identifier": "^1.0.3",
```
arv, line 21: Changes to a package.json usually comes with changes to package-lock.json too. Did you forget to include that?

packages/zero-pg/src/custom.ts
```diff
@@ -75,22 +78,168 @@ export class TransactionImpl<S extends Schema, TWrappedTransaction>
...
+          method.bind(txHolder),
+        ]),
+      ) as TableCRUD<TableSchema>;
+
+      return target[prop];
+    }
   }
+
+  return (dbTransaction: DBTransaction<unknown>) =>
+    new Proxy({}, new CRUDHandler(dbTransaction)) as SchemaCRUD<S>;
```
arv, line 139: Do we really need a Proxy?

For Replicache, the tx creation was hot and Proxies are not fast.

packages/zero-pg/src/web.ts
arv, line 1: It is not clear to me why this file is called web.ts?

packages/zero-pg/src/custom.ts
```diff
@@ -75,22 +78,168 @@ export class TransactionImpl<S extends Schema, TWrappedTransaction>
...
+): (dbTransaction: DBTransaction<unknown>) => SchemaCRUD<S> {
+  const schemaCRUDs: Record<string, TableCRUD<TableSchema>> = {};
+  for (const tableSchema of Object.values(schema.tables)) {
+    schemaCRUDs[tableSchema.name] = makeTableCRUD(tableSchema);
+  }
+
+  /**
+   * For users with very large schemas it is expensive to re-create
+   * all the CRUD mutators for each transaction. Instead, we create
+   * them all once up-front and then bind them to the transaction
```
arv, line 110: Looks more like they are lazily created?

arv (commented):
Generally LGTM but that proxy is giving some discomfort.

PR #3817, opened 2025-02-14
feat!(zql): `query.run` now returns a promise

tantaman:
`query.run` has to return a promise on the server (we're invoking PG) so it'll need to return a promise on the client as well so the code for custom mutators is isomorphic between the two.


arv (commented):
LGTM

PR #3818, opened 2025-02-14
feat(zero-pg): implement `makeSchemaQuery` to run queries on the server

tantaman:


packages/zero-pg/src/query.pg-test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  beforeEach(async () => {
+    pg = await testDBs.create('makeSchemaQuery-test');
+    await pg.unsafe(schemaSql);
+    await pg.unsafe(seedDataSql);
+
+    queryProvider = makeSchemaQuery(schema);
+  });
+
+  test('select', async () => {
```
tantaman, line 22: this is just to test the plumbing. Tests of query forms (related/where/exists/etc.) are in `z2s/test/chinook`

packages/zero-pg/src/query.ts
```diff
@@ -0,0 +1,98 @@
...
+    schema: TSchema,
+    tableName: TTable,
+    ast: AST,
+    format: Format | undefined,
+  ): Query<TSchema, TTable, TReturn> {
+    return new Z2SQuery(schema, tableName, this.#dbTransaction, ast, format);
+  }
+
+  async run(): Promise<HumanReadable<TReturn>> {
+    const sqlQuery = formatPg(compile(this._completeAst(), this.format));
```
arv, line 77: Should we cache this so we do not have to recompute it?

packages/zero-pg/src/query.ts
```diff
@@ -0,0 +1,98 @@
...
+    const result = await this.#dbTransaction.query(
+      sqlQuery.text,
+      sqlQuery.values,
+    );
+    if (Array.isArray(result)) {
+      return result as HumanReadable<TReturn>;
+    }
+    return [...result] as HumanReadable<TReturn>;
```
arv, line 85: Is this correct? How does this fit with `one()`?

tantaman: good catch.

arv (commented):
LGTM

I still wish we didn't have to depend on proxies

PR #3820, opened 2025-02-16
feat(zero-client): export `CustomMutatorDefs` and `CustomMutatorImpl` via `@rocicorp/zero/react`

cbnsndwch:
## Why

As a developer testing early support for custom mutators I need a way to provide typings for my custom mutator definitions when calling `new Zero` and `useZero`, so that intellisense works in my IDE

`CustomMutatorDefs` and `CustomMutatorImpl` are not currently exported via `@rocicorp/zero/react` and the default value of `undefined` for the `MD` type param in `Zero` and `useZero` make it impractical to 

## What 

- adds an optional second parameter `TArgs` to `CustomMutatorImpl` to allow custom mutators to specify the type of input data
- exports `CustomMutatorDefs` and `CustomMutatorImpl` from `@rocicorp/zero/react` for use in apps

### DIFF
--- a/packages/zero-client/src/client/custom.ts
+++ b/packages/zero-client/src/client/custom.ts
@@ -45,12 +45,12 @@ export type CustomMutatorDefs<S extends Schema> = {
   };
 };
 
-export type CustomMutatorImpl<S extends Schema> = (
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type CustomMutatorImpl<S extends Schema, TArgs = any> = (
   tx: Transaction<S>,
   // TODO: many args. See commit: 52657c2f934b4a458d628ea77e56ce92b61eb3c6 which did have many args.
   // The issue being that it will be a protocol change to support varargs.
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  args: any,
+  args: TArgs,
 ) => Promise<void>;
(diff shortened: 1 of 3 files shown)


cbnsndwch:
@aboodman 👀

packages/zero-client/src/client/custom.ts
```diff
@@ -45,12 +45,12 @@ export type CustomMutatorDefs<S extends Schema> = {
   };
 };
 
-export type CustomMutatorImpl<S extends Schema> = (
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type CustomMutatorImpl<S extends Schema, TArgs = any> = (
   tx: Transaction<S>,
   // TODO: many args. See commit: 52657c2f934b4a458d628ea77e56ce92b61eb3c6 which did have many args.
   // The issue being that it will be a protocol change to support varargs.
```
cbnsndwch, line 52: On this:

> // TODO: many args. See commit: 52657c2f934b4a458d628ea77e56ce92b61eb3c6 which did have many args.
> // The issue being that it will be a protocol change to support varargs.

I find that the pattern of a single object argument to queries and mutations helps in keeping code clean and self-contained. Implementors can define a single type/contract to receive input data, including an array of arbitrary length

PR #3822, opened 2025-02-16
feat(zero-client): export CustomMutatorDefs and CustomMutatorImpl from @rocicorp/zero

cbnsndwch:
## Why

As a developer testing early support for custom mutators I need a way to provide typings for my custom mutator definitions when calling `new Zero` and `useZero`, so that intellisense works in my IDE

`CustomMutatorDefs` and `CustomMutatorImpl` are not currently exported from `@rocicorp/zero` and the default value of `undefined` for the `MD` type param in `Zero` and `useZero` (from `@rocicorp/zero`) make it impractical to infer the type with ts-fu

## What 

- adds an optional second parameter `TArgs` to `CustomMutatorImpl` to allow custom mutators to specify the type of input data
- exports `CustomMutatorDefs` and `CustomMutatorImpl` from `@rocicorp/zero` for use in apps


packages/zero-client/src/client/custom.ts
```diff
@@ -45,12 +45,12 @@ export type CustomMutatorDefs<S extends Schema> = {
   };
 };
 
-export type CustomMutatorImpl<S extends Schema> = (
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type CustomMutatorImpl<S extends Schema, TArgs = any> = (
   tx: Transaction<S>,
   // TODO: many args. See commit: 52657c2f934b4a458d628ea77e56ce92b61eb3c6 which did have many args.
```
cbnsndwch, line 51: On this:

> // TODO: many args. See commit: 52657c2f934b4a458d628ea77e56ce92b61eb3c6 which did have many args.
> // The issue being that it will be a protocol change to support varargs.

I find that the pattern of a single object argument to queries and mutations helps in keeping code clean and self-contained. Implementors can define a single type/contract to receive input data, including an array of arbitrary length

tantaman: I think a single arg also encourage signatures for mutators that are more resilient.

eg.,

```ts
mutator(args: {
  arg1?: string;
  arg2?: string;
});
```

can have arguments removed without impacting backwards compatibility whereas positional arguments would be more rigid.

But this is up to @aboodman

packages/zero-client/src/mod.ts
```diff
@@ -93,6 +93,7 @@ export type {
 export type {Query, Row} from '../../zql/src/query/query.ts';
 export type {ResultType, TypedView} from '../../zql/src/query/typed-view.ts';
 export type {DBMutator, TableMutator} from './client/crud.ts';
+export type {CustomMutatorDefs, CustomMutatorImpl} from './client/custom.ts';
```
tantaman, line 96: where did you find yourself needing the `CustomMutatorImpl` type?

tantaman (approved):

packages/zero-client/src/client/custom.ts
```diff
@@ -45,12 +45,12 @@ export type CustomMutatorDefs<S extends Schema> = {
   };
 };
 
-export type CustomMutatorImpl<S extends Schema> = (
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type CustomMutatorImpl<S extends Schema, TArgs = any> = (
```
tantaman, line 49: `TArgs` should probably be `ReadOnlyJSONValue` but I can make that change since I think `TArgs` has the incorrect type in a few other places as well.

cbnsndwch: Here's an example of how I'm attempting to use them:

https://github.com/cbnsndwch/zero-sources/blob/develop/apps/zchat/app/zero/mutators.ts

PR #3832, opened 2025-02-19
!feat(zero-protocol): transition to pokeEnd.cookie

darkgnotic:
### ⚠️  PROTOCOL VERSION CHANGE

* This bumps the sync wire protocol from `/sync/v5/connect` to `/sync/v6/connect`
* The `view-syncers` must be fully rolled out before client code is rolled out

### Feature

* Makes `pokeStart.cookie` optional.
* Makes `pokeEnd.cookie` required.
* Server sends both
* Client only reads the latter


grgbkr (commented):
LGTM

PR #3833, opened 2025-02-19
fix(zbugs): comment edit was broken

aboodman:
feat(zql): Print out permissions when write authorization fails

Fixes https://bugs.rocicorp.dev/issue/3573


apps/zbugs/schema.ts
```diff
@@ -336,6 +336,10 @@ export const permissions: ReturnType<typeof definePermissions> =
               loggedInUserIsAdmin,
               and(loggedInUserIsCreator, canSeeComment),
             ],
+            postMutation: [
```
arv, line 339: Maybe define a canMutateComment and use that for insert, update, and delete?

arv (approved):
LGTM

PR #3834, opened 2025-02-19
wip(zero-cache): Mark desired query as inactive

arv:
and getInactiveDesiredQueries which returns all inactive queries in the order that we want to try to remove them.


packages/zero-cache/src/services/view-syncer/schema/cvr.ts
```diff
@@ -143,8 +143,9 @@ export type DesiresRow = {
   patchVersion: string;
   deleted: boolean | null;
   ttl: number | null;
-  expiresAt: number | null;
   inactivatedAt: number | null;
+  /** @deprecated */
```
arv, line 147: Now I feel bad that I split these into multiple PRs...

PR #3835, opened 2025-02-19
feat(zero-cache): copy before index (faster initial sync)

darkgnotic:
Copy tables / rows before creating indexes. On a 1GB dataset, this doubles the speed of initial sync, though improvements may be larger for tables with more complex indexes.

Before:
Screenshot 2025-02-18 at 18 46 27

After:
Screenshot 2025-02-19 at 08 27 52


cesara (commented):
wow

PR #3838, opened 2025-02-19
Allow `IS`/`IS NOT` comparison operators for `json` operands

jameshpalmer:
Change the `NoJsonSelector` type to allow `IS` and `IS NOT` operators on JSON operands


aboodman:
As discussed in discord I think we need to use deepEquals to make this work right in the query engine. Right fix for now is a doc change.

PR #3840, opened 2025-02-20
fix(zero-cache): handle query collisions

darkgnotic:
Handle the case when multiple queries (with different client hashes) are logically equivalent. This was not expected before but can reasonably happen after applying permissions. It is occasionally encountered when multiple queries are disallowed entirely.

https://bugs.rocicorp.dev/issue/3584

https://rocicorp.slack.com/archives/C08AHHFQMMY/p1739577732367319?thread_ts=1739485390.141439&cid=C08AHHFQMMY


arv (commented):
LGTM

PR #3843, opened 2025-02-20
chore(zero-cache): getInactiveQueries fix

arv:
getInactiveQueries needs to look at all clients in the client group since the eviction is based on the client group and not the client.


darkgnotic (commented):
Nice tests!

PR #3845, opened 2025-02-20
chore: update to esbuild 0.25.0

darkgnotic:
This is needed for installing packages via sst, which is at '0.25.0'.

### DIFF
--- a/packages/replicache-perf/package.json
+++ b/packages/replicache-perf/package.json
@@ -17,7 +17,7 @@
     "@types/command-line-usage": "^5.0.2",
     "command-line-args": "^6.0.1",
     "command-line-usage": "^7.0.3",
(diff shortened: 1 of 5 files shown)


arv (commented):
LGTM

packages/replicache-perf/package.json
```diff
@@ -17,7 +17,7 @@
     "@types/command-line-usage": "^5.0.2",
     "command-line-args": "^6.0.1",
     "command-line-usage": "^7.0.3",
-    "esbuild": "^0.24.2",
+    "esbuild": "^0.25.0",
```
arv, line 20: I thought we wanted to use `"esbuild": "0.25.0",`. In other words pin the version.

PR #3847, opened 2025-02-21
feat(zero-permissions): deploy permissions via a Lambda in the VPC

darkgnotic:
In the sandbox stage, deploy permission using a Lambda in the VPC, as an example of how permissions can be deployed to upstream databases that are not accessible outside of the AWS network.

This is largely how we would recommend users to deploy permissions. The only difference for the `mono` / `zbugs` setup is that because we deploy at head, we install "@rocicorp/zero" via a tarball, whereas users would simply reference their desired version in the function's `package.json` file.

For prod, continue using the local `Command` to deploy permissions, so that both approaches are exercised.

### DIFF
--- a/prod/functions/src/permissions.ts
+++ b/prod/functions/src/permissions.ts
@@ -1,0 +1,4 @@
+import { execSync } from "node:child_process";
+
+export const deploy = () =>
+  execSync("npx zero-deploy-permissions", { stdio: "inherit" });
--- a/prod/functions/tsconfig.json
+++ b/prod/functions/tsconfig.json
@@ -1,0 +1,8 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "lib": ["es2022", "dom"],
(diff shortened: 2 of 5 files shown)


prod/functions/src/permissions.ts
```diff
@@ -0,0 +1,4 @@
+import { execSync } from "node:child_process";
```
arv, line 1: strange formatting.

prod/functions/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "lib": ["es2022", "dom"],
```
arv, line 4: why do we need these?

prod/functions/src/permissions.ts
darkgnotic: Thank you. Fixed in https://github.com/rocicorp/mono/pull/3851

PR #3850, opened 2025-02-21
feat(zero-cache): Evict inactive queries 

arv:
We now evict inactive queries from the cache. This is done by keeping
track of the time a query was inactivated. When the number of rows in
the RowRecordCache is larger than the desired rows then we evict the
oldest inactive query. After that the run loop starts over and we might
evict another one.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/cvr-store.ts
+++ b/packages/zero-cache/src/services/view-syncer/cvr-store.ts
@@ -109,6 +109,7 @@ export class CVRStore {
   readonly #rowCache: RowRecordCache;
   readonly #loadAttemptIntervalMs: number;
   readonly #maxLoadAttempts: number;
+  #rowCount: number = 0;
 
   constructor(
     lc: LogContext,
@@ -642,6 +643,7 @@ export class CVRStore {
     };
     if (this.#pendingRowRecordUpdates.size) {
       const existingRowRecords = await this.getRowRecords();
+      this.#rowCount = existingRowRecords.size;
       for (const [id, row] of this.#pendingRowRecordUpdates.entries()) {
         if (this.#forceUpdates.has(id)) {
           continue;
@@ -716,12 +718,16 @@ export class CVRStore {
       stats.rows += this.#pendingRowRecordUpdates.size;
       return true;
     });
-    await this.#rowCache.apply(
+    this.#rowCount = await this.#rowCache.apply(
       this.#pendingRowRecordUpdates,
       newVersion,
       rowsFlushed,
     );
     return stats;
+  }
+
+  get rowCount(): number {
+    return this.#rowCount;
   }
 
   async flush(
--- a/packages/zero-cache/src/services/view-syncer/cvr.ts
+++ b/packages/zero-cache/src/services/view-syncer/cvr.ts
@@ -294,11 +294,14 @@ export class CVRConfigDrivenUpdater extends CVRUpdater {
     return getInactiveQueries(this._cvr);
   }
 
-  deleteDesiredQueries(clientID: string, queries: string[]): PatchToVersion[] {
+  deleteDesiredQueries(
+    clientID: string,
+    queryHashes: string[],
+  ): PatchToVersion[] {
     const patches: PatchToVersion[] = [];
     const client = this.#ensureClient(clientID);
     const current = new Set(client.desiredQueryIDs);
-    const unwanted = new Set(queries);
+    const unwanted = new Set(queryHashes);
     const remove = intersection(unwanted, current);
     if (remove.size === 0) {
       return patches;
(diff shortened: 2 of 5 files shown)


arv:
This is done on top of #3842 which has some questions for @darkgnotic which I want to resolve before moving along.

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -215,6 +231,16 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
             throw new ErrorForClient({kind: ErrorKind.ClientNotFound, message});
           }
 
+          // Check if any queries have expired.
+          if (await this.#removeExpiredQueries(lc, cvr)) {
```
darkgnotic, line 235: I think you can put these expire / evict checks in a separate timer-based loop, that either polls or gets scheduled intelligently (based on query expirations / changes).

this run loop is only called when the replica advances so it's not the best place to check (as you suspected).

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -262,6 +288,77 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
   }
 
+  /**
+   * @returns true if any queries were removed.
+   */
+  async #removeExpiredQueries(
+    lc: LogContext,
+    cvr: CVRSnapshot,
+  ): Promise<boolean> {
+    if (hasExpiredQueries(cvr)) {
```
darkgnotic, line 298: In your timer-based check, you can make it pretty efficient by checking the in-memory cvr state without acquiring the lock, and only acquiring the lock if changes need to be made (i.e. in order to call `#syncQueryPipelineSet()`)

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -262,6 +288,77 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
+
+      const inactiveQueries = getInactiveQueries(cvr);
+      if (inactiveQueries.length === 0) {
+        return false;
+      }
+
+      let anyEvicted = false;
+
+      // cvr = ... For #syncQueryPipelineSet().
+      cvr = await this.#updatePatchesForDesiredQueries(lc, cvr, updater => {
```
darkgnotic, line 324: Similarly, in the timer-based check you would delay / avoid acquiring the lock until you need to call `#updatePatchesForDesiredQueries()` here.

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -262,6 +288,77 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
+          // if we are still above the desired row count.
+          if (patches.length > 0) {
+            anyEvicted = true;
+            break;
+          }
+        }
+        return patches;
+      });
+
+      if (this.#pipelinesSynced) {
```
darkgnotic, line 354: Seems like we can push this check + sync call into `#updatePatchesForDesiredQueries()` because we always want to do it.

arv: The call to `#updatePatchesForDesiredQueries` is done inside an if statement in both cases.

packages/zero-cache/src/services/view-syncer/cvr-store.ts
```diff
@@ -211,7 +212,7 @@ export class CVRStore {
           EXTRACT(EPOCH FROM "ttl") * 1000 AS "ttl",
           "inactivatedAt"
           FROM ${this.#cvr('desires')}
-          WHERE "clientGroupID" = ${id} AND (deleted IS NULL OR deleted = FALSE)`,
+          WHERE "clientGroupID" = ${id}`,
```
tantaman, line 215: for line 205: `AND (deleted IS NULL OR deleted = FALSE)` could be `AND deleted IS DISTINCT FROM true`

packages/zero-cache/src/services/view-syncer/cvr.ts
```diff
@@ -211,8 +210,36 @@ export class CVRConfigDrivenUpdater extends CVRUpdater {
...
+      }
+
+      if (
+        ttl !== oldClientState.ttl &&
+        oldClientState.ttl !== undefined &&
+        ttl !== undefined &&
+        ttl >= oldClientState.ttl
+      ) {
+        needed.add(hash);
+      }
```
tantaman, line 240: What if the old client state had no ttl but the new state does have a ttl? Does the query record get revised with the new TTL even though it is not added to the `needed` set?

tantaman (commented):
We still have the old problem of leaking desired queries on tab close, correct?

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -129,6 +138,16 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
 
+  /**
+   * The {@linkcode maxRowCount} is used for the eviction of queries when the
+   * number of rows in the CVR exceeds this value. The eviction process will
+   * remove queries that are inactive.
+   */
+  maxRowCount: number;
+
+  #expiredQueriesTimer: ReturnType<typeof setTimeout> | 0 = 0;
+  #nextExpiredQueryTime: number | 0 = 0;
```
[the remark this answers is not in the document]
arv: Really good question. I have no idea :-)

arv:
Followup https://github.com/rocicorp/mono/pull/3913

PR #3856, opened 2025-02-22
chore(infra): sst ebs option

cesara:


prod/sst/sst.config.ts
```diff
@@ -49,10 +52,91 @@ export default $config({
...
+          role: ecsVolumeRole.name,
+          policyArn:
+            "arn:aws:iam::aws:policy/service-role/AmazonECSInfrastructureRolePolicyForVolumes",
+        },
+      );
+    }
+
+    const addEbsVolumeConfig = (
+      transform: any,
+      ecsVolumeRole: aws.iam.Role | undefined,
```
darkgnotic, line 92: Since this is a lambda within the lexical scope of `ecsVolumeRole`, and the latter is essentially constant, I think you can just reference it instead of passing it in (thereby simplifying the callers).

prod/sst/sst.config.ts
```diff
@@ -49,10 +52,91 @@ export default $config({
...
+    ) => {
+      return {
+        ...transform,
+        service: IS_EBS_STAGE
```
darkgnotic, line 96: I see how it's kind of tricky to recursively merge the stuff from the ebs volume.

However, we might be able to simplify this with the [`defu`](https://www.npmjs.com/package/defu) library which essentially does a recursive merge (we use this in our flags parser). 

Although I don't think it will "merge" the `taskDefinition` function, we only define that once so I think it will do the right thing.

So ideally it can look something like:

```ts
const EBS_TRANSFORM = !IS_EBS_STAGE ? {} : {
  service: {
    volumneConfiguration: { ... },
  },
  taskDefinition: (args: any) => { ... },
};


// Then in the actual service definition:
const replicationManager = cluster.addService(`replication-manager`, {
  ...
  transform: defu(EBS_TRANSFORM, {
    service: {
      healthCheckGracePeriodSeconds: 300, // same as health.startPeriod
    },
    ....
  }),
});
```

Can you see if this would work? I think it would be make this easier to read and maintain.

darkgnotic (commented):
So awesome to see this all come together. Nice work going under the covers to get the ebs stuff defined (without it being officially supported by sst).

prod/sst/sst.config.ts
```diff
@@ -232,19 +233,10 @@ export default $config({
...
         target: {
-          healthCheck: {
-            enabled: true,
-            path: "/keepalive",
-            protocol: "HTTP",
-            interval: 5,
-            healthyThreshold: 2,
-            timeout: 3,
-          },
+          ...BASE_TRANSFORM.target,
```
darkgnotic, line 239: Does this work?

```ts
transform: defu(
  EBS_TRANSFORM,
  BASE_TRANSFORM,
  {
    target: {
      stickiness: { ... }
    }
  },
);

cesara: good catch

darkgnotic (approved):

PR #3860, opened 2025-02-23
chore(shared): simplify the Queue class

darkgnotic:
`enqueue()` formerly returned a Promise that allowed the caller to wait for the item to be consumed.

However, no (application) code actually needs that API.

Since this class is used in some potentially hot paths, remove that functionality to both eliminate an unnecessary `Resolver` per invocation, as well as make the calling code more readable by not requiring the `void` qualifier.


arv (commented):
LGTM

PR #3861, opened 2025-02-23
chore(shared): avoid promise wrapping in the common case

darkgnotic:
A small optimization that avoids a `Promise.resolve()` wrapper for common case enqueueing.

`dequeue()` now returns `Promise<T> | T`, which makes a couple of call sights that relied `then()` slightly more complicated, but overall the reduction in Promise creation seems worth it.


darkgnotic:
FYI, this and #3860 together result in a small but noticeable reduction in steady state memory usage with the 200qps stress test.

chart (1)

arv (commented):
LGTM

PR #3865, opened 2025-02-25
chore(zero-cache): plumb an "appID" to code that reads `permissions`, `clients`, `schemaVersions`

darkgnotic:
Replaces the "zero" constant name with an `appID` parameter in code that reads the replicated `permissions`, `clients`, and `schemaVersions` tables.

This value plumbed from the top level config, and is currently hard-coded to "zero", so this PR in and of itself has no behavioral change.

This is in preparation for transitioning the existing `--shard-id` parameter to an `--app-id`. 

Context: https://www.notion.so/replicache/Zero-Upstream-Organization-1a33bed8954580e9a03fdddf81690fcc


arv (commented):
LGTM

PR #3873, opened 2025-02-26
chore(zero-cache): plumb an `appID` and use it for naming upstream PG schemas

darkgnotic:
Adds plumbing for naming upstream schemas by an `appID` rather than forcing `"zero"` and `"zero_{*}"`.

The `appID` value is still hard-coded to `"zero"` at the top-level, so this PR should not result in a behavioral change.

Configurability and migration is forthcoming.

https://www.notion.so/replicache/Zero-Upstream-Organization-1a33bed8954580e9a03fdddf81690fcc


arv (commented):
LGTM

:rubberstamp:

PR #3875, opened 2025-02-26
feat(zero-cache): scope the cvr schema within the appID

darkgnotic:
Changes the name of the CVR schema from `cvr_{SHARD_ID}` to `{APP_ID}_{SHARD_ID}/cvr`.

This allows zero-cache's from multiple apps and shards to use the same DB for the CVR.

The appID is still hardcoded to "zero", but will eventually be configurable.

Note that there is no migration for this change; new view-syncer use the new schema and notify clients with unknown CVRs to refresh. This is necessary because there can be multiple view-syncers, and new and old ones must be able to coexist.

https://www.notion.so/replicache/Zero-Upstream-Organization-1a33bed8954580e9a03fdddf81690fcc


arv (commented):
LGTM

PR #3878, opened 2025-02-27
feat(zero-pg): add the zero-pg package to `zero`

tantaman:
So users can `import {createPushHandler} from '@rocicorp/zero/pg'`

Usage example to follow in zbugs update(s).

### DIFF
--- a/packages/zero/package.json
+++ b/packages/zero/package.json
@@ -92,6 +92,10 @@
     "./change-protocol/v0": {
       "types": "./out/zero/src/change-protocol/v0.d.ts",
       "default": "./out/zero/src/change-protocol/v0.js"
+    },
+    "./pg": {
+      "types": "./out/zero-pg/src/mod.d.ts",
+      "default": "./out/pg.js"
     }
   },
   "bin": {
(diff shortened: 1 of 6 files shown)


packages/zero/tool/build.ts
```diff
@@ -102,6 +103,7 @@ async function buildZeroClient() {
         react: basePath('src/react.ts'),
         solid: basePath('src/solid.ts'),
         advanced: basePath('src/advanced.ts'),
+        pg: basePath('src/pg.ts'),
```
arv, line 106: This list is only for client entry points.

packages/zero/package.json
```diff
@@ -92,6 +92,10 @@
     "./change-protocol/v0": {
       "types": "./out/zero/src/change-protocol/v0.d.ts",
       "default": "./out/zero/src/change-protocol/v0.js"
+    },
+    "./pg": {
+      "types": "./out/zero-pg/src/mod.d.ts",
+      "default": "./out/pg.js"
```
arv, line 98: This should be out/zero/src/pg.js

arv (changes requested):

PR #3883, opened 2025-02-27
fix(zql): Remove split-push from exists

grgbkr:
Rework the logic in Exist that currently results in a "split push" to push a single Change.

### DIFF
--- a/packages/zql/src/ivm/exists.ts
+++ b/packages/zql/src/ivm/exists.ts
@@ -144,19 +144,23 @@ export class Exists implements Operator {
               size = this.#fetchSize(change.node);
             }
             if (size === 1) {
-              const type = this.#not ? 'remove' : 'add';
-              // The node for the remove pushed below will contain the child
-              // added by this change in its
-              // relationships[this.#relationshipName],
-              // so this child add needs to be sent first.  This balance
-              // is important for outputs doing ref counting,
-              if (type === 'remove') {
-                this.#output.push(change);
+              if (this.#not) {
+                this.#output.push({
+                  type: 'remove',
+                  node: {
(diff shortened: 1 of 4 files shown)


packages/zql/src/ivm/exists.ts
```diff
@@ -144,19 +144,27 @@ export class Exists implements Operator {
...
+                // pushed to output, the added child needs to be excluded from
+                // the remove being pushed to output (since the child has
+                // never been added to the output).
+                this.#output.push({
+                  type: 'remove',
+                  node: {
+                    row: change.node.row,
+                    relationships: {
+                      ...change.node.relationships,
+                      [this.#relationshipName]: () => [],
```
tantaman, line 158: I think it'd be clearer if the comment above `this.#output.push` was placed here where you're returning an empty relationship.

tantaman (commented):
LGTM. Unfortunate that this must be so nuanced. An explicit `split push` test case that checks the invariants and documents our reasoning (even links to your notion doc and zbugs issue) would help.

packages/zql/src/ivm/exists.ts
```diff
@@ -177,18 +185,28 @@ export class Exists implements Operator {
...
+              if (this.#not) {
+                this.#output.push({
+                  type: 'add',
+                  node: change.node,
+                });
+              } else {
+                // Since the remove child change currently being processed is
+                // not pushed to output, the removed child needs to be added to
+                // the remove being pushed to output.
+                this.#output.push({
```
arv, line 197: Can this be refactored to share the code with 'add'?

arv (commented):
LGTM

PR #3884, opened 2025-02-28
!feat(zero-cache): migrate from `--shard-id` to `--app-id` with app-isolated permissions

darkgnotic:
### ⚠️  BREAKING CHANGE

`--shard-id` and `--shard-publications` (or the equivalent `ZERO_SHARD_ID` and `ZERO_SHARD_PUBLICATIONS`) are now:
* `--app-id` (or `ZERO_APP_ID`)
* `--app-publications` (or `ZERO_APP_PUBLICATIONS`)

The `--app-id` or `ZERO_APP_ID` must be passed to `npx zero-deploy-permissions` in order deploy permissions to that app. (The default app id is "zero").

### Feature

Introduces the concept of **apps**, whereby a `zero-cache` instance/cluster is associate with an app, and multiple apps can coexist on the same upstream / cvr / cdc database, with each app having its own permissions. 
* The canonical example is a user-based app and an admin app that operate on the same upstream data.
* However, this separation can also be used for scenarios like previews (user request: https://discord.com/channels/830183651022471199/1341708504456761424)

**Sharding** is no longer exposed in the current API. However, under the covers, each app works within a single `0` shard, and support for multiple, partitioned shards will be introduced for data sharding in the future.

### Migration

Other than changing the names of the flags, no manual steps should be necessary.

Under the covers:
* If an app uses the default `"0"` shard id, then the new schema will be under the "zero" app and "0" shard number, resulting in an identical upstream schema. No changes occur.
* If an app uses a non-default shard id such as "prod", zero-cache will instantiate a new app under "prod_0". Once it has successfully completed initial-sync, it will automatically decommission the corresponding "zero_prod" shard.

Screenshot 2025-02-27 at 13 44 08

* TODO: Expose a tool like `npx zero-out` for manually decommissioning shards.

### Tangential / Related Fix

`npx zero-deploy-permissions` now detects if an app's set of publications is incomplete, which can happen when upstream is partially reset (e.g. using tools such as Prisma).

In this case, `npx zero-deploy-permissions` will detect the situation and proceed with a warning.

Screenshot 2025-02-27 at 09 50 13

(user report: https://discord.com/channels/830183651022471199/1332138203725041685/1344041259395977350)


darkgnotic:
Sorry that this PR is so yuge. I did try to break it up into smaller pieces, but this final step kind of needed to be all in one.

Luckily, it's mostly plumbing. Here are the points of interest:

* Usage documentation ([diff](https://github.com/rocicorp/mono/pull/3884/files#diff-1d1ba4163b52498a85ad332117db9714daa7fc8ecb16222e0c39207da8890fac))
* Consolidated location for naming: ([link](https://github.com/rocicorp/mono/pull/3884/files#diff-e4101e5c8ee7ba2d74f417f61b3a21fd3c67e7dc15cdb8a00923351eb34ab94b))
* Utility for decommissioning a shard: ([link](https://github.com/rocicorp/mono/pull/3884/files#diff-caec06ef4a2ee7655d6581664506c5eb6e3fe71c342ffa87b41fb842472e96d5))
* Logic for determining if / when to decommission a legacy shard: ([link](https://github.com/rocicorp/mono/pull/3884/files#diff-6b227ea4cd0b42e9f8049b777724b2e3e0072a5f7f744c118ca35be5a93a5c93))

packages/zero-cache/src/types/shards.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export const INVALID_APP_ID_MESSAGE =
+  'The App ID may only consist of lower-case letters, numbers, and the underscore character';
+
+function check(shard: ShardID): {appID: string; shardNum: number} {
+  const {appID, shardNum} = shard;
+  if (!ALLOWED_APP_ID_CHARACTERS.test(appID)) {
+    throw new Error(INVALID_APP_ID_MESSAGE);
+  }
+  assert(typeof shardNum === 'number');
```
arv, line 55: assertNumber is a bit more concise and gives a nicer error.

packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
```diff
@@ -120,9 +121,10 @@ export const replicationEventSchema = v.union(
 
 export type ReplicationEvent = v.Infer<typeof replicationEventSchema>;
 
-// Creates a function that appends `_SHARD_ID` to the input.
-export function append(shardID: string) {
-  return (name: string) => id(name + '_' + shardID);
+// Creates a function that appends `_{shard-num}` to the input and
+// quotes the result to be a valid identifier.
+function append(shardNum: number) {
+  return (name: string) => id(name + '_' + String(shardNum));
```
arv, line 127: `String()` is not really needed in cases like these.

arv (commented):
LGTM

PR #3887, opened 2025-02-28
fix(zql): Exist operator: remove cached size reuse across rows during push 

grgbkr:
The overlay solution to split push results in relationship's being inconsistent between Nodes at the same level of the tree during push.  This means its not correct to reuse cached sizes across rows during push. 

With this change initial hydration is really the only time that the cross row cache optimization is having an impact.

### DIFF
--- a/packages/zql/src/ivm/exists.ts
+++ b/packages/zql/src/ivm/exists.ts
@@ -19,14 +19,15 @@ import {
 import type {SourceSchema} from './schema.ts';
 import {first} from './stream.ts';
 
-type SizeStorageKey = `row/${string}/${string}`;
-type CacheStorageKey = `row/${string}`;
+type SizeStorageKeyPrefix = `row/${string}/`;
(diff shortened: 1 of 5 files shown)


packages/zql/src/ivm/exists.ts
```diff
@@ -97,125 +119,125 @@ export class Exists implements Operator {
...
-      case 'remove': {
-        const size = this.#getSize(change.node);
-        // If size is undefined, this operator has not output
-        // this row before and so it is unnecessary to output a remove for
-        // it.  Which is fortunate, since #fetchSize would
-        // not be able to fetch a Node for this change since it is
-        // removed from the source.
-        if (size === undefined) {
+    this.#inPush = true;
+    try {
```
grgbkr, line 123: git is doing a poor job with the diff here, very little has change here, it is just wrapped in a try / finally.

PR #3890, opened 2025-02-28
fix(zero-cache): detect mismatches between upstream and replica

darkgnotic:
In a multi-node setup, there is a scenario in which initial-sync happens but the service is killed before litestream has a chance to backup the newly-synced replica. If there is an old replica backed up, this will be out of sync with the state of upstream (when the service restarts), and this will often result in some invariant causing the system to be wedged.

Detect this situation by recording the replica version in upstream so that it can be verified at startup, and auto-reset if it is out of sync.


darkgnotic:
FYI, this is what would have automatically fixed the NVM gigabugs yesterday.

PR #3894, opened 2025-02-28
fix(zql): split push overlay logic for join

grgbkr:
Adds overlay logic to join for when join splits a push from its child input, in to multiple pushes to its output (because the child matches/matched multiple parents).

### DIFF
--- a/packages/zql/src/ivm/exists.ts
+++ b/packages/zql/src/ivm/exists.ts
@@ -196,11 +196,7 @@ export class Exists implements Operator {
             case 'remove': {
               let size = this.#getSize(change.node);
               if (size !== undefined) {
-                // Work around for issue https://bugs.rocicorp.dev/issue/3204
-                // assert(size > 0);
-                if (size === 0) {
-                  return;
-                }
+                assert(size > 0);
                 size--;
                 this.#setSize(change.node, size);
               } else {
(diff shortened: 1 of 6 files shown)


packages/zql/src/ivm/exists.ts
```diff
@@ -196,11 +196,7 @@ export class Exists implements Operator {
             case 'remove': {
               let size = this.#getSize(change.node);
               if (size !== undefined) {
-                // Work around for issue https://bugs.rocicorp.dev/issue/3204
-                // assert(size > 0);
-                if (size === 0) {
-                  return;
-                }
+                assert(size > 0);
```
grgbkr, line 199: It's been a long road to adding this asset back.

packages/zql/src/ivm/join.push.test.ts
```diff
@@ -5800,3 +5800,1796 @@ describe('joins with compound join keys', () => {
...
+              },
+            },
+            {
+              "relationships": {
+                "owner": [
+                  {
+                    "relationships": {},
+                    "row": {
+                      "id": "u1",
+                      "name": "Boogs",
```
grgbkr, line 6936: This is not right.

grgbkr: Fixed.

PR #3895, opened 2025-02-28
chore(zero-cache): Unflake expire tests

arv:
By injecting a fake setTimeout. (Unfortunately, vi.useFakeTimers() do not work with pg tests).


darkgnotic (commented):
Thank you!

PR #3896, opened 2025-02-28
feat(zero): Add ttl to useQuery and preload

arv:
Update zbugs to use ttl in useQuery and preload.


arv:
The layering of the ttl in zero-client is a bit odd. We want the ttl in materialize and preload. Preload passes an argument. But I do not really want to pass ttl as an argument in materialize (even though that would make sense). Instead we attach the ttl to the query and extract it in materialize...

Now that I write this, if we continue with attaching the ttl to the query, I'll change to use a WeakMap. Way less code and simpler to understand.

arv:
Closing this in favor of #3897

PR #3897, opened 2025-02-28
feat(zero): Add ttl to useQuery and preload

arv:
This moves the ttl to a parameter of materialize. This is a lot simpler.

Update zbugs to use ttl in useQuery and preload.


tantaman:
@arv / @aboodman  - wat do you think about allowing a default ttl on the `Zero` object?

```ts
const z = new Zero({defaultTtl: days(1)}
```

that is used when a ttl is not specified in `useQuery`.

Also -- is there a bound on how many queries can be active yet?

arv:
> Also -- is there a bound on how many queries can be active yet?

No.

The bound is 20,000 rows but active queries are never evicted.

PR #3901, opened 2025-03-01
feat(custom-mutators): handle `persist`, `refresh`, `pullEnd` rebases

tantaman:
Still a number of problems here:

1. [ ] Persist uses the memdag sync head for IVM for rebasing. This doesn't work if the memdag sync head is the same as the perdag sync head because client group mutations will be skipped in that case.
2. [ ] Refresh appears to change the hash of the sync head in a given client's memdag
3. [x] Replay mutations are not handled 

---

The central class in all of this is still `IVMSourceRepo`.

The main change is that we no longer refer to the IVM branches via global names. The branch to use in a rebase is forked off at the start of a transaction and passed though the `ReplicacheTransaction` object. This is safer as it allows concurrent  (interleaved) rebases. Idk if that happens in practice but it looked possible due to `pullEnd`, `persist`, and `refresh` all being able to start a rebase and mutators being async.

The other change is that callers to `IVMSourceRepo` must provide an expected hash for the branch they're retrieving. We fatal if the hash does not match except in the case of a `refresh` as we can not have a matching hash in the client dag for a refresh. Instead we fork and fast-forward the sync branch to handle `refresh`.

Please see the commentary on `IVMSourceRepo`:

```ts
/**
 * Provides handles to IVM sources at the `sync` and `main` heads to be used
 * by mutators and queries.
 *
 * `sync` tracks replicache's sync head.
 * `main` tracks replicache's main head.
 *
 * Initial mutations are run against the `main` head.
 * All UI queries also run against the `main` head.
 *
 * Rebases usually happen against the `sync` head.
 * Rebases can happen for three reasons:
 * 1. pullEnd
 * 2. persist
 * 3. refresh
 *
 * `pullEnd` and `persist` always rebase against the sync head.
 *
 * When a rebase is run, the caller doing the rebase receives a fork of the sync head.
 * Given that, a rebase does not change the sync head held by this class but
 * changes the fork used by the caller.
 *
 * The fork is discarded once the rebase completes. The `main`
 * head will be updated via `experimentalWatch` to allow active queries
 * on `main` to see the changes, if any.
 *
 * `refresh` is the special case that does not rebase against `sync`.
 * Refresh brings data from IDB into the current tab. The data in IDB may contain mutations from
 * other clients that are not yet in the sync head. This means
 * what we need to rebase our local changes onto is not the sync head but
 * some commit that is ahead of the sync head.
 *
 * To deal with this, `refresh` is handled by computing a diff from `sync` to `expectedHead`.
 * `sync` is forked and the diff is applied to the fork. Mutations are then
 * rebased against that fork.
 *
 * The two important methods on this class:
 * - advanceSyncHead
 * - getSourcesForTransaction
 *
 * - advanceSyncHead is called on `pullEnd` and is the only way sync head can be changed.
 * - getSourcesForTransaction is called by refresh, persist, pullEnd to get the IVM sources
 * to use in a rebase.
 */
```

### DIFF
--- a/apps/zbugs/api/index.ts
+++ b/apps/zbugs/api/index.ts
@@ -104,6 +104,7 @@ fastify.get<{
 });
 
 fastify.post('/api/push', async function (request, reply) {
+  console.log('INVOKING PUSH HANDLER');
   const response = await pushHandler(
     {
       authorization: request.headers.authorization,
--- a/apps/zbugs/src/pages/issue/issue-page.tsx
+++ b/apps/zbugs/src/pages/issue/issue-page.tsx
@@ -170,7 +170,7 @@ export function IssuePage({onReady}: {onReady: () => void}) {
   const [editing, setEditing] = useState<typeof displayed | null>(null);
   const [edits, setEdits] = useState<Partial<typeof displayed>>({});
   useEffect(() => {
-    if (displayed?.shortID !== undefined && idField !== 'shortID') {
+    if (displayed?.shortID != null && idField !== 'shortID') {
       navigate(links.issue(displayed), {
         replace: true,
         state: zbugsHistoryState,
--- a/packages/replicache/src/db/rebase.test.ts
+++ b/packages/replicache/src/db/rebase.test.ts
@@ -229,6 +229,7 @@ describe('rebaseMutationAndCommit', () => {
           new LogContext(),
           fixture.clientID,
           fixture.formatVersion,
+          undefined,
         ),
     );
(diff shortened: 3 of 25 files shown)


tantaman:
persist must be wrong in the case where the `memdagBaseSnapshot` is not newer then the `perdagBaseSnapshot`.

Might need to do one of:
a. keep an IVM source that tracks the perdag 🤔
b. create the `persist` source from the client's `sync` source by replaying all outstanding `clientGroupLocalMutations`

PR #3905, opened 2025-03-03
feat(zero): WebSocket close beacon

arv:
When Zero close is called we send a special close reason that is interpreted by the view syncer.

We send this close reason on pagehide when persisted is false. This websocket send goes through in unload when the tab is closed or reloaded. Different browsers deals with navigation differently.

https://replicache.notion.site/Close-Beacon-Yes-or-no-1ab3bed895458060bf8ad19603a2979f

In the view syncer, when we get this clean close we delete the client which means that it will no longer make queries desired by that client.

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -32,6 +32,7 @@ import {sleep, sleepWithAbort} from '../../../shared/src/sleep.ts';
 import * as valita from '../../../shared/src/valita.ts';
 import type {Writable} from '../../../shared/src/writable.ts';
 import type {ChangeDesiredQueriesMessage} from '../../../zero-protocol/src/change-desired-queries.ts';
+import type {CloseConnectionMessage} from '../../../zero-protocol/src/close-connection.ts';
 import type {ConnectedMessage} from '../../../zero-protocol/src/connect.ts';
 import {encodeSecProtocols} from '../../../zero-protocol/src/connect.ts';
 import type {DeleteClientsBody} from '../../../zero-protocol/src/delete-clients.ts';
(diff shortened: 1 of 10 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -1084,6 +1115,8 @@ export class Zero<
     this.#socket = ws;
     this.#socketResolver.resolve(ws);
 
+    getBrowserGlobal('window')?.addEventListener('pagehide', this.#onPageHide);
```
andrewcoelho, line 1118: @arv this breaks react-native, I'm getting error `getBrowserGlobal(\"window\")?.addEventListener is not a function (it is undefined)`

arv: @andrewcoelho Thanks for the report.

Strange... So react native has a `window` object but it is not an `EventTarget`?

```js
getBrowserGlobal('window')?.addEventListener?.('pagehide', this.#onPageHide);
```

Can you verify if that works on RN?

andrewcoelho: @arv yes that fixes it (had to add to the `getBrowserGlobal('window')?.removeEventListener` below as well. For context this is react-native with expo, not sure if expo is doing something with window since it's also designed to run on the web.

arv: I'll add a PR that fixes this. (unless you want to do it?)

andrewcoelho: sure will do

andrewcoelho: @arv: https://github.com/rocicorp/mono/pull/4037

PR #3908, opened 2025-03-03
feat(zero-cache): make initial-sync and vacuum faster with journal_mode=OFF

darkgnotic:
Make `journal_mode = OFF` work for real by temporarily disabling better-sqlite3's safe mode.

This brings the time it takes to VACUUM a 3GB db (on a MacBook) from ~10.7 seconds down to ~5.7 seconds.

Also print some freelist stats at startup to add visibility into how much space a VACUUM operation would safe. In the future we can consider adding freelist-based policies for maintenance vacuums.

### DIFF
--- a/packages/zero-cache/src/db/migration-lite.ts
+++ b/packages/zero-cache/src/db/migration-lite.ts
@@ -69,6 +69,7 @@ export async function runSchemaMigrations(
     randInt(0, Number.MAX_SAFE_INTEGER).toString(36),
   );
   const db = new Database(log, dbPath);
+  db.unsafeMode(true); // Enables journal_mode = OFF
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/db/migration-lite.ts
```diff
@@ -69,6 +69,7 @@ export async function runSchemaMigrations(
     randInt(0, Number.MAX_SAFE_INTEGER).toString(36),
   );
   const db = new Database(log, dbPath);
+  db.unsafeMode(true); // Enables journal_mode = OFF
```
tantaman, line 72: who knew. Nice find.

PR #3912, opened 2025-03-04
fix(zero-cache): Only print out permission policy when it changes, no…

aboodman:
…t every query.


aboodman:
This was causing a huge amount of logspam.

PR #3913, opened 2025-03-04
chore(zero-cache): followup changes from ttl pr

arv:


packages/zero-cache/src/services/view-syncer/cvr.ts
```diff
@@ -244,7 +244,7 @@ export class CVRConfigDrivenUpdater extends CVRUpdater {
       return patches;
     }
     const newVersion = this._ensureNewVersion();
-    client.desiredQueryIDs = [...union(current, needed)].sort(compareUTF8);
+    client.desiredQueryIDs = [...union(current, needed)].sort(stringCompare);
```
arv, line 247: query IDs are ascii

PR #3914, opened 2025-03-04
feat(zero): Inactive desired queries when deleting a client 

arv:
We used to delete the desired queries of a client when the client was
deleted. This was causing the queries and its rows to be removed from
the client group and when a client is re-added the rows had to be
fetched again.

Instead, when a client is deleted, we inactive the desired queries of
the deleted client. This way, the queries will use the ttl to determine
if the queries and its rows should be kept or removed.


darkgnotic (approved):
Nicely done!

PR #3922, opened 2025-03-05
fix(zero-client)!: Fix array view with multiple entries

arv:
BREAKING CHANGE!

The code assumed that one could never get two adds of the same row. However, this can happen when we have many to many relationships.

To deal with this we introduce a ref count to the array view. This allows us to keep track of how many times a row has been added to the view. When the ref count goes to zero we remove the row from the view.

This is a breaking change because `applyChange` now has a new required parameter `refCountMap`. It will require updates to vue-zero and vue-svelte.

https://bugs.rocicorp.dev/issue/3563


arv:
@stolinski @RobertoSnap @danielroe

FYI, once this gets into a release you update your libraries to pass `new WeakMap` like we do for React.

packages/zero-solid/src/solid-view.ts
```diff
@@ -30,6 +32,24 @@ type State = [Entry, QueryResultDetails];
...
+    return this.#map.get(unwrap(entry));
+  }
+  set(entry: Entry, refCount: number): void {
+    this.#map.set(unwrap(entry), refCount);
+  }
+  delete(entry: Entry): boolean {
+    return this.#map.delete(unwrap(entry));
+  }
+}
+
```
arv, line 52: @danielroe I think Vue needs to do this as well.

PR #3925, opened 2025-03-05
chore(zero-protocol): flag when PROTOCOL_VERSION might need to be bumped

darkgnotic:
This is similar to the analogous one in `ast.test.ts`


arv (approved):
LGTM 

Thanks for catching this

PR #3927, opened 2025-03-05
feat(shared): object mapping helpers

darkgnotic:
Per our conversation in https://rocicorp.slack.com/archives/C013XFG80JC/p1738806316413399?thread_ts=1738802020.370979&cid=C013XFG80JC

I'm using this in an upcoming PR but I figure it's worth carving out to reduce review overhead.

### DIFF
--- a/packages/shared/src/objects.ts
+++ b/packages/shared/src/objects.ts
@@ -1,0 +1,42 @@
+export function mapValues<T, U>(
+  input: Record<string, T>,
+  mapper: (value: T, index: number) => U,
+): Record<string, U> {
+  const output: Record<string, U> = {};
+  let i = 0;
+  for (const key in input) {
+    output[key] = mapper(input[key], i++);
+  }
+  return output;
+}
+
+/**
+ * Note: Use {@link mapValues()} if you do not need to change the keys, as it
+ *       is more efficient.
+ */
+export function mapEntries<T, U>(
+  input: Record<string, T>,
+  mapper: (
+    entry: [key: string, val: T],
+    index: number,
+  ) => [key: string, val: U],
+): Record<string, U> {
+  const output: Record<string, U> = {};
+  let i = 0;
+  for (const key in input) {
+    const [k, v] = mapper([key, input[key]], i++);
+    output[k] = v;
+  }
+  return output;
+}
+
+/**
+ * Note: Use {@link mapValues()} (preferred) or {@link mapEntries()} if possible,
+ *       as they are more efficient.
+ */
+export function mapAllEntries<T, U>(
+  input: Record<string, T>,
+  mapper: (entries: [key: string, val: T][]) => [key: string, val: U][],
+): Record<string, U> {
+  return Object.fromEntries(mapper(Object.entries(input)));
+}
--- a/packages/shared/src/objects.ts
+++ b/packages/shared/src/objects.ts
@@ -16,15 +16,12 @@ export function mapValues<T, U>(
  */
 export function mapEntries<T, U>(
   input: Record<string, T>,
-  mapper: (
-    entry: [key: string, val: T],
-    index: number,
-  ) => [key: string, val: U],
+  mapper: (key: string, val: T, index: number) => [key: string, val: U],
 ): Record<string, U> {
   const output: Record<string, U> = {};
   let i = 0;
   for (const key in input) {
-    const [k, v] = mapper([key, input[key]], i++);
+    const [k, v] = mapper(key, input[key], i++);
     output[k] = v;
   }
   return output;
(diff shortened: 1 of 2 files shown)


packages/shared/src/objects.ts
```diff
@@ -0,0 +1,42 @@
+export function mapValues<T, U>(
+  input: Record<string, T>,
+  mapper: (value: T, index: number) => U,
+): Record<string, U> {
+  const output: Record<string, U> = {};
+  let i = 0;
+  for (const key in input) {
```
darkgnotic, line 7: Any caveats about using `for ( ... in ... )`?

arv: Yup. You need to also do hasOwn because for-in includes enumerable properties on the prototype.

Here is the correct pattern:

https://github.com/rocicorp/mono/blob/bda4b03659881652ec0f29e4faa0c1e57deb1ca3/packages/shared/src/deep-clone.ts#L35-L36

arv:
I did a quick bench:

https://gist.github.com/arv/1b4e113724f6a14e2d4742bcc760d1fa

It seems like V8 got really good at optimizing away destructuring and the temp array of Object.entries

packages/shared/src/objects.ts
```diff
@@ -0,0 +1,39 @@
...
+    output[key] = mapper(input[key], i++);
```
arv, line 8: I'm questioning the value of passing an inded?

arv (commented):
LGTM

arv:
Based on the micro benchmark I think we should just use Object.values and Object.entries.

Sorry for the churn

darkgnotic:
> Based on the micro benchmark I think we should just use `Object.values` and `Object.entries`. Firefox and WebKit are slower with `entries` but not as drastically as V8.
> 
> Sorry for the churn.

Thanks for investigating. 

I'm curious though, should the benchmarks also include / measure the cost of calling `Object.fromEntries()` (compared to creating and populating a new `{}` Object)?

arv:
> I'm curious though, should the benchmarks also include / measure the cost of calling `Object.fromEntries()` (compared to creating and populating a new `{}` Object)?

Sure, it is easy enough to add

arv:
fromEntries does not seem very efficient, especially since we are doing `map` here.

```ts
describe.only('Object map performance', () => {
  const inc = (v: number) => v + 1;
  bench('using assignment', () => {
    const x: Record<string, number> = {};
    for (const [k, v] of Object.entries(obj)) {
      x[k] = inc(v);
    }
    if (typeof x !== 'object') throw new Error('Unexpected value');
  });

  bench('fromEntries', () => {
    const x = Object.fromEntries(
      Object.entries(obj).map(([k, v]) => [k, inc(v)]),
    );
    if (typeof x !== 'object') throw new Error('Unexpected value');
  });
});
```

```
npx vitest bench --watch
 RERUN  src/object.bench.ts x2 


 ✓  firefox (firefox)  src/object.bench.ts > Object map performance 1226ms
     name                     hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · using assignment  12,702.00  0.0000  3.0000  0.0787  0.0000  1.0000  1.0000  2.0000  ±8.61%     6351   fastest
   · fromEntries       10,280.00  0.0000  3.0000  0.0973  0.0000  1.0000  1.0000  2.0000  ±8.62%     5140

 ✓  webkit (webkit)  src/object.bench.ts > Object map performance 1213ms
     name                     hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · using assignment  15,528.00  0.0000  1.0000  0.0644  0.0000  1.0000  1.0000  1.0000  ±8.48%     7764   fastest
   · fromEntries       11,182.00  0.0000  1.0000  0.0894  0.0000  1.0000  1.0000  1.0000  ±8.37%     5591

 ✓  chromium (chromium)  src/object.bench.ts > Object map performance 1207ms
     name                    hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · using assignment  5,672.87  0.0000  0.5000  0.1763  0.2000  0.3000  0.4000  0.5000  ±1.40%     2837   fastest
   · fromEntries       4,033.19  0.1000  0.6000  0.2479  0.3000  0.4000  0.5000  0.5000  ±1.24%     2017

 BENCH  Summary

   chromium (chromium)  using assignment - src/object.bench.ts > Object map performance
    1.41x faster than fromEntries

   webkit (webkit)  using assignment - src/object.bench.ts > Object map performance
    1.39x faster than fromEntries

   firefox (firefox)  using assignment - src/object.bench.ts > Object map performance
    1.24x faster than fromEntries
```

PR #3929, opened 2025-03-06
chore(zero-cache): consolidate config updates into one method/flush

darkgnotic:
Assuming and taking advantage of the pattern that the `InitConnectionBody` is a superset of all smaller config updates, put everything in one method so that a config update always happens in once CVR flush.

(This will facilitate an upcoming change for schema detection, whereby another field will be added to the InitConnectionBody)


arv (commented):
LGTM

PR #3930, opened 2025-03-06
feat(zero-client): only track the `main` branch for IVM

tantaman:
`IVMSourceRepo` was removed as we no longer track the sync head and only track `main`. main will be forked and patched up as needed.

- refresh
- pullEnd
- persist

will all call `forkToHead` to get a copy of the IVM sources at the desired head.

`ivmSourceBranch.advance` is not yet called. It will be called in a future PR when we have the data required (expectedHash, newHash, diffs) to call it. ExperimentalWatch does not provided the two hashes we need so I'll be updating Replicache to call `advance` directly.


packages/zero-client/src/client/ivm-branch.test.ts
```diff
@@ -425,16 +340,150 @@ describe('advanceSyncHead', () => {
...
+        'mutator_name',
+        JSON.stringify([]),
+        null,
+        await dagStore.write(),
+        timestamp++,
+        'client-id',
+        FormatVersion.Latest,
+      );
+      await Promise.all(
+        diffs.map(async diff => {
```
arv, line 358: for of would have been simpler...

PR #3933, opened 2025-03-06
feat(replicache): add `zeroOption` to allow Zero to hook into Replicache events

tantaman:
zeroOption is a minimal interface that Replicache will call when:
  
  - pullEnd
  - persist
  - refresh
  - mutate
  
happen. Replicache will invoke
  
- `advance` whenever it's main head moves forward.
- `getTxData` to fork IVM sources whenever it needs to rebase
- `init` on construction to populate the IVM sources

Invoking the above three methods in `persist.ts`, `refresh.ts`, `replicache-impl.ts::mutate`, `replicache-impl.ts::maybeEndPull` will happen in the next PR.

### DIFF
--- a/packages/replicache/src/replicache-options.ts
+++ b/packages/replicache/src/replicache-options.ts
@@ -15,7 +15,7 @@ import type {Store} from './dag/store.ts';
 
 export interface ReplicacheOptions<
   MD extends MutatorDefs,
-  TZeroData = unknown,
+  TZeroData extends ZeroTxData = ZeroTxData,
 > {
   /**
    * This is the URL to the server endpoint dealing with the push updates. See
@@ -239,11 +239,13 @@ export interface ReplicacheOptions<
   zero?: ZeroOption<TZeroData> | undefined;
 }
 
+export interface ZeroTxData {}
+
 /**
  * Minimal interface that Replicache needs to communicate with Zero.
  * Prevents us from creating any direct dependencies on Zero.
  */
-export type ZeroOption<T> = {
+export type ZeroOption<T extends ZeroTxData> = {
   /**
    * Allow Zero to initialize its IVM state from the given hash and store.
    */
--- a/packages/replicache/src/replicache-options.ts
+++ b/packages/replicache/src/replicache-options.ts
@@ -5,7 +5,6 @@ import type {Puller} from './puller.ts';
 import type {Pusher} from './pusher.ts';
 import type {MutatorDefs, RequestOptions} from './types.ts';
 import type {Hash} from './hash.ts';
-import type {MaybePromise} from '../../shared/src/types.ts';
 import type {InternalDiff} from './btree/node.ts';
 import type {Store} from './dag/store.ts';
 
@@ -13,10 +12,7 @@ import type {Store} from './dag/store.ts';
  * The options passed to {@link Replicache}.
  */
 
-export interface ReplicacheOptions<
-  MD extends MutatorDefs,
-  TZeroData extends ZeroTxData = ZeroTxData,
-> {
+export interface ReplicacheOptions<MD extends MutatorDefs> {
   /**
    * This is the URL to the server endpoint dealing with the push updates. See
    * [Push Endpoint Reference](https://doc.replicache.dev/reference/server-push) for more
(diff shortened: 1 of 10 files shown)


packages/replicache/src/replicache-options.ts
```diff
@@ -4,12 +4,19 @@ import type {StoreProvider} from './kv/store.ts';
...
+import type {Store} from './dag/store.ts';
 
 /**
  * The options passed to {@link Replicache}.
  */
 
-export interface ReplicacheOptions<MD extends MutatorDefs> {
+export interface ReplicacheOptions<
+  MD extends MutatorDefs,
+  TZeroData extends ZeroTxData = ZeroTxData,
```
arv, line 18: I think it would be better to create a new interface that extends ReplicacheOptions. ReplicacheImpl can use that type. That way we do not have to change the public API of replicache.

arv: I think you can either add one more option here:

https://github.com/rocicorp/mono/blob/de356bdedc85e76a6622d7c110f5ed5d8e152402/packages/replicache/src/replicache-impl.ts#L388-L391

or you can extend `ReplicacheImplOptions` which is all internal stuff the Zero and Reflect use.

packages/replicache/src/replicache-options.ts
```diff
@@ -223,4 +230,39 @@ export interface ReplicacheOptions<MD extends MutatorDefs> {
...
+
+export interface ZeroTxData {}
+
+/**
+ * Minimal interface that Replicache needs to communicate with Zero.
+ * Prevents us from creating any direct dependencies on Zero.
+ */
+export type ZeroOption<T extends ZeroTxData> = {
+  /**
+   * Allow Zero to initialize its IVM state from the given hash and store.
```
arv, line 250: store -> dag store for clarity

packages/replicache/src/transactions.ts
```diff
@@ -271,7 +272,8 @@ export class SubscriptionTransactionWrapper implements ReadTransaction {
  * {@link ReplicacheOptions.mutators} and allows read and write operations on the
  * database.
  */
-export interface WriteTransaction extends ReadTransaction {
+export interface WriteTransaction<TZeroData extends ZeroTxData = ZeroTxData>
```
arv, line 275: I would prefer if we do not change Replicache public types.

For Reflect we have:

https://github.com/rocicorp/reflect-archive/blob/209cecc2c899bfbe16e57f32dcad1be8f75cf024/packages/reflect-shared/src/types.ts#L21

packages/zero-client/src/client/replicache-types.ts
```diff
@@ -34,7 +35,8 @@ export interface ReadTransaction extends ReplicacheReadTransaction {
   readonly env?: Env | undefined;
 }
 
-export interface WriteTransaction extends ReplicacheWriteTransaction {
+export interface WriteTransaction<T extends ZeroTxData = ZeroTxData>
```
arv, line 38: I thought we did this... I don't yet understand why we have to change the Replicache interface.

arv (changes requested):

packages/zero-client/src/client/replicache-types.ts
tantaman: When Replicache starts a rebase it'll ask Zero (via zeroOption) to create a fork of the IVM sources at the given head. This fork will be added to `WriteTransaction` so `Zero` can use it in it's mutators.

Previously I assumed there were a finite (or at least small) set of heads and that Zero could look them up by name rather than having to pass the fork along in the Replicache transaction. It turns out there isn't a small set of heads being rebased onto. Looking up by name also opens us up to concurrency issues.

tantaman: Let me put up the PR that actually uses this argument.

arv: I think I see the big picture better now. I still wish we could keep the Replicache public interfaces as they are.

I have a few tricks up my sleeves that would allow this to work without changing the interface. I think it is OK to add public properties (using symbols) to the implementation without changing the interface for example. Other options include using weakmaps as sidetables.

tantaman: hmm, maybe I can add it to the args going to `MutatorImpl` and all `Zero` mutators will require this argument.

packages/replicache/src/transactions.ts
tantaman: We have the same for Zero:

https://github.com/rocicorp/mono/blob/743c95499e12f0f66119fc819921da85031d4263/packages/zero-client/src/client/custom.ts#L70-L80

https://github.com/rocicorp/mono/blob/743c95499e12f0f66119fc819921da85031d4263/packages/zero-client/src/client/custom.ts#L90-L98

but somehow we need to plumb through the correct IVM fork. Only Replicache has the information required to create that fork (base head hash, new head hash) and when to fork (when rebase begins).

tantaman:
@arv - I removed all changes to the Replicache public interface.

tantaman:
> @arv - I removed all changes to the Replicache public interface.

well not quite `readonly zeroData: ZeroTxData | undefined;` is still on `interface WriteTransaction`

Let me try swapping that to a symbol.

arv (approved):
LGTM

Thanks

PR #3934, opened 2025-03-06
feat(custom-mutators): maintain and fork ivm through pull, mutate, refresh

tantaman:
Persist requires some more invasive changes and will be in a separate PR.

Calls to `advance` are the key thing to review. `advance` replaces our use of `experimentalWatch` to update the main IVM sources. The main difference from experimentalWatch is that `advance` sends in the expected hash we're advancing from and the hash we're advancing to. This allows `advance` to set an invariant that we always advance in order. The IVMBranch also keeps track of its current hash so it can compute diffs to use to patch forks in `forkToHead(hash)`.

I'm surprised we never ran into problems with `experimentalWatch` as I would imagine there would have been cases where a `refresh` diff against HEAD-1 jumps ahead of a `pullEnd` diff against HEAD-0 (due to awaits) or vice-versa.

### DIFF
--- a/packages/replicache/src/persist/persist.ts
+++ b/packages/replicache/src/persist/persist.ts
@@ -36,6 +36,7 @@ import {
   setClient,
 } from './clients.ts';
 import {GatherMemoryOnlyVisitor} from './gather-mem-only-visitor.ts';
+import type {ZeroOption} from '../replicache-options.ts';
 
 type FormatVersion = Enum<typeof FormatVersion>;
 
@@ -65,6 +66,7 @@ export async function persistDD31(
   mutators: MutatorDefs,
   closed: () => boolean,
   formatVersion: FormatVersion,
+  _getZeroData: ZeroOption['getTxData'] | undefined,
   onGatherMemOnlyChunksForTest = () => Promise.resolve(),
 ): Promise<void> {
   if (closed()) {
--- a/packages/replicache/src/persist/refresh.ts
+++ b/packages/replicache/src/persist/refresh.ts
@@ -55,6 +55,7 @@ type RefreshResult =
       type: 'complete';
       diffs: DiffsMap;
       newPerdagClientHeadHash: Hash;
+      oldHead: Hash;
       newHead: Hash;
     };
 
@@ -72,7 +73,7 @@ export async function refresh(
   closed: () => boolean,
   formatVersion: FormatVersion,
   zero: ZeroOption | undefined,
-): Promise<[Hash, DiffsMap] | undefined> {
+): Promise<{oldHead: Hash; newHead: Hash; diffs: DiffsMap} | undefined> {
   if (closed()) {
     return;
   }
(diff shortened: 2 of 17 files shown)


packages/replicache/src/persist/persist.ts
```diff
@@ -65,6 +66,7 @@ export async function persistDD31(
   mutators: MutatorDefs,
   closed: () => boolean,
   formatVersion: FormatVersion,
+  _getZeroData: ZeroOption['getTxData'] | undefined,
```
tantaman, line 69: to be used in the next PR. Getting `persist` to work requires changes to `LazyRead` and `LazyStore` which wouldn't fit well to be reviewed in this context.

arv:
Looking good. There are some failing tests here

I'm not sure if the perf benchmark is a real issue or not?

packages/zero-client/src/client/zero.ts
```diff
@@ -483,11 +482,24 @@ export class Zero<
       licenseKey: 'zero-client-static-key',
       kvStore,
     };
+
+    this.#zeroContext = new ZeroContext(
+      new LogContext(logOptions.logLevel, {}, logOptions.logSink),
```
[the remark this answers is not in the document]
tantaman: yeah, fixed.

tantaman:
@arv - I'm confused by the bencher output. It says this PR is slower but running `bench` locally I see this:

# On this branch
CleanShot 2025-03-11 at 12 16 46


# On Main
CleanShot 2025-03-11 at 12 17 44


This reads to me that this PR is more than 2x faster rather than being slower [as seen in the bencher report](https://bencher.dev/perf/zero?branches=90134553-1cf7-4d7e-b5d4-415127ae107d&heads=be8041bb-2b10-47b1-b366-c4f0bcc3ed1f&testbeds=50a21deb-8296-4657-b7ad-5a4002b77d68&benchmarks=b46723ed-a942-4b77-ad67-e5557cfd9304&measures=fc802276-c2ed-462d-9fa5-0c1fcf2deb8e&start_time=1739117527000&end_time=1741709527000&report=813df625-9385-4e29-abea-feabf386c808&lower_boundary=true&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&plots_page=1&clear=true) :|

arv:
I think the perf runner was probably slowed down for some other reason. It is running on ec2 and even though it is a dedicated instance it might still be affected by other things.

We should maybe change the runner to use relative performance or find a physical machine.

packages/replicache/src/persist/refresh.ts
```diff
@@ -289,8 +304,17 @@ export async function refresh(
     }
     return undefined;
   }
+
+  // Advance zero here before setting refresh hashes
+  // since we must advance before delegating control of the microtask
```
grgbkr, line 309: Why must we advance before delegating control of the microtask loop?

arv (commented):
LGTM

PR #3935, opened 2025-03-06
feat(zero-cache): support replication of tables with REPLICA IDENTITY FULL

darkgnotic:
When a table is configured with `REPLICA IDENTITY FULL`, the full rows are sent as the "keys", and all of the columns are considered the "keyColumns". 

However, these rows/columns cannot safely be used as keys because they may contain `NULL` values.

To handle these replication messages, create keys from the full rows using the primary key / unique index information from the replicated table schema.

User request: https://discord.com/channels/830183651022471199/1288232858795769917/1347172642507063368

### DIFF
--- a/packages/zero-cache/src/services/replicator/change-processor.ts
+++ b/packages/zero-cache/src/services/replicator/change-processor.ts
@@ -8,7 +8,11 @@ import {
   createIndexStatement,
   createTableStatement,
 } from '../../db/create.ts';
-import {listIndexes, listTables} from '../../db/lite-tables.ts';
+import {
+  computeZqlSpecs,
+  listIndexes,
+  listTables,
+} from '../../db/lite-tables.ts';
 import {
   mapPostgresToLite,
   mapPostgresToLiteColumn,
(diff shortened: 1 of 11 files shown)


packages/zero-cache/src/services/replicator/change-processor.ts
```diff
@@ -323,11 +338,31 @@ class TransactionProcessor {
...
+    if (!keyColumns?.length) {
+      throw new Error(
+        `Cannot replicate table "${relation.name}" without a PRIMARY KEY or UNIQUE INDEX`,
+      );
+    }
+    // For the common case (replica identity default), the row is already the
+    // key for deletes and updates, in which case a new object can be avoided.
+    return numCols === keyColumns.length
+      ? row
+      : Object.fromEntries(keyColumns.map(col => [col, row[col]]));
```
tantaman, line 358: nit: 

```ts
keyColumns.reduce((l, r) => l[r] = row[r], {});
```

would be less allocations in what I assume to be a hot path.

tantaman (commented):
lgtm.

Another option is to use `IS` for comparisons so `null` will correctly match then we don't have to do the key extraction.

`INSERT OR REPLACE` would need to change to two statements, however.

PR #3936, opened 2025-03-06
refactor(zql): move splitting of edits up to source

grgbkr:
This slightly simplifies take and join logic, but more importantly it avoids having to add fairly complex overlay logic in take and join to correctly handle split pushes.  Instead the overlay logic for edits split into a remove and add is centralized at the source layer.  

This is done by gathering the join keys and partition keys in builder and passing them to the source connection as a set called splitEditKeys.  If the an edit modifies the values of any of these keys, then the source will push it as a remove followed by an add instead of as an edit.

Added asserts in join and take ensuring no unexpected edits are received.

### DIFF
--- a/packages/zql/src/builder/builder.test.ts
+++ b/packages/zql/src/builder/builder.test.ts
@@ -684,6 +684,50 @@ test('self-join edit', () => {
               {
                 "relationships": {},
                 "row": {
+                  "id": 1,
+                  "name": "aaron",
+                  "recruiterID": null,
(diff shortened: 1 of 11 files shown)


tantaman (approved):
Clever.

Mutating source state is still unsettling to me. The memory-source's b-tree is technically persistent. Wonder if there's something we can do in the future where we track "timestamps" / "versions" that "fetch" can use to get the correct view.

PR #3937, opened 2025-03-07
fix(zero-cache): support "resumptive" replication

darkgnotic:
Support introducing rows that were not originally synced to the replica during initial sync via `UPDATE` operations. This is termed "resumptive replication" and is used for operations in which new tables/rows are introduced into the replica after initial sync, including:
* when altering a publication to include new tables or columns
* [resharding](https://www.notion.so/replicache/Zero-Sharding-1903bed89545803a9345dffc1e781048?pvs=4#1913bed8954580458202fb378ffffaba) (future feature)

In such operations, the rows in the new tables are "touched" with an innocuous update to push the data through the replication stream, e.g.

```sql
BEGIN;
ALTER PUBLICATION my_zero_pub ADD TABLE my_table;
UPDATE my_table SET id = id;
COMMIT;
```

User report: https://discord.com/channels/830183651022471199/1344646746407698483

### Note

This codifies the requirement that the `update` messages in the ChangeSource protocol contain the *full* contents of the updated row, and not just the updated values. This is how the Postgres Change Source already works, but it was not documented for the general (i.e. custom) change source protocol.  (@cbnsndwch)

This is now documented, and unit tests updated to reflect the codified requirement.

### DIFF
--- a/packages/zero-cache/src/integration/integration.pg-test.ts
+++ b/packages/zero-cache/src/integration/integration.pg-test.ts
@@ -768,6 +768,11 @@ describe('integration', {timeout: 30000}, () => {
               new: {
                 id: 'bar',
                 ['far_id']: 'not_baz',
+                b: true,
+                j1: {foo: 'bar\u0000'},
+                j2: true,
+                j3: 123,
+                j4: 'string',
               },
               key: null,
             },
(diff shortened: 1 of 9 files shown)


grgbkr:
LGTM

PR #3945, opened 2025-03-07
feat(zero-protocol): structural schema compatibility detection

darkgnotic:
Replaces the manual `schemaVersion`  numbering scheme with structural schema compatibility detection.

* A `ClientSchema` message is derived from the tables, columns, and types of the `schema.ts` declaration
* The normalized hash of the message is part of the IDB key.
* The `initConnectionMessage` contains an optional `clientSchema` field where the client declares its schema
  * This is only sent for the very first request (when baseCookie is `null`). 
  * Because it is stored with the CVR, it need not be sent for non-initial requests, as all clients in a group have the same client schema
 * The view-syncer checks a CVR's `clientSchema` against the replicated tables when the syncer is first initialized and on every schema change. Helpful error messages are generated, including:
   * If there are no tables, a message suggesting that replication is not working at all
   * If there are missing tables, or columns, the ones that are present
   * If a table is not syncable because it doesn't have a primary key
   * If a column is not syncable because its data type is not supported
 * The `schemaVersion` field in CRUD mutators is now optional. The connection level checking of the schema is sufficient.
   * For custom mutators, zero is agnostic to the protocol / schema of the custom mutations.

Next: 
* Get rid of the `schema.version` number. This will be in a separate PR for readability.


darkgnotic:
@tantaman There's one custom mutator test that's failing and I couldn't figure out where the schemaVersion value (7 ???) is coming from. A pointer would be appreciated!

Screenshot 2025-03-07 at 13 57 52

packages/zero-cache/src/services/view-syncer/client-handler.ts
```diff
@@ -141,7 +141,7 @@ export class ClientHandler {
     const pokeID = versionToCookie(tentativeVersion);
     const lc = this.#lc.withContext('pokeID', pokeID);
 
-    if (schemaVersions) {
+    if (schemaVersions && this.#schemaVersion) {
```
grgbkr, line 144: Should there be a TODO to remove this once we've fully migrated?

packages/zero-client/src/client/zero.ts
```diff
@@ -465,10 +470,13 @@ export class Zero<
 
     this.storageKey = storageKey ?? '';
 
+    const {clientSchema, hash} = clientSchemaFrom(schema);
+    this.#clientSchema = clientSchema;
+
     const replicacheOptions: ReplicacheOptions<WithCRUD<MutatorDefs>> = {
-      // The schema stored in IDB is dependent upon both the application schema
+      // The schema stored in IDB is dependent upon both the ClientSchema
```
grgbkr, line 477: We could reuse the IDB cache if the schema change just added new tables... but its probably not worth the complexity.

grgbkr:
LGTM.  This is a very nice quality of life change.

grgbkr:
Is the plan to also use clientSchema to limit what columns are synced to the client?

packages/zero-cache/src/services/mutagen/mutagen.ts
```diff
@@ -404,11 +406,13 @@ async function checkSchemaVersionAndIncrementLastMutationID(
     });
   }
 
-  assert(supportedVersionRange.length === 1);
-  throwErrorForClientIfSchemaVersionNotSupported(
-    schemaVersion,
-    supportedVersionRange[0],
-  );
+  if (schemaVersion !== undefined && supportedVersionRange !== undefined) {
```
grgbkr, line 409: Should there be a TODO to remove this once we've fully migrated?

packages/zero-protocol/src/client-schema.ts
```diff
@@ -0,0 +1,50 @@
...
+  columns: v.record(columnSchemaSchema),
+});
+
+export type TableSchema = v.Infer<typeof tableSchemaSchema>;
+
+export const clientSchemaSchema = v.object({
+  tables: v.record(tableSchemaSchema),
+});
+
+export type ClientSchema = v.Infer<typeof clientSchemaSchema>;
```
grgbkr, line 30: Idea: should we store the schema used to create permission as part of the permissions data in the DB?  Then we could validate that the permissions schema is compatible with the replicated schema?

If we do, perhaps this should be called ZqlSchema rather than ClientSchema.

packages/zero-cache/src/services/mutagen/mutagen.ts
```diff
@@ -44,7 +44,7 @@ export interface Mutagen {
   processMutation(
```
grgbkr, line 44: There is a narrow race with CRUD mutators.

Client c1 is connected with clientSchema X that is compatible with DB schema.
DB schema is updated such that it is no longer compatible with clientSchema X. 
Before this schema change is replicated c1 pushes a mutation using clientSchema X, which mutagen 
will try to execute against the now incompatible DB schema.

Not sure this is worth addressing given mutagen is expected to go away soon.

grgbkr: If this race occurs and the schema change was incompatible with the mutation, likely the mutation will just fail with an error and the client be disconnected.  It will try to reconnect and repush.  When the schema change replicates, the client will be disconnected with an error that will cause it to reload, hopefully getting the updated schema.

aboodman: Yeah this is a form of race we already knew about.

PR #3946, opened 2025-03-08
!feat(zero-cache): farewell, schema version

darkgnotic:
### ⚠️ Breaking Change

`createSchema()` no longer takes a version. The `schemaVersion` is no more.

> _ever pervasive_
> _so often misunderstood_
> _missed you will not be_

### DIFF
--- a/packages/zero-schema/src/builder/schema-builder.ts
+++ b/packages/zero-schema/src/builder/schema-builder.ts
@@ -14,7 +14,6 @@ import type {Relationships} from './relationship-builder.ts';
 import {type TableBuilderWithColumns} from './table-builder.ts';
 
 export type Schema = {
-  readonly version: number;
   readonly tables: {readonly [table: string]: TableSchema};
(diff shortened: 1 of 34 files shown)


grgbkr (commented):
Beautiful.  LGTM

packages/zero-schema/src/builder/schema-builder.ts
```diff
@@ -24,29 +23,16 @@ export type Schema = {
...
+ * createSchema({rsdfgafg: table('users')...}, {sdfd: relationships(users, ...)})
  * ```
- *
- * @param version The version of the schema. Only needs to be incremented
- * when the backend Postgres schema moves forward in a way that is not
- * compatible with the frontend. As in, if:
- * 1. Columns are removed
- * 2. Optional columns are made required
- *
- * Adding columns, adding tables, adding relationships, making
```
grgbkr, line 36: This documentation was not quite right.

It should be:

Adding optional columns

PR #3950, opened 2025-03-10
fix(zero-cache): Deal with deleted clients in CVR store

arv:
We keep the queries around for deleted clients in the CVR store until the query expires. However, the load code was not handling the case where the client was deleted correctly.


darkgnotic (commented):
LGTM

PR #3952, opened 2025-03-10
feat(zero): Print out a warning when a query takes a long time to

aboodman:
materialize on client.

### DIFF
--- a/packages/zero-client/src/client/context.ts
+++ b/packages/zero-client/src/client/context.ts
@@ -1,3 +1,4 @@
+import type {LogContext} from '@rocicorp/logger';
 import type {NoIndexDiff} from '../../../replicache/src/btree/node.ts';
 import {assert, unreachable} from '../../../shared/src/asserts.ts';
 import type {AST} from '../../../zero-protocol/src/ast.ts';
@@ -35,16 +36,22 @@ export class ZeroContext implements QueryDelegate {
   readonly #batchViewUpdates: (applyViewUpdates: () => void) => void;
   readonly #commitListeners: Set<CommitListener> = new Set();
 
+  readonly slowMaterializeThreshold: number;
+  readonly lc: LogContext;
   readonly staticQueryParameters = undefined;
 
   constructor(
+    lc: LogContext,
     mainSources: IVMSourceBranch,
     addQuery: AddQuery,
     batchViewUpdates: (applyViewUpdates: () => void) => void,
+    slowMaterializeThreshold: number,
   ) {
     this.#mainSources = mainSources;
     this.#addQuery = addQuery;
     this.#batchViewUpdates = batchViewUpdates;
+    this.lc = lc;
+    this.slowMaterializeThreshold = slowMaterializeThreshold;
   }
 
   getSource(name: string): Source | undefined {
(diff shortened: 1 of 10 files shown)


packages/zero-client/src/client/context.ts
```diff
@@ -34,16 +37,29 @@ export class ZeroContext implements QueryDelegate {
   readonly #batchViewUpdates: (applyViewUpdates: () => void) => void;
   readonly #commitListeners: Set<CommitListener> = new Set();
 
+  readonly slowMaterializationTheshold: number | undefined;
```
arv, line 40: typo

packages/zero-client/src/client/context.ts
```diff
@@ -34,16 +37,29 @@ export class ZeroContext implements QueryDelegate {
...
     this.#batchViewUpdates = batchViewUpdates;
+    this.slowMaterializationTheshold = slowMaterializationTheshold;
+  }
+
+  get lc() {
+    return this.#lc;
+  }
+
+  get slowMaterializationThreshold() {
+    return this.slowMaterializationTheshold;
```
arv, line 62: odd.

packages/zero-client/src/client/context.ts
```diff
@@ -34,16 +37,29 @@ export class ZeroContext implements QueryDelegate {
...
+    slowMaterializationTheshold: number | undefined,
   ) {
+    this.#lc = lc;
     this.#mainSources = mainSources;
     this.#addQuery = addQuery;
     this.#batchViewUpdates = batchViewUpdates;
+    this.slowMaterializationTheshold = slowMaterializationTheshold;
+  }
+
+  get lc() {
```
arv, line 57: Can you make this data properties instead?

packages/zero-client/src/client/context.ts
```diff
@@ -34,16 +37,29 @@ export class ZeroContext implements QueryDelegate {
...
+    return this.#lc;
+  }
+
+  get slowMaterializationThreshold() {
```
arv, line 61: Can you make these data properties instead?

packages/zero-client/src/client/context.ts
arv, line 57: Do we really need to make this part of the interface?

packages/zql/src/query/query-impl.query.test.ts
```diff
@@ -30,7 +30,7 @@ import {schema} from './test/test-schemas.ts';
  * write by hand.
  */
 
-const lc = createSilentLogContext();
+const lc = new LogContext('info');
```
arv, line 33: Undo?

packages/zql/src/query/test/query-delegate.ts
```diff
@@ -39,6 +39,14 @@ export class QueryDelegateImpl implements QueryDelegate {
...
   }
 
+  get lc() {
+    return lc;
+  }
+
+  get slowMaterializationThreshold() {
+    return undefined;
+  }
+
```
arv, line 49: Maybe refactor this to:


```ts
onSlowMaterialization(duration: number, ast: ...): void;
```

Then the LogContext does not have to be part of the interface and the actual logging is handled by the delegate.

packages/zero-client/src/client/context.ts
aboodman: ? I don't see the typo. Changed name anyway.

packages/zero-client/src/client/context.ts
aboodman: Indeed fixed.

packages/zero-client/src/client/context.ts
aboodman: I tried it by alternately passing the lc into `QueryImpl` and it's a massive yak shave to pass it through everywhere.

packages/zql/src/query/test/query-delegate.ts
aboodman: Good idea, will do this!

packages/zero-client/src/client/context.ts
```diff
@@ -35,16 +36,22 @@ export class ZeroContext implements QueryDelegate {
   readonly #batchViewUpdates: (applyViewUpdates: () => void) => void;
   readonly #commitListeners: Set<CommitListener> = new Set();
 
+  readonly slowMaterializeThreshold: number;
+  readonly lc: LogContext;
```
arv, line 40: These two can be private now

arv (commented):
LGTM

PR #3954, opened 2025-03-10
chore(zero): Minor nits

arv:


packages/zero-client/src/client/query-manager.ts
```diff
@@ -198,15 +196,13 @@ export class QueryManager {
...
     };
   }
 
-  #remove(astHash: string, gotCallback: GotCallback | undefined) {
-    const entry = must(this.#queries.get(astHash));
+  #remove(entry: Entry, astHash: string, gotCallback: GotCallback | undefined) {
     if (gotCallback) {
-      const index = entry.gotCallbacks.indexOf(gotCallback);
-      entry.gotCallbacks.splice(index, 1);
+      entry.gotCallbacks.delete(gotCallback);
```
arv, line 205: Sorry. This is not right. I will undo this.

PR #3956, opened 2025-03-10
fix(zero-client): Fix regression in got callback

arv:
If the same function object is added multiple times it should be called multiple times.


grgbkr:
What PR introduced the regression?

arv:
#3954

grgbkr:
LGTM

PR #3957, opened 2025-03-11
test(zql): impl `push` tests for pg vs zql vs zqlite

tantaman:
The tests that compare ZQL vs ZQLite vs PG output only tested hydration.

Add `push` tests that push:

- add
- remove

events into sources.

`edit` to come shortly.

---

The motivation to do this now is to support testing the tldraw fix so we can have the most confidence that it works in all environments.

### DIFF
--- a/packages/z2s/src/compiler.ts
+++ b/packages/z2s/src/compiler.ts
@@ -106,9 +106,11 @@ function relationshipSubquery(
     ) as ${sql.ident(relationship.subquery.alias)}`;
   }
   return sql`(
-    SELECT ${format?.singular ? sql`` : sql`array_agg`}(row_to_json(${sql.ident(
-      `inner_${relationship.subquery.alias}`,
-    )})) FROM (${select(
+    SELECT ${
+      format?.singular ? sql`` : sql`COALESCE(array_agg`
+    }(row_to_json(${sql.ident(`inner_${relationship.subquery.alias}`)})) ${
+      format?.singular ? sql`` : sql`, ARRAY[]::json[])`
+    } FROM (${select(
       relationship.subquery,
       format,
       correlate(
--- a/packages/z2s/src/test/chinook/chinook.pg-test.ts
+++ b/packages/z2s/src/test/chinook/chinook.pg-test.ts
@@ -11,7 +11,8 @@
  * so that the test can be reproduced.
  */
 
-import {beforeAll, describe, expect, test} from 'vitest';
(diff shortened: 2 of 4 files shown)


packages/z2s/src/compiler.ts
```diff
@@ -106,9 +106,11 @@ function relationshipSubquery(
     ) as ${sql.ident(relationship.subquery.alias)}`;
   }
   return sql`(
-    SELECT ${format?.singular ? sql`` : sql`array_agg`}(row_to_json(${sql.ident(
-      `inner_${relationship.subquery.alias}`,
-    )})) FROM (${select(
+    SELECT ${
+      format?.singular ? sql`` : sql`COALESCE(array_agg`
+    }(row_to_json(${sql.ident(`inner_${relationship.subquery.alias}`)})) ${
+      format?.singular ? sql`` : sql`, ARRAY[]::json[])`
```
tantaman, line 112: found a bug in the compiler. empty relationships should be returned as `[]` not `null`.

packages/z2s/src/test/chinook/chinook.pg-test.ts
```diff
@@ -185,7 +219,221 @@ async function checkZqlAndSql(
...
+    );
+    expect(zqliteMaterialized.data).toEqualNullish(pgResult);
+    expect(zqlMaterialized.data).toEqualNullish(pgResult);
+  }
+}
+
+// Edits rows that are in the result set to random values
+// re-runs pg query after each operation.
+// Randomly choses a table on each iteration.
+function checkEditToRandom() {}
```
tantaman, line 375: coming in next pr

packages/z2s/src/test/chinook/chinook.pg-test.ts
```diff
@@ -185,7 +219,221 @@ async function checkZqlAndSql(
...
+          )}`,
+        );
+      }
+
+      const pgResult = await sql.unsafe(
+        sqlQuery.text,
+        sqlQuery.values as JSONValue[],
+      );
+      // TODO: relationships return `undefined` from ZQL and `null` from PG
+      expect(zqliteMaterialized.data).toEqualNullish(pgResult);
```
tantaman, line 330: `toEqualNullish` is because ZQL is returning `undefined` for empty `one` relationships and `z2s` returns null.

grgbkr:
Very cool.  LGTM

PR #3959, opened 2025-03-11
feat(zero-cache): Add a server-side warning for slow hydrations.

aboodman:
Add a server-side slow hydrate warning too.

### DIFF
--- a/packages/zqlite/src/db.test.ts
+++ b/packages/zqlite/src/db.test.ts
@@ -33,17 +33,17 @@ test('slow queries are logged', () => {
     [
       'warn',
       {class: 'Database', path: ':memory:', method: 'pragma'},
-      ['Slow query', 0],
+      ['Slow SQLite query', 0],
     ],
     [
       'warn',
       {class: 'Database', path: ':memory:', method: 'exec'},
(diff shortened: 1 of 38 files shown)


packages/zqlite/src/db.test.ts
```diff
@@ -102,7 +102,7 @@ test('slow queries are logged', () => {
         method: 'iterate',
         type: 'total',
       },
-      ['Slow query', 200],
+      ['Slow SQLite query', 200],
```
tantaman, line 105: Not quite.

`method: iterate` includes the cost of the downstream IVM pipeline rooted at the given SQLite query.

So

```ts
issue.related('labels').related('comments').related('owner');
```

could log:

```
SLOW QUERY `SELECT * FROM issue` METHOD `iterate`
```

and that includes the total cost of the entire pipeline.

```
SLOW QUERY `SELECT * FROM comment WHERE issue_id = ?`
```

would be the total time to iterate the rows coming out of the `comment` branch of that query.

tantaman: It confuses everyone so we can also remove the `iterate` logging.

PR #3961, opened 2025-03-11
feat(custom-mutators): patch IVM sources in `persist.ts`, allow providing a `sourceRead` to `LazyRead`

tantaman:
# Problem:

Persist.ts persists the mutations from the current client onto the perdag. Mutations can do reads so we need to create IVM sources at the same head we'll be rebasing onto.

To get IVM into the right state, we diff from `ivmMain.hash` to `perdagHead.hash`.

When doing this diff, `perdagHead` is not loaded into the memdag yet so `LazyStore` must do a "source read" against the persistent store. See below "_getSourceRead()":

https://github.com/rocicorp/mono/blob/148b1608c353e49c934cf6edd26950eeafb03bc3/packages/replicache/src/dag/lazy-store.ts#L249-L254

The problem with this is that we already have a write transaction open against the persistent store when going to do the diff.

https://github.com/rocicorp/mono/blob/148b1608c353e49c934cf6edd26950eeafb03bc3/packages/replicache/src/persist/persist.ts#L145-L154

When the LazyStore tries to load the `perdagHead` from the persistent store, it opens a new IDB read transaction which completes on the next tick of the event loop. Because of this releasing of the event loop, the outer write transaction closes and all further writes fail.

```ts
persist.perdagWriteTx {
   ivm.computeDiff {
     lazyStore.getSourceRead {
        await openIdbReadTransaction // at this point `persist.perdadgWriteTx` auto-commits
     }
   }

   // all code that write to the perdag after this point fails due to the auto-commit of the write tx.
}
```

The simplest solution to this problem (implemented here) seems to be to pass a `sourceRead` to `LazyRead`. If `sourceRead` is provided, `lazyRead` uses that when it needs to fall back rather than trying to open a new transaction.

### DIFF
--- a/packages/replicache/src/dag/lazy-store.ts
+++ b/packages/replicache/src/dag/lazy-store.ts
@@ -159,7 +159,7 @@ export class LazyStore implements Store {
     this.#assertValidHash = assertValidHash;
   }
 
-  async read(): Promise<LazyRead> {
+  async read(sourceRead?: Read): Promise<LazyRead> {
     const release = await this.#rwLock.read();
     return new LazyRead(
       this.#heads,
@@ -168,6 +168,7 @@ export class LazyStore implements Store {
       this.#sourceStore,
       release,
       this.#assertValidHash,
+      sourceRead,
     );
   }
 
@@ -215,6 +216,7 @@ export class LazyRead implements Read {
   readonly #release: () => void;
   #closed = false;
   readonly assertValidHash: (hash: Hash) => void;
+  readonly #sourceReadOwnedByCaller: boolean;
 
   constructor(
     heads: Map<string, Hash>,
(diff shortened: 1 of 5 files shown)


packages/replicache/src/dag/lazy-store.ts
```diff
@@ -159,7 +159,7 @@ export class LazyStore implements Store {
     this.#assertValidHash = assertValidHash;
   }
 
-  async read(): Promise<LazyRead> {
+  async read(sourceRead?: Read): Promise<LazyRead> {
```
grgbkr, line 162: Would be good to add a test for this.

tantaman:
There's a bug in Replicache's `persist`.

Still investigating but the perdag LMID does not move forward in the case where the memdag snapshot is behind the perdag snapshot.

Here:

https://github.com/rocicorp/mono/blob/ca0ff42e3ca2986208951cff2faa399f97346f43/packages/replicache/src/persist/persist.ts#L217-L226

Replicache is rebasing mutations that have already been rebased into `newMainClientGroupHeadHash` before 🤔

packages/replicache/src/persist/persist.ts
```diff
@@ -144,6 +144,11 @@ export async function persistDD31(
   }
 
   let memdagBaseSnapshotPersisted = false;
+  const zeroDataForMemdagBaseSnapshot =
+    getZeroData === undefined
+      ? undefined
+      : await getZeroData('rebase', memdagBaseSnapshot.chunk.hash);
```
arv, line 150:
```suggestion
    getZeroData && await getZeroData('rebase', memdagBaseSnapshot.chunk.hash);
```

arv (approved):
LGTM

PR #3962, opened 2025-03-11
feat(zero-client): 100x speedup: lazy creation of queries and crud for mutators

tantaman:
When users have really large schemas it gets expensive to create all the `schemCRUD` and `schemaQuery` objects up-front.

# Before:
CleanShot 2025-03-11 at 14 46 50

# After:
CleanShot 2025-03-11 at 14 47 01

### DIFF
--- a/packages/zero-client/src/client/custom.ts
+++ b/packages/zero-client/src/client/custom.ts
@@ -97,8 +97,6 @@ export class TransactionImpl implements Transaction<Schema> {
     this.clientID = repTx.clientID;
     this.mutationID = repTx.mutationID;
     this.reason = repTx.reason === 'initial' ? 'optimistic' : 'rebase';
-    // ~ Note: we will likely need to leverage proxies one day to create
-    // ~ crud mutators and queries on demand for users with very large schemas.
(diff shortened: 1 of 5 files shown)


arv:
FYI, there is a way to run vitest bench so that it does the comparison between the branches.

```
  --compare <filename>                                  Benchmark output file to compare against 
  --outputJson <filename>                               Benchmark output file 
```

packages/zero-client/src/client/custom.ts
```diff
@@ -131,11 +148,50 @@ function makeSchemaCRUD(
...
-  for (const [name] of Object.entries(schema.tables)) {
-    mutate[name] = makeTableCRUD(schema, name, tx, ivmBranch);
-  }
-  return mutate;
+  // Only creates the CRUD mutators on demand
+  // rather than creating them all up-front for each mutation.
+  return new Proxy(
+    {},
+    {
+      get(target: Record<string, TableCRUD<TableSchema>>, prop: string) {
```
arv, line 156: Do we want to support enumeration too?

arv (commented):
LGTM

PR #3963, opened 2025-03-11
fix(zql): take split push fixes

grgbkr:
In several cases Take splits a received push change in to two output push changes.  
These fixes ensure that when a fetch is performed by an output receiving a push change from Take the fetch results reflect the push changes received so far, and no more.

For take this can be achieved by careful ordering of the two push changes and the update of take state. 

Updates all take test which output multiple push changes to check the results of fetching during each push change.


tantaman (commented):
makes sense. lgtm.

PR #3964, opened 2025-03-12
fix(zbugs): Fix a few react errors.

aboodman:
1. defaultProps is deprecated
2. Cannot call setState() from within render. These callbacks were calling setState internally. I'm not sure what the React way (tm) to do this is, since I can't know inside a component what a callback I pass does. But this works.


arv (commented):
LGTM

PR #3968, opened 2025-03-12
preview(zbugs): custom mutators

tantaman:


apps/zbugs/api/push.ts
```diff
@@ -38,9 +40,42 @@ class Transaction implements DBTransaction<postgres.TransactionSql> {
...
 const mutatorSql = postgres(process.env.ZERO_UPSTREAM_DB as string);
 
+const transporter = nodemailer.createTransport({
+  host: 'smtp.ethereal.email',
+  port: 587,
+  auth: {
+    user: 'micaela35@ethereal.email',
+    pass: 'cq8t7nRzYW61YjmX56',
+  },
+});
```
tantaman, line 50: this'll be removed. Wanted to check mutators w/ code that can only run on the server.

tantaman, line 50: this'll be removed. Wanted to check mutators w/ code that can only run on the server.

apps/zbugs/shared/mutators.ts
```diff
@@ -0,0 +1,218 @@
...
+        created,
+        creatorID,
+        modified,
+        open: true,
+        visibility: 'public',
+      });
+    },
+
+    async update(tx, change: UpdateValue<typeof schema.tables.issue>) {
+      await assertIsAdminOrCreator(tx, tx.query.issue, change.id);
```
aboodman, line 58: how does the app get a new auth token.

apps/zbugs/shared/validators.ts
```diff
@@ -0,0 +1,70 @@
...
+import {jwtVerify, type JWK, type JWTPayload} from 'jose';
+import type {schema} from './schema.ts';
+import {must} from '../../../packages/shared/src/must.ts';
+import {assert} from '../../../packages/shared/src/asserts.ts';
+
+const publicJwk = JSON.parse(
+  process?.env?.VITE_PUBLIC_JWK ?? import.meta.env.VITE_PUBLIC_JWK,
+) as JWK;
+
+export async function verifyToken(
```
aboodman, line 11: Few things here:

- kinda weird to re-parse this constantly. can we find a nice pattern to not do that?
- in case where token is *not* encrypted, should we skip verification? this prevents a little complexity of having to get the public key to the client.
- as discussed i think we should get rid of `userID` to simplify the model. I guess technically we should deprecate it to avoid a breaking change for now. Kind of a different task but maybe you could do as long as you're in there.

aboodman (commented):
Looking good. Sorry for rambly meeting! I'm in sleep debt. Fun catching up :).

apps/zbugs/shared/mutators.ts
tantaman: today `mutagen` throws an auth error that the zero-client handles, allowing the developer to refresh the auth token.

We'll need to add an error type for the custom mutator protocol to indicate auth problems.

PR #3972, opened 2025-03-12
fix(zero-client): Send updated TTL when changed 

arv:
When we use `useQuery` (both Solid and React) we check if we already have a view for the query. If we do we reuse the view but update the ttl as needed.

To do this we have to add updateTTL to the AdvancedQuery. This in turn calls `updateServerQuery` on the `QueryDelegate` which in turn calls `update` on the `QueryManager`.


arv:
I need to clean this up further but I think it works.

arv:
I kind of want to pipe through the clientID (somehow) to Solid so we can reuse views like we do in React.

packages/zero-client/src/client/context.ts
```diff
@@ -43,12 +46,12 @@ export class ZeroContext implements QueryDelegate {
   constructor(
     lc: LogContext,
     mainSources: IVMSourceBranch,
-    addQuery: AddQuery,
+    queryManager: QueryManager,
```
aboodman, line 49: I think maybe `ZeroContext` isn't buying us much and should just be folded into `Zero`. I get why `UpdateManager` exists - it encapsulates some logic. But `ZeroContext` doesn't really seem to play a role that's clearly separate from `Zero`.

packages/zero-client/src/client/query-manager.ts
```diff
@@ -202,6 +185,34 @@ export class QueryManager {
     };
   }
 
+  update(ast: AST, ttl: TTL) {
```
aboodman, line 188: Cool. It seems like this could play a role when queries get stable IDs separate from their hash. Then we could update the AST too.

aboodman (approved):

packages/zero-client/src/client/context.ts
```diff
@@ -14,13 +14,16 @@ import type {
...
 import {ENTITIES_KEY_PREFIX, sourceNameFromKey} from './keys.ts';
+import type {QueryManager} from './query-manager.ts';
 
 export type AddQuery = (
   ast: AST,
   ttl: TTL,
   gotCallback: GotCallback | undefined,
 ) => () => void;
 
+export type UpdateQuery = (ast: AST, ttl: TTL) => void;
```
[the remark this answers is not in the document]
arv: I had to bring it back because initialization ordering.

PR #3973, opened 2025-03-12
chore(vitest): allow `console.log` through

tantaman:
@arv  - is this ok to flip?


arv:
Can you not override this on the command line for the cases where you want to see the log?

arv:
Do you want them on the ci?

tantaman:
> Can you not override this on the command line for the cases where you want to see the log?

good point.

arv:
I think --silent=false is not working.

Feel free to remove `silent: true`

PR #3977, opened 2025-03-13
fix(zql): Fix ref count for edit in applyChange

arv:
The edit branch was incomplete.

### Singular

We replace the old object with the new object transferring the ref count.

### Not Singular

If we edit the value of a non primary column, we transfer the ref count to the new object.

If we edit the value of a primary column, we need to decrement the ref count of the old object and increment the ref count of the new object.

### DIFF
--- a/packages/zql/src/ivm/view-apply-change.ts
+++ b/packages/zql/src/ivm/view-apply-change.ts
@@ -301,27 +301,19 @@ export function applyChange(
               schema.compareRows,
             );
             let rc = 1;
-            let newEntry: Entry;
+            const newEntry = makeEntryPreserveRelationships(
+              change.node.row,
+              oldEntry,
+              format.relationships,
+            );
             let deleteCount = 0;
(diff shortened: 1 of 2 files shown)


packages/zql/src/ivm/view-apply-change.ts
```diff
@@ -278,7 +284,14 @@ export function applyChange(
           );
           assert(found, 'node does not exists');
           const oldEntry = view[pos];
-          view.splice(pos, 1);
+          const rc = must(refCountMap.get(oldEntry));
```
grgbkr, line 287: I think this code is repeated for remove, so maybe a little helper can be factored out.

grgbkr (commented):
LGTM

packages/zql/src/ivm/view-apply-change.ts
arv: Good catch

PR #3978, opened 2025-03-13
fix(zql): enable `upsert` for memory and table source

tantaman:


packages/zqlite/src/table-source.ts
```diff
@@ -321,21 +331,39 @@ export class TableSource implements Source {
...
 
-  push(change: SourceChange): void {
+  push(change: SourceChange | SourceChangeSet): void {
     for (const _ of this.genPush(change)) {
       // Nothing to do.
     }
   }
 
-  *genPush(change: SourceChange) {
+  *genPush(change: SourceChange | SourceChangeSet) {
```
arv, line 340: Can you add SourceChangeSet to SourceChange? Is there a reason not to?

arv (commented):
LGTM

PR #3979, opened 2025-03-13
feat(zql): add astToZQL function

arv:
This adds a function that takes an AST and returns a string representing the code needed to build that AST using the Query interface.

This is going to be used for the inspector tool.

### DIFF
--- a/packages/z2s/src/test/data-gen.ts
+++ b/packages/z2s/src/test/data-gen.ts
@@ -1,5 +1,0 @@
-import type {Row} from '../../../zero-protocol/src/data.ts';
-
-export type Dataset = {
-  [table: string]: Row[];
-};
--- a/packages/zql/src/query/ast-to-zql.test.ts
+++ b/packages/zql/src/query/ast-to-zql.test.ts
@@ -1,0 +1,518 @@
+import {en, Faker, generateMersenne53Randomizer} from '@faker-js/faker';
+import {expect, test} from 'vitest';
+import {type AST} from '../../../zero-protocol/src/ast.ts';
+import {astToZQL} from './ast-to-zql.ts';
+import {staticQuery} from './static-query.ts';
+import {generateQuery} from './test/query-gen.ts';
+import {generateSchema} from './test/schema-gen.ts';
+import {ast} from './test/util.ts';
+
+test('simple table selection', () => {
+  const ast: AST = {
+    table: 'issue',
+  };
+  expect(astToZQL(ast)).toMatchInlineSnapshot(`""`);
+});
+
+test('simple where condition with equality', () => {
+  const ast: AST = {
+    table: 'issue',
+    where: {
+      type: 'simple',
+      left: {type: 'column', name: 'id'},
+      op: '=',
+      right: {type: 'literal', value: 123},
+    },
+  };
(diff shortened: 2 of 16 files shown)


packages/z2s/src/test/data-gen.ts
arv, line 1: I ended up moving these from z2s to zql.

packages/zql/src/query/ast-to-zql.test.ts
```diff
@@ -0,0 +1,518 @@
...
+      'schema',
+      'tableName',
+      `return staticQuery(schema, tableName)${code}`,
+    )(staticQuery, schema, ast(q).table);
+    expect(ast(q2)).toEqual(ast(q));
+  }
+
+  expect(codes).toMatchInlineSnapshot(`
+    [
+      ".where('nudge', 'IS NOT', false)",
```
arv, line 506: generateQuery doesn't generate very advanced queries but it is still really cool to able to do this.

packages/zql/src/query/ast-to-zql.ts
```diff
@@ -0,0 +1,275 @@
...
+  }
+  return String(literal.value);
+}
+
+function transformParameter(param: Parameter): string {
+  const fieldStr = Array.isArray(param.field)
+    ? `[${param.field.map(f => `'${f}'`).join(', ')}]`
+    : `'${param.field}'`;
+
+  return `authParam(${fieldStr})`;
```
arv, line 274: What should we do here?

packages/zql/src/query/ast-to-zql.ts
```diff
@@ -0,0 +1,275 @@
...
+    return `.where(({exists, not}) => not(exists('${relationship}')))`;
+  }
+  args.add('not');
+  args.add('exists');
+
+  return `not(exists('${relationship}')))`;
+}
+
+function extractRelationshipName(related: CorrelatedSubquery): string {
+  // Try to extract the relationship name from the alias
```
arv, line 204: I'm not really sure what the alias is used for.

packages/zql/src/query/ast-to-zql.ts
```diff
@@ -0,0 +1,275 @@
...
+  const {alias} = related.subquery;
+  if (alias && alias.startsWith(SUBQ_PREFIX)) {
+    return alias.substring(SUBQ_PREFIX.length);
+  }
+
+  // Fallback: use the table name
+  return related.subquery.table;
```
tantaman, line 211: There should always be an alias for a subquery. Should be safe to throw here.

packages/zql/src/query/ast-to-zql.ts
```diff
@@ -0,0 +1,275 @@
...
+ * This is useful for debugging and understanding queries.
+ *
+ * @example
+ * ```
+ * const ast = query.issue.where('id', '=', 123)[astForTestingSymbol];
+ * console.log(astToZQL(ast)); // outputs: .where('id', '=', 123)
+ * ```
+ */
+export function astToZQL(ast: AST): string {
+  let code = '';
```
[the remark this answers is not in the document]
arv: I had that but if you look at the test in how the code is evaluated I think this is more flexible.

PR #3981, opened 2025-03-13
chore(custom-mutators): read-your-writes

tantaman:
custom mutators couldn't read their own writes making the interactive transaction not so great.

This now forks main and applies the writes to that fork while in an `initial` mutation. Main remains untouched until the mutation completes giving us transaction isolation.


packages/zero-client/src/client/custom.test.ts
```diff
@@ -128,3 +135,158 @@ test('custom mutators write to the local store', async () => {
   issues = await z.query.issue.run();
   expect(issues.length).toEqual(0);
 });
+
+test('custom mutators can query the local store during an optimistic mutation', async () => {
+  const z = zeroForTest({
+    schema,
+    mutators: {
+      issue: {
+        create: async (tx, args: InsertValue<typeof schema.tables.issue>) => {
```
arv, line 144: OT, does the following work now:

```js
mutators: doSomething(tx) { ... {
```

or is the 2 level definition still required?

packages/zero-client/src/client/custom.test.ts
```diff
@@ -128,3 +135,158 @@ test('custom mutators write to the local store', async () => {
...
+    await z.mutate.issue.createAndReadCreated({
+      id: '1',
+      title: 'foo',
+      description: '',
+      closed: false,
+    });
+
+    const issue = must(await z.query.issue.where('id', '1').one().run());
+    expect(issue.title).toEqual('foo updated');
+    expect(issue.description).toEqual('updated');
```
arv, line 260: cool

arv (approved):
LGTM

packages/zero-client/src/client/custom.test.ts
tantaman: It doesn't. I'll fix that today.

PR #3982, opened 2025-03-14
chore(zero-cache): restore support for partial updates

darkgnotic:
The change to support resumptive replication (#3937) assumed that Postgres always sends the full row for UPDATE messages. However, this is not true for rows with large values that are TOASTed, in which unchanged values are sent with an "unchanged" tag, but the value is otherwise not included:

https://www.postgresql.org/docs/current/protocol-logicalrep-message-formats.html#PROTOCOL-LOGICALREP-MESSAGE-FORMATS-TUPLEDATA

To restore support for updates of these rows, revert to using `UPDATE` statements by default, and only falling back to an `INSERT` / `UPSERT` if the former did not modify any rows (i.e. resumptive replication).

It is a TODO to figure out how to make resumptive replication work for rows with TOASTed values.

### DIFF
--- a/packages/zero-cache/src/services/replicator/change-processor.ts
+++ b/packages/zero-cache/src/services/replicator/change-processor.ts
@@ -395,7 +395,7 @@ class TransactionProcessor {
 
   // Updates by default are applied as UPDATE commands to support partial
   // row specifications from the change source. In particular, this is needed
-  // to handle updates for which large TOAST'd values are not sent:
+  // to handle updates for which large TOASTed values are not sent:
   //
   // https://www.postgresql.org/docs/current/protocol-logicalrep-message-formats.html#PROTOCOL-LOGICALREP-MESSAGE-FORMATS-TUPLEDATA
   //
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/services/replicator/change-processor.ts
```diff
@@ -393,36 +393,55 @@ class TransactionProcessor {
...
+        SET ${setExprs.join(',')}
+        WHERE ${conds.join(' AND ')}
+      `,
+      [...Object.values(row), ...Object.values(currKey)],
+    );
+
+    // If the UPDATE did not affect any rows, perform an UPSERT of the
+    // new row for resumptive replication.
+    if (changes === 0) {
+      this.#upsert(table, row);
```
grgbkr, line 443: Should we try to detect if any of the columns in this case are TOASTed and error?

darkgnotic: Yeah, I'll be present something in the upcoming resharding brainstorm.

PR #3998, opened 2025-03-14
chore(mutators): make auth token available on the client transactions

tantaman:
This reduces the amount of boilerplate we need to write by allowing the server and client to share more code.

### DIFF
--- a/packages/replicache/src/replicache-impl.ts
+++ b/packages/replicache/src/replicache-impl.ts
@@ -215,7 +215,7 @@ export class ReplicacheImpl<MD extends MutatorDefs = {}> {
   pushURL: string;
 
   /** The authorization token used when doing a push request. */
-  auth: string;
+  #auth: string;
 
   /** The name of the Replicache database. Populated by {@link ReplicacheOptions#name}. */
   readonly name: string;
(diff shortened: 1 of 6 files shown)


packages/replicache/src/replicache-impl.ts
```diff
@@ -424,7 +436,7 @@ export class ReplicacheImpl<MD extends MutatorDefs = {}> {
       onClientsDeleted = () => {},
     } = implOptions;
     this.#zero = implOptions.zero;
-    this.auth = auth ?? '';
+    this.#auth = auth ?? '';
```
tantaman, line 439: why not undefined? 🤔

arv: If I recall correctly it was only so that the type is always string.

packages/replicache/src/replicache-options.ts
```diff
@@ -264,6 +267,8 @@ export type ZeroReadOptions = {
  * Prevents us from creating any direct dependencies on Zero.
  */
 export interface ZeroOption {
+  set auth(auth: string);
```
arv, line 270: maybe this should just be `auth: string` in the type. Using a setter seems like an implementation detail

arv (approved):
LGTM

PR #4000, opened 2025-03-14
fix(zero-cache): ensure lmid changes are synced when there are no desired queries

grgbkr:
Fixes https://bugs.rocicorp.dev/issue/3628


darkgnotic (commented):
LGTM

PR #4003, opened 2025-03-14
chore(zero-cache): add support for unparsed json values in the change stream

darkgnotic:
This is the first step to adding support for unparsed JSON in the change stream, which is the equivalent of the optimization done for initial-sync (#3919), applied to incremental replication.

Avoiding JSON parsing and instead ferrying values directly as JSON strings from Postgres (through the change-streamer, to the replicator,) to the replica improves performance by 20 ~ 25% in the change-streamer and 25 ~ 30% in the replicator:
 
parsed vs stringified JSON

Rolling this out is a two-step process. 

1. In the first step (this PR), an option is introduced into the change protocol to declare, per transaction, the format of the JSON values in the comprising changes, and a client (i.e. view-syncer) supporting this protocol connects to the change-streamer at `v1`. The server (replication-manager), however, starts off supporting both `v0` and `v1`, and thus does not apply the operation, since `v0` clients do not handle it correctly.
2. In a subsequent rollout, the server (replication-manager) will stop supporting `v0`. With the guarantee that only `v1` clients can connect to the change stream, the replication-manager can safely configure its postgres client to skip the parsing of JSON/JSONB values and instead ferry them directly as strings through the various layers to the SQLite replica.

@cbnsndwch if your stack (replication-manager and view-syncers) is fully rolled out with this change, you can begin formatting your changes with stringified objects and `json: 's'` to get the performance gains.

### DIFF
--- a/packages/zero-cache/src/types/lite.ts
+++ b/packages/zero-cache/src/types/lite.ts
@@ -20,6 +20,11 @@ function columnType(col: string, table: LiteTableSpec) {
   return spec.dataType;
 }
 
+export const JSON_STRINGIFIED = 's';
+export const JSON_PARSED = 'p';
+
+export type JSONFormat = typeof JSON_STRINGIFIED | typeof JSON_PARSED;
+
 /**
  * Creates a LiteRow from the supplied RowValue. A copy of the `row`
  * is made only if a value conversion is performed.
@@ -27,6 +32,7 @@ function columnType(col: string, table: LiteTableSpec) {
 export function liteRow(
   row: RowValue,
   table: LiteTableSpec,
+  jsonFormat: JSONFormat,
 ): {row: LiteRow; numCols: number} {
   let copyNeeded = false;
   let numCols = 0;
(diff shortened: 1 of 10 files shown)


packages/zero-cache/src/types/lite.ts
```diff
@@ -76,14 +82,14 @@ export function liteValues(
...
-  jsonAsString?: 'json-as-string',
+  jsonFormat: JSONFormat,
 ): LiteValueType {
   if (val instanceof Uint8Array || val === null) {
     return val;
   }
   const valueType = dataTypeToZqlValueType(pgType);
   if (valueType === 'json') {
-    if (jsonAsString && typeof val === 'string') {
+    if (jsonFormat === JSON_STRINGIFIED && typeof val === 'string') {
```
tantaman, line 92: Should `typeof val === 'string'` always be the case when `jsonFormat === JSON_STRINGIFIED`? In other words, is it an error if typeof val is not a string here?

If not, there will be edge cases around JSON.stringified strings.

CleanShot 2025-03-19 at 15 01 14

darkgnotic: Currently this is the case, but the conditional is meant to handle future non-"JSON" data types that end up being represented as objects or arrays (e.g. array types like `text[]`). These (I believe) will map to the "json" ValueType but will arrive from postgres as a `string[]` which we'll then have to manually stringify.

PR #4005, opened 2025-03-15
chore(zero-protocol): bump min protocol to v6 and clean up obsolete code

darkgnotic:
Bumps to minimum supported `/sync/v.../connect` wire protocol version to `v6`, released in zero v0.16, which made the `pokeStart.cookie` optional and in favor of `pokeEnd.cookie`. (#3832)

With that minimum supported version, the `cookie` field can be removed from the `PokeStart` message and the server no longer sends it.

The logic which skips empty flushes (to avoid quadratic writes) is now also unconditional, since clients started supporting that protocol from v5. (#3735)


arv (commented):
LGTM

PR #4007, opened 2025-03-15
chore: Use the playwright docker container

arv:


arv:
This takes the playwright step from 47s to 26s... but the `--changed` is not working so we end up spending 1m24s in the stead of 47s...

PR #4012, opened 2025-03-17
fix(zero): Fix another issue with view ref counting

arv:


tantaman (commented):
I'm having a hard time finding the exact functionality that changed. It looks like a refactor.

Also, any tests for the case being fixed?

PR #4013, opened 2025-03-17
 fix(zero): Fix issue with view ref counting

arv:
`applyChange` no longer stores the ref count in a `WeakMap` side table. Instead it uses `Entry & {[refCountSymbol]: number}`.

The main motivation for this change is so that bindings like Solid and Vue do not need to unwrap things.

This marks the `applyChange` function that uses a `RefCountMap` as deprecated.


tantaman (commented):
👍

grgbkr (commented):
LGTM

PR #4015, opened 2025-03-17
chore(zero-cache): add test to verify timestamptz behavior in mutagen

darkgnotic:


packages/zero-cache/src/services/mutagen/mutagen.pg-test.ts
```diff
@@ -740,6 +742,8 @@ describe('processMutation', {timeout: 15000}, () => {
                 value: {
                   id: '1',
                   num: 23.45,
+                  time1: 1742246216309,
+                  time2: '2025-03-17T21:18:42.792Z',
```
tantaman, line 746: The client would never send a string for a timestamp field, correct? It should always be sending an int.

darkgnotic: Yes, I believe our crud mutators work that way (but the server side has code to handle either).

PR #4016, opened 2025-03-17
chore(zero-cache): reduce allocations when making keys for replica identity full

darkgnotic:
As suggested in https://github.com/rocicorp/mono/pull/3935/files#r1993637868


tantaman (commented):
👍

PR #4017, opened 2025-03-18
chore: Enable Firefox again?

arv:


arv:
https://github.com/vitest-dev/vitest/issues/7377

PR #4020, opened 2025-03-18
fix(zql): fix `single output already exists`

grgbkr:
`applyViewChange` asserts that when format is singular, there is only ever a single node.  When this assert fails the error message is `single output already exists`. 

This change https://github.com/rocicorp/mono/commit/b453073597931d53d12d259f64ba14f05dab1c88 introduced a regression that caused this assert to fail for queries or subqueries with limit 1 (and thus format singular), when the view had one node for the query/subquery, and a new node was pushed that was less than the existing node, and thus should replace it.

The above change made Take first push the `add` of the new node, temporarily increasing its output size to `limit + 1`, before pushing the `remove` of the old  node, brining its size back to `limit`.  This was done because it allowed take to maintain the correct push/fetch semantics without any additional overlay state.  However, it caused `applyViewChange` to fail this assert when format is singular as it would see an increase to size 2.

This change updates Take to maintain the invariant that its output size is always <= limit, even mid split-push.  It introduces a minimal amount of overlay state to achieve this.

This likely also fixes: https://bugs.rocicorp.dev/issue/3630

### DIFF
--- a/packages/zql/src/builder/builder.test.ts
+++ b/packages/zql/src/builder/builder.test.ts
@@ -700,28 +700,6 @@ test('self-join edit', () => {
               {
                 "relationships": {},
                 "row": {
-                  "id": 2,
-                  "name": "erik",
-                  "recruiterID": 1,
-                },
-              },
-            ],
-          },
-          "row": {
-            "id": 3,
-            "name": "greg",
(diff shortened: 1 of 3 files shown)


tantaman (commented):
lgtm

PR #4021, opened 2025-03-18
chore: Force pg-types version to 4

arv:
By modifying package-lock.json, we can force the version of pg-types


darkgnotic (commented):
Thanks @arv.

PR #4022, opened 2025-03-18
chore(zqlite,z2s): map to/from server names in z2s

tantaman:
z2s needs to understand how to map to/from server names as `pg` will be using server names for its columns and tables. `z2s.compile` now takes the schema as a parameter so it can map names as it builds queries.

The `z2s` test suite is an integration test that:

1. Creates a PG DB
2. Uses `initialSync` to create a replica

Due to (2), that replica will be using server names so this PR also adds support for server names in zqlite tests. This more faithfully reproduces our production setup.


darkgnotic (commented):
LGTM

PR #4025, opened 2025-03-18
chore(zero-protocol): bump protocol to v8 to safely drop support for v5

darkgnotic:
This should have been together with #4005


grgbkr:
LGTM

PR #4035, opened 2025-03-19
feat(zbugs): swap to custom mutators

tantaman:
We're ready to port zbugs to custom mutators.

Missing features:
- threading mutation errors from the api server back to the client
- adding an auth error type to the protocol so the api server can cause the token to be refreshed on the client

Bugs:
- `related` does not do type conversion in `z2s` (zql 2 sql). E.g., dates come back as strings rather than milliseconds when doing reads, which contain `related` calls, on the server via ZQL. This doesn't impact zbugs at the moment.

# A puzzle:

Setting `created` and `modified` for an issue in the custom mutator does indeed cause an i-loop until the API server responds.

The steps are as follows:
1. Mutation runs to create the issue
2. UI navigates to the issue page
3. `useQuery` runs to set up `j/k` queries
4. This writes a new query
5. This causes the server to poke
6. This causes a rebase of outstanding mutations
7. Rebasing the issue create mutation updates created and modified again
8. This updates the query for `j/k` since `start` takes in the `issue` row which has a new `created` and `modified` time.
9. This writes a new query
10. Back to step 5

https://github.com/user-attachments/assets/d5f6aa37-f4d7-4fdb-9662-3bae74080a28

# 🤔
- one option is to not rebase mutations if only queries change and no data changes. This doesn't fully prevent the issue, however. A query that returns data can cause this loop.
- memoize the input to the `j/k` queries based on `issue.id` so we ignore created/modified time changes
- definitely add some brakes if queries are cycling too fast, throw an error.

### DIFF
--- a/apps/zbugs/api/push.ts
+++ b/apps/zbugs/api/push.ts
@@ -6,7 +6,8 @@ import {
   type Row,
 } from '@rocicorp/zero/pg';
 import postgres, {type JSONValue} from 'postgres';
-import {schema} from '../schema.ts';
+import {schema} from '../shared/schema.ts';
+import {createMutators} from '../shared/mutators.ts';
 
 class Connection implements DBConnection<postgres.TransactionSql> {
   readonly #pg: postgres.Sql;
@@ -40,6 +41,6 @@ const mutatorSql = postgres(process.env.ZERO_UPSTREAM_DB as string);
 
 export const pushHandler: PushHandler = createPushHandler({
   dbConnectionProvider: () => new Connection(mutatorSql),
-  mutators: {},
+  mutators: createMutators(process.env.VITE_PUBLIC_JWK as string),
   schema,
 });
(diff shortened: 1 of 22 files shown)


arv (approved):
LGTM 

Very nice

arv:
Setting random values in the mutation body is not a good idea. Rebasing should be idempotent.

I do agree that we might need to add some checks to prevent ilooping.

aboodman:
> Setting random values in the mutation body is not a good idea.

I disagree with this. I like the fact that mutations can be impure. However...

> Rebasing should be idempotent.

This feels like a clue. It is definitely wrong/weird for the created/modified dates in the UI to change because of rebase.

OTOH, it is _also_ wrong/weird to pass them to the server since they should not be used by the server.

Hm.

===

On option is to add the concept of defaults that has been discussed: https://bugs.rocicorp.dev/issue/3465. This would allow a structured way in the system to handle created/modified everywhere. I am not sure if this is a complete solution (I am not sure if other types of parameters would have same problem as created/modified here, but would not be fixed by the defaults proposal). I guess I also just don't like the idea of fixing an API problem by adding a totally different concept, rather than fixing the API.

===

For now let's leave it as-is. I agree it's the best solution for now.

PR #4047, opened 2025-03-20
Update create-canary.js

arv:


packages/zero/tool/create-canary.js
```diff
@@ -99,7 +99,7 @@ function bumpCanaryVersion(version) {
 
 // To do a maintenance/cherry-pick release:
 // - create a maintenance release from tag you want to patch, like
-//   `maint/zero/vX.Y`
+//   `main/zero/vX.Y`
```
aboodman, line 102: CleanShot 2025-03-20 at 04 49 57@2x

It was right as-is. "maint" is short for "maintenance".

arv: hehe, I'll undo

PR #4058, opened 2025-03-20
chore(zero-cache): rewrite logical replication on postgres.js

darkgnotic:
This grafts the protocol parsing logic from [pg-logical-replication](https://github.com/kibae/pg-logical-replication/tree/main/src/output-plugins/pgoutput) onto a connection managed by a `postgres.js` client rather than a `pg` client.

The benefits of this rewrite include:
* **Better connection handling**: postgres.js supports an `ssl: prefer` option that automatically detects if ssl is supported. We currently have to mimic this on the `pg` client by trying SSL first and falling back to non-SSL
* **Flow control**: The new implementation uses the postgres.js `.readable()` API to pipeline the replication stream to the application's change Subscription, allowing the change-streamer to exert backpressure on the postgres. (This reuses the flow control mechanism for the websocket connection to custom change sources.)
* **Direct type parser configuration**: Type parsing is now directly configured instead of using the static `pg-types` library, avoiding library linkage mishaps that can arise in certain build environments (e.g. https://discord.com/channels/830183651022471199/1350643925836365917)

This PR only defines the new library. It will replace the status quo in a subsequent PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27016,6 +27016,7 @@
       },
       "devDependencies": {
         "@faker-js/faker": "^9.6.0",
+        "@rocicorp/eslint-config": "^0.7.0",
         "@rocicorp/prettier-config": "^0.3.0",
         "@vitest/runner": "3.0.8",
         "vitest": "3.0.8"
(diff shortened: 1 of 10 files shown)


arv (commented):
LGTM

PR #4059, opened 2025-03-20
feat(zero-cache): add flow control to logical replication (a.k.a. farewell, "pg")

darkgnotic:
Migrate to the new flow-control-capable logical replication library (#4058) backed by a postgres.js client.

This makes logical replication flow-control capable. In a subsequent CL, the change-streamer will be modified to exert back-pressure when the number of uncommitted changes exceeds a certain threshold, which will now work for both custom change sources (via websocket back-pressure) and the pg change-source (via this logical replication library).

The other behavioral change is that acks sent to postgres no longer use the misguided (LSN + 1) method (which is what the [pg-logical-replication library does](https://github.com/kibae/pg-logical-replication/blob/c55abddc62eadd61bd38922037ecb7a1469fa8c3/src/logical-replication-service.ts#L189)). This practice is based on a [misreading of the postgres docs](https://www.dolthub.com/blog/2024-03-08-postgres-logical-replication/), and has been a source of confusion and bugs with the `confirmed_flush_lsn` sometimes being reset on '0/1'. Unit tests confirm that postgres resumes as desired when the `confirmed_flush_lsn` is the `commitEndLsn` itself; no "+ 1" is necessary.

Other resulting benefits / cleanup:
* The SSL / not-SSL retry loop is eliminated, as postgres.js's `ssl: prefer` feature applies to logical replication too. There is only one library used to connect to postgres now.
* We will avoid module mismatch situations that can occur in certain build environments where multiple instances of "pg-types" result in the misconfiguration of the logical replication type parser (e.g. https://discord.com/channels/830183651022471199/1350643925836365917)


arv (commented):
Woho!

LGTM

PR #4060, opened 2025-03-20
chore(run-query): handle server name mappings

tantaman:
The ZQL query provided to the run-query script is expected to use client names since ZQL queries are always crafted using client names.

Given that, `run-query` needs to map the AST to server names since the replica file will be using server names.

https://discord.com/channels/830183651022471199/1352273830596317217/1352292869284692090


darkgnotic (approved):

packages/zero-cache/src/scripts/run-query.ts
```diff
@@ -67,12 +80,21 @@ const host: QueryDelegate = {
       testLogConfig,
       '',
       db,
-      name,
-      schema.tables[name].columns,
-      schema.tables[name].primaryKey,
+      serverTableName,
+      Object.fromEntries(
```
arv, line 84: @darkgnotic added a helper for this common pattern:

https://github.com/rocicorp/mono/blob/4d9838dd48c78e73db2119a06865d41cf4c36d61/packages/shared/src/objects.ts#L25

arv (commented):
LGTM

Thanks.

PR #4063, opened 2025-03-21
feat(zero-cache): protect the replication-manager with back pressure

darkgnotic:
Protect the `replication-manager` from being overwhelmed by a highly active replication stream by exerting back pressure from the change-streamer when the queued changes exceeds a threshold.

Currently this threshold is based on a simple count. It can later be improved to be memory-size based. This works for both postgres and custom change sources.

Loadtests confirm that a change-streamer without this protection eventually crashes, while the back-pressure-protected change-streamer continues humming along.

chart (2)

Without back-pressure:

Screenshot 2025-03-20 at 19 34 45

With back-pressure:

Screenshot 2025-03-20 at 19 35 02

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
+++ b/packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
@@ -367,6 +367,8 @@ class ChangeStreamerImpl implements ChangeStreamerService {
 
           if (type === 'commit') {
             watermark = null;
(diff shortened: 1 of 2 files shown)


darkgnotic:
FYI, I've verified this with the load harness; as a followup I will think about how to unit test this with continuous integration.

arv (commented):
LGTM

Very nice

PR #4065, opened 2025-03-21
feat(zero): Add npx ast-to-zql binary.

arv:
Usage:

```bash
cat ast.json | npx ast-to-zql
```


arv:
@darkgnotic I wanted to support:

```sh
npx ast-to-zql ast.json
```

but I'm not sure how the options system supports positional args?

arv:
This also moves the code into `packages/ast-to-zql`... Putting things in tools does not work if we want to put things as part of the npm package.

darkgnotic (commented):
LGTM

PR #4066, opened 2025-03-21
chore(zero-cache): Make maxRowCount configurable

arv:
https://bugs.rocicorp.dev/issue/3633


erikmunson:
🙏 🥳

packages/zero-cache/src/config/zero-config.test.ts
```diff
@@ -289,6 +289,11 @@ test('zero-cache --help', () => {
                                                                 to reduce the amount of heap memory used during initial sync (e.g. for tables                     
                                                                 with large rows).                                                                                 
                                                                                                                                                                   
+     --max-row-count number                                     optional
```
darkgnotic, line 292: Some suggestions for names that might be clearer to the reader:
* `--target-client-row-count`
* `--max-client-row-count`

A couple of concepts that would be good to convey are (1) this is a per-client parameter and not global to the server, and (2) it's not a hard max, and clients can go over it if all of the queries are active.

Thoughts?

darkgnotic (commented):
LGTM

packages/zero-cache/src/config/zero-config.test.ts
arv: I like `--target-client-row-count`

I will also take another stab the description

PR #4067, opened 2025-03-21
chore(mutators): add an `OutboundStream` for `pusher`

tantaman:
Just like `ViewSyncer` returns an `OutboundStream` which is proxied down to the client, follow the same pattern for passing mutation responses back to the client.

https://github.com/rocicorp/mono/blob/06697185aea4770c5aa7063e2f2012d0136ebe5f/packages/zero-cache/src/workers/connection.ts#L184-L192

### DIFF
--- a/packages/zero-cache/src/services/mutagen/pusher.ts
+++ b/packages/zero-cache/src/services/mutagen/pusher.ts
@@ -1,15 +1,26 @@
 import type {LogContext} from '@rocicorp/logger';
 import {must} from '../../../../shared/src/must.ts';
 import {Queue} from '../../../../shared/src/queue.ts';
-import * as ErrorKind from '../../../../zero-protocol/src/error-kind-enum.ts';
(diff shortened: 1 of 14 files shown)


packages/zero-cache/src/services/mutagen/pusher.ts
```diff
@@ -45,8 +57,28 @@ export class PusherService implements Service {
...
 
-  enqueuePush(push: PushBody, jwt: string | undefined) {
+  enqueuePush(
+    clientID: string,
+    wsID: string,
+    push: PushBody,
+    jwt: string | undefined,
+  ): HandlerResult {
+    const downstream: Subscription<Downstream> | undefined =
+      this.#pusher.maybeInitConnection(clientID, wsID);
```
tantaman, line 67: returns a downstream if this is the first push we've received for the given `clientID` and `wsID` combo.

PR #4073, opened 2025-03-22
chore(z2s): type conversion in select

grgbkr:


packages/zero-schema/src/builder/table-builder.ts
```diff
@@ -43,6 +43,22 @@ export function json<T extends ReadonlyJSONValue = ReadonlyJSONValue>() {
...
 
+export function date<T extends ReadonlyJSONValue = ReadonlyJSONValue>() {
+  return new ColumnBuilder({
+    type: 'date',
+    optional: false,
+    customType: null as unknown as T,
+  });
+}
+
+export function timestamp<T extends ReadonlyJSONValue = ReadonlyJSONValue>() {
```
tantaman, line 54: `T extends number` and maybe even default it to an alias

`T extends number = Timestamp`

packages/z2s/src/compiler.ts
```diff
@@ -397,4 +398,14 @@ export class Compiler {
     const mapped = this.#nameMapper.tableName(table);
     return sql.ident(mapped);
   }
+
+  #selectCol(table: string, column: string, type: ValueType) {
```
tantaman, line 402: Might need bigint handling, bigint-json. And we'll want to double-check `numeric`. We represent it as a number but PG might convert it to a string when serializing to JSON.

grgbkr: bigint and numeric get converted as we need

image

image

PR #4074, opened 2025-03-22
chore(zero-cache): fix add-then-remove edge case

darkgnotic:
Fix an edge case that was discovered in a stress test, whereby a row that was not previously in the CVR was added during an IVM advancement, and then removed in a different batch (of 10,000 rows).


arv (commented):
LGTM

PR #4075, opened 2025-03-22
chore(zero-cache): better handling of large upstream transactions

darkgnotic:
Fix two issues discovered while stress-testing the change-streamer with very large transactions.

1. If the replication stream disconnects, the change-streamer will loop around and reconnect. Before doing so, wait for any pending changes to flush to the change db in order to resume from the correct watermark. Otherwise, a duplicate key error can occur:

Screenshot 2025-03-21 at 19 30 56

2. The replication stream may disconnect if the change-streamer is applying back pressure and thus not responding to the periodic keepalives from postgres. To signal to postgres that the subscriber is still alive, manually send keepalive responses if none have been sent for longer than postgres's 30-second keepalive interval. This prevents postgres from disconnecting when the replication-manager is busy processing messages.

Screenshot 2025-03-21 at 20 07 51

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/schema/types.ts
+++ b/packages/zero-cache/src/services/view-syncer/schema/types.ts
@@ -52,14 +52,14 @@ export function cmpVersions(
   return a === null && b === null
     ? 0
     : a === null
-      ? -1
-      : b === null
-        ? 1
-        : a.stateVersion < b.stateVersion
-          ? -1
-          : a.stateVersion > b.stateVersion
-            ? 1
-            : (a.minorVersion ?? 0) - (b.minorVersion ?? 0);
+    ? -1
+    : b === null
+    ? 1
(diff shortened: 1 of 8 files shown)


packages/zero-cache/src/services/view-syncer/schema/types.ts
```diff
@@ -52,14 +52,14 @@ export function cmpVersions(
...
-            ? 1
-            : (a.minorVersion ?? 0) - (b.minorVersion ?? 0);
+    ? -1
+    : b === null
+    ? 1
+    : a.stateVersion < b.stateVersion
+    ? -1
+    : a.stateVersion > b.stateVersion
+    ? 1
+    : (a.minorVersion ?? 0) - (b.minorVersion ?? 0);
```
arv, line 62: I think we had the wrong version of prettier at some point.

arv (commented):
LGTM

PR #4081, opened 2025-03-24
feat(zbugs): Add discord notifications.

aboodman:
There are a few issues here though:

1. We are awaiting on the notification within the PG tx :-|. I had to do this because the notification wasn't getting sent if I didn't. Need to figure out what's going on there.

2. I observed that if there was not a matching mutator name, the mutation retries forever. It should be silently skipped.

At the same time refactored how created/modified dates are set. This seems slightly nicer?

### DIFF
--- a/apps/zbugs/README.md
+++ b/apps/zbugs/README.md
@@ -36,15 +36,6 @@ ZERO_UPSTREAM_DB = "postgresql://user:password@127.0.0.1:6434/postgres"
 # Where to send custom mutations
 ZERO_PUSH_URL = "http://localhost:5173/api/push"
 
-# A separate Postgres database we use to store CVRs. CVRs (client view records)
-# keep track of which clients have which data. This is how we know what diff to
-# send on reconnect. It can be same database as above, but it makes most sense
-# for it to be a separate "database" in the same postgres "cluster".
-ZERO_CVR_DB = "postgresql://user:password@127.0.0.1:6435/postgres"
-
-# Yet another Postgres database which we used to store a replication log.
-ZERO_CHANGE_DB = "postgresql://user:password@127.0.0.1:6435/postgres"
-
 # Place to store the SQLite data zero-cache maintains. This can be lost, but if
 # it is, zero-cache will have to re-replicate next time it starts up.
 ZERO_REPLICA_FILE = "/tmp/zbugs-sync-replica.db"
@@ -79,6 +70,10 @@ PRIVATE_JWK = ""
 VITE_PUBLIC_SERVER="http://localhost:4848"
 # See comment on `ZERO_AUTH_JWK`
 VITE_PUBLIC_JWK=''
+
+# Discord webhook to send notifications to. Not required. Notifications won't
+# be sent if absent.
+DISCORD_WEBHOOK_URL=''
 ```
 
 Then start the server:
--- a/apps/zbugs/api/discord.ts
+++ b/apps/zbugs/api/discord.ts
@@ -1,0 +1,36 @@
+export async function postToDiscord({
+  title,
+  message,
+  link,
+}: {
+  title: string;
+  message: string;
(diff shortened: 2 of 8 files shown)


apps/zbugs/api/server-mutators.ts
```diff
@@ -0,0 +1,143 @@
...
+        });
+        void notify(tx, v, {kind: 'create-issue', issueID: id});
+      },
+
+      async update(tx, update: UpdateValue<typeof schema.tables.issue>) {
+        await mutators.issue.update(tx, {
+          ...update,
+          modified: Date.now(),
+        });
+        await notify(tx, v, {
```
tantaman, line 43: Can you schedule this to run outside the tx on the next event loop tick?

apps/zbugs/README.md
```diff
@@ -36,15 +36,6 @@ ZERO_UPSTREAM_DB = "postgresql://user:password@127.0.0.1:6434/postgres"
 # Where to send custom mutations
 ZERO_PUSH_URL = "http://localhost:5173/api/push"
 
-# A separate Postgres database we use to store CVRs. CVRs (client view records)
-# keep track of which clients have which data. This is how we know what diff to
-# send on reconnect. It can be same database as above, but it makes most sense
-# for it to be a separate "database" in the same postgres "cluster".
-ZERO_CVR_DB = "postgresql://user:password@127.0.0.1:6435/postgres"
```
tantaman, line 43: I guess we default these to `upstream` now if they're not present?

apps/zbugs/api/discord.ts
```diff
@@ -0,0 +1,36 @@
...
+  link,
+}: {
+  title: string;
+  message: string;
+  link: string;
+}) {
+  const content = `**${title}**\n${message}\n<${link}>`;
+  const url = process.env.DISCORD_WEBHOOK_URL;
+  if (!url) {
+    console.log('No discord URL, skipping posting message', content);
```
tantaman, line 13: nit: log context.

apps/zbugs/api/notify.ts
```diff
@@ -0,0 +1,153 @@
...
+        message: [issue.title, emoji].join('\n'),
+        link: `https://bugs.rocicorp.dev/issue/${issue.shortID}`,
+      });
+      break;
+    }
+
+    case 'add-emoji-to-comment': {
+      const {commentID, emoji} = args;
+      const comment = await tx.query.comment.where('id', commentID).one().run();
+      assert(comment);
```
tantaman, line 120: I think we should add a `onex / oneMust` that collapses this pattern.

apps/zbugs/api/server-mutators.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      async create(tx, {id, title, description}: CreateIssueArgs) {
+        await mutators.issue.create(tx, {
+          id,
+          title,
+          description,
+          created: Date.now(),
+          modified: Date.now(),
+        });
+        void notify(tx, v, {kind: 'create-issue', issueID: id});
```
tantaman, line 35: thought you had to await for the pushToDiscord to work?

apps/zbugs/api/notify.ts
```diff
@@ -0,0 +1,153 @@
...
+        message: [issue.title, clip(comment)].join('\n'),
+        link: `https://bugs.rocicorp.dev/issue/${issue.shortID}#comment-${commentID}`,
+      });
+      break;
+    }
+  }
+}
+
+function clip(s: string) {
+  return s.length > 255 ? s.slice(0, 255) + '...' : s;
```
tantaman, line 152: nit: 

```ts
s.length > 258 ? s.slice(0, 255) + '...' : s;
```

So you don't end up with things like:

`Hi Mar...` for `Hi Mark`

PR #4082, opened 2025-03-24
chore(zero-cache): rename max-row-count to target-client-row-count

arv:
Rename configuration property to clarify it refers to target number of rows for each client, not a hard limit.


arv:
We also need to update the docs.

packages/zero-cache/src/config/zero-config.ts
```diff
@@ -458,10 +458,15 @@ export const zeroOptions = {
...
 
-  maxRowCount: {
+  targetClientRowCount: {
     type: v.number().optional(),
     desc: [
-      'The target number of rows to keep in the client side cache.',
+      'The target number of rows to keep per client in the client side cache.',
+      'This limit is a soft limit. Active queries are allowed to',
+      'use more rows than this limit, but if the cache grows to',
+      'this size, zero-cache will evict inactive queries from the cache',
```
darkgnotic, line 467: Nice. Another iteration:

> The target number of rows to keep per client in the client side cache.
> This limit is a soft limit. When the number of rows in the cache exceeds
> this limit, zero-cache will evict inactive queries in order of ttl-based expiration.
> Active queries, on the other hand, are never evicted and are allowed to use more
> rows than the limit.

PR #4083, opened 2025-03-24
chore: Add test:watch and check-types:watch to root

arv:
You can now run `npm run test:watch` and `npm run check-types:watch` from the mono root and it will do the right thing. to watch for changes and re-run tests and type checks.


cesara (commented):
lgtm

PR #4084, opened 2025-03-24
fix(zero-cache): Fix too large timer exception

arv:
NodeJS timers have a maximum limit of 2^31-1 milliseconds, which is about 24.8 days. This commit fixes the issue by setting the timer to 1h which will then reschedule the expiration work again.

https://bugs.rocicorp.dev/issue/3645


darkgnotic (commented):
LGTM

Thank you!

PR #4086, opened 2025-03-24
chore(zero-client): Deprecate AdvancedQuery and friends

arv:
- `AdvancedQuery` got folded into `Query`
- `ZeroAdvancedOptions` got folded into `ZeroOptions`
  - `maxRecentQueries` is flagged as deprecated in favor of TTL.

The export `@rocicorp/zero/advanced` is still there but will be removed in the future.


packages/zql/src/query/query.ts
arv, line 1: This is the important changes to the public API.

packages/zql/src/query/query.ts
```diff
@@ -108,6 +109,10 @@ export interface Query<
   TTable extends keyof TSchema['tables'] & string,
   TReturn = PullRow<TTable, TSchema>,
 > {
+  readonly format: Format;
+  hash(): string;
+  updateTTL(ttl: TTL): void;
```
arv, line 114: This one is the most unfortunate one. My gut feeling is that we should hide these from the common usage...

One suggestion is to have:

```ts
interface Query {
  delegate?: QueryDelegate
}
```

but this is not optimal either.

or we can do 

```ts
export getQueryDelegate(q: Query): QueryDelegate | undefined;
```

grgbkr (approved):
LGTM

packages/zql/src/query/query-impl.ts
```diff
@@ -98,32 +97,25 @@ export function staticParam(
...
+  constructor(schema: TSchema, tableName: TTable, ast: AST, format: Format) {
     this.#schema = schema;
     this.#tableName = tableName;
     this.#ast = ast;
-    this.#format = format ?? {singular: false, relationships: {}};
-  }
-
-  get format(): Format {
-    return this.#format;
+    this.format = format;
```
grgbkr, line 118: I think caused a regression.  Nate is hitting errors because of an undefined format.  

image

One way format can end up undefined here are these calls https://github.com/rocicorp/mono/blob/bf275086f4c906b6cb7e801a5eaa740e9570f3e8/packages/zql/src/query/query-impl.ts#L420
https://github.com/rocicorp/mono/blob/bf275086f4c906b6cb7e801a5eaa740e9570f3e8/packages/zql/src/query/query-impl.ts#L456

unsure if there are others.

Not sure why TypeScript isn't catching this.

arv: Thanks for the fix

PR #4087, opened 2025-03-24
chore(mutators): close connection on fatal errors

tantaman:
Problem 1:
- custom mutators did not close the connection on `InvalidPush` errors (OOO mutation, bad push version).

Problem 2:
- custom mutators would resolve client promises for a mutation on `InvalidPush` errors. This was wrong since these mutations will be retried later. No longer resolve these promises. We do that by not sending a response for mutations that fail with `InvalidPush` errors and instead close the connection.

--- 

Removed 'token' error type as that is unused. Left protocol version unchanged since it was unused.

### DIFF
--- a/packages/zero-cache/src/services/mutagen/pusher.test.ts
+++ b/packages/zero-cache/src/services/mutagen/pusher.test.ts
@@ -566,6 +566,112 @@ describe('pusher streaming', () => {
       });
     }
   });
+
+  test('fails the stream on ooo mutations', async () => {
+    const fetch = (global.fetch = vi.fn());
+    const oooResponse: PushResponse = {
+      mutations: [
+        {
+          id: {clientID, id: 3},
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/services/mutagen/pusher.test.ts
```diff
@@ -566,6 +566,112 @@ describe('pusher streaming', () => {
...
+
+    const result = pusher.enqueuePush(
+      clientID,
+      wsID,
+      makePush(2, clientID),
+      'jwt',
+    );
+    expect(result.type).toBe('stream');
+
+    if (result.type === 'stream') {
```
arv, line 607: this is fine but next time you can use assert to do the type narrowing so you can reduce indentation.

arv (commented):
LGTM

PR #4088, opened 2025-03-24
chore(mutators): deal with extra properties in mutator responses

tantaman:
Adds a `JSON` field to `error` and `ok` responses to allow the user to return any sort of extra data required.


tantaman:
Maybe we can use a placeholder field for user data that is of type JSONValue.

arv:
How is this value read on the client?

arv (commented):
LGTM

But I'm not really sure what the motivation is?

PR #4089, opened 2025-03-24
feat(zero-cache): protect zero-cache with client-side back-pressure

darkgnotic:
Use TCP backpressure to protect the zero-cache against memory blowup from either slow downstream connections or over-zealous upstream messages (e.g. DOS).

### Downstream

By simulating a large download (#4080) over a slow connection (#4076), the load harness is able to create a scenario in which a zero-cache (without downstream backpressure) crashes with an OOM.
  
chart (4)

With back-pressure applied, the zero-cache eventually succeeds in the streaming the large download to the client:

chart (5)

This works by having IVM wait for downstream messages to be flushed to the connection before computing the next batch of data to be sent. For mutations, the Pusher similarly waits for responses to be sent before computing the next batch of results.

### Upstream

Upstream flow control is similarly enforced by only pulling messages from the connection after the previous messages have been processed (modulo an internal buffer in the TCP channel).

This works out of the box for the View Syncer as most commands block on being processed in the lock.

There is a TODO added to the Pusher to cap the size of the mutation queue to prevent memory blowup.

### DIFF
--- a/packages/zero-cache/src/services/mutagen/pusher.ts
+++ b/packages/zero-cache/src/services/mutagen/pusher.ts
@@ -257,7 +257,7 @@ class PushWorker {
         }
       }
     }
-    // Wait for results to get sent downstream go avoid queueing up
+    // Wait for results to be sent downstream to avoid queueing up
     // an arbitrary amount of responses in memory.
     await Promise.allSettled(pushed);
   }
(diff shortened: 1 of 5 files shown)


tantaman:
I don't think this breaks our invariant that IVM must advance linearly. As in, each change to an IVM source is fully consumed before another change is allowed to be applied. Even though the asynchrony added here should only be spreading the work over a larger time horizon, I do worry a bit about that invariant breaking. Do we have any tests to check this on the server?

darkgnotic:
(Also, we added awaits to ivm in the time-slicing change a couple months ago.)

PR #4091, opened 2025-03-24
chore(zbugs): prefix private `api` files with `_`

tantaman:
vercel was routing to these files if the user typed them in. E.g., `http://bugs.rocicorp.dev/api/discord`

CleanShot 2025-03-24 at 19 41 10@2x

`_` is the vercel way to hide a script from filesystem routing.


aboodman:
I am confused what is happening here. vercel.json is rewriting `/api/*` to
`/api` which should send everything to the fastify stuff in index.ts. How
is vercel's file-based routing coming into play?

PR #4093, opened 2025-03-25
chore(zero): update to dotenvx

cesara:


packages/zero/src/zero-cache-dev.ts
```diff
@@ -13,7 +13,7 @@ import {
 } from '../../zero-cache/src/config/zero-config.ts';
 import {deployPermissionsOptions} from '../../zero-cache/src/scripts/permissions.ts';
 
-const deployPermissionsScript = 'zero-deploy-permissions';
+const deployPermissionsScript = 'zero-deploy-permissions -p shared/schema.ts';
```
tantaman, line 16: Should this change? `shared/schema.ts` is zbugs specific.

darkgnotic: Agree, this should be reverted.

The caller should instead pass in the environment variable `ZERO_SCHEMA_PATH=shared/schema.ts`

cesara: ok i'll fix in a seperate PR

PR #4094, opened 2025-03-25
chore(zero-cache,pg): test type conversion on write

tantaman:
All of these tests fail. We need to convert types on write.

@darkgnotic  - the CRUD test failure surprised me because I know we have people inserting epochs into date columns.


darkgnotic:
We handle timestamp and timestamptz here: 

https://github.com/rocicorp/mono/blob/e9c6b7e9c958536b19a120f02ababf0db4b507d8/packages/zero-cache/src/types/pg.ts#L111

But I guess we didn't add date and time (cuz those types are weird?  😜)

darkgnotic:
But I think this is the same quandary... if we don't have control of the client, do we have to figure out the types and then convert them manually?

PR #4095, opened 2025-03-25
chore(zbugs): email issue subscribers in `dev`

tantaman:
Needs a production transport but before we do that we should support unsubscribing.

- `gatherRecipients` drops to raw SQL. I think this is a useful demonstration of how to do that in a custom mutator.
- add a `userInternal` table as we do not want to expose user email addresses over sync. The other option is to add `email` to the `user` table and change the publication to not include it.


apps/zbugs/api/_notify.ts
```diff
@@ -69,15 +75,23 @@ export async function notify(
   console.log('modifierUser', modifierUser);
   assert(modifierUser);
 
+  const recipients = await gatherRecipients(
+    // TODO: we shouldn't have to assert and cast this given this is called from `server-mutators`.
+    tx as ServerTransaction<Schema, postgres.TransactionSql>,
```
tantaman, line 80: this is a papercut I missed.

apps/zbugs/docker/share_sql/base.sql
```diff
@@ -18,6 +18,11 @@ CREATE TABLE "user" (
     "githubID" INTEGER NOT NULL
 );
 
+CREATE TABLE "userInternal" (
```
aboodman, line 21: Can we use the publication approach instead?

PR #4096, opened 2025-03-25
chore(mutators): register read queries in mutators

tantaman:
Problem:

> if a user checks permissions on the client in a custom mutator, nothing prevents the queries that provided the permission data from being deregistered before the mutator is rebased.
>
> It’d be a weird interaction to:
>
> 1. Have the mutator succeed
> 2. Navigate to a new page
> 3. Have the mutator rebase and fail because the queries it used are gone

This implement's @arv's idea:

> That is a good point. We should keep queries in the mutators alive until mutator has been confirmed
>
> I don’t think we need to change ref counting… we just need to add the queries to the desired set when the mutator is run the first time and removed when the mutator is acknowledged

https://rocicorp.slack.com/archives/C013XFG80JC/p1742815429158529

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -623,6 +623,7 @@ export class Zero<
       rep.experimentalWatch.bind(rep),
       maxRecentQueries,
     );
+    this.#mutationTracker.setQueryManager(this.#queryManager);
     this.#clientToServer = clientToServer(schema.tables);
 
     this.#deleteClientsManager = new DeleteClientsManager(
--- a/packages/shared/src/sentinels.ts
+++ b/packages/shared/src/sentinels.ts
@@ -1,0 +1,1 @@
+export function emptyFunction() {}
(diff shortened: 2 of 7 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -477,7 +478,9 @@ export class Zero<
             mutatorOrMutators,
             schema,
             slowMaterializeThreshold,
-          );
+            // Replicache expects mutators to only be able to return JSON
+            // but Zero wraps the return with: `{server?: Promise<MutationResult>, client?: T}`
+          ) as () => MutatorReturn;
```
tantaman, line 483: This is unfortunate. We could relax the Replicache return type constraint on mutators.

tantaman:
Going to go a different route for this which is to:

Prevent de-registering of any queries while mutations are pending.

`QueryManager` will queue up requests to change desired queries until pending mutations = 0.

PR #4097, opened 2025-03-25
feat(mutators)!: nest server result in client promise for custom muta…

tantaman:
…tors

We cannot tack properties onto `Promise` since any `async` function above us in the stack will undo that change by wrapping the promise in a new one.

----

The change --

The return type for a custom mutator was:

```ts
Promise<void> & {server: Promise<MutationResult>}
```

Now it is:

```ts
Promise<{server?: Promise<MutationResult>}>
```

### DIFF
--- a/packages/zero-client/src/client/custom.test.ts
+++ b/packages/zero-client/src/client/custom.test.ts
@@ -1,12 +1,12 @@
+import * as sinon from 'sinon';
 import {beforeEach, describe, expect, expectTypeOf, test} from 'vitest';
 import {schema} from '../../../zql/src/query/test/test-schemas.ts';
 import {
   TransactionImpl,
   type CustomMutatorDefs,
   type MakeCustomMutatorInterfaces,
-  type PromiseWithServerResult,
 } from './custom.ts';
-import {zeroForTest} from './test-utils.ts';
+import {MockSocket, zeroForTest} from './test-utils.ts';
 import type {InsertValue, Transaction} from '../../../zql/src/mutate/custom.ts';
 import {IVMSourceBranch} from './ivm-branch.ts';
 import {createDb} from './test/create-db.ts';
@@ -15,6 +15,7 @@ import type {WriteTransaction} from './replicache-types.ts';
 import {zeroData} from '../../../replicache/src/transactions.ts';
 import {must} from '../../../shared/src/must.ts';
 import type {MutationResult} from '../../../zero-protocol/src/push.ts';
+import * as ConnectionState from './connection-state-enum.ts';
 
 type Schema = typeof schema;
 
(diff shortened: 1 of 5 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -477,7 +478,9 @@ export class Zero<
             mutatorOrMutators,
             schema,
             slowMaterializeThreshold,
-          );
+            // Replicache expects mutators to only be able to return JSON
+            // but Zero wraps the return with: `{server?: Promise<MutationResult>, client?: T}`
+          ) as () => MutatorReturn;
```
tantaman, line 483: this is unfortunate. We could widen the replicache type.

packages/zero-client/src/client/custom.test.ts
```diff
@@ -344,4 +345,71 @@ describe('rebasing custom mutators', () => {
 
     expect(mutationRun).toEqual(true);
   });
+
+  test('waiting for server results', async () => {
+    sinon.replace(
```
arv, line 350: try to stick with vi in new tests..

```js
vi.stubGlobal('WebSocket', MockSocket as unknown as typeof WebSocket);
```

packages/zero-client/src/client/custom.test.ts
```diff
@@ -344,4 +345,71 @@ describe('rebasing custom mutators', () => {
...
+        {
+          id: {clientID: z.clientID, id: 2},
+          result: {
+            error: 'app',
+          },
+        },
+      ],
+    });
+
+    sinon.restore();
```
arv, line 410: the restore part also needs to be done in an afterEach.

packages/zero-client/src/client/custom.ts
```diff
@@ -62,7 +62,7 @@ export type MakeCustomMutatorInterfaces<
     tx: Transaction<S>,
     ...args: infer Args
   ) => Promise<void>
-    ? (...args: Args) => PromiseWithServerResult<void, MutationResult>
+    ? (...args: Args) => Promise<{server?: Promise<MutationResult>}>
```
arv, line 65: Can we name this type... It is a bit ugly.

packages/zero-client/src/client/zero.ts
arv: Yes. I don't think we serialize the result to JSON anywhere so this  type restriction was not necessary.

arv (commented):
LGTM with comments

I liked the old thing better... I understand that we lose the server promise if we unpack the promise but is that really a problem. If you want it you need to get it before you unpack it.

I'm sure you ran into a real problem but I don't see it at the moment. Can you show me some code where this becomes problematic?

tantaman:
I'm sure you ran into a real problem but I don't see it at the moment. Can you show me some code where this becomes problematic?

This line: 

https://github.com/rocicorp/mono/blob/64d4c8b94222c04b5c41249820bbe812e0f2e8e5/packages/replicache/src/replicache-impl.ts#L1518

Maybe we can stick the `server` result back on the promise returned by `replicache-impl::#mutate`

arv:
The "solution" is to not use async functions 😨! However, this is pretty deep down in an async call stack so it does not seem like a good idea.

PR #4098, opened 2025-03-25
feat(zero-cache): support ALTER SCHEMA ddl changes

darkgnotic:
Adds support for `ALTER SCHEMA ...` ddl changes, which essentially map to one or more table renames on the replica.

This is necessary to implement (one strategy of) non-disruptive resyncs, but is technically also a user-usable feature given that non-public schemas are supported.

Also clean up / simplify the trigger logic as it is now based on a simple diff. The new format deprecates the unused `table` and `index` fields and does not constrain the `tag` values, thus improving forwards compatibility.


arv (commented):
LGTM

PR #4099, opened 2025-03-26
chore(mutators): await pushed messages before failing the stream

tantaman:
If we've enqueued messages to the stream, wait for those messages to be consumed before closing the connection.

Previously a failure in a batch of mutations could cause the connection to close before the success responses ever got sent.

```
m1 : success
m2 : success
m3 : success
m4 : fail
```

^-- this would close the stream before responses to m1 -> m3 could be processed.


darkgnotic (approved):
Nice!

PR #4100, opened 2025-03-26
chore(zero-client): call `push` when reconnecting

tantaman:
Pending mutations were being pushed on page load but not on reconnect of the websocket after initial page load.

Found this by injecting code to fail mutations 50% of the time.


arv (approved):
LGTM with missing tests added.

arv (commented):
LGTM

PR #4105, opened 2025-03-26
chore(mutators): do not send the client `mutation-success` responses

tantaman:
We do not want the client to think a mutation is completed on the server before the client has received the corresponding pokes from the server that contain the state changes of the mutation.

This updates `mutation-tracker` to only resolve mutations as successful when the lmid moves forward.

---

There is also a race for transient errors. A mutation that receives an `app` failure could be skipped and a new lmid poked down before the `push-response` is received. Looks like the only recourse we have is to write all mutation results to the DB.

### DIFF
--- a/packages/shared/src/sentinels.ts
+++ b/packages/shared/src/sentinels.ts
@@ -1,1 +1,2 @@
 export function emptyFunction() {}
+export const emptyObject = Object.freeze({});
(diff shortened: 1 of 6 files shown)


tantaman:
Actually we can solve this by having the client track two promises for each mutation but exposing it to the user as a single promise that resolves with both resolve.

PR #4106, opened 2025-03-26
feat(zero-client): Add onError option to zero client

arv:


arv:
I was thinking that the onError callback should have a more structured type. Something along the lines of:

```ts
onError: (...args: ErrorArgs): void


type ErrorArgs = 
  [type: 'server', error: ServerErrorKind],
  [type: 'a', error: ...],
  [type: 'b', error: ...],
  ...,
}
```

One way to enforce this in the code is to change to use a ZeroLogContext in zero client and make the error function of that enforce that there error messages are type safe.

WDYT?

arv:
@grgbkr ^^^

tantaman:
@aboodman and I were discussing how to deal with authentication errors when calling custom mutators and landed on `onError` being the best way to do it and getting rid of the `auth` callback that we have now.

Is the last blocking thing here a review from @grgbkr ?

arv:
I was also thinking about a structured API. Right now this PR calls this with `...args: unknown[]` which is less than optimal.

PR #4110, opened 2025-03-26
chore(z2s): type conversions for query results

grgbkr:
This make z2s output TEXT containing json in order to minimize the chance of any type converters installed on the customers pg client library incorreclty changing the results.

It also correctly converts `timestamp`, `timestamptz` and `date` columns to `number` in the json output.  To be able to do this we need to add more information to the zero schema.  Currently these column types are all represented simply as `type: 'number'` in zero schema.  To determine that these numbers are actually timestamps/dates we'd have to fetch the upstream db's schema, and since the api server may be a stateless lambda this might need to be done effectively on every push.  Instead we disambiguate these types in the zero schema.

'bigint' handling to be done in a follow up.

This change has backwards compatibility, it will still work with schema's that use 'number' for timestamps and dates, these schema's just won't work with z2s.


packages/zero-schema/src/builder/table-builder.ts
```diff
@@ -43,6 +43,22 @@ export function json<T extends ReadonlyJSONValue = ReadonlyJSONValue>() {
   });
 }
 
+export function date<T extends number = number>() {
```
grgbkr, line 46: We could get away with just a 'timestamp' time that could cover all the Date/Time Types (https://www.postgresql.org/docs/current/datatype-datetime.html), including date, time, timestamp, timestamptz.  We don't currently support time.  

I'm torn... I don't think of a date as a timestamp, so it feels a bit more intuitive to have a separate date type.   Also time is even less like a timestamp (though you'd represent it as millis from midnight).

packages/z2s/src/compiler.ts
```diff
@@ -15,12 +16,21 @@ import {
...
+} from '../../zero-schema/src/table-schema.ts';
 import type {Format} from '../../zql/src/ivm/view.ts';
 import {sql} from './sql.ts';
 
 type Tables = Record<string, TableSchema>;
 
+const ZQL_RESULT_KEY = 'zql_result';
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function extractZqlResult(pgResult: any): JSONValue {
+  return JSON.parse(pgResult[0][ZQL_RESULT_KEY]);
```
tantaman, line 31: We need the BigintJson variant, correct?

grgbkr: Yes plan to do this in a follow up.  I'm still uncertain if with z2s we should throw if a BigInt value is > MAX_SAFE_INTEGER

PR #4117, opened 2025-03-27
chore: Run pg tests again

arv:
vitest doesn't support nested workspaces, so we need to run the pg tests from the root workspace.


grgbkr:
Thanks @arv

PR #4121, opened 2025-03-27
chore(z2s): convert args internally via SQL in Mutagen

tantaman:
All args are now passed to sql queries as text. We then convert that text to the appropriate type. This is to bypass any specific type conversions a user may have installed that would not be compatible with Zero.

Mutagen can now write to timestamp columns.

### DIFF
--- a/packages/z2s/src/sql.test.ts
+++ b/packages/z2s/src/sql.test.ts
@@ -41,9 +41,11 @@ describe('json arg packing', () => {
       ),
     ).toMatchInlineSnapshot(`
       {
-        "text": "SELECT * FROM "user" WHERE "id" = ($1->>0)::numeric",
+        "text": "SELECT * FROM "user" WHERE "id" = ($1::json->>0)::numeric",
         "values": [
-          "[1]",
+          [
+            1,
+          ],
         ],
       }
     `);
@@ -57,9 +59,11 @@ describe('json arg packing', () => {
       ),
     ).toMatchInlineSnapshot(`
       {
-        "text": "SELECT * FROM "user" WHERE "id" = ($1->>0)::numeric OR "other_id" = ($1->>0)::numeric",
+        "text": "SELECT * FROM "user" WHERE "id" = ($1::json->>0)::numeric OR "other_id" = ($1::json->>0)::numeric",
         "values": [
-          "[1]",
+          [
+            1,
+          ],
         ],
       }
     `);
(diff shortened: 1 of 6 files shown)


packages/zero-pg/src/test/schema.ts
```diff
@@ -55,10 +64,25 @@ CREATE TABLE "compoundPk" (
...
   c TEXT,
   PRIMARY KEY (a, b)
-);`;
+);
+
+CREATE TABLE "dateTypes" (
+  "ts" TIMESTAMP,
+  "tstz" TIMESTAMPTZ,
+  "tswtz" TIMESTAMP WITH TIME ZONE,
+  "tswotz" TIMESTAMP WITHOUT TIME ZONE,
```
grgbkr, line 73: Would be good to also test DATE.

packages/z2s/src/sql.test.ts
```diff
@@ -72,9 +76,37 @@ describe('json arg packing', () => {
       ),
     ).toMatchInlineSnapshot(`
       {
-        "text": "SELECT * FROM "foo" WHERE "a" = $1->0 OR "b" = ($1->>1)::numeric OR "c" = $1->>2 OR "d" = ($1->>3)::boolean",
+        "text": "SELECT * FROM "foo" WHERE "a" = $1::json->0 OR "b" = ($1::json->>1)::numeric OR "c" = $1::json->>2 OR "d" = ($1::json->>3)::boolean",
```
grgbkr, line 79: I'm concerned that casting to json multiple times may parse the text to json multiple times... 

What if instead of packing all the args into a single json array, we pass them each as a separate text arg?

tantaman: true, we don't need the json indirection.

packages/z2s/src/sql.ts
```diff
@@ -91,23 +103,31 @@ class JsonPackedFormat implements FormatConfig {
...
     return {
       placeholder: this.#createPlaceholder(this.#seen.size, value),
-      value: value.value,
+      value: stringify(value.type, value.value),
     };
   };
 
-  #createPlaceholder(index: number, value: JsonPackArg) {
+  #createPlaceholder(index: number, value: SqlConvertArg) {
+    // Ok, so what is with all the `::text` casts
```
grgbkr, line 111: Ugh fun.

PR #4124, opened 2025-03-28
feat(zero-cache): non-disruptive resync

darkgnotic:
`zero-cache` previously used a constant replication slot named `{app-id}_{shard-num}`, requiring a previously synced `zero-cache` instance to be shut down (and the replication slot dropped and re-created) in order for the new `zero-cache` to perform initial sync. This can result in a lot of downtime when resyncing large databases.

With non-disruptive resync, an initial-syncing `zero-cache` creates a replication slot of the form `{app-id}_{shard-num}_{timestamp}` and records it in a `replicas` upstream table when initial-sync completes. This allows the old replication slot to continue servicing the old `zero-cache` while the new one performs initial-sync.

Once initial-sync is complete, all subscribers to any replication slot for the shard are shutdown, and all but the new replication slot are cleaned up. The new replica then begins its subscription on the new (and sole remaining) replication slot, resulting in a non-disruptive handoff similar to the usual replication-manager restart.

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/decommission.ts
+++ b/packages/zero-cache/src/services/change-source/pg/decommission.ts
@@ -12,18 +12,28 @@ export async function decommissionShard(
 
   lc.info?.(`Decommissioning zero shard ${shard}`);
   await db.begin(async tx => {
-    await tx.unsafe(dropShard(appID, shardID));
-    lc.debug?.(`Dropped upstream shard schema ${shard} and event triggers`);
-
+    // Kill the active_pid's on existing slots before altering publications,
+    // as deleting a publication associated with an existing subscriber causes
+    // weirdness; the active_pid becomes null and thus unable to be terminated.
     const slots = await tx<{pid: string | null}[]>`
     SELECT pg_terminate_backend(active_pid), active_pid as pid
-      FROM pg_replication_slots WHERE slot_name = ${shard}`;
+      FROM pg_replication_slots 
+      WHERE slot_name = ${shard} 
+         OR slot_name LIKE ${shard + '_%'}`;
     if (slots.length > 0) {
       if (slots[0].pid !== null) {
         lc.info?.(`signaled subscriber ${slots[0].pid} to shut down`);
       }
-      await tx`SELECT pg_drop_replication_slot(${shard})`;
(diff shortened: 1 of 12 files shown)


packages/zero-cache/src/services/change-source/pg/decommission.ts
```diff
@@ -12,18 +12,28 @@ export async function decommissionShard(
...
-
+    // Kill the active_pid's on existing slots before altering publications,
+    // as deleting a publication associated with an existing subscriber causes
+    // weirdness; the active_pid becomes null and thus unable to be terminated.
     const slots = await tx<{pid: string | null}[]>`
     SELECT pg_terminate_backend(active_pid), active_pid as pid
-      FROM pg_replication_slots WHERE slot_name = ${shard}`;
+      FROM pg_replication_slots 
+      WHERE slot_name = ${shard} 
+         OR slot_name LIKE ${shard + '_%'}`;
```
darkgnotic, line 22: @elledienne FYI, the format of replication slot names will be variable when non-disruptive restarts rolls out. You may have to update your previews cleanup logic accordingly.

packages/zero-cache/src/services/change-source/pg/decommission.ts
```diff
@@ -12,18 +12,30 @@ export async function decommissionShard(
...
     if (slots.length > 0) {
       if (slots[0].pid !== null) {
         lc.info?.(`signaled subscriber ${slots[0].pid} to shut down`);
       }
-      await tx`SELECT pg_drop_replication_slot(${shard})`;
-      lc.debug?.(`Dropped replication slot ${shard}`);
+      // Escape underscores for the LIKE expression.
+      const slotExpression = `${appID}_${shardID}_%`.replaceAll('_', '\\_');
```
darkgnotic, line 28: @elledienne FYI, the format of replication slot names will be variable when non-disruptive restarts rolls out. You may have to update your previews cleanup logic accordingly.

elledienne: Thank you so much for the ping on this one!

erikmunson:
THE WORK OF ANGELS 🙏 🥳 

thanks so much for knocking this out its gonna have a very big impact on availability

PR #4125, opened 2025-03-28
chore(z2s): run z2s compiler tests again postgres.js and node-postgres

grgbkr:
It should be possible to extend this approach to the chinook tests as well.

### DIFF
--- a/packages/z2s/src/compiler.pg-test.ts
+++ b/packages/z2s/src/compiler.pg-test.ts
@@ -1,4 +1,4 @@
-import {beforeAll, describe, expect, test} from 'vitest';
+import {afterAll, beforeAll, describe, expect, test} from 'vitest';
 import {testLogConfig} from '../../otel/src/test-log-config.ts';
 import type {JSONValue} from '../../shared/src/json.ts';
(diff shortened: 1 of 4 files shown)


packages/z2s/src/compiler.pg-test.ts
```diff
@@ -215,118 +234,483 @@ function createdAtToMillis(row: Record<string, unknown>) {
...
-    const sqlQuery = formatPg(compile(ast(query), schema.tables, query.format));
-    const pgResult = extractZqlResult(
-      await pg.unsafe(sqlQuery.text, sqlQuery.values as JSONValue[]),
-    );
-    expect(
-      mapResultToClientNames(await query.run(), schema, 'issue'),
-    ).toEqualPg(pgResult);
-  });
+    test('nested related with where clauses', async () => {
+      // TODO Change to filter comments on createdAt once
```
grgbkr, line 583: @tantaman Filtering by createdAt didn't work on node-postgres, got an error that 1000 was an invalid type for the input.  This should be fixed by your upcoming change to pass args as text and explicitly convert them.

PR #4129, opened 2025-03-28
chore(z2s): use new string encoding for bindings in z2s

tantaman:


grgbkr:
I think you need to update compiler-bigint.pg-test.ts as well.

tantaman:
> I think you need to update compiler-bigint.pg-test.ts as well.

Thanks. Not sure how I missed that in my `find-all`

packages/z2s/src/compiler-bigint.pg-test.ts
```diff
@@ -254,7 +254,7 @@ describe('compiling ZQL to SQL', () => {
     test('bigint exceeds safe range', async () => {
       const query = issueQuery.related('comments');
       const c = compile(ast(query), schema.tables);
-      const sqlQuery = formatPg(c);
+      const sqlQuery = formatPgInternalConvert(c);
```
grgbkr, line 257: Would be good to add a test case for a where condition involving a bigint column.

grgbkr (approved):
LGTM

packages/z2s/src/compiler.pg-test.ts
```diff
@@ -587,7 +593,7 @@ describe('compiling ZQL to SQL', () => {
         .related('comments', q =>
           q.where('text', 'ILIKE', '%2%').related('author'),
         );
-      const sqlQuery = formatPg(
+      const sqlQuery = formatPgInternalConvert(
```
grgbkr, line 596: Would be good to update the comment where filter to be on `createdAt` instead of text so that we exercise a timestamp filter.

tantaman:
added tests to compare against bigints and dates. Added `.0` to division of timestamp in order to maintain precision going into and coming out of the DB.


CleanShot 2025-03-28 at 13 58 47

grgbkr:
> added tests to compare against bigints and dates. Added `.0` to division of timestamp in order to maintain precision going into and coming out of the DB.
> 
> CleanShot 2025-03-28 at 13 58 47

Good catch.

PR #4132, opened 2025-03-28
fix(zero-cache): fix handling of column default value changes

darkgnotic:
Even though SQLite does not support non-constant defaults, zero can support _changes_ of column defaults to arbitrary expressions, e.g.

```sql
ALTER TABLE foo ALTER COLUMN bar SET DEFAULT now();
```

because changing a column's defaults in Postgres does not modify existing values.

https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-DESC-SET-DROP-DEFAULT

> The new default value will only apply in subsequent INSERT or UPDATE commands; it does not cause rows already in the table to change.

and so column values will continue to come through the replication stream.

This was handled correctly most of the time; the `change-streamer` ignores changes to column defaults. 

However, if a change to a column default was followed by a subsequent change that _is_ relevant to zero (e.g. a change to nullability), the default of the new column was incorrectly triggering an unsupported change error.

The update column logic in the replicator itself is now fixed to ignore defaults.

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/change-source.end-to-mid.pg-test.ts
+++ b/packages/zero-cache/src/services/change-source/pg/change-source.end-to-mid.pg-test.ts
@@ -520,6 +520,57 @@ describe('change-source/pg/end-to-mid-test', {timeout: 30000}, () => {
       ],
     ],
     [
+      'change column default and set not null',
+      `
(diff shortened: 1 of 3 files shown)


tantaman (commented):
lgtm

PR #4137, opened 2025-03-29
chore(zero): DeepMerge custom mutator type.

aboodman:
The intersection operator was creating overloaded functions.


arv (approved):
This could use a type test.

PR #4148, opened 2025-03-31
feat(mutators)!: move `sever` promise up a level

tantaman:
Nobody was a fan of:

```ts
type Result = Promise<{
  server: Promise<MutationResult>
}>
```

Refactor things so we can do:

```ts
type Result = Promise<void> & {server: Promise<MutationResult>}
```

Or, if we add 1 more layer of wrapping in Zero, we can do:

```ts
type Result = {client: Promise<void>, server: Promise<void>};
```

although it seems more appropriate to return the "thing that should be awaited" as a top level promise.

Why was this so hard to pull off? Because the mutator implementation is buried deep within an async callstack and each async function will wrap its result in a promise automatically. We also needed information about the `mutationID`, in order to create the server promise, which can only be gotten from deep inside the call stack from `withWrite`.

Example of the issue:
```ts
// Custom mutator impl. is async because:
// 1. We need to share code with server which would have async ZQL calls in the mutator
// 2. Could call to async APIs in browser
const createIssue = async (): Promise<void> => { ... };

// basic structure of replicache-impl::#mutate
async #mutate(mutatorImpl) {
  await pendingNotifications;
  return withWrite(dag, async write => {
    const tx = new WriteTransaction(write.getMutationID())
    
    // `result` is the promise returned by `createIssue`, unwrapped
    const result = await mutatorImpl();
    
    // but it is re-wrapped into a promise on return
    // due to #mutate being an async fn
    return result;
  })
  
  // We returned `withWrite` above (a promise).
  // Could we have awaited it? Sure, but the return
  // would be re-wrapped in a promise again.
  // Resulting back again in `Promise<{server: Promise<MutatorResult>}>`
  
  // Could we make this function not async?
  // Yes, but `promise.then` re-wraps too.
}
```

Because `mutationID` comes from `withWrite` we couldn't do something trivial like:

```ts
// zero.ts
const wrappedMutator = (args) => {
  const server = trackMutation();
  const result = mutator(args);
  return {
    client: result,
    server,
  };
};
```

### DIFF
--- a/packages/replicache/src/replicache-impl.ts
+++ b/packages/replicache/src/replicache-impl.ts
@@ -76,7 +76,11 @@ import {refresh} from './persist/refresh.ts';
 import {ProcessScheduler} from './process-scheduler.ts';
 import type {Puller} from './puller.ts';
 import {type Pusher, PushError} from './pusher.ts';
-import type {ReplicacheOptions, ZeroOption} from './replicache-options.ts';
+import type {
+  MutationTrackingData,
+  ReplicacheOptions,
+  ZeroOption,
+} from './replicache-options.ts';
 import {
   getKVStoreProvider,
   httpStatusUnauthorized,
@@ -122,6 +126,7 @@ import {
   withWrite,
   withWriteNoImplicitCommit,
 } from './with-transactions.ts';
+import type {DiffsMap} from './sync/diff.ts';
 
 declare const TESTING: boolean;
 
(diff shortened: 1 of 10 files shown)


arv (commented):
LGTM

packages/replicache/src/replicache-impl.ts
```diff
@@ -1450,8 +1455,27 @@ export class ReplicacheImpl<MD extends MutatorDefs = {}> {
...
+
+      const result = this.#mutate(
+        trackingData,
+        name,
+        mutatorImpl,
+        args,
+        performance.now(),
+      );
+
+      if (trackingData) {
```
tantaman, line 1472: > It is not clear to me how this refactoring prevents the wrapping of the result? We are still returning result from the async function.

@arv - we add the `server` field here where we're no longer in an async funcion.

arv: OK. I was referring the the changes to `#mutate`.

PR #4149, opened 2025-03-31
feat(zero-cache): eliminate quadratic writes/pokes (take 2)

darkgnotic:
This builds on (and completes) the intent of #3736, which is to avoid flushing the CVR and sending pokes to clients for which an advancement did not result in any material changes.

The hole in the logic of the original PR was that an advancement is associated with a new version, and that version was eagerly set in the CVRUpdater, thus counting as a "material change" and bypassing the flush-skipping logic. The unit test also verified this incompletely by not properly waiting for the no-op change to propagate to the replica and view-syncer before committing the affecting change.

In this fix, the CVR update / put is delayed until checking for material changes (i.e. to the rows or query tables), correctly implementing the intent of the first PR. The test is also corrected to verify that the no empty pokes are sent.

User report: https://discord.com/channels/830183651022471199/1356104289184907334/1356354680518611208

### DIFF
--- a/packages/zero-cache/src/integration/integration.pg-test.ts
+++ b/packages/zero-cache/src/integration/integration.pg-test.ts
@@ -497,6 +497,12 @@ describe('integration', {timeout: 30000}, () => {
     }
   }
 
+  async function expectNoPokes(client: Queue<unknown>) {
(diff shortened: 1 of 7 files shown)


arv (commented):
LGTM

PR #4152, opened 2025-04-01
chore: fetch upstream schema to disambiguate types

grgbkr:
This is necessary to correctly support uuid and enumerations (without adding even more metadata to schema.ts).

Also aims to fix collation for uuid and enumerations. In postgres uuids by default are sorted by byte and enumerations by order of enum value declaration in enum creation.  To order the same as zql we cast to text and compare as text with `COLLATION "ucs_basic"`.

Removes the `timestamp` and `date` types from schema.ts.

Remove the sql from Inspector as now compiling the sql requires fetching schema from upstream, which is going to take some work to restore to the Inspector. 

Immediate follow ups needed:
Fix sql.test.ts and sql.pg-test.ts.
Add tests for uuid and enumeration.

Eventual follow up:
Add validation of the server schema vs the schema.ts.
Add sql back to Inspector.


packages/zero-pg/src/schema.ts
```diff
@@ -0,0 +1,82 @@
...
+  )) as Iterable<ServerSchemaRow>;
+
+  const serverSchema: ServerSchema = {};
+  for (const row of results) {
+    serverSchema[
+      row.schema === 'public' ? row.table : `${row.schema}.${row.table}`
+    ] = {
+      [row.column]: {
+        type: row.type,
+        isEnum: row.enum.toLowerCase().startsWith('f'),
```
tantaman, line 74: shouldn't this be `row.enum.toLowerCase().startsWith('t')`?

packages/zero-pg/src/schema.ts
```diff
@@ -0,0 +1,82 @@
...
+  const inClause = sql.join(
+    schemaTablePairs.map(
+      ([schema, table]) => sql`(${schema}::text, ${table})::text`,
+    ),
+    ',',
+  );
+
+  const query = sql`
+      SELECT
+          c.table_schema::text as schema,
```
tantaman, line 47: Its worth a comment why we use `::text`

packages/zero-pg/src/schema.ts
```diff
@@ -0,0 +1,82 @@
...
+    schemaTablePairs.push(schemaTablePair);
+    serverToClient[`${schemaTablePair[0]}.${schemaTablePair[1]}`] = name;
+  }
+
+  if (schemaTablePairs.length === 0) {
+    return {}; // No pairs to query for
+  }
+  const inClause = sql.join(
+    schemaTablePairs.map(
+      ([schema, table]) => sql`(${schema}::text, ${table})::text`,
```
tantaman, line 40: Missing a `::text` cast?

`(${schema}::text, ${table}::text)::text`

packages/zero-pg/src/schema.ts
```diff
@@ -0,0 +1,82 @@
...
+  column: string;
+  type: string;
+  enum: string;
+};
+
+export async function getServerSchema<S extends Schema>(
+  dbTransaction: DBTransaction<unknown>,
+  schema: S,
+): Promise<ServerSchema> {
+  const serverToClient: Record<string, string> = {};
```
tantaman, line 18: I'm not seeing where you read from this map.

packages/zero-pg/src/schema.ts
```diff
@@ -0,0 +1,82 @@
...
+      row.schema === 'public' ? row.table : `${row.schema}.${row.table}`
+    ] = {
+      [row.column]: {
+        type: row.type,
+        isEnum: row.enum.toLowerCase().startsWith('f'),
+      },
+    };
+  }
+
+  // Validate that schema is a valid subset of serverSchema
```
tantaman, line 79: nit: // Validate that schema.ts is a valid subset of serverSchema and types are compatible.

packages/z2s/src/sql.ts
```diff
@@ -24,32 +24,63 @@ export function formatSqlite(sql: SQLQuery) {
...
-  type: ValueType,
+export function sqlConvertPluralLiteralArg(
+  type: PluralLiteralType,
+  value: PluralLiteralType[],
+): SQLQuery {
+  const arg: SqlConvertArg = {
+    [sqlConvert]: 'literal',
+    type,
+    value,
+    plural: false,
```
tantaman, line 68: Maybe this'll make sense to me as I review but `plural: false` inside of `sqlConvertPluralLiteralArg` is surprising.

packages/z2s/src/sql.ts
```diff
@@ -73,26 +104,7 @@ class ReusingFormat implements FormatConfig {
...
-    case 'timestamp':
-      return (arg.value as number).toString();
-    case 'string':
-      return arg.value as string;
-    case 'null':
-      return 'null';
-    default:
-      unreachable(arg.type);
-  }
+  return JSON.stringify(arg.value);
```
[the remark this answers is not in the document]
grgbkr: great catch, not on purpose

packages/zero-pg/src/schema.ts
grgbkr: ah the one ::text should be in the parens

packages/zero-pg/src/schema.ts
```diff
@@ -0,0 +1,82 @@
...
+  const {text, values} = formatPg(query);
+  const results: Iterable<ServerSchemaRow> = (await dbTransaction.query(
+    text,
+    values,
+  )) as Iterable<ServerSchemaRow>;
+
+  const serverSchema: ServerSchema = {};
+  for (const row of results) {
+    serverSchema[
+      row.schema === 'public' ? row.table : `${row.schema}.${row.table}`
```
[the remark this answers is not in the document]
grgbkr: The idea is to have this match what nameMapper will produce for server names.

packages/z2s/src/compiler.ts
```diff
@@ -369,32 +377,101 @@ export class Compiler {
...
+              this.#nameMapper.columnName(table, otherValuePos.name)
+            ],
+            valuePos.value,
+            plural,
+          );
+        case 'literal': {
+          assert(plural === Array.isArray(valuePos.value));
+          if (Array.isArray(valuePos.value)) {
+            if (valuePos.value.length > 0) {
+              return sqlConvertPluralLiteralArg(
```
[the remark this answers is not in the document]
grgbkr: Maybe we address it when the feature gets added. 

If we add handling for it, would be hard to test since our types don't allow for it now.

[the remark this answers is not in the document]
grgbkr: Also the above example would go in the column case above, this branch is literal compared to literal.

PR #4154, opened 2025-04-01
fix(zql): When using one in the relation we should add a limit 1

arv:
This adds a limit 1 to the query when using one in the relation. This is fixes an issue where we end up sending multiple rows down where only one is expected.

### DIFF
--- a/packages/zero-schema/src/table-schema.ts
+++ b/packages/zero-schema/src/table-schema.ts
@@ -114,16 +114,14 @@ export function atLeastOne<T>(arr: readonly T[]): AtLeastOne<T> {
   return arr as AtLeastOne<T>;
 }
 
-export function isOneHop(r: Relationship) {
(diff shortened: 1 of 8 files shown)


arv:
The type test is not working yet... looking closer.

I also want to add a runtime test that triggers the invalid behavior.

packages/zero-schema/src/table-schema.ts
```diff
@@ -114,18 +114,16 @@ export function atLeastOne<T>(arr: readonly T[]): AtLeastOne<T> {
   return arr as AtLeastOne<T>;
 }
 
-export function isOneHop(r: Relationship) {
+export function isOneHop(r: Relationship): r is readonly [Connection] {
```
tantaman, line 117: 👍

tantaman (approved):
lgtm

PR #4157, opened 2025-04-01
feat(zero): Add `npx analyze-query`

tantaman:
following the pattern from: https://github.com/rocicorp/mono/pull/4065

CleanShot 2025-04-01 at 13 34 34

### DIFF
--- a/packages/analyze-query/src/bin.ts
+++ b/packages/analyze-query/src/bin.ts
@@ -2,46 +2,46 @@
 import chalk from 'chalk';
 import '@dotenvx/dotenvx/config';
 
-import {testLogConfig} from '../../../otel/src/test-log-config.ts';
-import {createSilentLogContext} from '../../../shared/src/logging-test-utils.ts';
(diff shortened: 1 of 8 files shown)


packages/analyze-query/src/bin.ts
```diff
@@ -153,7 +153,7 @@ function runQuery(queryString: string): [number, number] {
   suppressError.q = q;
   suppressError.z = z;
 
-  eval(`q = ${queryString};`);
+  eval(`q = z.query.${queryString};`);
```
arv, line 156: Try to use new function instead... eval prevents all kind of optimizations.

arv (changes requested):

arv (commented):
LGTM

packages/analyze-query/src/bin.ts
arv: https://github.com/rocicorp/mono/pull/4167

PR #4166, opened 2025-04-02
fix(zql): fix potentially undefined format

grgbkr:
Bit unclear on why TypeScript didn't catch that the implementations of this abstract method did not except undefined for Format but the abstract signature did. 

Regression was introduced here https://github.com/rocicorp/mono/commit/19e5a333e8e43e4e50e8b46336939013d21c322b

### DIFF
--- a/packages/zql/src/query/query-impl.ts
+++ b/packages/zql/src/query/query-impl.ts
@@ -140,7 +140,7 @@ export abstract class AbstractQuery<
     schema: TSchema,
     table: TTable,
     ast: AST,
-    format: Format | undefined,
+    format: Format,
   ): AbstractQuery<TSchema, TTable, TReturn>;
 
   one(): Query<TSchema, TTable, TReturn | undefined> {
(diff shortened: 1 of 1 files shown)


arv:
I guess the override logic in TS is very relaxed. The following is a simplified case:

```ts
abstract class B {
  abstract m(x: number | undefined): void;
  concrete() {
    this.m(undefined);
  }
}

class C extends B {
  override m(x: number): void {
    if (x === undefined) throw 'boo';
     
  }
}

const c: B = new C();
c.m(undefined);
```

arv (commented):
LGTM

grgbkr:
Method pa

> ```ts
> abstract class B {
>   abstract m(x: number | undefined): void;
>   concrete() {
>     this.m(undefined);
>   }
> }
> 
> class C extends B {
>   override m(x: number): void {
>     if (x === undefined) throw 'boo';
>      
>   }
> }
> ```

Surprising.  It's the same with interfaces

```
interface B {
 m(x: number | undefined): void;
}

class C implements B {
   m(x: number): void {
    if (x === undefined) throw 'boo';
     
  }
}
```


This just seems wrong method parameter types should be contravariant, not covariant.

PR #4167, opened 2025-04-03
fix(analyze-query): fix the analyze-query command

arv:
Use Function instead of eval which exposed a bug where we were not awaiting the result.

### DIFF
--- a/packages/analyze-query/src/bin-analyze.ts
+++ b/packages/analyze-query/src/bin-analyze.ts
@@ -15,6 +15,7 @@ import {
   type AST,
   type CompoundKey,
 } from '../../zero-protocol/src/ast.ts';
+import type {Schema} from '../../zero-schema/src/builder/schema-builder.ts';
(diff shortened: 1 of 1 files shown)


packages/analyze-query/src/bin-analyze.ts
```diff
@@ -149,13 +149,12 @@ function runQuery(queryString: string): [number, number] {
...
-  suppressError.q = q;
-  suppressError.z = z;
 
-  eval(`q = z.query.${queryString};`);
+  const f = new Function('z', `return z.query.${queryString};`);
+  const q: Query<Schema, string, PullRow<string, Schema>> = f(z);
 
   const start = performance.now();
-  q.run();
+  await q.run();
```
arv, line 157: Maybe this was not really a problem because we use Promise.resolve for QueryImpl.

tantaman:
thanks

PR #4168, opened 2025-04-03
chore(mutators): allow top level mutators on the server

tantaman:
https://discord.com/channels/830183651022471199/1288232858795769917/1357366018812412085

### DIFF
--- a/packages/zero-pg/tsconfig.json
+++ b/packages/zero-pg/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "lib": ["ESNext.Array", "ESNext"]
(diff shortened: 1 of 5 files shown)


packages/zero-pg/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "lib": ["ESNext.Array", "ESNext"]
+    "lib": ["ESNext.Array", "ESNext", "DOM"]
```
arv, line 4: What DOM api are we using here?

arv (commented):
LGTM

packages/zero-pg/tsconfig.json
arv: tricky

PR #4169, opened 2025-04-03
chore(zero-cache): fix streams test to not rely on node defaults

darkgnotic:
Node v22 has different default high watermark defaults: https://nodejs.org/en/blog/announcements/v22-release-announce#stream-default-high-water-mark

Change the test to not rely on the default value.


arv (approved):
LGTM

PR #4170, opened 2025-04-03
chore(zero-cache): reduce flakiness in test

darkgnotic:
The dropping of a replication slot is non-transactional, and sometimes it takes a bit for pg to update the `pg_replication_slots` table.

Swap the order of the PG calls and add a `sleep()` to reduce false positive failures.


grgbkr:
LGTM

PR #4173, opened 2025-04-04
feat(zero-cache): send pings to check and signal websocket liveness

darkgnotic:
Send pings at 30 second intervals to keep websockets alive in the face of connection idle timeouts (which defaults to 60 seconds on AWS).

Also use them to determine liveness, closing connections if not.

This is used both for the replicator <-> change-streamer connection and for custom change sources.

User report: https://discord.com/channels/830183651022471199/1357355649905987635/1357392133815931122

### DIFF
--- a/packages/zero-cache/src/types/ws.ts
+++ b/packages/zero-cache/src/types/ws.ts
@@ -22,3 +22,59 @@ export function closeWithError(
   // https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close#reason
   ws.close(code, elide(errMsg, 123));
 }
+
+export function sendPingsForLiveness(
+  lc: LogContext,
+  ws: WebSocket,
+  intervalMs: number,
+) {
+  let alive = true;
+  ws.on('pong', () => (alive = true));
+
+  let heartbeatTimer: NodeJS.Timeout | undefined;
+  function startHeartBeats() {
+    heartbeatTimer = setInterval(() => {
+      if (!alive) {
+        lc.warn?.(
+          `socket@${ws.url} did not respond to heartbeat. Terminating...`,
+        );
+        ws.terminate();
+        return;
+      }
+
+      alive = false;
+      ws.ping();
+    }, intervalMs);
+  }
+
+  if (ws.readyState === ws.CONNECTING) {
+    ws.on('open', () => startHeartBeats());
+  } else if (ws.readyState === ws.OPEN) {
+    startHeartBeats();
+  }
+
+  ws.once('close', () => clearInterval(heartbeatTimer));
+}
+
+export function expectPingsForLiveness(
+  lc: LogContext,
+  ws: WebSocket,
+  intervalMs: number,
+  timeoutBufferMs = 3_000,
+) {
+  let missedPingTimer: NodeJS.Timeout | undefined;
+
+  function expectNextPing() {
+    clearTimeout(missedPingTimer);
+
+    missedPingTimer = setTimeout(() => {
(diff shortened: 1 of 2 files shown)


darkgnotic:
@tjenkinson I've adapted your draft PR (https://github.com/rocicorp/mono/pull/4172)

Will this work for you?

packages/zero-cache/src/types/ws.ts
```diff
@@ -22,3 +22,59 @@ export function closeWithError(
...
+        return;
+      }
+
+      alive = false;
+      ws.ping();
+    }, intervalMs);
+  }
+
+  if (ws.readyState === ws.CONNECTING) {
+    ws.on('open', () => startHeartBeats());
```
tjenkinson, line 51:
```suggestion
    ws.once('open', () => startHeartBeats());
```
Bit safer?

packages/zero-cache/src/types/ws.ts
```diff
@@ -22,3 +22,59 @@ export function closeWithError(
...
+  } else if (ws.readyState === ws.OPEN) {
```
tjenkinson, line 52: Nice catch

packages/zero-cache/src/types/ws.ts
```diff
@@ -22,3 +22,59 @@ export function closeWithError(
...
+    clearTimeout(missedPingTimer);
+
+    missedPingTimer = setTimeout(() => {
+      lc.warn?.(`socket@${ws.url} did not send heartbeat. Terminating...`);
+      ws.terminate();
+    }, intervalMs + timeoutBufferMs);
+  }
+
+  ws.on('ping', expectNextPing);
+  ws.on('close', () => clearTimeout(missedPingTimer));
```
tjenkinson, line 77:
```suggestion
  ws.once('close', () => clearTimeout(missedPingTimer));
```

tjenkinson (approved):
Nice! Lgtm thanks 🙏

>Send pings at 30 second intervals to keep websockets alive in the face of connection idle timeouts (which default to 60 seconds on AWS and GCP).

On GCP it refers to a [connection timeout, which defaults to 5 minutes](https://cloud.google.com/run/docs/triggering/websockets#client-reconnects). Is the idle timeout something else? Did you find it mentioned somewhere? Might explain why with the pings alone it seems to be detecting the close and the termination due to heartbeat isn't happening (although I think we should still keep that logic in)

Can deploy this pr next week and check it works

darkgnotic:
> Nice! Lgtm thanks 🙏
> 
> > Send pings at 30 second intervals to keep websockets alive in the face of connection idle timeouts (which default to 60 seconds on AWS and GCP).
> 
> On GCP it refers to a [connection timeout, which defaults to 5 minutes](https://cloud.google.com/run/docs/triggering/websockets#client-reconnects). Is the idle timeout something else? Did you find it mentioned somewhere? Might explain why with the pings alone it seems to be detecting the close and the termination due to heartbeat isn't happening (although I think we should still keep that logic in)
> 
> Can deploy this pr next week and check it works

Ah, interesting. I was mistaken about GCP then. The idle timeout is an AWS thing: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#:~:text=idle_timeout.timeout_seconds

The fact the GCP will close the connection anyway (and that the client doesn't get the signal that the connection closed) means that it will take up to the ping-interval (plus buffer) to determine that the connection closed. Not the end of the world but not ideal.

Maybe the way to solve that is to explicitly close the connection (from the server) before GCP closes it, e.g. with an optional max-connection-lifetime-seconds parameter. Presumably the client _would_ get that close signal and thus be able to reconnect immediately.  🤔

PR #4178, opened 2025-04-07
fix(z2s): return `undefined` for empty result set and `one` format

tantaman:
https://bugs.rocicorp.dev/issue/3715

### DIFF
--- a/packages/zero-pg/src/query.ts
+++ b/packages/zero-pg/src/query.ts
@@ -122,10 +122,19 @@ export class Z2SQuery<
         ),
       );
     this.#query = sqlQuery;
-    const result = extractZqlResult(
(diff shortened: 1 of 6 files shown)


packages/zero-pg/src/query.ts
```diff
@@ -122,10 +122,19 @@ export class Z2SQuery<
         ),
       );
     this.#query = sqlQuery;
-    const result = extractZqlResult(
-      await this.#dbTransaction.query(sqlQuery.text, sqlQuery.values),
+    const pgIterableResult = await this.#dbTransaction.query(
```
grgbkr, line 125: Would be great to document what we expect the shape of pgIterableResult to be.

grgbkr:
LGTM

PR #4179, opened 2025-04-07
feat!(mutators): re-add types to `push-processor`

tantaman:
Users do not know what types to provide and keep getting errors about missing properties.

https://bugs.rocicorp.dev/issue/3704?f=1&d=2
https://bugs.rocicorp.dev/issue/3711
https://discord.com/channels/830183651022471199/1357260393176895508/1357835830256734381

While not fully typed (as in, not asking for the exact shape of the query string and body) we at least now specify that the query params need to be a record or a URLSearchParams object. We also specify that the request body must be provided as JSON.

### DIFF
--- a/packages/zero-pg/src/web.ts
+++ b/packages/zero-pg/src/web.ts
@@ -30,6 +30,7 @@ import {MutationAlreadyProcessedError} from '../../zero-cache/src/services/mutag
 import type {ServerSchema} from '../../z2s/src/schema.ts';
 import {getServerSchema} from './schema.ts';
 import {assert} from '../../shared/src/asserts.ts';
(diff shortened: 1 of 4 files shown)


packages/zero-pg/src/web.ts
```diff
@@ -64,10 +65,13 @@ export class PushProcessor<
 
   async process(
     mutators: MD,
-    queryString: unknown,
-    body: unknown,
+    queryString: URLSearchParams | Record<string, string>,
```
aboodman, line 68: I think it would be good to add some documentation too to this method (and generally any public entrypoints).

aboodman: (I mean comment-based docs that will show up in intellisense)

PR #4183, opened 2025-04-08
feat(zql): make queries thenable

tantaman:
Users can now:

```ts
await z.query.issue.limit(5);
```

rather than

```ts
await z.query.issue.limit(5).run();
```


arv:
I LOVE IT!

I did a quick look at weather `extends Promise` is preferred over `implements PromiseLike` but it seems like then-able is the clear winner due to promise wrapping all over the place.

PR #4188, opened 2025-04-09
chore(zero-client): Doc dev server and JSDoc for zero

arv:
This started as adding JSDoc to zero-client Query but ended up adding a simple doc server generated with Co-Pilot.

Usage

cd packages/zero-client
npm run docs:server

Then go ahead and open http://localhost:3000 in your browser.

Edit your JSDoc and see changes on save.

### DIFF
--- a/packages/zql/src/query/query.ts
+++ b/packages/zql/src/query/query.ts
@@ -124,6 +124,7 @@ export type Row<T extends TableSchema | Query<ZeroSchema, string, any>> =
  * Remember that any `view` returned by `materialize` must be destroyed.
  *
  * A query can be run as a 1-shot query by awaiting it. E.g.,
+ *
  * ```ts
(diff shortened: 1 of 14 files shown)


packages/zql/src/query/query.ts
```diff
@@ -206,28 +302,107 @@ export interface Query<
     ) => Query<TSchema, string>,
   ): Query<TSchema, TTable, TReturn>;
 
+  /**
+   * Skips the rows of the query until row matches the given row. If opts is
+   * provided, it determines whether the match is inclusive.
+   *
+   * @param row The row to start from. This is a partial row object and only the provided
```
tantaman, line 309: I'm 99% certain that the `orderBy` fields must be present in the row.

tantaman: https://bugs.rocicorp.dev/issue/3722

PR #4193, opened 2025-04-09
chore(test): use `zpg` instead of manual compiler invocation

tantaman:
`zpg` removes some manual steps when using the z2s compiler


arv (commented):
LGTM

PR #4199, opened 2025-04-09
chore(zero-cache): allow custom change source to force an auto-reset in initial sync

darkgnotic:
Handle the change-source protocol message:

```js
 ['control', {tag: 'reset-required', message: ...}]
```

during initial sync of custom change sources, responding by auto-resetting and running initial sync from scratch.

User request: https://discord.com/channels/830183651022471199/1356728897085112411/1359618257983766609

@cbnsndwch 

@DAlperin

### DIFF
--- a/packages/zero-cache/src/services/change-source/custom/change-source.ts
+++ b/packages/zero-cache/src/services/change-source/custom/change-source.ts
@@ -185,6 +185,7 @@ export async function initialSync(
 
       case 'status':
         break; // Ignored
+      // @ts-expect-error: falls through if the tag is not 'reset-required
       case 'control': {
         const {tag, message} = change[1];
         if (tag === 'reset-required') {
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/services/change-source/custom/change-source.ts
```diff
@@ -182,7 +185,17 @@ export async function initialSync(
...
         break; // Ignored
-      case 'control':
+      case 'control': {
+        const {tag, message} = change[1];
+        if (tag === 'reset-required') {
+          throw new AutoResetSignal(
+            message ?? 'auto-reset signaled by change source',
+          );
+        }
+        throw new Error(
```
arv, line 195: you can fall through here if you want.

arv (approved):
LGTM

darkgnotic:
> Looks like I don't have access to the Discord chat with the user request. Would love to get the context that prompted the change
> 
> @darkgnotic

This is an api for the change-source implementation to reject incoming requests that have a `&watermark=...` that is no longer available (e.g. gc'ed, compacted, etc.). By returning the `reset-required` control message, the change source signals to the subscriber that they need to reset (e.g. perform initial-sync) from the current head. the zero-cache will respond by sending a watermark-less request to perform initial sync.

Does that clarify?

cbnsndwch:
Yup! Thanks!

PR #4200, opened 2025-04-09
chore(zql)!: Deprecate Query updateTTL

arv:
BREAKING CHANGE. Bindings need to be updated.

Instead of hanging this of off the Query we add it to the View that materialize returns. This allows us to remove the updateTTL method from the Query interface and the QueryImpl class.


tantaman:
lgtm. If a query with the given hash already has a TTL that is longer than the one provided to `updateTTL`, is `updateTTL` a no-op?

arv:
> lgtm. If a query with the given hash already has a TTL that is longer than the one provided to `updateTTL`, is `updateTTL` a no-op?

Right, we should check that so we do not send a websocket message

PR #4202, opened 2025-04-09
feat(mutators)!: no longer reject server promise on errors

tantaman:
Discussed with @aboodman --

Any errors returned by the remote endpoint now resolve, rather than reject, the promise and set an error field on the result.


arv (approved):
LGTM

PR #4205, opened 2025-04-10
chore(zero-solid): rename useQuery to createQuery

arv:
The naming conventions for SolidJS is to name things createX over useX.


arv:
This also moves the doc server to packages/zero since I wanted to see how these changes rendered.

PR #4211, opened 2025-04-10
feat(zero-cache): option to run the zero-cache lazily

darkgnotic:
### Feature

Adds a `--run-lazily` (or `ZERO_RUN_LAZILY`) option to delay starting the zero-cache processes until the receiving the first (websocket) request:

`--run-lazily` usage documentation:

```
        Delay starting the zero-cache processes until the first request.

        Note: This works as expected in single-node mode. While it is technically usable
        in a multi-node setup, there is a bootstrapping complication in that the
        view-syncer must first restore the replica from litestream before connecting to
        the replication-manager (to know where to continue replication from). If the
        replication-manager has never run, there will be no replica file to restore, and
        the view-syncer will fail to start up, never connecting to the replication-manager.

        As such, it is not recommended to run a replication-manager lazily.
```

User request: https://discord.com/channels/830183651022471199/1359461732518592532/1359836126885445662

### Implementation

* The websocket handoff framework was modified to allow the handoff logic to invoke a callback instead of immediately returning the Worker to dispatch the websocket to.
* The top-level runner now spawns a Worker on demand, upon the first call to `getWorker()`. 
* These Workers are spawned eagerly if the `--run-lazily` option is not set (i.e. to achieve the status quo) .
* Those that are not spawned eagerly are spawned upon the first call to `getWorker()`, which happens when receiving a request for that worker.

### DIFF
--- a/packages/shared/src/options.ts
+++ b/packages/shared/src/options.ts
@@ -104,14 +104,13 @@ export type Group = Record<string, Option>;
 export type Options = Record<string, Group | Option>;
 
 /** Unwrap the Value type from an Option<V>. */
-type ValueOf<T extends Option> =
-  T extends v.Optional<infer V>
-    ? V | undefined
-    : T extends v.Type<infer V>
-      ? V
-      : T extends WrappedOptionType
-        ? ValueOf<T['type']>
-        : never;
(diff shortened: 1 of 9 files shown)


arv (commented):
LGTM

Should it maybe be an error to have this in a multi node setup?

PR #4212, opened 2025-04-11
fix(zero-cache): resync when publications are missing

darkgnotic:
Recover from upstream corruption that stopped being detected when resyncs were made to be non-disruptive.

Extra checks are performed before initial sync and replication subscription to ensure that the expected publications still exist upstream. If they are missing, a full (disruptive) resync is used to recover.

User report: https://discord.com/channels/830183651022471199/1358765533545369842/1359925435902726336


arv (commented):
LGTM

PR #4213, opened 2025-04-11
chore(infra): SST with ECS/EC2

cesara:
This config allows for the launching on ECS tasks on top of EC2 containers


prod/sst-ecs-ec2/sst.config.ts
```diff
@@ -0,0 +1,144 @@
...
+      ZERO_LOG_FORMAT: 'json',
+      ZERO_REPLICA_FILE: 'sync-replica.db',
+      ZERO_LITESTREAM_BACKUP_URL: replicationBucket.name.apply(
+        name => `s3://${name}/backup/20250319-00`,
+      ),
+      ZERO_IMAGE_URL: process.env.ZERO_IMAGE_URL!,
+      ZERO_APP_ID: process.env.ZERO_APP_ID || 'zero',
+    };
+
+    //Create the replication manager service
```
darkgnotic, line 87: Nit: Add space after '//'

prod/sst-ecs-ec2/sst.config.ts
```diff
@@ -0,0 +1,144 @@
...
+    return {
+      name: process.env.APP_NAME || 'zero',
+      removal: input?.stage === 'production' ? 'retain' : 'remove',
+      home: 'aws',
+      region: process.env.AWS_REGION || 'us-east-1',
+      providers: {command: '1.0.2', docker: true},
+    };
+  },
+  async run() {
+    const {networkConfig} = await import('./infra/network');
```
darkgnotic, line 32: Can you not use normal imports at the top of the file? 

I think I see them in the other `sst.config.ts`.

darkgnotic (approved):
This is amazing work. Mad props!   💪 

While there are a lot of parallels between the original `sst.config.ts` and the new one, I can see how it would be difficult to try to turn them into one config with a fargate vs ec2 "plugin" type structure. So I think leaving them as completely separate trees seems reasonable.

In terms of naming / organization, most of the code here isn't sst anymore and perhaps we should no longer be naming our folders by the deployment mechanism. What do you think of something like:

```
prod/ec2/
prod/ec2/sst.config.ts
prod/ec2/infra/...
prod/fargate/sst.config.ts   <-- formerly known as prod/sst/sst.config.ts
prod/functions/...
```

darkgnotic:
> 
> ```
> prod/ec2/
> prod/ec2/sst.config.ts
> prod/ec2/infra/...
> prod/fargate/sst.config.ts   <-- formerly known as prod/sst/sst.config.ts
> prod/functions/...
> ```

And if you like this, feel free to rename the `prod/sst/...` stuff in a separate PR.   😸

prod/sst-ecs-ec2/sst.config.ts
cesara: with the latest version of sst, you have to do imports this way 
https://github.com/sst/sst/issues/5616

PR #4214, opened 2025-04-11
chore(zero-pg): Support Request param ion PushProcessor

arv:
Almost all the framework (Next.js, SolidStart, TanStack Start) pass in a Request in their POST handlers. This is a standard Web API interface that is popular in FE servers.


arv:
This makes Next.js, SolidStart, TanStack Start one liners:

```ts
// Next.js
export function POST(request: Request) {
  return processor.process(createMutators(), request);
}


// SolidStart
export function POST({request}: APIEvent) {
  return processor.process(createMutators(), request);
}

// TanStack Start
export const APIRoute = createAPIFileRoute('/api/push')({
  POST: ({request}) => processor.process(createMutators(), request)
});
```

PR #4216, opened 2025-04-11
Fix issue with timestamps and `timestamptz` data types

tjenkinson:
Currently if you have a `timestamptz` column, and the server timezone is `Europe/London`, when a timestamp is written it's an hour out 😱, because they query ends up being

```sql
INSERT INTO "table" ("id","createdAt") VALUES ($1::text::character varying, to_timestamp($2::text::bigint / 1000.0) AT TIME ZONE 'UTC')
```

where
- `to_timestamp($2::text::bigint / 1000.0)` returns a timestamp with timezone, where the timezone is `Europe/London`
- `AT TIME ZONE 'UTC'` then converts it to a timestamp without the timezone, representing the time in UTC
- Then the insert happens into the column that does want a timezone. It interprets the timestamp as being a timestamp in `Europe/London`, not `UTC`, and therefore it ends up being an hour out

So in this PR I've adjusted it so that if the column is a `timestamptz` we don't convert it to a timezoneless timestamp in UTC. If the column is a `timestamp`, then we do ensure it's converted to UTC (although I wonder if this is the right thing to do?)

I've done a quick test in our app and this fixes it, but not 100% sure if this is correct

### Some queries I ran when debugging this
```sql
select to_timestamp(1::text::bigint / 1000.0);
--> 1970-01-01 01:00:00.001+01
```

```sql
select to_timestamp(1::text::bigint / 1000.0) AT TIME ZONE 'UTC';
--> 1970-01-01 00:00:00.001
```

### DIFF
--- a/packages/z2s/src/compiler.output.test.ts
+++ b/packages/z2s/src/compiler.output.test.ts
@@ -280,7 +280,7 @@ test('compile with timestamp (with timezone)', () => {
     ),
   ).toMatchInlineSnapshot(`
     {
-      "text": "SELECT COALESCE(json_agg(row_to_json("root")), '[]'::json)::text as "zql_result" FROM (SELECT "timestampsTable"."id","timestampsTable"."timestampWithTz",EXTRACT(EPOCH FROM "timestampsTable"."timestampWithoutTz"::timestamp AT TIME ZONE 'UTC') * 1000 as "timestampWithoutTz" FROM "timestampsTable" WHERE "timestampWithTz" = $1::text::timestampz   )"root"",
+      "text": "SELECT COALESCE(json_agg(row_to_json("root")), '[]'::json)::text as "zql_result" FROM (SELECT "timestampsTable"."id","timestampsTable"."timestampWithTz",EXTRACT(EPOCH FROM "timestampsTable"."timestampWithoutTz") * 1000 as "timestampWithoutTz" FROM "timestampsTable" WHERE "timestampWithTz" = $1::text::timestampz   )"root"",
       "values": [
         ""abc"",
       ],
@@ -305,7 +305,7 @@ test('compile with timestamp (without timezone)', () => {
     ),
   ).toMatchInlineSnapshot(`
     {
-      "text": "SELECT COALESCE(json_agg(row_to_json("root")), '[]'::json)::text as "zql_result" FROM (SELECT "timestampsTable"."id","timestampsTable"."timestampWithTz",EXTRACT(EPOCH FROM "timestampsTable"."timestampWithoutTz"::timestamp AT TIME ZONE 'UTC') * 1000 as "timestampWithoutTz" FROM "timestampsTable" WHERE "timestampWithoutTz" = to_timestamp($1::text::bigint / 1000.0) AT TIME ZONE 'UTC'   )"root"",
+      "text": "SELECT COALESCE(json_agg(row_to_json("root")), '[]'::json)::text as "zql_result" FROM (SELECT "timestampsTable"."id","timestampsTable"."timestampWithTz",EXTRACT(EPOCH FROM "timestampsTable"."timestampWithoutTz") * 1000 as "timestampWithoutTz" FROM "timestampsTable" WHERE "timestampWithoutTz" = to_timestamp($1::text::bigint / 1000.0) AT TIME ZONE 'UTC'   )"root"",
       "values": [
         ""abc"",
       ],
@@ -957,7 +957,7 @@ test('related thru junction edge', () => {
     {
       "text": "SELECT COALESCE(json_agg(row_to_json("root")), '[]'::json)::text as "zql_result" FROM (SELECT (
             SELECT COALESCE(json_agg(row_to_json("inner_labels")), '[]'::json) FROM (SELECT "table_1"."id","table_1"."name" FROM "issue_label" as "issueLabel" JOIN "label" as "table_1" ON "issueLabel"."label_id" = "table_1"."id" WHERE ("issue"."id" = "issueLabel"."issue_id")    ) "inner_labels"
-          ) as "labels","issue"."id","issue"."title","issue"."description","issue"."closed","issue"."ownerId",EXTRACT(EPOCH FROM "issue"."created"::timestamp AT TIME ZONE 'UTC') * 1000 as "created" FROM "issue"    )"root"",
+          ) as "labels","issue"."id","issue"."title","issue"."description","issue"."closed","issue"."ownerId",EXTRACT(EPOCH FROM "issue"."created") * 1000 as "created" FROM "issue"    )"root"",
       "values": [],
     }
   `);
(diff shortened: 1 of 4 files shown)


tantaman:
Thanks for this. This is indeed a problem whenever the postgres server has a timezone set to anything but UTC since a column with a time zone will mark the input value as being recorded at the server's time zone. So, as you observed, the UTC epoch we insert gets marked with the Eurpoe/London timezone, skewing it 1 hour.

```
user@127.0.0.1:postgres> show timezone;
+----------+
| TimeZone |
|----------|
| EST      |
+----------+
user@127.0.0.1:postgres> insert into foo values (TO_TIMESTAMP(1744392199));
INSERT 0 1
Time: 0.004s
user@127.0.0.1:postgres> SELECT * FROM foo;
+------------------------+
| a                      |
|------------------------|
| 2025-04-11 12:23:19-05 |
+------------------------+
user@127.0.0.1:postgres> insert into foo values (TO_TIMESTAMP(1744392199) AT TIME ZONE 'UTC');
INSERT 0 1
Time: 0.004s
user@127.0.0.1:postgres> SELECT * FROM foo;
+------------------------+
| a                      |
|------------------------|
| 2025-04-11 12:23:19-05 |
| 2025-04-11 17:23:19-05 |
+------------------------+
```

Same timestamp (1744392199), but adjusted differently by the table.

> If the column is a timestamp [without time zone], then we do ensure it's converted to UTC (although I wonder if this is the right thing to do?)

I think that depends on the assumption the developer has about a timezoneless column. Unfortunately we cannot know what the developer intends to store in a timezoneless column.

- If we store it as `TO_TIMESTAMP(1744392199)` and the developer is expecting UTC times then everything we insert will be off by +/- the server's timezone offset
- If we store it as `TO_TIMESTAMP(1744392199) AT TIME ZONE 'UTC'` and the developer is expecting times in their server's time zone... we'll also be off by the same amount but in a different direction.

packages/z2s/src/compiler.ts
```diff
@@ -596,20 +596,29 @@ export class Compiler {
...
-        column,
-      )}::timestamp AT TIME ZONE 'UTC') * 1000 as ${sql.ident(column)}`;
+        serverType === 'timestamp with time zone'
+      ) {
+        return sql`EXTRACT(EPOCH FROM ${sql.ident(
+          table,
+        )}.${this.#mapColumnNoAlias(
+          table,
+          column,
+        )}::timestamp) * 1000 as ${sql.ident(column)}`;
```
tantaman, line 620: I think this should just be:

```suggestion
        return sql`EXTRACT(EPOCH FROM ${sql.ident(
          table,
        )}.${this.#mapColumnNoAlias(
          table,
          column,
        )}) * 1000 as ${sql.ident(column)}`;
```

as that round-trips correctly (1744392199 in and 1744392199 out):

```
user@127.0.0.1:postgres> create table foo (a timestamp with time zone);
CREATE TABLE
Time: 0.028s
user@127.0.0.1:postgres> INSERT INTO foo VALUES (to_timestamp(1744392199));
INSERT 0 1
Time: 0.004s
user@127.0.0.1:postgres> select * from foo;
+------------------------+
| a                      |
|------------------------|
| 2025-04-11 12:23:19-05 |
+------------------------+
SELECT 1
Time: 0.006s
user@127.0.0.1:postgres> SELECT EXTRACT(EPOCH FROM "a") FROM foo;
+-------------------+
| extract           |
|-------------------|
| 1744392199.000000 |
+-------------------+
SELECT 1
Time: 0.008s
user@127.0.0.1:postgres>
```

packages/z2s/src/compiler.ts
```diff
@@ -596,20 +596,29 @@ export class Compiler {
...
+        serverType === 'date' ||
         serverType === 'timestamp' ||
+        serverType === 'timestamp without time zone'
+      ) {
+        return sql`EXTRACT(EPOCH FROM ${sql.ident(
+          table,
+        )}.${this.#mapColumnNoAlias(
+          table,
+          column,
+        )}::timestamp AT TIME ZONE 'UTC') * 1000 as ${sql.ident(column)}`;
```
tantaman, line 610: Should be:

```suggestion
        return sql`EXTRACT(EPOCH FROM ${sql.ident(
          table,
        )}.${this.#mapColumnNoAlias(
          table,
          column,
        )}) * 1000 as ${sql.ident(column)}`;
```

round-trips correctly:

```
user@127.0.0.1:postgres> CREATE TABLE bar (a timestamp without time zone);
CREATE TABLE
Time: 0.028s
user@127.0.0.1:postgres> INSERT INTO bar VALUES(to_timestamp(1744392199) AT TIME ZONE 'UTC');
INSERT 0 1
Time: 0.011s
user@127.0.0.1:postgres> SELECT * FROM bar;
+---------------------+
| a                   |
|---------------------|
| 2025-04-11 17:23:19 |
+---------------------+
SELECT 1
Time: 0.004s
user@127.0.0.1:postgres> SELECT EXTRACT(EPOCH FROM "a") FROM bar;
+-------------------+
| extract           |
|-------------------|
| 1744392199.000000 |
+-------------------+
SELECT 1
Time: 0.005s
user@127.0.0.1:postgres>
```

The reason is that `EXTRACT(EPOCH FROM` is already returning an epoch time which is at UTC.

tantaman:
> The epoch time number written in should always be the one that comes back?

correct

packages/z2s/src/compiler.ts
tjenkinson: this is just because the `::timestamp` cast is redundant?

packages/z2s/src/compiler.ts
tjenkinson: ah yeh so this one the `::timestamp` cast was stripping the timezone, so then it was being treated as local timezone and converted to UTC, and the epoch taken from that /0\

packages/z2s/src/compiler.ts
tjenkinson: and I guess `EPOCH FROM` with a `timestamp` just treats it as UTC

packages/z2s/src/compiler.ts
tjenkinson: wait no there was no timezone, so that converted it to a `timestamptz`, interpreting as a timestamp in the user local timezone 🤯

packages/z2s/src/compiler.ts
tjenkinson: sorry read it again. This was a `timestamptz`, and the cast was stripping the tz, and then we were treating it as local time and converting that to utc

packages/z2s/src/compiler.output.test.ts
```diff
@@ -130,6 +139,11 @@ const serverSchema: ServerSchema = {
     id: {type: 'text', isEnum: false},
     status: {type: 'statusEnum', isEnum: true},
   },
+  'timestampsTable': {
+    id: {type: 'text', isEnum: false},
+    timestampWithoutTz: {type: 'timestamp', isEnum: false},
+    timestampWithTz: {type: 'timestampz', isEnum: false},
```
tjenkinson, line 145: oops

packages/z2s/src/compiler.ts
```diff
@@ -596,20 +596,29 @@ export class Compiler {
...
       ];
     const serverType = serverColumnSchema.type;
-    if (
-      !serverColumnSchema.isEnum &&
-      (serverType === 'date' ||
+    if (!serverColumnSchema.isEnum) {
+      if (
+        serverType === 'date' ||
         serverType === 'timestamp' ||
+        serverType === 'timestamp without time zone'
```
grgbkr, line 603: I believe these two branches are the same again, and can be merged.

grgbkr (approved):
LGTM.  Thanks so much for identifying this and putting together a fix.

packages/z2s/src/compiler.ts
tjenkinson: ah yep. done

PR #4217, opened 2025-04-11
chore(zql): Fix missing NOT LIKE operators in type

arv:
By using the one from the protocol these now have a single source of truth.


tantaman:
lgtm. Clean 🧽

PR #4218, opened 2025-04-11
feat!(zbugs): remove all write permissions

tantaman:
zubgs strictly uses custom mutators. We do not need write permissions anymore.


arv (commented):
LGTM

🎉

PR #4219, opened 2025-04-11
feat(mutators): allow custom headers and query string params

tantaman:
- We previously had no way to initialize the pusher service with custom parameters from the client. This adds an `initConnection` method to `pusher` and extra arguments to the `initConnection` protocol message. 
- Responses to `initConnection` previously could only return a single value but now we must return two: the pusher stream and the view syncer stream.
- The logic to determine if a user had opted into custom mutators was based around the presence of `pushURL` in the config but this can now come from the client itself.

### DIFF
--- a/packages/zero-cache/src/services/mutagen/pusher.test.ts
+++ b/packages/zero-cache/src/services/mutagen/pusher.test.ts
@@ -20,7 +20,7 @@ const config = {
 };
 
 const clientID = 'test-cid';
-const wsID = 'test-wsid';
+// const wsID = 'test-wsid';
 describe('combine pushes', () => {
   test('empty array', () => {
     const [pushes, terminate] = combinePushes([]);
@@ -33,14 +33,17 @@ describe('combine pushes', () => {
       {
         push: makePush(1),
         jwt: 'a',
+        clientID,
       },
       {
         push: makePush(1),
         jwt: 'a',
+        clientID,
       },
       {
         push: makePush(1),
         jwt: 'a',
+        clientID,
       },
     ]);
     expect(pushes).toHaveLength(1);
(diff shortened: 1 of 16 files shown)


packages/zero-client/src/client/options.ts
```diff
@@ -77,7 +77,27 @@ export interface ZeroOptions<
...
+   * your API server. If you would like specific headers to be available
+   * to your API server when it receives a custom mutation, you can
+   * specify them here.
+   */
+  pushParams?:
+    | {
+        headers?: Record<string, string> | undefined;
+        queryParams?: Record<string, string> | undefined;
+      }
+    | undefined;
```
tantaman, line 100: @aboodman - wdyt about exposing the headers and query params to tack onto pushes here?

```ts
const z = new Zero({
  pushParams: {
    header: { 'some-header': '...' },
   queryParams: {'some-query-param': '...' },
  }
});
```

packages/zero-cache/src/services/mutagen/pusher.test.ts
```diff
@@ -20,7 +20,7 @@ const config = {
 };
 
 const clientID = 'test-cid';
-const wsID = 'test-wsid';
+// const wsID = 'test-wsid';
```
tantaman, line 23: note to self: need to update these tests to call `initConnection` now

aboodman:
I think let's just allow users to specify `pushURL` in the constructor.

We already allow them (require them) to specify `serverURL`. It doesn't seem like a difference security-wise. And they can already control output connections allowed via CSP.

thdxr:
this PR is great
it's fantastic, never seen a PR like this
they're saying this PR should be merged immediately

PR #4220, opened 2025-04-11
chore(zero-cache): handle socket closures during websocket handoff

darkgnotic:
Discovered by the load harness. The socket handle is `undefined` if it was closed before is it handed off to the subprocess.

Failing to check this will otherwise crash the server.


darkgnotic:
> LGTM
> 
> I'm not sure if we want add a test for this?

Good point. I'll do that!

PR #4221, opened 2025-04-12
fix(z2s): junction relationship fixes

grgbkr:
Reported on discord: https://discord.com/channels/830183651022471199/1288232858795769917/1360371961141661936

Relationships nested in a junction relationship were not working correctly.

Additionally, `#selectCol` conversion logic for date and timestamp types were not being applied to columns of junction relationships.

### DIFF
--- a/packages/z2s/src/compiler.ts
+++ b/packages/z2s/src/compiler.ts
@@ -224,9 +224,12 @@ export class Compiler {
     if (relationship.hidden) {
       const [join, lastAlias, lastLimit, lastTable] =
         this.makeJunctionJoin(relationship);
-      assert(relationship.subquery.related);
-      const nestedAst = relationship.subquery.related[0].subquery;
 
+      assert(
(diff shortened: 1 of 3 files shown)


packages/z2s/src/compiler.ts
```diff
@@ -256,46 +266,55 @@ export class Compiler {
...
-      )}) ${sql.ident(`inner_${relationship.subquery.alias}`)}
+      )}) ${sql.ident(innerAlias)}
     ) as ${sql.ident(relationship.subquery.alias)}`;
   }
 
   pullTablesForJunction(
     relationship: CorrelatedSubquery,
-    tables: [string, Correlation, number | undefined][] = [],
-  ) {
+  ): [
```
grgbkr, line 290: @tantaman This seemed to be trying to support an arbitrary chain of hidden relationships, but afaik we only support a single junction hop.  Was my misunderstanding what this was doing?

PR #4224, opened 2025-04-14
Put db in `UTC+1` for integration tests

tjenkinson:
Followup for https://github.com/rocicorp/mono/pull/4216

I tested (`zql-integration-tests`) with the patch in the above PR reverted and things broke, so this adds some coverage

### DIFF
--- a/packages/zero-cache/test/pg-container-setup.ts
+++ b/packages/zero-cache/test/pg-container-setup.ts
@@ -1,6 +1,6 @@
 import {PostgreSqlContainer} from '@testcontainers/postgresql';
 
-export function runPostgresContainer(image: string) {
+export function runPostgresContainer(image: string, timezone: string) {
   return async ({provide}) => {
     const container = await new PostgreSqlContainer(image)
       .withCommand([
@@ -8,7 +8,7 @@ export function runPostgresContainer(image: string) {
         '-c',
         'wal_level=logical',
         '-c',
-        'timezone=UTC+1',
+        `timezone=${timezone}`,
       ])
       .start();
 
--- a/packages/zero-cache/test/pg-container-setup.ts
+++ b/packages/zero-cache/test/pg-container-setup.ts
@@ -3,7 +3,13 @@ import {PostgreSqlContainer} from '@testcontainers/postgresql';
 export function runPostgresContainer(image: string) {
   return async ({provide}) => {
     const container = await new PostgreSqlContainer(image)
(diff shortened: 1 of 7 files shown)


packages/zero-cache/test/pg-container-setup.ts
```diff
@@ -3,7 +3,13 @@ import {PostgreSqlContainer} from '@testcontainers/postgresql';
 export function runPostgresContainer(image: string) {
   return async ({provide}) => {
     const container = await new PostgreSqlContainer(image)
-      .withCommand(['postgres', '-c', 'wal_level=logical'])
+      .withCommand([
+        'postgres',
+        '-c',
+        'wal_level=logical',
+        '-c',
+        'timezone=UTC+1',
```
tantaman, line 11: @darkgnotic - wdyt? Run every test at UTC+1 or just 1 of the PG containers?

packages/zql/src/query/test/test-schemas.ts
```diff
@@ -198,6 +198,7 @@ CREATE TABLE IF NOT EXISTS "comments" (
   "authorId" TEXT NOT NULL,
   "issue_id" TEXT NOT NULL,
   "text" TEXT NOT NULL,
+  -- not TIMESTAMPTZ so that we're checking both TIMESTAMP and TIMESTAMPTZ behaviour
   "createdAt" TIMESTAMP NOT NULL
```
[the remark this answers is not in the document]
tantaman: Yeah, I think this is fine for the test schema since we want to try various types.

packages/zero-cache/test/pg-container-setup.ts
darkgnotic: Just 1 of them sounds like a good idea. Otherwise we wouldn't catch the issue of them being all incorrect.

tjenkinson: sure will update

tjenkinson: done

tantaman:
> Bit confused because in sql.ts createPlaceholder there's also some handling where it's plural and that's missing the AT TIME ZONE 'UTC' for the timestamp cases. Maybe it needs it?

Checking this out. Been on spring break over the last week.

tantaman:
Some of the test failures are having too much precision from Postgres. That's not too bad.

CleanShot 2025-04-21 at 12 19 34



The real worrying ones are the ones with completely different values before the decimal part.


CleanShot 2025-04-21 at 12 19 45

PR #4225, opened 2025-04-14
fix(zero): TTL is of by 1000x in some cases

arv:
https://bugs.rocicorp.dev/issue/3729

The SQL type for ttl is INTERVAL which when set as a number means 1s.

There was one code path where the conversion from ms to s was not done.


darkgnotic (commented):
LGTM

darned units.

arv:
Yeah, I kind of wish I had just used float64 in the first place.

PR #4226, opened 2025-04-14
feat(zero)!: Add options to Query run

arv:
BREAKING CHANGE!

```ts
type RunOptions = {
  type: 'complete' | 'unknown';
}

interface Query<...> {
  ...
  run(options?: RunOptions): Promise<...>;
}
```

Defaults to `complete` for all queries except for the query used inside
custom mutations which defaults to `unknown`.

This is breaking because the `run` method of the `Query` class now waits
for `complete` results by default. This can lead to the promise returned
never resolving if the `server` is not set (or we are offline).

Another gotcha is that if you call run with `unknown` you get both
unknown and complete results but there is no way to tell which is
which.

### DIFF
--- a/packages/analyze-query/src/bin-analyze.ts
+++ b/packages/analyze-query/src/bin-analyze.ts
@@ -171,6 +171,7 @@ const host: QueryDelegate = {
   normalizeRunOptions(options) {
     return options ?? DEFAULT_RUN_OPTIONS_COMPLETE;
   },
+  defaultQueryComplete: true,
 };
 
 let start: number;
--- a/packages/analyze-query/src/bin-analyze.ts
+++ b/packages/analyze-query/src/bin-analyze.ts
@@ -1,10 +1,14 @@
 /* eslint-disable no-console */
 import '@dotenvx/dotenvx/config';
 import chalk from 'chalk';
+import {astToZQL} from '../../ast-to-zql/src/ast-to-zql.ts';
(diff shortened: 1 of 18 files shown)


arv:
This WIP does not deal with the offline case which is problematic with `run({resultType: 'complete'})`

https://www.notion.so/replicache/Query-run-1d53bed895458059abc5f9cfd4fc061d?pvs=4#1d53bed89545809183e2fb0780c7f263

aboodman:
The main reason for this API shape in my mind is consistency with the `{type}` property on the details return value from `useQuery`.

I can see it feels a little awkward but it feels worth it to me for the consistency?

arv:
> I guess the field name should actually be `type` not `resultType` to align with: https://zero.rocicorp.dev/docs/reading-data#completeness.

At one point I wanted to change `type` to `status`.

https://bugs.rocicorp.dev/issue/3489

https://github.com/danielroe/zero-vue/pull/24#issuecomment-2624237985

https://github.com/danielroe/zero-vue/blob/13ac47e0b4926333815e66aeec39c5ef9ad43c73/src/query.ts#L22-L25

but if we want these to align then I do not think `status` works.

I also do not particularly like using `unknown` as a flag to run.

```ts
await q.run({type: 'unknown'});
const [issues, {type}] = useQuery(q);
if (type === 'unknown') {
  ...
}
```

but I do think aligning them has value.

making these changes...

PR #4228, opened 2025-04-14
feat(zero-cache)!: only allow `pushURL` to be configured and not headers

tantaman:
https://rocicorp.slack.com/archives/C013XFG80JC/p1744651637357629


arv:
This PR was superseded by only allowing passing query params

PR #4233, opened 2025-04-16
Change the default type of `run()` to 'partial'.

aboodman:


aboodman:
This makes it no longer a breaking change and I think is also just a better default.

arv (commented):
This change is not complete.

https://github.com/rocicorp/mono/pull/4234

PR #4234, opened 2025-04-16
chore(zero-client)!: Change run default to unknown

arv:
BREAKING CHANGE -- This is a breaking change from a previous change but it restores the old 0.18 behavior of running the default query to unknown.

This simplifies the code since run always defaults to unknown now.

We still throw if you try to pass complete to a run inside a custom mutator.


aboodman:
Thanks!

PR #4237, opened 2025-04-16
chore(zbugs): fix zbugs dev script and documentation

jevakallio:
I had some trouble getting the `zbugs` app running from a clean clone.

- The docs said to run the `zero` script instead of the the `zero-cache-dev`
  - This started the server, but does not deploy permissions
- I tried deploying permissions with `npx zero-deploy-permissions` but ran into two problems
  - The local repository `.bin` scripts were not linked (fixed by https://github.com/rocicorp/mono/commit/437e9b1d6cabe56e961f95753d6d973ce64fb81b)
  - The schema file is not in the default path `./schema.ts` (fixed specifying by the `--schema-path` param)

I elected to fix this by running the dev server instead and specifying the schema path there.

I'm not sure why you'd want to run the production runner instead of the `zero-cache-dev` server. If that was intentional, the documentation should document the `zero-deploy-permissions` step instead.

In both cases, you still have to first build the workspace packages and re-install to get npm workspaces to link the binaries.


tantaman:
Thanks. I think things had just gotten stale.

aboodman (approved):
Thank you @jevakallio

PR #4244, opened 2025-04-22
chore(zero-cache): log context tracker

tantaman:
Simplify managing log context so it
  is available without threading it through a callstack.
 E.g., 
  ```ts
  app.get('/foo', (req, res) => {
    withContext({userId: req.user.id}, () => {
       someFn();
    });
  });
 
  // someFn can be arbitrarily nested
  // and still access the context.
  function someFn() {
   log(`User ID: ${getContext().userId}`);
  }
  ```


darkgnotic:
TIL about [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage)  😁 

Personally I like explicitly threading context objects through the call stack (I am a fan of how Golang does it too) but am open to adapting to what the preferred style is in the JS/TS world. The AsyncLocalStorage API doesn't require threading the argument through the callstack, but does require wrapping each context-adding step with a lambda. Is it just a matter of style preference or are there other benefits to doing it that way? (I'm not objecting, but rather curious)

For my benefit, can you also flesh out how we would migrate our logging to work with this? Would we instrument the two log sinks in the [logging file](https://github.com/rocicorp/mono/blob/6d055c04a3a21b37a5188f6484297cddbd82e3e1/packages/zero-cache/src/server/logging.ts#L13) to get the Context object from AsyncLocalStorage, and then replace all threaded LogContext code to instead call a `getLogger()` method or something?

tantaman:
>  Is it just a matter of style preference or are there other benefits to doing it that way?

Preference in order to simplify getting context values to every place that needs them.

A problem we have today with `lc` is that there are a number of classes that take the `lc` in their constructor. This isn't quite right as callers to methods on that class may have additional context they want to provide. They can't do that if the class is using its own internal `lc`.

> Personally I like explicitly threading context objects through the call stack 

@aboodman prefers this as well so I'll abandon this PR.

> Would we instrument the two log sinks in the [logging file](https://github.com/rocicorp/mono/blob/6d055c04a3a21b37a5188f6484297cddbd82e3e1/packages/zero-cache/src/server/logging.ts#L13) to get the Context object from AsyncLocalStorage, and then replace all threaded LogContext code to instead call a getLogger() method or something?

That's right.


---

Random: some more exotic languages support compile time checks of context as though context values are implicit function parameters -- https://tomasp.net/coeffects/ & https://github.com/facebook/hhvm/blob/master/hphp/hack/doc/HIPs/contexts_and_coeffects.md

darkgnotic:
> A problem we have today with `lc` is that there are a number of classes that take the `lc` in their constructor. This isn't quite right as callers to methods on that class may have additional context they want to provide. They can't do that if the class is using its own internal `lc`.
> 

Yeah, this is an interesting quandary. Often the classes that take an `lc` are services that both (1) run their own continuous background logic (in which case they need an `lc` with the values passed to the constructor) and (2) respond to invocations that have more context. Maybe we need an API that merges context from the callstack with "object" context. But I see that there's a slight impedance mismatch in that the LogContext is both context and logger/sink, and the "merge" api doesn't really make sense for the latter.

PR #4245, opened 2025-04-22
chore(zero-cache): more debug logs

tantaman:
JB is hitting a case where a refresh of a tab never produces any logs at all. We're bailing out before we get to any logger calls, pokes, hydrations, etc.

Theories:
- an exception is thrown and swallowed
- `if (client?.wsID !== wsID) {` is always true causing us to early return
- we never acquire the cvr lock and thus never run `initConnection`

this adds more debug logs up the call stack to figure out where things break down.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/view-syncer.ts
+++ b/packages/zero-cache/src/services/view-syncer/view-syncer.ts
@@ -399,6 +399,7 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
     ctx: SyncContext,
     initConnectionMessage: InitConnectionMessage,
   ): Source<Downstream> {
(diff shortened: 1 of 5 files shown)


packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -617,8 +620,8 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
         }
 
         // Apply requested patches.
+        lc.debug?.(`applying ${desiredQueriesPatch?.length} query patches`);
```
tantaman, line 623: his `har` file has `desiredQueriesPatches` so this shouldn't be the issue but moving the log out of the `if` just in case.

darkgnotic (commented):
LGTM

PR #4247, opened 2025-04-22
chore(zero-cache): more logs to debug stuck client groups

tantaman:
This existing log line is never prtined:
CleanShot 2025-04-22 at 11 40 52

Either:
a. we have a deadlock
b. the view syncer is shut down

There is 1 more place `if (!this.#stateChanges.active) {` where we bail before calling that log line which would indicate (b) as being the cause.


darkgnotic (commented):
LGTM

PR #4248, opened 2025-04-22
chore(zero-cache): debugging around lock acquisition in the view syncer

tantaman:
- prints if we're stuck waiting on a lock for more than 1 second
- emits the stack track of the stuck acquirer
- prints the name of the task in front of the acquirer that is blocking it

### DIFF
--- a/packages/zero-cache/src/lock.ts
+++ b/packages/zero-cache/src/lock.ts
@@ -3,7 +3,7 @@ import {resolver} from '@rocicorp/resolver';
 
 let id = 0;
 export class Lock {
-  #lockP: Promise<void> | null = null;
+  #lockP: {promise: Promise<void>; name: string} | null = null;
   readonly #lc: LogContext | undefined;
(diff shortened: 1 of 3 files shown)


packages/zero-cache/src/lock.ts
```diff
@@ -0,0 +1,96 @@
...
+      ?.withContext('id', this.#id);
+  }
+
+  async lock(taskName: string): Promise<() => void> {
+    const previous = this.#lockP;
+    const {promise, resolve} = resolver();
+    this.#lockP = {promise, name: taskName};
+
+    let waiting = true;
+    const acquisitionStack = new Error('Lock acquisition stack');
```
tantaman, line 25: nit: we should only acquire `acquisitionStack` if the log level is set to debug.

tantaman:
@grgbkr - I'll let you do what you like with this since you're taking over the bug.

PR #4250, opened 2025-04-22
fix(zero): deal with receiving many responses for the same mutation

tantaman:
Return "mutation already processed" as an error from the server. In mutation tracker, no longer throw if an ephemeral id is missing for a mutation that says it has already been processed.

CleanShot 2025-04-22 at 16 36 00

### DIFF
--- a/packages/zero-client/src/client/mutation-tracker.ts
+++ b/packages/zero-client/src/client/mutation-tracker.ts
@@ -183,18 +183,31 @@ export class MutationTracker {
       mid.clientID === this.#clientID,
       'received mutation for the wrong client',
     );
+
     this.#lc.error?.(
       OnErrorKind.Mutation,
       `Mutation ${mid.id} returned an error`,
       error,
     );
+
     const ephemeralID = this.#ephemeralIDsByMutationID.get(mid.id);
+    if (!ephemeralID && error.error === 'alreadyProcessed') {
+      return;
+    }
+
+    // Each tab sends all mutations for the client group
+    // and the server responds back to the individual client that actually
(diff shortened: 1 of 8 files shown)


packages/zero-client/src/client/mutation-tracker.ts
```diff
@@ -183,18 +183,31 @@ export class MutationTracker {
...
     );
+
     this.#lc.error?.(
       OnErrorKind.Mutation,
       `Mutation ${mid.id} returned an error`,
       error,
     );
+
     const ephemeralID = this.#ephemeralIDsByMutationID.get(mid.id);
+    if (!ephemeralID && error.error === 'alreadyProcessed') {
```
aboodman, line 194: Let me see if I understand what's going on here:

* Each tab sends all pending mutations
  * This means that the same mutation can be sent twice
  * This means that api server can find that a mutation has already been processed when multiple tabs are open
* The server sends information about completion back to the client that originally ran the mutation (by looking at the mutation's clientID presumably?)
* Because the server can end up running the same mutation twice it can end up sending two responses to the originating client:
  * It can return a success/error message first
  * Then a "mutation already processed" response
* This change treats the "mutation already processed" result as a first-class error in the zero-cache<->api-server and client<->zero-cache protocols.
* If `MutationTracker` finds that it gets a push response, but does't have the corresponding promise pending, it used to assert. 
* With this change it only asserts for cases other than `MutationAlreadyProcessed`.

Assuming this is all correct, I like this change. It makes the three cases mutation processing can end in explicit.

tantaman: Yep, that's right.

PR #4251, opened 2025-04-22
feat(zero-pg)!: support custom database implementations in PushProcessor

erikmunson:
### Update 4/24/25
The final public API ended up slightly different from the original proposal in this PR description, rationale captured in PR comments. The new API looks like this:
```typescript
const processor = new PushProcessor(
  new ZQLDatabaseProvider(
    new ZQLPostgresJSAdapter(postgres(process.env.ZERO_UPSTREAM_DB as string)),
    schema,
  ),
);
```
-----------------

Discord thread for full context behind this design direction: https://discord.com/channels/830183651022471199/1362583784377024673

This pulls apart the ZQL/PG and PushProcessor layers of the mutation stack, which previously were mingled together:
- `PushProcessor` is a canonical implementation of the push protocol/algo, with no knowledge of database implementations. Going forward it could easily be used to implement Push for a non-pg upstream like Mongo.
- `DatabaseProvider` specifies how to give PushProcessor the implementation details it needs to do it's work, including passing an arbitrary "transaction" object into mutator functions and calling "hooks" for the db-specific parts of Push. Currently the only hook is updating the LMID.
- `ZQLPGDatabaseProvider` is an implementation of ZQL-on-the-server mutations, using Postgres.js as the db driver. This is 1:1 the same as what existed before when used with the new PushProcessor:

```typescript
const processor = new PushProcessor(
  new ZQLPGDatabaseProvider(
    postgres(process.env.ZERO_UPSTREAM_DB as string),
    schema,
  ),
);
```

Users in the community who want to use ZQL in their server mutators with a different DB driver (e.g. `ZQLPrismaDatabaseProvider`) can create their own DB provider, instantiating ZQL components using newly exported helper functions like `makeServerTransaction`. ZQLPGDatabaseProvider serves as a reference implementation for this pattern.

Alternatively, users can define their own fully custom database implementation that does not use ZQL at all. For example in my app I'll be creating a `DrizzleDatabaseProvider` that does not instantiate a ZQL server transaction, and passes the drizzle client directly into my server mutators instead of using the `ServerTransaction` interface.

### Future work
- I left all of the push processor stuff in `@rocicorp/zero/pg`, but given the new layering it probably should live in a non-pg-specific package. the natural name for this would be `@rocicorp/zero/server`, which could also hold future custom query helpers when those exist, but it looks like that package name is taken by the zero-cache server entrypoint right now
- The exported helpers for creating ZQL transactions are a bit low level, and the pluggable database stuff on ServerTransaction was left as-is to avoid making this refactor more invasive. I think this is totally fine for now especially since there's a reference impl showing how to build a ZQL db provider, but I can imagine a future refactor where the exports from `/zero/pg` are more carefully crafted to create a nicer experience for folks writing custom ZQL providers, and perhaps for folks who want to instantiate a standalone ZQL transaction on the server outside the context of a PushProcessor


packages/zero-pg/src/push-processor.ts
```diff
@@ -277,52 +244,43 @@ export class PushProcessor<
...
-    `Incremented LMID. Received: ${receivedMutationID}. New: ${lastMutationID}`,
-  );
 }
 
 class OutOfOrderMutation extends Error {
   constructor(
     clientID: string,
     receivedMutationID: number,
-    lastMutationID: bigint,
+    lastMutationID: number | bigint,
```
tantaman, line 283: why'd you end up needing to widen the type here?

packages/zero-pg/src/zql-pg-provider.ts
```diff
@@ -0,0 +1,110 @@
...
+      const zeroTx = await makeServerTransaction(
+        dbTx,
+        transactionInput.clientID,
+        transactionInput.mutationID,
+        this.#schema,
+        this.#mutate,
+        this.#query,
+      );
+
+      return callback(zeroTx, {
```
tantaman, line 89: It is still not possible to handle the case where the DBProvider does not support running ZQL server side, correct?

packages/zero-pg/src/zql-pg-provider.ts
```diff
@@ -0,0 +1,110 @@
...
+        async updateClientMutationID() {
```
tantaman, line 90: the hooks felt weird to me but I don't see any other way to enable custom DBs.

packages/zero-pg/src/push-processor.ts
```diff
@@ -218,53 +189,49 @@ export class PushProcessor<
...
-      await checkAndIncrementLastMutationID(
-        this.#lc,
-        dbTx,
-        params.schema,
-        req.clientGroupID,
-        m.clientID,
-        m.id,
-      );
+    return this.#dbProvider.transaction(
+      async (dbTx, transactionHooks): Promise<MutationResponse> => {
```
tantaman, line 193: nit: this is the `zeroTx`

```suggestion
      async (zeroTx, transactionHooks): Promise<MutationResponse> => {
```

packages/zero-pg/src/zql-pg-provider.ts
tantaman: @cbnsndwch -- this is the next area we would need to refactor so the server side zero transaction does not require `mutate` or `query`.

In situations where `mutate` and `query` are missing the server side mutator would need to read and write to the DB directly.

@erikmunson, cbnsndwch is working on Mongo support.

tantaman:
Thanks for doing this. It is a bit of extra work to create a `DatabaseProvider` now. 

I wonder if we can simplify that a bit more for those who are still using postgres but bindings other than postgres.js.

packages/zero-pg/src/push-processor.ts
erikmunson: not strictly necessary, the thinking here was that storing the mutation id as a bigint vs. a number is actually (maybe?) an implementation detail of the upstream database, not a core requirement of the push interface.

was thinking about the mongo example where storing a bigint would requiring stringifying / parsing on the integrator side and it might be desirable to store mutation ids as JSON numbers instead. possibly over-designed and could back it out with trivial effort

(fyi @cbnsndwch in case you have any opinions)

packages/zero-pg/src/zql-pg-provider.ts
erikmunson: yea, @aboodman and i spent some time noodling on this in the original discord thread, haven't been able to think of anything else either.

conceptually though, i think it's worth re-stating that DatabaseProvider is really the lowest level interface/layer to get DB integrations into PushProcessor, and my expectation/hope is that in practice anybody using PushProcessor with a commonly used DB will grab an already implemented DatabaseProvider off the shelf. in other words this is more of an extension/integration API, not an end-user API.

in that context having a clear set of hooks that we expect integrators to implement feels totally fine and normal, and in fact is consistent with the spirit of how the custom change source protocol messages work for replicating upstreams.

erikmunson: fwiw the other interface design here that we considered — and i actually prototyped in the first commit on this PR — involved putting these hooks onto the transaction object itself instead of having them defined separately ([discord message w/ example](https://discord.com/channels/830183651022471199/1362583784377024673/1363625832660336691)).

while in theory it was nice to only have one "thing" (the transaction object), in practice it felt awkward because it created a sort of subtyped version of whatever transaction object the integrator was trying to build, and it also meant that end users working with the tx object in their mutators would've seen all the hook functions as autocompleted methods on the object. didn't feel as clean as this to me.

packages/zero-pg/src/zql-pg-provider.ts
erikmunson: if i'm understanding your question, i think the answer is yes: it is possible to use this without ZQL server side. (so not correct? 🙃 😆 )

the "transaction" object returned in this callback is typed such that it is _completely_ generic — it can be literally anything. Conceptually it is a "transaction" for mutators to use but what that means is up to the integrator to define.

PushProcessor's type signature expects that this callback is called with something as the "transaction" object, and that the hooks are all implemented, but as long as both those things are true it does not get involved with what's going on in the tx itself. it just passes through the tx to all the correct places and flows the type of the tx object from the DB provider all the way to the mutator functions so they get the appropriate type.

In this reference implementation of ZQL + PG, the transaction happens to be a ZQL server transaction, but in the mongo example @cbnsndwch could initially create a `MongoDatabaseProvider` whose transaction object was just the actual mongodb tx object itself. this would of course not allow for ZQL inside the mutators, if a mongo implementation of ZQL became available it could be used in an implementation like this (e.g. `ZQLMongoDatabaseProvider`) to wire in a ZQL transaction backed by a mongo client.

the original use case that prompted this PR is that i use Drizzle on my backend, and have a large footprint of shared backend code (e.g. lots of aysnc background jobs for data processing) which i want to share Drizzle code with. for me, sharing ZQL code across client and server is a non-goal, and I'd prefer to opt out of using the ZQL implementation on the server, passing in my drizzle tx object instead. What I will do in my codebase is implement a `DrizzleDatabaseProvider`, where the transaction object is just my regular Drizzle client tx object, and the hook implementation for updating LMID is done via drizzle's query builder instead of in raw SQL.

packages/zero-pg/src/push-processor.ts
erikmunson: see my answer to your question in the other thread: https://github.com/rocicorp/mono/pull/4251/files#r2056417403

this is actually not a zero tx, it is a completely abstract "transaction" that can be anything defined by the database provider. PushProcessor's job is to flow that integrator-defined transaction through to the mutator functions with the correct types, as part of implementing the push protocol/algo

erikmunson:
> I wonder if we can simplify that a bit more for those who are still using postgres but bindings other than postgres.js.

def agree there's a ton of opportunity to make end users' lives easier for different permutations of db clients. were you thinking of this in relation to use of the ZQL implementation, or using PG decoupled from ZQL?

my thinking here is that `DatabaseProvider` is now the very lowest level of integration, which all other things will be built atop. there are a ton of permutations of different ways people will likely want to use this, e.g. just off the top of my head:
- ZQL with `postgres.js` as the driver
- ZQL with `pg` as the driver
- ZQL with Prisma as the driver
- ZQL with Drizzle as the driver
- All of the above db drivers _without_ ZQL, directly using the db client (this is my use case personally)
- ZQL with a non-PG upstream like mongo
- Non-PG upstream without ZQL
- etc. etc.

so it's hard to pick upfront, without motivating customer examples, exactly how to slice these things into composable blocks that can be clicked together. or at least hard for me at the moment.

`DatabaseProvider` unlocks the ability to incrementally solve the problem of layering and composability for all those combinations, i bet there's something quite nice we could do some sort of ZQL PG provider that is able to accept PG clients of any type. We're actually really close to that in the current `ZQLPGDatabaseProvider` implementation — the abstraction around a `Transaction` object is still a thing so in theory we could abstract that class a little more than I initially did and make it so `ZQLPGDatabaseProvider` accepts any `Transaction` class implementation, and there is a separate `PostgresJSTransaction` that gets passed into it, perhaps wrapping that composition together into a `ZQLPostgresJSDatabaseProvider` or something. LMK if you want me to explore/prototype that.

i was hesitant to put too much thought into the ZQL / PG layers on top of the DatabaseProvider layer because I am not personally using that part, so was concerned my design ideas would be detached from reality. happy to give something like what i described a shot though.

erikmunson:
@tantaman ok take a peek at the quick changes i did in [c25cf30](https://github.com/rocicorp/mono/pull/4251/commits/c25cf30d770aa5ccc11872cc09052bb1bc368804) and [d8d4042](https://github.com/rocicorp/mono/pull/4251/commits/d8d4042c2391139435ea8c2f53f8d81d563285e4). roughly:
- Bring back the abstract `DBConnection` interface for the ZQL layer that @aboodman originally had before this PR
- Re-layer the ZQL part of the layering so that there two composable parts:
  - a `ZQLDatabaseProvider` which implements the update LMID hook and builds the ZQL stuff, and accepts an abstract DB connection
  - a `ZQLPostgresJSAdapter` which _exclusively_ implements the abstract DBConnection and DBTransaction, against the PostgresJS client
- Have consumers compose these like so:
```typescript
const processor = new PushProcessor(
  new ZQLDatabaseProvider(
    new ZQLPostgresJSAdapter(postgres(process.env.ZERO_UPSTREAM_DB as string)),
    schema,
  ),
);
```

in the first commit, I also tried an approach where the composition of database provider + ZQL client adapter was wrapped up in one wrapper function, which looked like this:
```typescript
const processor = new PushProcessor(
  makeZQLPostgresJSDatabaseProvider(
    postgres(process.env.ZERO_UPSTREAM_DB as string),
    schema,
  ),
```

I didn't personally like this quite as much — it felt like too much bundling of composable stuff for not enough actual DX benefit, and to me composing the provider + client adapter feels pretty clean on the end user side. but i don't have a strong opinion and if you like the general pattern here i'm happy to revert back to the wrapper function api.

i think `zql-provider.ts` and `zql-postgresjs-provider.ts` now both feel pretty clean, and all that someone would need to write in order to swap out PostgresJS for another DB client library would be the very small amount of code in `zql-postgresjs-provider.ts`. Or if they wanted to go even more custom they could copy-paste `ZQLDatabaseProvider` and do their own thing with the `makeServerTransaction` helper.

apps/zbugs/server/push-handler.ts
```diff
@@ -1,14 +1,19 @@
...
 import {createServerMutators, type PostCommitTask} from './server-mutators.ts';
 import type {AuthData} from '../shared/auth.ts';
-import {connectionProvider} from '@rocicorp/zero/pg';
 import type {ReadonlyJSONValue} from '@rocicorp/zero';
 
 const processor = new PushProcessor(
-  schema,
-  connectionProvider(postgres(process.env.ZERO_UPSTREAM_DB as string)),
+  new ZQLDatabaseProvider(
+    new ZQLPostgresJSAdapter(postgres(process.env.ZERO_UPSTREAM_DB as string)),
```
tantaman, line 14: This looks good. We have a few users that use `pg.js` instead of `postgres.js` so this keeps things simple for them.

tantaman:
working through some weird bug where zbugs won't start with this change o_O

```shell
VITE_PUBLIC_SERVER="http://127.0.0.1:4848" npm run dev

> zbugs@0.0.0 dev
> vite

failed to load config from /Users/mlaw/workspace/mono/apps/zbugs/vite.config.ts
error when starting dev server:
file:///Users/mlaw/workspace/mono/apps/zbugs/node_modules/.vite-temp/vite.config.ts.timestamp-1745593032659-de02ec48b7bed.mjs:64
  ZQLDatabaseProvider,
  ^^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module 'file:///Users/mlaw/workspace/mono/packages/zero/out/zero/src/pg.js' does not provide an export named 'ZQLDatabaseProvider'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:134:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:217:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
    at async loadConfigFromBundledFile (file:///Users/mlaw/workspace/mono/node_modules/vite/dist/node/chunks/dep-glQox-ep.js:54618:15)
    at async bundleAndLoadConfigFile (file:///Users/mlaw/workspace/mono/node_modules/vite/dist/node/chunks/dep-glQox-ep.js:54461:22)
    at async loadConfigFromFile (file:///Users/mlaw/workspace/mono/node_modules/vite/dist/node/chunks/dep-glQox-ep.js:54423:44)
    at async resolveConfig (file:///Users/mlaw/workspace/mono/node_modules/vite/dist/node/chunks/dep-glQox-ep.js:53931:24)
    at async _createServer (file:///Users/mlaw/workspace/mono/node_modules/vite/dist/node/chunks/dep-glQox-ep.js:43843:18)
    at async CAC.<anonymous> (file:///Users/mlaw/workspace/mono/node_modules/vite/dist/node/cli.js:750:20)
npm error Lifecycle script `dev` failed with error:
npm error Error: command failed
```

erikmunson:
whoa weird — i just tried to repro but mine works. I wonder if your machine has a stale version of the zero-pg build output? i have been running `npm run check-types` and `npm run test` when testing code, so i wonder if my machine already has some output that yours is missing.

tantaman:
Yeah, I forget that we have to build zero-pg before running zbugs in dev. Everything else works without having to build first 🤷‍♂️

erikmunson:
👍 👍 appreciate your help dude!!

packages/zero-pg/src/push-processor.ts
cbnsndwch: for easier compatibility I'd say that might as well be a string in the first place

mongo does not have a native BigInt data type. Libraries like mongoose do some behind the scenes conversions but `PushProcessor` implementations that would use the driver directly would need to handle the conversion

PR #4252, opened 2025-04-22
fix(zero-cache): fix incorrect shutdown under initial CPU load

erikmunson:
Discord thread w/ original context: https://discord.com/channels/830183651022471199/1361448930058965023

Recently I ran into a pathological shutdown loop in prod, where my replication manager was doing an expensive operation against sqlite and then immediately shutting itself down, even though nothing seemed to actually be wrong. Because this expensive sqlite operation was being done as part of catching up to upstream on every node startup, it was infinitely looping through failed restarts. Logs revealed that shutdown was being triggered by HeartbeatMonitor, which believed it had not received a keepalive request from the load balancer recently. This seemed to be false — the LB was sending keepalives at the normal rate.

The theory at the time was that something was funky with the ordering of the timers and request handlers in the HeartbeatMonitor class, such that after the expensive operations completed the state got out of sync with reality and triggered an erroneous shutdown. To test this hypothesis we increased the replication manager's CPU cores from 2 to 8, in hopes that the process running the HeartbeatMonitor would have enough CPU cycles available to avoid getting into this corrupted state while other expensive work was happening. This did in fact solve the infinite shutdown loop and appears to have been directionally correct.

Looking more closely at how/why this could happen, it seems very likely that what we saw was due to the fact that I/O events are processed *after* timers in the Node.js event loop: https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick

If true, the explanation for the behavior we saw during this incident would be that after CPU became available again, first the timer to check the stop interval fired (the "timers" phase), checked a very stale lastHeartbeat timestamp, and believed it was appropriate to shut down. This pre-empted the I/O event callbacks from the keepalive network request, which would have been processed in the later "poll" phase of the event loop and caused the lastHeartbeat timestamp to be bumped up to a much more recent time that would not have triggered a shutdown.

This change attempts to protect against that pathological case by wrapping interval checks with a `setImmediate`. When called in early phases of the event loop, `setImmediate` allegedly executes in the later `check` phase of the same turn, after all outstanding I/O events have been processed. This should in theory mean the ordering aligns with how we'd like this logic to behave.

cc @darkgnotic

### DIFF
--- a/packages/zero-cache/src/services/life-cycle.ts
+++ b/packages/zero-cache/src/services/life-cycle.ts
@@ -1,15 +1,15 @@
-import { LogContext } from '@rocicorp/logger'
-import { resolver } from '@rocicorp/resolver'
-import type { IncomingHttpHeaders } from 'node:http'
-import { pid } from 'node:process'
-import type { EventEmitter } from 'stream'
+import {LogContext} from '@rocicorp/logger';
+import {resolver} from '@rocicorp/resolver';
+import type {IncomingHttpHeaders} from 'node:http';
+import {pid} from 'node:process';
+import type {EventEmitter} from 'stream';
 import {
   singleProcessMode,
   type Subprocess,
   type Worker,
-} from '../types/processes.ts'
-import { RunningState } from './running-state.ts'
-import type { SingletonService } from './service.ts'
+} from '../types/processes.ts';
+import {RunningState} from './running-state.ts';
+import type {SingletonService} from './service.ts';
 
 /**
  * * `user-facing` workers serve external requests and are the first to
@@ -333,10 +333,10 @@ export class HeartbeatMonitor {
       const timeSinceLastHeartbeat = Date.now() - this.#lastHeartbeat;
       if (timeSinceLastHeartbeat >= this.#stopInterval) {
         this.#lc.info?.(
-        `last heartbeat received ${
-          timeSinceLastHeartbeat / 1000
-        } seconds ago. draining.`,
-      );
+          `last heartbeat received ${
+            timeSinceLastHeartbeat / 1000
+          } seconds ago. draining.`,
+        );
         process.kill(process.pid, GRACEFUL_SHUTDOWN[0]);
       }
     });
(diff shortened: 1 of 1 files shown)


packages/zero-cache/src/services/life-cycle.ts
```diff
@@ -300,26 +301,51 @@ export class HeartbeatMonitor {
...
+    // - When CPU becomes available and the event loop wakes up, this stop interval
+    //   check is run *before* the keepalive request is processed. The value of
+    //   this.#lastHeartbeat is now very stale, and erroneously triggers a shutdown
+    //   even though keepalive requests were about to be processed and update
+    //   this.#lastHeartbeat. Downtime ensues.
+    //
+    // To avoid this, we push the check out to a phase of the event loop *after*
+    // I/O events are processed, using setImmediate(). This ensures we see
+    // a value for this.#lastHeartbeat that reflects any keepalive requests
+    // that came in during the current event loop turn.
```
darkgnotic, line 330: This is a great explanation. Can you add a link to the documentation here for reference?

https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick#setimmediate-vs-settimeout

darkgnotic (approved):
Thank you for researching the innards of I/O and event timers in NodeJS, and for coming up with a clever solution!

When you merge this, can you prefix the commit description with `fix(zero-cache): ...` to facilitate writing up release notes?

packages/zero-cache/src/services/life-cycle.ts
erikmunson: [done](https://github.com/rocicorp/mono/pull/4252/commits/5b9dff352a1923dffffbba8c9008b9aaba10e2f5) ✅

erikmunson:
> When you merge this, can you prefix the commit description with fix(zero-cache): ... to facilitate writing up release notes?

I don't have merge permissions so i think you will have to merge and add the appropriate commit name/prefix — just updated the branch with the latest upstream commits and CI passed, so i think you can merge when ready. thx for the review 🙏

erikmunson:
also re: the naming of the changelog — while this fix was prompted by replication-manager crashes, it does technically fix this same erroneous shutdown case on view-syncers as well. you and i just believe it's less likely to occur on view syncers in practice because of the way processes are laid out relative to CPU cores on view syncers (i.e. lower chance of locking up all cores). If somebody is running view syncers with very few cores, especially in single node configuration, this could definitely happen with view syncer keepalives.

PR #4254, opened 2025-04-23
chore(zero-cache): configure metrics collection

tantaman:
Configures a metrics exporter so we can start collecting top level metrics in addition to traces.


darkgnotic (commented):
LGTM

PR #4256, opened 2025-04-23
fix(zero-cache): fix blackholing when pushing on a closed websocket

darkgnotic:


packages/zero-cache/src/workers/connection.ts
```diff
@@ -350,6 +350,9 @@ function send(
     lc.debug?.(`Dropping outbound message on ws (state: ${ws.readyState})`, {
       dropped: data,
     });
+    if (callback !== 'ignore-backpressure') {
+      callback(new Error('websocket closed'));
```
darkgnotic, line 354: Without this, the pipeline can get stuck on line 293 above.

aboodman: Can we find a way to test this? And more importantly should we try and detect if work is stalled in this way for long periods and throw an error/exit?

PR #4257, opened 2025-04-23
fix(zero-cache): keep microsecond precision across timezones

tantaman:
@darkgnotic - this fixes the issue where `PreciseDate` does not return microsecond precision for timezones that are not utc.

Weirdly, dbs that are not UTC are returning `BigInt` rather than `number`

(this is the output for pg-17 tests which are UTC+1)

CleanShot 2025-04-23 at 13 42 32

have not tracked that down yet as the value returned by the converter is a number.


darkgnotic:
Weird. Do you know what string came from PG when it was not UTC? (e.g. what `timestamp` was passed in to `timestampToFpMillis(timestamp: string)`)

If that timestamp did not have the microsecond precision, then it's most likely a quirk of pg. But it would be nice to know that (and rule out some bug in our string parsing / conversion logic).

tantaman:
> If that timestamp did not have the microsecond precision, then it's most likely a quirk of pg. But it would be nice to know that (and rule out some bug in our string parsing / conversion logic).

@darkgnotic - it is a problem with PreciseDate.

CleanShot 2025-04-24 at 09 52 15
CleanShot 2025-04-24 at 09 52 26

Looks like we're following the ISO string format for timezones that are not UTC. Fixing.

tantaman:
CleanShot 2025-05-05 at 11 10 24

darkgnotic (approved):
LGTM (once tests are adjusted of course)

PR #4258, opened 2025-04-23
chore(zero-cache): top level metrics to track for zero-cache

tantaman:
Top level metrics to bump.

- The counters can be visualized as requests per day/hour/minute/second/whatever by the visualization tool (e.g. Grafana)
- Histograms can be partitioned by attributes. E.g., timing per websocket message type
- up/down counters provide us with point-in-time values for how much of a thing we have. Useful for uncovering leaks and/or correlating against periods of slowness. Slow server while there are 1,000 active clients? Ok. Slow while there is 1 active client? 😬

There is `https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node` which we can add to presumably get top level os/fs/http stats.


darkgnotic (commented):
❤️

PR #4260, opened 2025-04-23
feat(zero-cache): faster litestream restores

darkgnotic:
Switch to the [rocicorp fork of litestream](https://github.com/rocicorp/litestream) and enable its [multipart download](https://github.com/rocicorp/litestream/pull/1) feature.

The default settings are those tested by @erikmunson to work the best when running from an AWS container. This was verified to speed up the restore of a 200GB replica from ~35 min to ~5 min.


erikmunson:
yas

CANNOT WAIT to try this out. transformative perf improvement for folks with larger dbs like myself. thanks for seeing this one all the way through.

PR #4263, opened 2025-04-25
feat(zero-cache): continuous changelog cleanup

darkgnotic:
### Feature

Continuous changelog cleanup based on `litestream` watermark monitoring in the `change-streamer`

### Implementation

Continuous changelog cleanup leverages a feature called "watermark tracking" in the rocicorp fork of litestream: https://github.com/rocicorp/litestream/pull/3

A `BackupMonitor` service periodically polls the prometheus metrics endpoint exported by litestream to track the last replicated value of the `"_zero.replicationState"."stateVersion"` cell, configured via the `watermark-table` and `watermark-column` litestream options, respectively.

When a new watermark value has been replicated, a cleanup is scheduled to happen at or after the `replication time` + `restore duration`, the latter added to ensure that a `zero-cache` restoring the replica is able to catch up from the change log even if the restore takes a long time.

The change log purging logic is the same as that which is performed when a subscriber reports the watermark that it's replica was restored at.


packages/zero-cache/src/server/change-streamer.ts
```diff
@@ -46,10 +49,12 @@ export default async function runWorker(
   const shard = getShardConfig(config);
 
   let changeStreamer: ChangeStreamerService | undefined;
+  let initialSyncTime: number | undefined;
 
   for (const first of [true, false]) {
     try {
       // Note: This performs initial sync of the replica if necessary.
+      const start = Date.now();
```
tantaman, line 57: any reason to prefer `Date.now()` over `performance.now()`?

packages/zero-cache/src/server/main.ts
```diff
@@ -96,7 +96,18 @@ export default async function runWorker(
...
+        config,
+        runChangeStreamer ? 1 : 10,
+        3000,
+      );
+      if (!config.litestream.restoreDurationMsEstimate && restoreElapsedMs) {
+        internalFlags.push(
+          '--litestream-restore-duration-ms-estimate',
+          String(restoreElapsedMs),
+        );
+      }
```
tantaman, line 110: Is it a problem if this value gets out of date? E.g., for a zero-cache that lives a very long time such that the replica size grows by a non-negligible amount.

packages/zero-cache/src/services/change-streamer/backup-monitor.ts
```diff
@@ -0,0 +1,132 @@
...
+// purging changelog records as soon as those watermarks are replicated would
+// result in the zero-cache not being able to catch up from minute 0 once it
+// has finished restoring the replica.
+//
+// Consequently, cleanup is delayed by an interval equivalent to the time it
+// takes for the litestream restore, which is calculated by the runner (when
+// it performed the restore or the initial sync). In addition to this delay,
+// the ChangeStreamerService itself adds at least 30 seconds of padding before
+// purging records, ensuring all active subscribers are past the watermark
+// before purging preceding records.
```
tantaman, line 35: How hard would it be to track if there is an active restore happening and prevent purging until the restore completes? Maybe via a record in the upstream DB?

Is the timing of operations really consistent enough to not have to worry about this?

packages/zero-cache/src/server/change-streamer.ts
darkgnotic: Not really ... this is mostly out of habit. Is the latter better than the former (other than having higher precision)? Please educate me!   😄

packages/zero-cache/src/server/main.ts
darkgnotic: This is a good point. We should think about this, and maybe factor in your next thought / question.

packages/zero-cache/src/services/change-streamer/backup-monitor.ts
darkgnotic: I like this train of thought. Some ideas:
- Some row in the Change DB, which the change-streamer can check and/or lock on before purging records
- Would have to be careful to handle / distinguish orphaned rows
- Or ... maybe connect to the change-streamer via some different endpoint (like a pre-subscribe connection) to indicate that a subscription is coming soon and change-logs should not be purged until then.

I think the latter might be the cleanest. I will explore this more. I agree that the timing stuff is a bit "fuzzy" and would be nice to solidify. Thank you for the suggestion!

PR #4265, opened 2025-04-25
feat(zero-cache): Add statz endpoint.

aboodman:
Also remove script as I don't think we need both.


packages/zero-cache/src/services/statz.ts
```diff
@@ -0,0 +1,359 @@
...
+    }
+    seenQueries.add(queryHash);
+
+    try {
+      const zql = clientAST.table + astToZQL(clientAST);
+      out.write('HASH:' + queryHash);
+      out.write('ZQL:' + zql + '\n');
+    } catch (e) {
+      parseFailures.push(queryHash);
+    }
```
tantaman, line 228: Not sure we should dump these. It could be quite a lot of data.

aboodman: Yeah :-/. Let's see how it goes. Seems useful.

aboodman:
@darkgnotic can you review the config changes?

PR #4266, opened 2025-04-25
feat(zql): empty `related('x', q => q.one())` calls are now `null`

tantaman:


grgbkr:
Does this enable us to get rid of or tighten up toEqualPg?

tantaman:
> Does this enable us to get rid of or tighten up toEqualPg?

Yes, we can tighten it. It is still used to strip out the `ref count` symbols.

PR #4267, opened 2025-04-25
chore(zql): `run` now needs an option for "complete" vs "unknown"

tantaman:
The default of `unknown` causes `run` to return no data in tests.

related: #4234

### DIFF
--- a/packages/zql-integration-tests/src/helpers/runner.ts
+++ b/packages/zql-integration-tests/src/helpers/runner.ts
@@ -252,7 +252,9 @@ function makeTest<TSchema extends Schema>(
       zqlSchema,
       ast(queries.sqlite).table,
     );
-    const memoryResult = await queries.memory;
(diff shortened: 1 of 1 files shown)


tantaman:
> I am not sure how this changed? The default behavior of unknown should be the same as what it was doing before.

@aboodman - Hmm, you're right. The sources should have their data so `unknown` shouldn't return empty results. Looking into this some more.

tantaman:
https://github.com/rocicorp/mono/pull/4234 also causes some of the compiler tests to time out 🤔

tantaman:
Yeah, the tests were calling `QueryDelegate` incorrectly. Fix is here: https://github.com/rocicorp/mono/pull/4269

tantaman:
Maybe the tests didn't get run in CI? They were using the old vitest setup before everything was updated to independent vitest configs per postgres db version.

PR #4278, opened 2025-04-28
chore(zbugs): Introduce a prisma schema

arv:
This is working now.

Steps that were done.

- Let Prisma use inspection to generate schema.prisma.
- Modified schema.prisma to match what we had before
- let shared/schema.ts import from the generated file
  - most notably how we had to manually define the Issue labels relation
- Migrations:
  - The db had a lot of fields as optional but our zero schema had them as required
  - Change User role to an enum in the db too.


apps/zbugs/prisma/schema.prisma
```diff
@@ -0,0 +1,128 @@
...
+  testJson    Json?   @default("{}") @ignore
+
+  assignee    User?        @relation("issue_assigneeIDToUser", fields: [assigneeID], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  comments    Comment[]
+  creator     User         @relation("issue_creatorIDToUser", fields: [creatorID], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  emoji       Emoji[]
+  viewState   ViewState[]
+  issueLabels IssueLabel[] @ignore
+
+  labels Label[] @relation("IssueLabel")
```
arv, line 67: This seems to work...

apps/zbugs/prisma/schema.prisma
```diff
@@ -0,0 +1,128 @@
...
+  @@id([labelID, issueID])
+  @@index([issueID], map: "issueLabel_issueid_idx")
+  @@map("issueLabel")
+}
+
+model Label {
+  id   String @id @db.VarChar
+  name String @db.VarChar
+
+  issues Issue[] @relation("IssueLabel") @ignore
```
arv, line 89: ...paired with this

apps/zbugs/prisma/generated/zero/schema.ts
```diff
@@ -0,0 +1,220 @@
...
+    viewState: many({
+      sourceField: ['id'],
+      destField: [],
+      destSchema: viewStateTable,
+    }),
+    labels: many({
+      sourceField: ['id'],
+      destField: [],
+      destSchema: labelTable,
+    }),
```
arv, line 166: but this is not the expected generated code.

arv (commented):
@elledienne I'll look at the generator to see how it models this.

aboodman:
Did you compare to Drizzle? It’s more popular among our user base.

a (phone)

arv:
Drizzle is all runtime which I find a bit off putting. Prisma is codegen so there is zero overhead.

However, at this point Prisma cannot use predefined join/junction tables. Prisma can work with them but they have to have a very specific name and shape `_IssueToLabel` with fields A` and `B`.

arv:
We should also create a GitHub Action to deploy the migrations 

https://www.prisma.io/docs/orm/prisma-client/deployment/deploy-database-changes-with-prisma-migrate#deploying-database-changes-using-github-actions

arv:
I would like to do the same exercise with Drizzle

PR #4280, opened 2025-04-28
feature(infra): otel sidecar

cesara:


prod/sst/otel.ts
```diff
@@ -0,0 +1,184 @@
+/* eslint-disable */
+/// <reference path="./.sst/platform/config.d.ts" />
+interface ContainerDefinition {
+  name: string;
+  image?: string;
```
tantaman, line 5: What happens if `image` is not specified?

prod/sst/otel.ts
```diff
@@ -0,0 +1,184 @@
...
+          hostname: zero-sandbox
+          api:
+            key: ${config.apiKey}
+            site: datadoghq.com
+      service:
+        pipelines:
+          traces:
+            receivers: [otlp]
+            processors: [resourcedetection/env, batch/traces]
+            exporters: [datadog/connector, datadog/api]
```
tantaman, line 141: I think this is reasonable for a first cut. My guess is we'll quickly get users that want to use a different exporter but may as well wait until that happens before getting more complicated.

prod/sst/sst.config.ts
```diff
@@ -156,106 +155,146 @@ export default $config({
...
+            maxCapacity: 10,
+          },
+        }),
+        // Set this to `true` to make SST wait for the view-syncer to be deployed
+        // before proceeding (to permissions deployment, etc.). This makes the deployment
+        // take a lot longer and is only necessary if there is an AST format change.
+        wait: false,
+      },
+      {
+        apiKey: process.env.DATADOG_API_KEY,
```
tantaman, line 293: can we add service w/o otel if the datadog API key is missing? Just want to make sure users can deploy Zero without having to have datadog set up.

prod/sst/sst.config.ts
```diff
@@ -155,8 +154,8 @@ export default $config({
           },
         };
 
-    // Replication Manager Service
-    const replicationManager = cluster.addService(`replication-manager`, {
+    const replicationManager = addServiceWithOtel(`replication-manager`, {
```
darkgnotic, line 157: Noiiiice!!! This is a beautiful API.

It would be nice try to figure out how to export the `otel` as a package, but if it's impossible users can just copy-and-paste the `otel.ts` file. Maybe figure this out as another step.

prod/sst/otel.ts
```diff
@@ -0,0 +1,236 @@
...
+  } else {
+    if (!image) {
+      throw new Error('Either `containers` or top-level `image` is required');
+    }
+
+    containers = [{name: serviceName, image, health, environment, logging}];
+  }
+
+  const otelConfig = {
+    apiKey: process.env.DATADOG_API_KEY!,
```
darkgnotic, line 222: So I guess the idea is what we give users an API key to use and they set it in their env.

On the Datadog side did we ever figure out a way to tag incoming data based on the key? If not, is there a way to inject something like a `customer-id` (or even the api key itself so that there isn't another variable for users to deal with) into the otel data via a processor or exporter step in the pipeline?

darkgnotic (approved):

prod/sst/otel.ts
cesara: ya we can do this here 
```
          trace_statements:
            - context: resource
              statements:
                - set(attributes["datadog.host.name"], "${config.appName}")
                - set(attributes["datadog.host.version"], "${config.appVersion}")
 ```

prod/sst/sst.config.ts
cesara: ya ill move this as a next step

prod/sst/otel.ts
cesara: its sometimes optional because a image can be defined a level above the container definition

PR #4282, opened 2025-04-29
fix(zero-cache): fix incorrect suppression of row-del patches

darkgnotic:
User reports:
*

Fixes the failure to send row-delete patches in certain scenarios (when IVM generates an add and subsequent remove of a row that will ultimately be removed). In these cases, the row-delete patches were tagged with the wrong `patchVersion`, causing them in to be incorrectly classified as "old" patches and filtered before being sent.


grgbkr:
LGTM!

PR #4285, opened 2025-04-29
feat(zero-cache): avoid sending duplicate row patches

darkgnotic:
1. Tracks the row patches that have been sent and avoids resending redundant patches. This may save significant bandwidth during hydrations for queries with lots of joins.
2. Fixes an edge case to accompany https://github.com/rocicorp/mono/pull/4282: when an existing row is deleted, it will result in a new `patchVersion`, but if it is subsequently re-added, the patchVersion in the CVR will be restored. In this case, however, the `toVersion` used for filtering downstream patches should not move backwards, or else it will be undesirably filtered.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/cvr.ts
+++ b/packages/zero-cache/src/services/view-syncer/cvr.ts
@@ -23,6 +23,7 @@ import {type CVRFlushStats, type CVRStore} from './cvr-store.ts';
 import {KeyColumns} from './key-columns.ts';
 import {
   cmpVersions,
+  maxVersion,
   oneAfter,
   versionString,
   type CVRVersion,
@@ -438,6 +439,11 @@ type Hash = string;
 export type Column = string;
 export type RefCounts = Record<Hash, number>;
 
+type RowPatchInfo = {
+  op: 'put' | 'del';
+  toVersion: CVRVersion;
+};
+
 /**
  * A {@link CVRQueryDrivenUpdater} is used for updating a CVR after making queries.
  * The caller should invoke:
@@ -458,6 +464,7 @@ export class CVRQueryDrivenUpdater extends CVRUpdater {
     rowIDString,
   );
   readonly #replacedRows = new CustomKeyMap<RowID, boolean>(rowIDString);
+  readonly #lastPatches = new CustomKeyMap<RowID, RowPatchInfo>(rowIDString);
 
   #existingRows: Promise<Iterable<RowRecord>> | undefined = undefined;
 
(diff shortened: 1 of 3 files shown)


packages/zero-cache/src/services/view-syncer/cvr.ts
```diff
@@ -710,29 +717,42 @@ export class CVRQueryDrivenUpdater extends CVRUpdater {
...
+                id,
+              },
+              toVersion,
+            });
+            this.#lastPatches.set(id, {op: 'del', toVersion});
+          }
+        }
+      } else if (contents) {
+        // dedupe
+        if (lastPatch?.op !== 'put') {
```
grgbkr, line 744: What if version/contents have changed?  Then we should dedupe the push right?

darkgnotic: Good question. The invariant assumed here is that we are always reading from a single point in the replica, so all row-puts must have identical versions / contents. This allows us to not have to store the contents in memory (and the mapping only contains metadata).

With the hashing / checksumming thing were perusing, we could add some sanity checking to verify this invariant.

darkgnotic: Ah, it'll help to keep in mind that in an advancement, only the latest version of a row will be pushed. So, for example, if PG sends us:

- `update a v1`
- `update a  v2`
- `insert b  v1`
- `update b  v2`
- `update a  v3`

the advancement will only process two changes: [`set a v3`, `set b v2`].

grgbkr: I also suspect this logic is not quite right.  https://github.com/rocicorp/mono/blob/ad9f90e2c26ce8822e706135f42ff9b9d2f149ac/packages/zero-cache/src/services/view-syncer/view-syncer.ts#L1177

Taking a step back.  

IVM's output is a tree.  The same row (as identified by row id) can appear multiple places in this tree.  When a change is pushed into an IVM pipeline, multiple changes are output, these changes update the tree one node at a time (i.e. one subtree at a time).  Mid processing of these multiple change output by the IVM pipeline the same row (as identified by row id) can be in different states in different parts of the output tree (i.e. it can exist at old version in this subtree, exist at new version in this subree, or be deleted in this subtree).  Once all output changes are consumed these inconsistencies are resolved, i.e. the row should be at the same version (the new version, or deleted) everywhere in the tree.

darkgnotic: Thank you. This should be addressed in https://github.com/rocicorp/mono/pull/4289.  🙏

PR #4287, opened 2025-04-29
fix(zero-cache): add explicit dependency on `@opentelemetry/resources`

tantaman:
We depend on this but did not list it in package.json

https://discord.com/channels/830183651022471199/1366445174057599006/1366449323428282378


arv:
LGTM

I'm a bit surprised that this happens... we have a script that is supposed to validate the dependencies... but this happened before so the script is not running on PR changes?

PR #4289, opened 2025-04-29
chore(zero-cache): account for multiple row versions with deduping

darkgnotic:
Fix the row patch deduping logic in #4285 to handle the possibility that multiple versions of a row can be produced by the IVM pipeline.

Also add more end-to-end tests for these multiple-row-appearance scenarios.


grgbkr:
LGTM!

PR #4291, opened 2025-04-30
fix(z2s): handle self join, not exists

tantaman:
When server name mappings were added, the compiler got way out of hand. Fixing the `self join` bug would have made everything even more complicated because the fix required introducing aliasing for every table. This PR refactors the compiler so table and column mapping and aliasing only happen in a single place. Qualified column names are also passed everywhere via a `QualifiedColumn` structure so there is never confusion about what "string" a function is dealing with.

---

Two bugs were found when adding more tests to z2s:

1. `self join`
2. `not exists`

# Self Join

Self join failed because we did not auto-alias tables.

Before auto-aliasing we would generate this SQL:

```sql
SELECT (SELECT * FROM employee WHERE employee.reports_to = employee.id) as reports_to FROM employee;
```

which would return no results for the sub-select since the sub-query never references the outer-query.

With auto-aliasing we generate this instead:

```sql
SELECT (SELECT * FROM employee as employee_1 WHERE employee_1.reports_to = employee_0.id) as reports_to FROM employee as employee_0
```

which correctly compares the child to the parent.

# Not Exists

For not exists, `NOT EXISTS` forgot to apply the correlation. We pass the correlation to `NOT EXISTS` now.

### DIFF
--- a/packages/z2s/src/compiler.output.test.ts
+++ b/packages/z2s/src/compiler.output.test.ts
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/naming-convention */
-import {expect, test} from 'vitest';
-import {Compiler} from './compiler.ts';
+import {beforeEach, expect, test} from 'vitest';
 import {formatPgInternalConvert} from './sql.ts';
 import {
   boolean,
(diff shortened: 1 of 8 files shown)


packages/z2s/src/compiler.output.test.ts
```diff
@@ -366,7 +438,7 @@ test('any', () => {
 //   const compiler = new Compiler(schema.tables, serverSchema);
 //   expect(
 //     formatPgInternalConvert(
-//       compiler.valuePosition(
+//       valuePosition(
```
grgbkr, line 441: Should we restore or delete these commented out tests?

grgbkr (commented):
LGTM!

PR #4292, opened 2025-04-30
fix(zql): Do not allow NOT EXISTS in client queries

arv:
`NOT EXISTS` can't work reliably on client because:
- Zero only syncs a subset of data to the client
- Client can't distinguish between non-existent rows vs. non-synced rows
- Correct `NOT EXISTS` requires complete knowledge of data

Added `assertNoNotExists()` function to validate conditions and throw
errors when `NOT EXISTS` is used in client queries. Updated `query-impl.ts`
to check conditions after DNF conversion.

https://bugs.rocicorp.dev/issue/3438

### DIFF
--- a/packages/zero-pg/src/query.ts
+++ b/packages/zero-pg/src/query.ts
@@ -5,7 +5,11 @@ import type {AST} from '../../zero-protocol/src/ast.ts';
 import type {Schema} from '../../zero-schema/src/builder/schema-builder.ts';
 import type {Format} from '../../zql/src/ivm/view.ts';
 import type {DBTransaction, SchemaQuery} from '../../zql/src/mutate/custom.ts';
-import {AbstractQuery, defaultFormat} from '../../zql/src/query/query-impl.ts';
+import {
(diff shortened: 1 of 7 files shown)


packages/zero-pg/src/query.ts
```diff
@@ -82,15 +86,13 @@ export class ZPGQuery<
     ast: AST,
     format: Format,
   ) {
-    super(schema, tableName, ast, format);
+    super(schema, tableName, ast, format, 'permissions');
```
tantaman, line 89: note to self: new system for `api server`?

https://bugs.rocicorp.dev/issue/3791

packages/zql/src/query/query-impl.ts
```diff
@@ -122,6 +123,9 @@ export const SUBQ_PREFIX = 'zsubq_';
 
 export const defaultFormat = {singular: false, relationships: {}} as const;
 
+export const systemSymbol = Symbol();
```
tantaman, line 126: this looks unused

PR #4299, opened 2025-05-01
!feat(zero-cache): automatic replication-manager discovery / routing

darkgnotic:
### Fargate / most multi-node configurations

To update a multi-node configuration without disruption:
1. roll out new `replication-manager`
2. roll out new `view-syncer`, replacing the:
   * `ZERO_CHANGE_STREAMER_URI=http://{host}` option with
   * `ZERO_CHANGE_STREAMER_MODE=discover`
3. when `view-syncer`s are rollback-safe, remove the internal load balancer that was previously used for `view-syncer` to `replication-manager` routing

### Single-node configurations

Single-node configurations are unaffected

### Uncommon multi-node configurations

For container setups in which the process does not have access to the externally visible ip address or port (e.g. using docker a.k.a. "bridge" mode networking), an external routing or proxying mechanism is still needed. In such configurations, add the `ZERO_CHANGE_STREAMER_ADDRESS={host}` option to the `replication-manager`, where `{host}` is the hostname that was formerly part of the `ZER_CHANGE_STREAMER_URI`, e.g.
* `view-syncer`: `CHANGE_STREAMER_MODE=discover`
* `replication-manager`: `CHANGE_STREAMER_ADDRESS=internal-prod-repmgr-125468.us-east-1.elb.amazonaws.com`

## Feature

The discovery and routing of the `replication-manager` is now facilitated by the Postgres Change DB, using the same row-level locking mechanism used to enforce single-writer access to the change log.

This obviates the need for an external addressing or proxying mechanism such as Service Discovery, Service Connect, or an Internal Load Balancer.

Screenshot 2025-04-30 at 17 57 06


darkgnotic:
We had to revert this (1) to keep it out of 0.19 and (2) because of an issue of how it selects IP addresses on MacBooks. I'll resend when we're past 0.19.

arv:
FYI

https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-scope-and--to-draw-attention-to-breaking-change

`!` after `)`

darkgnotic:
> FYI
> 
> https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-scope-and--to-draw-attention-to-breaking-change
> 
> `!` after `)`

ha! I've been formatting it wrong this whole time 😅  ... thank you.

PR #4300, opened 2025-05-01
feat(analyze-query): normalizeConfig to default CVR to upstream if not specified

tantaman:


arv (commented):
LGTM

@darkgnotic FYI

PR #4303, opened 2025-05-01
feat!(zero-client): reject server mutator promise in all error cases

tantaman:
server errors used to resolve and client errors would reject.

now they both reject.


arv (commented):
LGTM

FYI

https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-scope-and--to-draw-attention-to-breaking-change

`!` after `)`

PR #4304, opened 2025-05-01
fix(ast-to-zql): Output junciton relationships (i.e. hidden).

grgbkr:


packages/ast-to-zql/src/ast-to-zql.ts
```diff
@@ -36,7 +36,12 @@ export function astToZQL(ast: AST): string {
   // Handle related subqueries
   if (ast.related && ast.related.length > 0) {
     for (const related of ast.related) {
-      if (!related.hidden) {
+      if (related.hidden) {
+        const nestedRelated = related.subquery.related?.[0];
+        if (nestedRelated) {
```
grgbkr, line 41: The alliterative is to throw an error here 'Malformed AST hidden relationship X missing second relationship', but I'm leaning towards being more tolerant in this debug tool.

arv (approved):
LGTM

PR #4305, opened 2025-05-01
feat(zero-client): server and client promise separation

tantaman:


packages/replicache/src/replicache-impl.ts
```diff
@@ -1449,13 +1449,19 @@ export class ReplicacheImpl<MD extends MutatorDefs = {}> {
   #register<Return extends ReadonlyJSONValue | void, Args extends JSONValue>(
     name: string,
     mutatorImpl: (tx: WriteTransaction, args?: Args) => MaybePromise<Return>,
-  ): (args?: Args) => Promise<Return> {
+  ): (
+    args?: Args,
+  ) => Promise<Return> | {client: Promise<Return>; server: Promise<unknown>} {
```
aboodman, line 1454: I think this should only return:

```
{client: Promise<Return>; server: Promise<Return>}
```

Is there a reason I'm not understanding to keep the `Promise<Return> |`  return option?

tantaman:
Just because in vanilla replicache it would be a Promise<TReturn> but when
used by Zero it is the client, server promises.

aboodman:
> Just because in vanilla replicache it would be a Promise<TReturn> but when used by Zero it is the client, server promises.

I don't understand this. Let's talk on zoom.

packages/replicache/src/replicache-impl.ts
tantaman: I suppose we can do whatever we want with this return type because Replicache forcibly casts it later anyway:

https://github.com/rocicorp/mono/blob/8b97014070fc8333ffdfa548e7caee29674b921d/packages/replicache/src/replicache-impl.ts#L1491-L1493

and `Zero` does a similar thing by casting `rep.mutate` to `any` 

https://github.com/rocicorp/mono/blob/8b97014070fc8333ffdfa548e7caee29674b921d/packages/zero-client/src/client/zero.ts#L621

then assigning to the public interface.

https://github.com/rocicorp/mono/blob/8b97014070fc8333ffdfa548e7caee29674b921d/packages/zero-client/src/client/zero.ts#L646

---

`Promise<Return> | {client: Promise<Return>; server: Promise<unknown>}` would let us tighten these unsafe casts.

tantaman: The unsafe casts are there so the Replicache interface can remain unchanged.

packages/zero-client/src/client/custom.test.ts
```diff
@@ -172,17 +166,17 @@ test('custom mutators write to the local store', async () => {
...
     createdAt: 1743018138477,
-  });
+  }).client;
 
   await z.markQueryAsGot(z.query.issue);
   let issues = await z.query.issue;
   expect(issues[0].title).toEqual('foo');
 
-  await z.mutate.issue.setTitle({id: '1', title: 'bar'});
+  await z.mutate.issue.setTitle({id: '1', title: 'bar'}).client;
```
tantaman, line 175: @aboodman - this is going to break people pretty hard as all their `await z.mutate.issue.setTitle({id: '1', title: 'bar'});` will no longer work correctly but there will be no type errors to tell them.

packages/replicache/src/replicache-impl.ts
```diff
@@ -1470,8 +1476,17 @@ export class ReplicacheImpl<MD extends MutatorDefs = {}> {
       );
 
       if (trackingData) {
-        (result as unknown as Record<string, unknown>).server =
-          trackingData.serverPromise;
+        return {
+          client: result,
+          server: trackingData.serverPromise,
+          then: (onFulfilled, onRejected) => {
```
tantaman, line 1482: catch and finally intentionally left off. The user will get type errors if they try to call those directly. They'll get type errors trying to use `then` too but not with `await` (which calls then).

packages/replicache/src/replicache-impl.ts
```diff
@@ -1470,8 +1476,17 @@ export class ReplicacheImpl<MD extends MutatorDefs = {}> {
...
+            this.#lc.warn?.(
```
aboodman, line 1483: Beautiful!

PR #4307, opened 2025-05-01
chore(zero-client): remove ErrorKind and simplify onError

arv:
The type of OnError is now:

```ts
type OnError = (message: string, ...rest: unknown[]) => void;
```

Having the first argument as a string is intended to provide a short error message that can be readable. The other parameters may be anything but in practice they are usually an Error object or a JSON value


aboodman:
> The other parameters may be anything but in practice they are usually an Error object or a JSON value

But how do you know what to do with the ...rest params? If you actually want a string.

arv:
> > The other parameters may be anything but in practice they are usually an Error object or a JSON value
> 
> But how do you know what to do with the ...rest params? If you actually want a string.

You can use JSON.stringify or toString.

I'm not really sure I understand what you are trying to solve here?

aboodman:
> > > The other parameters may be anything but in practice they are usually an Error object or a JSON value
> > 
> > 
> > But how do you know what to do with the ...rest params? If you actually want a string.
> 
> You can use JSON.stringify or toString.
> 
> I'm not really sure I understand what you are trying to solve here?

If the implementor of `onError` doesn't know what types will come through `...rest` how can they know whether to use `JSON.stringify()` or `toString()` or what.

PR #4312, opened 2025-05-02
chore(zbugs): Allowing issue creation without adding a description

alexhking:


tantaman:
ty this bugged me so much 😅

PR #4313, opened 2025-05-02
chore(bench): remove setup and teardown, invert metric

tantaman:
For some reason `setup` time is contributing to the total benchmark time giving us bad results. Doing setup once results in 10x faster benchmarks as we're now only measuring the thing we intend to measure.

More importantly, the benchmarks now provide meaningful information when changing the value of `N`. Before, the time was dominated by setup time so it was hard to figure out how a query scaled as `N` changed.

Discovered this while investigating a user report: https://bugs.rocicorp.dev/issue/3793

Before:
CleanShot 2025-05-02 at 13 08 46

After:
CleanShot 2025-05-02 at 13 07 19

# Invert Metric
https://github.com/rocicorp/mono/pull/4313/commits/91c4058dd538b765f504a069f9f08042c37abe87

Bencher expects throughput (e.g., QPS). Vitest outputs a timing per run. Divide 1 second by time per run to get QPS.

### DIFF
--- a/packages/shared/src/tool/vitest-perf-json-to-bmf.ts
+++ b/packages/shared/src/tool/vitest-perf-json-to-bmf.ts
@@ -166,10 +166,14 @@ function convertVitestToBMF(vitestOutput: VitestOutput): BMFMetric {
         const metricName = `${group.fullName} > ${benchmark.name}`;
 
         bmf[metricName] = {
+          // benchmark outputs are the time in milliseconds it took to run.
+          // If we want to convert to throughput, we need to divide 1000 by the time.
+          // This is the inverse of the time taken.
+          // So if it took 100ms to run, the throughput is 1000 / 100 = 10
           throughput: {
-            value: benchmark.mean,
-            ['lower_value']: benchmark.min,
-            ['upper_value']: benchmark.max,
+            value: 1000 / benchmark.mean,
+            ['lower_value']: 1000 / benchmark.min,
+            ['upper_value']: 1000 / benchmark.max,
           },
         };
       });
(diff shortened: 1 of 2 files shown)


tantaman:
I don't get the bencher report. This vastly improved the speed rather than slowing it down when run locally.

tantaman:
ok, found `vitest-perf-json-to-bmf` which may be the issue.

tantaman:
this was the issue -- https://github.com/rocicorp/mono/pull/4313/commits/91c4058dd538b765f504a069f9f08042c37abe87


the metric was inverted.

tantaman:
perfect. This is much closer to what is happening locally.

CleanShot 2025-05-02 at 16 01 53

arv:
The vitest bench is broken in a few ways. It has been on my list to fix vitest. There is currently no way to have teardown/setup between the runs within a single bench test.

I have played around with this in the past and I know how to hack it but I should just fix it upstream.

packages/shared/src/tool/vitest-perf-json-to-bmf.ts
```diff
@@ -166,10 +166,14 @@ function convertVitestToBMF(vitestOutput: VitestOutput): BMFMetric {
...
         bmf[metricName] = {
+          // benchmark outputs are the time in milliseconds it took to run.
+          // If we want to convert to throughput, we need to divide 1000 by the time.
+          // This is the inverse of the time taken.
+          // So if it took 100ms to run, the throughput is 1000 / 100 = 10
           throughput: {
-            value: benchmark.mean,
-            ['lower_value']: benchmark.min,
-            ['upper_value']: benchmark.max,
+            value: 1000 / benchmark.mean,
```
arv, line 174: The json does have a `hz` field but since we also convert min and max to hertz I think this is better.

arv (commented):
LGTM

PR #4317, opened 2025-05-02
feature(zbugs): email notify

cesara:


apps/zbugs/server/email.ts
```diff
@@ -0,0 +1,67 @@
+import nodemailer from 'nodemailer';
+
+async function getTransport() {
+  if (process.env.LOOPS_EMAIL_API_KEY !== undefined) {
+    const transport = nodemailer.createTransport({
```
aboodman, line 5: Surprised to see that we're using smtp directly.

This seems very likely to be a pita. Can we just use loops' REST API.

apps/zbugs/server/email.ts
```diff
@@ -0,0 +1,67 @@
...
+  recipients: string[];
+  title: string;
+  message: string;
+  link: string;
+}) {
+  const transport = await getTransport();
+  if ((transport as any).isLoops) {
+    await transport.sendMail({
+      from: 'no-reply@roci.dev',
+      to: recipients.join(', '),
```
tantaman, line 47: We don't want to reveal the email addresses of users to one another so we should do 1 email per `to` or use `bcc`. Maybe `loops.so` has a specific API for this case?

apps/zbugs/package.json
```diff
@@ -60,6 +60,7 @@
     "@types/js-cookie": "^3.0.6",
     "@types/k6": "^0.53.2",
     "@types/node": "^20.8.4",
+    "@types/nodemailer": "^6.4.17",
```
tantaman, line 63: you can rm this now.

tantaman (approved):
lgtm

PR #4321, opened 2025-05-03
fix(zero-cache): fix websocket liveness check to account for messages

darkgnotic:
When a websocket is processing a lot of messages (e.g. a large changelog catchup at the beginning of a change-streamer subscription), it is possible for the initial ping message to be backed up behind those messages, and not processed until it is too late.

Fix the liveness checking logic to accept either pings or messages as a sign of liveness.

User report:

https://discord.com/channels/830183651022471199/1367606405481365674/1368063265594282004


arv (commented):
LGTM

Test?

PR #4322, opened 2025-05-03
fix(zero-cache): handle exceptions in the eviction timer

darkgnotic:
User report:

https://discord.com/channels/830183651022471199/1367957671621296190/1368294818676998215

Catch and handle exceptions thrown from CVR operations in the eviction timer. Otherwise it can crash the server.


arv (commented):
LGTM

Test?

PR #4325, opened 2025-05-05
chore(zero-cache)!: remove `--tenants-json`; simplify

darkgnotic:
Farewell, `--tenants-json`.

Removes the multi-tenant feature of `zero-cache` and simplifies that layer of the code.

* We are back to one config definition (`zero-config.ts`)
* `/statsz` is handled at the top-level and not in the internal HttpServices
* Only the top-level `ZeroDispatcher` listens on the main port. The `zero-cache` child process only does websocket handoff to its children (i.e. sync workers) and does not consume a port itself.
* The `change-streamer` still runs on its own port, but can be dispatched via the top-level in the future for lazy startup of the replication manager. 
  * However, top-level dispatch of change streams is disabled in single-node mode for security (as it is an internal API).

Note that the entry point `runner` package (which used to be called `multi`) is still useful for internal process management, such as the `--lazy-startup` feature, and in the future, faster auto resets and permanent volume support. In other words, preserving this layer affords some process management logic that is otherwise slow or difficult to configure at the container platform level.


packages/zero-cache/src/server/worker-dispatcher.ts
```diff
@@ -0,0 +1,141 @@
...
+import type {IncomingMessageSubset} from '../types/http.ts';
+import type {Worker} from '../types/processes.ts';
+import {
+  createWebSocketHandoffHandler,
+  type Handoff,
+  type WebSocketHandoffHandler,
+} from '../types/websocket-handoff.ts';
+import {getConnectParams} from '../workers/connect-params.ts';
+
+export class WorkerDispatcher implements Service {
```
darkgnotic, line 18: Note: This was moved from `services/dispatcher/dispatcher.ts`. 

(git didn't figure that out, at least in my diff).

arv (commented):
LGTM

PR #4327, opened 2025-05-05
fix(zql): O(logn) point queries

tantaman:
No more full table scans to do a point query by primary key on the client.

https://bugs.rocicorp.dev/issue/3793

Before:
CleanShot 2025-05-05 at 20 53 33@2x

After:
CleanShot 2025-05-05 at 14 33 37

### DIFF
--- a/packages/zql/src/ivm/constraint.test.ts
+++ b/packages/zql/src/ivm/constraint.test.ts
@@ -1,0 +1,352 @@
+import {describe, test, expect} from 'vitest';
+import {
+  pullSimpleAndComponents,
+  primaryKeyConstraintFromFilters,
+} from './constraint.ts';
+import type {SimpleOperator} from '../../../zero-protocol/src/ast.ts';
+
+describe('constraint', () => {
+  describe('pullSimpleAndComponents', () => {
+    test.each([
+      {
+        name: 'pulls simple conditions from AND',
+        condition: {
+          type: 'and',
+          conditions: [
+            {
+              type: 'simple',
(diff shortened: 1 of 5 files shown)


tantaman:
nit -- we can move this computation to the static connection level

arv:
LGTM

packages/zql/src/ivm/constraint.test.ts
```diff
@@ -0,0 +1,352 @@
...
+      expect(result).toEqual(expected);
+    });
+  });
+});
+
+test('returns undefined for operators other than equality', () => {
+  const operators: SimpleOperator[] = [
+    '>',
+    '<',
+    '>=',
```
arv, line 323: Can we not allow `>=`?

tantaman: Not sure. The existing constraint logic does look to create a `startAt` and scan from there so potentially we can.

packages/zql/src/ivm/constraint.ts
```diff
@@ -42,6 +47,89 @@ export function constraintMatchesPrimaryKey(
...
+): SimpleCondition[] {
+  if (condition.type === 'and') {
+    return condition.conditions.flatMap(pullSimpleAndComponents);
+  }
+
+  if (condition.type === 'simple') {
+    return [condition];
+  }
+
+  if (condition.type === 'or' && condition.conditions.length === 1) {
```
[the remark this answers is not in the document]
tantaman: I wanted this to be able to work regardless of input format.

PR #4330, opened 2025-05-05
chore(zero-cache): consolidate config normalization logic into one place

darkgnotic:
(with the exception of `bin-analyze`  😉 )

Also adds network interface lookup which will be used for change-streamer discovery. Currently it is simply logged for analysis of interface names in different runtimes.


aboodman:
> (with the exception of bin-analyze 😉 )

why with this exception?

darkgnotic:
> > (with the exception of bin-analyze 😉 )
> 
> why with this exception?

`bin-analyze` relies on a single option (the cvr.db). There's a lot more stuff getting normalized now, and I wanted to keep that interface clean and manageable, and avoid having to optionalize the fields that bin-analyze doesn't use every time we add more normalization logic for the rest of the server.

Maybe there's a way to have our cake and eat it too?

PR #4333, opened 2025-05-06
chore(zql): harness to benchmark pushes

tantaman:
Seemed most efficient for me to knock this out while I was in there.

CleanShot 2025-05-06 at 10 44 02

There are many more benchmarks to write but for another time.

Pushing into a limited query inside the bound is a 40% reduction in speed for zql but not zqlite... I wonder if the time to convert rows to server names via table mapper is dominating??? Seems odd.

`tinybench` seems riddled with bugs (it does not respect any warmup and iterations options, it includes setup and tear down time in its benchmark). Considering swapping to: https://github.com/evanwashere/mitata or opening PRs against tinybench.

### DIFF
--- a/packages/zql-integration-tests/src/helpers/runner.ts
+++ b/packages/zql-integration-tests/src/helpers/runner.ts
@@ -46,6 +46,7 @@ import type {TableSchema} from '../../../zero-schema/src/table-schema.ts';
 import os from 'node:os';
 import path from 'node:path';
 import fs from 'node:fs/promises';
+import type {SourceChange} from '../../../zql/src/ivm/source.ts';
 
 const lc = createSilentLogContext();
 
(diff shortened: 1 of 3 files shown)


grgbkr:
Neat.  LGTM

arv:
I think tinybench is fine (afaict) but setup/teardown for each iteration is not working in the vitest wrapper.

I think:
- fixing vitest (upstream)
- hack vitest to make setup/teardown per iteration work
- change to tinybench directly
- change to something else?
are all valid options

arv:
To be a bit clearer. I think vitest's usage of tinybench is not working as intended.

packages/zql-integration-tests/src/helpers/runner.ts
```diff
@@ -304,49 +317,188 @@ export async function createVitests<TSchema extends Schema>(
...
+  delegates: Delegates;
+  dbs: DBs<TSchema>;
+  generatePush: PushGenerator | undefined;
 }) {
-  const zql = createQuery(queries.memory);
-  const zqlite = createQuery(queries.sqlite);
-  const pg = createQuery(queries.pg);
-  return [doBench('zql', zql), doBench('zqlite', zqlite), doBench('zpg', pg)];
+  switch (type) {
+    case 'push': {
```
arv, line 387: maybe makePushBenchmark and makeHydrationBenchmark since they do not share any code.

packages/zql-integration-tests/src/helpers/runner.ts
```diff
@@ -304,49 +317,188 @@ export async function createVitests<TSchema extends Schema>(
...
+      // materialize before starting the benchmark. We only want to time push
+      // and not the initial hydration.
+      const zqliteView = zqlite.materialize();
+      const zqlView = zql.materialize();
+      try {
+        benchPush('zql', newDelegates, must(generatePush));
+        benchPush('zqlite', newDelegates, must(generatePush));
+      } finally {
+        zqliteView.destroy();
+        zqlView.destroy();
```
arv, line 424: Are you sure that the benchmarks runs before this? I would have thought benchPush sets up the benchmark but does not actually run the benchmark.

arv (commented):
I'm not sure this is doing what you think it is doing. Please see comments.

packages/zql-integration-tests/src/helpers/runner.ts
tantaman: Yeah, you're probably right. I'll check.

PR #4335, opened 2025-05-06
feat(zero-cache)!: automatic replication-manager discovery / routing

darkgnotic:
### Fargate / most multi-node configurations (e.g. host / awsvpc networking)

To update a multi-node configuration without disruption:
1. roll out new `replication-manager`
2. roll out new `view-syncer`, replacing the:
   * `ZERO_CHANGE_STREAMER_URI=http://{host}` option with
   * `ZERO_CHANGE_STREAMER_MODE=discover`
3. when `view-syncer`s are rollback-safe, remove the internal load balancer that was previously used for `view-syncer` to `replication-manager` routing

### Single-node configurations

Single-node configurations are unaffected

### Uncommon multi-node configurations

For container setups in which the process does not have access to the externally visible ip address or port (e.g. using docker a.k.a. "bridge" mode networking), an external routing or proxying mechanism is still needed. In such configurations, add the `ZERO_CHANGE_STREAMER_ADDRESS={host}` option to the `replication-manager`, where `{host}` is the hostname that was formerly part of `ZERO_CHANGE_STREAMER_URI`, e.g.

Before:
* `view-syncer`: `ZERO_CHANGE_STREAMER_URI=http://internal-prod-repmgr-125468.us-east-1.elb.amazonaws.com`

After:
* `view-syncer`: `ZERO_CHANGE_STREAMER_MODE=discover`
* `replication-manager`: `ZERO_CHANGE_STREAMER_ADDRESS=internal-prod-repmgr-125468.us-east-1.elb.amazonaws.com`

Note: For configurations that continue to use an explicit load-balancing mechanism, the `replication-manager` health check should be configured to on the `/keepalive` path, and not the root `/` path.

## Feature

The discovery and routing of the `replication-manager` is now facilitated by the Postgres Change DB, using the same row-level locking mechanism used to enforce single-writer access to the change log.

This obviates the need for an external addressing or proxying mechanism such as Service Discovery, Service Connect, or an Internal Load Balancer.

Screenshot 2025-04-30 at 17 57 06


packages/zero-cache/src/services/http-service.ts
```diff
@@ -55,7 +75,10 @@ export class HttpService implements Service {
   }
 
   async run(): Promise<void> {
-    await this.start();
+    // Check if start() was already called.
+    if (this.#fastify.addresses().length === 0) {
+      await this.start();
+    }
```
arv, line 81: If start was already called, this will stop? Is that the intended behavior?

packages/zero-cache/src/services/change-streamer/schema/tables.ts
```diff
@@ -51,15 +51,25 @@ export type ReplicationState = {
...
 `;
 }
 
+export async function discoverChangeStreamerAddress(
+  shard: ShardID,
+  sql: PostgresDB,
+): Promise<string | null> {
+  const result = await sql<{ownerAddress: string | null}[]>/*sql*/ `
+    SELECT "ownerAddress" FROM ${sql(cdcSchema(shard))}."replicationState"`;
+  return result[0].ownerAddress;
```
arv, line 70: At some point it would make sense to start using `values()` more.

packages/zero-cache/src/services/change-streamer/schema/tables.ts
```diff
@@ -100,10 +110,10 @@ export async function setupCDCTables(
   await db.unsafe(createTables(shard));
 }
 
-export async function markResetRequired(db: PostgresDB, shard: ShardID) {
+export async function markResetRequired(sql: PostgresDB, shard: ShardID) {
   const schema = cdcSchema(shard);
-  await db`
-  UPDATE ${db(schema)}."replicationConfig"
+  await sql`
+  UPDATE ${sql(schema)}."replicationConfig"
```
arv, line 116: OOC, how do you get VSCode to syntax highlight these?

arv (commented):
LGTM

Very nice

packages/zero-cache/src/services/change-streamer/schema/tables.ts
darkgnotic: Sorry I missed this question!

I'm using this (on Matt's recommendation):

https://marketplace.visualstudio.com/items?itemName=frigus02.vscode-sql-tagged-template-literals-syntax-only

PR #4343, opened 2025-05-07
chore(zbugs): leverage `and`'s filtering of `undefined`

tantaman:
Stumbled on this pattern while working through custom queries.

Elegant.

### DIFF
--- a/apps/zbugs/src/pages/list/list-page.tsx
+++ b/apps/zbugs/src/pages/list/list-page.tsx
@@ -56,53 +56,41 @@ export function ListPage({onReady}: {onReady: () => void}) {
     qs.get('sort')?.toLowerCase() === 'created' ? 'created' : 'modified';
   const sortDirection =
     qs.get('sortDir')?.toLowerCase() === 'asc' ? 'asc' : 'desc';
+  const open =
+    status === 'open' ? true : status === 'closed' ? false : undefined;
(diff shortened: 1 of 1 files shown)


apps/zbugs/src/pages/list/list-page.tsx
```diff
@@ -56,54 +56,42 @@ export function ListPage({onReady}: {onReady: () => void}) {
...
-
-  if (textFilter) {
-    q = q.where(({or, cmp, exists}) =>
-      or(
-        cmp('title', 'ILIKE', `%${escapeLike(textFilter)}%`),
-        cmp('description', 'ILIKE', `%${escapeLike(textFilter)}%`),
-        exists('comments', q =>
-          q.where('body', 'ILIKE', `%${escapeLike(textFilter)}%`),
-        ),
+    .related('viewState', q => q.where('userID', z.userID).one())
```
aboodman, line 68: It's a cool pattern, and this is the best "one-line" version of this that I've seen so far, but honestly I prefer the code as multiple statements.

I like that the query system is composable so that you can construct the query iteratively with normal, simple procedural code.

aboodman: With that I would *almost* be on your side. That's pretty compact.

I think I would still probably write it as multiple statements myself but not certain.

PR #4344, opened 2025-05-07
feat(zero-cache)!: improved backup protocol and change-log cleanup logic

darkgnotic:
## ⚠️  Deployment notes

* The new `replication-manager` must be rolled out before `view-syncer`s are rolled out
* The `ZERO_LITESTREAM_BACKUP_URL` is no longer needed by `view-syncers`. 
  * Specifying it is fine but it will be ignored.
  * This means that initiating a resync (by updating the backup url) only requires updating the `replication-manager`. The `view-syncers` will pick this up and reset when signaled by the `replication-manager`; there is no need to update them manually.

## Feature

Better backup protocol and cleanup logic

This feature was inspired by @tantaman's idea for more reliably pausing change-log cleanup when a `litestream` restore is in progress: https://github.com/rocicorp/mono/pull/4263#discussion_r2060297645

* The `replication-manager` (i.e. `change-streamer`) exports a new `/snapshot` websocket handler used to "reserve" a snapshot.
* `view-syncers` make this request before restoring the replica from the backup. In particular, the `replication-manager` immediately sends a downstream message indicating the backup-url from which to restore. This obviates the need for the `view-syncers` to be configured with backup-url, replacing the previous config coordination point with a runtime protocol.
* The request is kept open while the replica is restored and prepared. While a "reservation" is in progress, the backup monitor refrains from scheduling any watermark cleanup.
* When the `view-syncer` finishes its restore and subscribes to the `change-streamer`, the corresponding reservation is closed. This allows the backup monitor to resume watermark cleanup scheduling.
* In addition, the duration for which the reservation lasted is used to increase the cleanup delay, if approprate, as a secondary precaution against premature change log cleanup.

### Miscellaneous cleanup

* The BackupMonitor implementation was simplified to perform all logic in a single `setInterval()` callback (previously there was a Timer for every watermark encountered).
* On each interval, the latest qualifying watermark is scheduled for cleanup. Earlier watermarks are discarded.

### DIFF
--- a/.github/workflows/sst-gigabugs-deploy.yml
+++ b/.github/workflows/sst-gigabugs-deploy.yml
@@ -2,7 +2,7 @@ name: Build and Deploy Gigabugs Zero-Cache AWS SST
 
 on:
   push:
-    branches: [darkgnotic/better-backup] # [gigabugs]
+    branches: [gigabugs]
 env:
   # Setting an environment variable with the value of a configuration variable
   ECR_IMAGE_ZERO_CACHE: zero-zgigabugs
--- a/.github/workflows/sst-gigabugs-deploy.yml
+++ b/.github/workflows/sst-gigabugs-deploy.yml
@@ -2,7 +2,7 @@ name: Build and Deploy Gigabugs Zero-Cache AWS SST
 
 on:
   push:
-    branches: [gigabugs]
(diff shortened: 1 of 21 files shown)


tantaman (commented):
lgtm. Clever use of websockets to hold the reservation.

PR #4359, opened 2025-05-10
chore(zero-cache): avoid sending the `columns` field in pg MessageRelations

darkgnotic:
The `columns` field can be large for tables with many columns. They are not used downstream, so removing them from the change-stream can significantly reduce the amount of downstream I/O (i.e. change-db writes, change-streamer -> subscriber websocket)


arv (commented):
LGTM

PR #4362, opened 2025-05-12
fix(zero-cache): fix ordering of columns in replicated indexes

darkgnotic:
User report:

https://discord.com/channels/830183651022471199/1370898291721506876/1371170647241265293

Fixes replicated indexes to respect the column ordering of the upstream index.

Tangential: Also corrects the logic that excludes `INCLUDE`d (i.e. stored but not indexed) columns. 

### ⚠️ NOTE

To fully apply this fix, a resync is required. 

A resync is thus recommended for any app with compound indexes (including primary keys).

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/change-source.end-to-mid.pg-test.ts
+++ b/packages/zero-cache/src/services/change-source/pg/change-source.end-to-mid.pg-test.ts
@@ -166,9 +166,9 @@ describe('change-source/pg/end-to-mid-test', {timeout: 30000}, () => {
       ],
       [
         {
-          columns: {id: 'ASC'},
-          name: 'my.baz_pkey',
           tableName: 'my.baz',
+          name: 'my.baz_pkey',
+          columns: {id: 'ASC'},
           unique: true,
         },
       ],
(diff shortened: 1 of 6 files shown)


arv (commented):
LGTM

(but I'm not sure I understand the SQL changes and how they fix the issue)

darkgnotic:
> LGTM
> 
> (but I'm not sure I understand the SQL changes and how they fix the issue)

Sorry, it was probably confusing to bundle two fixes, but fixing the tests revealed the other issue so I felt like I had no choice.  😁 

The `DISTINCT` is what was causing the problem (the second change):

Screenshot 2025-05-14 at 13 00 29

The preceding query goes through the trouble of ensuring that the columns arrive in the same order as the index key:

```sql
JOIN LATERAL (
      SELECT pg_attribute.attname as name, col.index_pos as pos
        FROM UNNEST( (pg_index.indkey::smallint[])[:pg_index.indnkeyatts - 1] ) 
          WITH ORDINALITY as col(table_pos, index_pos)
        JOIN pg_attribute ON attrelid = pg_index.indrelid AND attnum = col.table_pos
    ) AS index_column ON true
```

only to be reordered by `DISTINCT`, which I guess has the side effect of alphabetizing keys.

The first change in the diff has to do with excluding columns that are stored (via the `INCLUDE` keyword) but not otherwise indexed. Those aren't relevant to zero and we don't want to dismiss the index if that stored column is generated.

PR #4365, opened 2025-05-13
chore(shared): Add literalUnion helper

arv:
This is such a common pattern that it should be a helper.


darkgnotic (commented):
Thank you! This is so much better.

PR #4367, opened 2025-05-13
chore(zero-cache): allow Transaction pool users to block on write tasks

darkgnotic:
Refactors the TransactionPool to allow callers to optionally block on write tasks.

This will be used in an upcoming change that allows the ChangeDB to exert back-pressure in the face of a large replication stream backlog. (Revealed by stress tests while investigating https://discord.com/channels/830183651022471199/1370057107814088796/1370245009219260599).

A related optimization in this change is to only await the result of the last statement, rather than holding all of the statements in memory. This allows most completed statements to be garbage collected; only the result of the last statement need be awaited to satisfy transactional semantics.


arv (approved):
LGTM

PR #4368, opened 2025-05-13
fix(analyze-query): `from` support in permissions, `schema_path` config arg

tantaman:
- the schema path arg did not follow the same format as `deploy permissions`
- analyze-query would throw when permissions used columns or tables with different server names. This is because the permissions are already in server-name form.


erikmunson:
yay thanks matt!!

arv (commented):
LGTM

PR #4369, opened 2025-05-13
feature(zero-cache): otel auto instrumentation

cesara:


packages/zero-cache/src/server/logging.ts
```diff
@@ -13,12 +13,9 @@ import {OtelLogSink} from './otel-log-sink.ts';
...
   const consoleSink =
     config.format === 'json' ? consoleJsonLogSink : consoleLogSink;
-  if (config.logCollector) {
-    const otelSink = new OtelLogSink(config);
-    return new CompositeLogSink([otelSink, consoleSink]);
-  }
-
-  return consoleSink;
+  // Always create OtelLogSink for debugging
+  const otelSink = new OtelLogSink();
```
tantaman, line 17: can we check if the otel env vars are set and only create the sink if so? No need to log to OTEL if OTEL is not enabled.

packages/zero-cache/src/server/otel-log-sink.ts
```diff
@@ -1,26 +1,24 @@
...
     // this is a no-op if already started
-    startOtel(endpoints);
+    startOtelAuto();
     this.#logger = logs.getLogger('zero-cache');
   }
 
-  log(level: LogLevel, context: Context | undefined, ...args: unknown[]): void {
+  log(
+    level: LogLevel,
+    _context: Context | undefined,
```
tantaman, line 19: we need to add the context to the `payload.attributes` otherwise we'll lose all the context that has been set for the logger.

```ts
 if (context) {
    payload.attributes = context as AnyValueMap;
  }
```

packages/zero-cache/src/server/otel-log-sink.ts
```diff
@@ -35,30 +33,11 @@ export class OtelLogSink implements LogSink {
...
-      severityNumber: toErrorNum(level),
       severityText: level,
       body: message,
-      timestamp: Date.now() * 1_000_000, // nanoseconds
     };
-    if (context) {
-      payload.attributes = context as AnyValueMap;
-    }
 
+    // eslint-disable-next-line no-console
```
tantaman, line 40: remove?

packages/zero-cache/src/server/otel-log-sink.ts
```diff
@@ -35,30 +33,11 @@ export class OtelLogSink implements LogSink {
     }
 
     const payload: LogRecord = {
```
tantaman, line 35: why no severity?

packages/zero-cache/src/server/otel-start.ts
```diff
@@ -3,82 +3,92 @@ import {OTLPMetricExporter} from '@opentelemetry/exporter-metrics-otlp-http';
...
+import {getNodeAutoInstrumentations} from '@opentelemetry/auto-instrumentations-node';
 
-export type OtelEndpoints = {
-  traceCollector?: string | undefined;
-  metricCollector?: string | undefined;
-  logCollector?: string | undefined;
-};
+class OtelManager {
+  private static _instance: OtelManager;
+  private _started = false;
```
tantaman, line 29: nit: preference is for the JS `#` for private fields
```suggestion
  static #instance: OtelManager;
  #started = false;
```

packages/zero-cache/src/server/otel-log-sink.ts
cesara: readded

packages/zero-cache/src/server/otel-log-sink.ts
cesara: added context

packages/zero-cache/src/server/otel-start.ts
```diff
@@ -3,82 +3,92 @@ import {OTLPMetricExporter} from '@opentelemetry/exporter-metrics-otlp-http';
...
-  const logRecordProcessors: LogRecordProcessor[] = [];
-  const resource = resourceFromAttributes({
-    [ATTR_SERVICE_NAME]: 'syncer',
-    [ATTR_SERVICE_VERSION]: version,
-  });
+  startOtelAuto() {
+    if (this._started) {
+      return;
+    }
+    this._started = true;
```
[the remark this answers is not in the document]
cesara: tested and works

packages/zero/src/zero-cache-dev.ts
```diff
@@ -117,19 +117,27 @@ async function main() {
...
-        env: {
-          // Set some low defaults so as to use fewer resources and not trip up,
-          // e.g. developers sharing a database.
-          ['ZERO_NUM_SYNC_WORKERS']: '3',
-          ['ZERO_CVR_MAX_CONNS']: '6',
-          ['ZERO_UPSTREAM_MAX_CONNS']: '6',
-          // But let the developer override any of these dev defaults.
-          ...process.env,
+      zeroCacheProcess = spawn(
+        'npx',
```
tantaman, line 121: missed this. We can't call `npx` directly. If the user used a different package manager then this call will fail.

e.g.: https://bugs.rocicorp.dev/issue/3888

PR #4371, opened 2025-05-13
fix(zero-cache): add flow control for large change DB transactions

darkgnotic:
Respond to back pressure from the change db for very large transactions by awaiting (the execution of) every 10,000th statement. This builds off of the new Promise API for Transaction Pool writes in https://github.com/rocicorp/mono/pull/4367.

This was verified in a stress test that performs a 2M row update. Previously this would result in crashing the replication-manager with an OOM, and now it successfully process the transaction as quickly as the change-db is able to store the changes.

User report:

https://discord.com/channels/830183651022471199/1370057107814088796/1370245009219260599

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
+++ b/packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
@@ -357,8 +357,12 @@ class ChangeStreamerImpl implements ChangeStreamerService {
 
           if (type === 'commit') {
             watermark = null;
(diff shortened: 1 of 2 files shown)


darkgnotic:
> LGTM
> 
> Not sure if this is something we should test? I'm totally fine without a test for this.

Yeah, I was thinking along similar lines. I did verify it with a stress test locally (which is how I discovered the issue), but this sort of thing (where it either works or it eventually OOMS) is hard to test in an unit test.

PR #4376, opened 2025-05-14
chore(zero-cache): logic for parsing a pg COPY stream

darkgnotic:
This will be the basis for migrating initial-sync to use postgres COPY streams rather than cursor batching, allowing TCP backpressure to adapt the amount of buffered data as it arrives.

* TextTransform: implements the parsing of the [COPY text format](https://www.postgresql.org/docs/current/sql-copy.html) as it arrives in a NodeJS Readable stream. (Surprisingly, I couldn't find a suitable streaming TSV parser out there)
* RowTransform: applies the Postgres-client configured parsers to convert the text values to JS values, as it is done for non-streaming PG queries as well as for the replication stream.

This is logic and tests only. Integration into initial-sync is forthcoming.


darkgnotic:
btw, the diff doesn't display the test because it contains binary characters, but it is a pretty good / thorough test

please let me know if you can think of any edge cases that we should add to the test.

packages/zero-cache/src/db/pg-copy.ts
```diff
@@ -0,0 +1,164 @@
...
+          append(text.substring(l, r));
+        }
+        l = r + 1;
+      };
+
+      for (; r < text.length; r++) {
+        const ch = text[r];
+        if (this.#escaped) {
+          switch (ch) {
+            case 'N':
```
arv, line 51: Blast from the past!

Use `charCodeAt` for these kind of things. It will use a SMIs which is stack allocated instead of a string which is heap allocated and give a nice perf increase.

packages/zero-cache/src/db/pg-copy.ts
```diff
@@ -0,0 +1,164 @@
...
+  ) {
+    try {
+      const text = chunk.toString(encoding);
+
+      let l = 0;
+      let r = 0;
+
+      const append = (text: string) => {
+        this.#currCol ??= '';
+        this.#currCol += text;
```
arv, line 37: micro optimize?

packages/zero-cache/src/db/pg-copy.ts
```diff
@@ -0,0 +1,164 @@
...
+    const typeParsers = await getTypeParsers(db);
+    const columnParsers = columns.map(c =>
+      typeParsers.getTypeParser(c.typeOID),
+    );
+    return new RowTransform(columnParsers);
+  }
+
+  readonly #columnParsers: TypeParser[];
+
+  private constructor(columnParsers: TypeParser[]) {
```
arv, line 140: Why this dance with static create and private constructor?

arv (commented):
LGTM

packages/zero-cache/src/db/pg-copy.ts
darkgnotic: > micro optimize?

Can you elaborate? I want to learn.  🙏

packages/zero-cache/src/db/pg-copy.ts
darkgnotic: The static create potentially blocks on I/O.

packages/zero-cache/src/db/pg-copy.ts
darkgnotic: > Blast from the past!
> 
> Use `charCodeAt` for these kind of things. It will use a SMIs which is stack allocated instead of a string which is heap allocated and give a nice perf increase.

Got it. Thank you!

packages/zero-cache/src/db/pg-copy.ts
```diff
@@ -0,0 +1,164 @@
+import {Transform} from 'node:stream';
```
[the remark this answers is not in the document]
darkgnotic: Oh, I don't know; I haven't looked at web streams. I would be interested in switching to the latter if it is better in some way.

PR #4380, opened 2025-05-14
docs: improve client auth JSDoc

Gerbuuun:
- Clarified why Zero client keeps retrying tokens that don't change
- Explicitly mention `sub` claim in JWT for `userID` field
- Explicitly mention JWT

### DIFF
--- a/packages/zero-client/src/client/options.ts
+++ b/packages/zero-client/src/client/options.ts
@@ -35,13 +35,13 @@ export interface ZeroOptions<
    *
    * Token validation behavior:
    * 1. **For function providers:**  
-   *    When Zero-cache reports that a token is invalid (expired, malformed,
+   *    When zero-cache reports that a token is invalid (expired, malformed,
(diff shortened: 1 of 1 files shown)


packages/zero-client/src/client/options.ts
```diff
@@ -28,15 +28,22 @@ export interface ZeroOptions<
...
-   * 2. if a string token was provided Zero will continue to retry with the
-   *    provided token.
+   * A JWT to identify and authenticate the user. Can be provided as either:
+   * - A string containing the JWT token
+   * - A function that returns a JWT token
+   * - `undefined` if there is no logged in user
+   *
+   * Token validation behavior:
+   * 1. **For function providers:**  
+   *    When Zero-cache reports that a token is invalid (expired, malformed,
```
aboodman, line 38: Use "`zero-cache`" (with the backticks and lower-case) when referring to this component.

PR #4381, opened 2025-05-14
fix(zql): return after first row for point queries (60x speedup)

tantaman:
**60x faster** (of course this depends on the size of the collection. ~3,500 rows in this case)

CleanShot 2025-05-14 at 19 39 13

Before:
CleanShot 2025-05-14 at 19 38 32

After:
CleanShot 2025-05-14 at 19 37 07


grgbkr:
LGTM.

PR #4382, opened 2025-05-15
feat(zero-cache): memory-adaptive implementation of initial-sync

darkgnotic:
Replaces the previous cursor-batched table copying logic in initial sync with a streaming implementation using postgres `COPY TO` streams. 

Each table copy worker now downloads rows via a COPY stream and uses NodeJS streaming to regulate memory usage via TCP backpressure on upstream. It is no longer necessary to tune a "batch size" in order to avoid out of memory errors for large tables / rows.

User reports:
* https://discord.com/channels/830183651022471199/1372283029644640290/1372284231128055939
* https://discord.com/channels/830183651022471199/1372086577072836618/1372087049779281970

### DIFF
--- a/packages/zero-cache/src/config/zero-config.ts
+++ b/packages/zero-cache/src/config/zero-config.ts
@@ -520,18 +520,9 @@ export const zeroOptions = {
       type: v.number().default(5),
       desc: [
         `The number of parallel workers used to copy tables during initial sync.`,
-        `Each worker copies a single table at a time, fetching rows in batches of`,
-        `of {bold initial-sync-row-batch-size}.`,
-      ],
-    },
-
-    rowBatchSize: {
-      type: v.number().default(10_000),
-      desc: [
-        `The number of rows each table copy worker fetches at a time during`,
-        `initial sync. This can be increased to speed up initial sync, or decreased`,
-        `to reduce the amount of heap memory used during initial sync (e.g. for tables`,
-        `with large rows).`,
+        `As each worker establishes a db connection, the upstream database must be`,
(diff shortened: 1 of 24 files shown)


packages/zero-cache/src/config/zero-config.ts
```diff
@@ -520,18 +520,9 @@ export const zeroOptions = {
       type: v.number().default(5),
       desc: [
         `The number of parallel workers used to copy tables during initial sync.`,
-        `Each worker copies a single table at a time, fetching rows in batches of`,
-        `of {bold initial-sync-row-batch-size}.`,
-      ],
-    },
-
-    rowBatchSize: {
```
arv, line 528: I guess this is considered a breaking change now?

packages/zero-cache/src/services/change-source/pg/initial-sync.ts
```diff
@@ -322,56 +331,70 @@ async function copy(
...
+
+  const valuesPerRow = columnSpecs.length + 1; // includes _0_version column
+  // Accumulates row values up to INSERT_BATCH_SIZE batchedRows.
+  let values: LiteValueType[] = [];
+  let batchedRows = 0;
+
+  await pipeline(
+    await from.unsafe(`COPY (${selectStmt}) TO STDOUT`).readable(),
+    new TextTransform(),
+    await RowTransform.create(dbClient, columnSpecs),
```
arv, line 344: Now I see why you wanted to have create/private constructor :-)

arv (commented):
LGTM

PR #4383, opened 2025-05-15
chore(zero-cache): micro-optimize TextTransform used in COPY

darkgnotic:
10% ~ 15% performance improvement

Based on helpful feedback from @arv

* use `charCodeAt()` to reduce string allocation
* reduce / inline function calls in the main loop

Also simplify internal representation to use the NULL byte to signify a `null` column value, eliminating the need for `undefined` / `null` checks on the member variables.


arv (commented):
LGTM

PR #4385, opened 2025-05-15
chore(zero-cache): CVR: handle both custom and regular queries [cust-queries]

tantaman:
- call out that `queryHash` is the client query hash
- make `clientAST` nullable (not used for custom queries)
- add fields for query name and args. This is the custom query case.

Custom queries do not persist an AST and will call the API server [every time before they are hydrated](https://www.notion.so/replicache/Custom-Queries-1773bed8954580e59573df47f699deb3?pvs=4#1f13bed8954580508e04fdc6dd3f99d5). This is the case since a given client connection may be established with updated / changed auth data.

The queries returned by the API server will:
1. Replace existing pipelines for the client group if their server ast hash differs
2. Be added to the pipelines for the client group if there is not yet a pipeline with a given server ast hash
3. Cause pipelines to be destroyed for the client group if they are running and:
   a. have the same client query hash
   b. are not present in the set of returned server hashes


packages/zero-cache/src/services/view-syncer/cvr.pg-test.ts
```diff
@@ -2621,6 +2675,16 @@ describe('view-syncer/cvr', () => {
       ),
     ).toMatchInlineSnapshot(`
       [
+        {
```
tantaman, line 2678: @darkgnotic - idk why this would have moved with this change and if that matters.

tantaman: it seems more correct that `stateVersion 19z` is now output before `stateVersion 1aa`

packages/zero-cache/src/services/view-syncer/schema/cvr.ts
```diff
@@ -108,8 +110,10 @@ function createQueriesTable(shard: ShardID) {
   return `
 CREATE TABLE ${schema(shard)}.queries (
   "clientGroupID"         TEXT,
-  "queryHash"             TEXT,
-  "clientAST"             JSONB NOT NULL,
+  "queryHash"             TEXT, -- this is the hash of the client query AST
+  "clientAST"             JSONB, -- this is nullable as custom queries will not persist an AST
+  "queryName"             TEXT, -- the name of the query if it is a custom query
+  "queryArgs"             JSONB, -- the arguments of the query if it is a custom query
```
darkgnotic, line 116: Probably not an issue, here, but a couple of caveats about storing data as `JSONB`:
- Fields are reordered, so the dev won't be able to rely on the fact that they object that they send on the wire comes out with the fields in the same order on the api-server
- The NULL byte is not allowed in `JSONB`.

The latter issue was actually a problem for downstream replication, where a user had NULL bytes in their JSON field that needed to be preserved: https://github.com/rocicorp/mono/pull/3810

Not sure if it will ever matter for custom queries args, but it's not impossible.

So I guess my take is that, unless we need specific JSONB properties (indexing, etc.), it might be better to store things as `JSON`, so that we preserve the original data to the greatest extent possible.

tantaman: Thanks for calling that out. Will switch to `json`.

PR #4386, opened 2025-05-15
perf(zql): replace DNF normalization with condition simplification 

grgbkr:
https://github.com/rocicorp/mono/commit/b9c9188be0443a53a2e6c9f6f3d0a53e5b906154 enabled zero to correctly and efficiently process queries with conditions not in DNF. 

DNF can lead to performance problems and even queries-of-death as in worse case it can lead to an exponential increase in condition size.  See:  https://bugs.rocicorp.dev/issue/3569

This is showing a nice throughput improvement on some of our existing benchmarks:
image


tantaman (commented):
lgtm

PR #4390, opened 2025-05-16
feat(zero): Treat PG Arrays as JSON

arv:
# Support for Postgres array types

This PR introduces support for Postgres array types in our schema system. Here's what's included:

## Key changes

- Added support for Postgres array types which are represented as JSON in the Zero schema
- Arrays in SQLite are stored as TEXT with a new `|TEXT_ARRAY` attribute, using the same pattern as our existing attributes for enums and non-null types
- Added proper type handling for array columns in schema compatibility checks
- Enhanced schema fetching to detect array types and their element types

## Implementation details

- Modified `ServerColumnSchema` to include an `isArray` flag alongside the existing `isEnum` flag
- Updated the schema detection in Postgres to properly identify array types and their element types
- Added support in `dataTypeToZqlValueType` to handle array types properly
- Added helper functions `isEnum()` and `isArray()` for type checking
- Extended type system to prevent comparison operations on array columns
- Added tests for array type handling

## Challenges addressed

The main complexity was properly handling enum arrays, which required detecting:
1. Whether a column is an array
2. Whether the element type of that array is an enum

This required querying the Postgres type system more deeply to get information about element types.

## Testing

- Added unit tests for array columns in schema builder
- Updated existing tests to include the new `elemPgTypeClass: null` field
- Added type checking tests to verify proper typing of array fields


aboodman:
Do we want to store in SQLite as json to support future JSON operators?

arv:
> Do we want to store in SQLite as json to support future JSON operators?

It doesn't matter. SQLite is dynamically typed and TEXT and JSON are the same. The only real difference is that ORMs might do something special for JSON columns.

packages/zero-cache/src/services/change-source/protocol/version.test.ts
```diff
@@ -38,9 +38,9 @@ test('protocol versions', () => {
   // Then update the version number of the `CHANGE_SOURCE_PATH`
   // in current and export it appropriately as the new version
   // in `mod.ts`.
-  t(current, '2v7p7445ss8ce', '/changes/v0/stream');
+  t(current, 'hfkz08pmaavz', '/changes/v0/stream');
```
arv, line 41: The ColumnSpec got a new optional field.

packages/z2s/src/compiler.output.test.ts
```diff
@@ -33,7 +34,9 @@ const user = table('user')
   .columns({
     id: string(),
     name: string(),
+    nameArray: json<string[]>(),
```
tantaman, line 37: is an array of json possible in PG? E.g., `json<json[]>`. Granted it seems like an insane edge case.

packages/zero-cache/src/db/lite-tables.ts
```diff
@@ -60,12 +63,19 @@ export function listTables(db: Database): LiteTableSpec[] {
       tables.push(table);
     }
 
+    const elemPgTypeClass = isArray(col.type)
+      ? isEnum(col.type)
+        ? PostgresTypeClass.Enum
+        : PostgresTypeClass.Base
```
tantaman, line 69: Base is the super type of all type in PG or something?

packages/zero-cache/src/db/specs.ts
```diff
@@ -26,6 +26,12 @@ export const columnSpec = v.object({
   pos: v.number(),
   dataType: v.string(),
   pgTypeClass: pgTypeClassSchema.optional(),
+
+  // If the column is an array, this will be the type of the
+  // elements in the array. If the column is not an array,
+  // this will be null.
+  elemPgTypeClass: pgTypeClassSchema.nullable().optional(),
```
tantaman, line 33: and we've punted on supporting arrays of arrays, correct?

packages/zql-integration-tests/src/helpers/runner.ts
```diff
@@ -30,24 +49,6 @@ import {
...
-import type {SourceSchema} from '../../../zql/src/ivm/schema.ts';
-import type {Node} from '../../../zql/src/ivm/data.ts';
-import type {Change} from '../../../zql/src/ivm/change.ts';
-import {must} from '../../../shared/src/must.ts';
-import {
-  clientToServer,
-  type NameMapper,
-} from '../../../zero-schema/src/name-mapper.ts';
-import type {Writable} from '../../../shared/src/writable.ts';
-import type {TableSchema} from '../../../zero-schema/src/table-schema.ts';
```
tantaman, line 45: does your prettier plugin auto-sort imports?

tantaman (commented):
still reviewing but very cool. How far down the priority list are array operators?

packages/zero-cache/src/db/specs.ts
arv: pg does not support array of array. it supports multi dimensional arrays which is slightly different. Those are strange in pg to say the least

packages/zero-cache/src/db/lite-tables.ts
arv: This is not entirely correct. I should add a comment... there are more types and this fells a bit unclean

packages/z2s/src/compiler.output.test.ts
arv: From zero's point of view it is just json. I haven't tested if that maps cleanly... Or at least I don't remember

PR #4392, opened 2025-05-16
chore(zero): remove `ast` from downstream query patch messages [cust-queries]

tantaman:
1. It is extra information that does not need to be sent
2. Custom queries will not send their AST to the client

**edit** -- double check this is safe given we do push some queries back up on reconnect which could have had their AST fields nuked via poke.

### DIFF
--- a/packages/zero-protocol/src/queries-patch.ts
+++ b/packages/zero-protocol/src/queries-patch.ts
@@ -4,8 +4,14 @@ import {astSchema} from './ast.ts';
 export const putOpSchema = v.object({
   op: v.literal('put'),
   hash: v.string(),
-  ast: astSchema,
+  // Deprecated. The server does not need to poke the AST back down
+  // to the client.
+  ast: astSchema.optional(),
   ttl: v.number().optional(),
+});
+
+export const upPutOpSchema = putOpSchema.extend({
+  ast: astSchema,
 });
 
 const delOpSchema = v.object({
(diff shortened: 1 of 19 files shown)


packages/zero-protocol/src/queries-patch.ts
```diff
@@ -4,10 +4,13 @@ import {astSchema} from './ast.ts';
 export const putOpSchema = v.object({
   op: v.literal('put'),
   hash: v.string(),
-  ast: astSchema,
```
darkgnotic, line 7: This is a tricky one. We're bumpgin the PROTOCOL_VERSION (which is correct), but not the MIN_SERVER_SUPPORTED_PROTOCOL_VERSION, which means the server will still accept connections from old clients (that expect the `ast` field), and for them the server will return a response that fails to parse on the client side.

Best thing to do would be to make `ast` optional, with a comment that it's deprecated, and when we later bump the `MIN_SERVER_SUPPORTED_PROTOCOL_VERSION` to 17, then we can delete the field.

darkgnotic: Also, to clarify, we should not bump the `MIN_SERVER_SUPPORTED_PROTOCOL` in this change as that would result in a period of time in which the new server cannot talk to any old clients.

tantaman:
Made AST optional and marked as deprecated.

PR #4394, opened 2025-05-18
Add support for `sslmode=no-verify`

typedrat:
Instead of trying to cover every query-string option related to SSL, this just unblocks my use case, [Neon Local](https://neon.tech/docs/local/neon-local).


aboodman:
LGTM - @darkgnotic can you review on monday.

PR #4395, opened 2025-05-19
chore(zero-cache): add a unit test that triggers the COPY stream PG hang

darkgnotic:
Adds a unit test to the `initial-sync` test suite that triggers the Postgres-hanging bug that happens in the `COPY ... TO` implementation that has been reverted. The hang reproduces when reviving the change, and will be used to vet that the next `COPY ... TO`-based implementation is not susceptible to (or rather, successfully works around) the issue.


arv (commented):
LGTM

PR #4396, opened 2025-05-19
chore(zero-cache): track the (approx) size of each row in the copy transform

darkgnotic:
This will be used in the next iteration of the copy-stream implementation, for size-based buffering logic.

Also, small optimization of reusing the array of strings (from the `TextTransform`) for the outputs of the `RowTransform`, to save one array allocation per row.


arv (commented):
LGTM

PR #4399, opened 2025-05-19
chore(zero-cache): CopyRunner substitute for TransactionPool

darkgnotic:
CopyRunner is a special-case "drop-in" replacement for TransactionPool specifically intended for running a single `COPY` command and either reusing or discarding the connection based on whether it is healthy after the copy.

This works around a bug in Postgres in which certain types / sequences of `COPY` commands cause it to be unresponsive on the connection after it completes.

The unit test recreates the Postgres hang issue on the TransactionPool, and verifies that the CopyRunner successfully processes the same logic.


arv (commented):
LGTM

PR #4401, opened 2025-05-20
fix(zql): pk lookup on `related`: `related(x, q => q.where('id', y))`

tantaman:
Primary key filters in `related` calls would fail and return no results.

```ts
z.issue.related(
  'comments',
  q => q.where('id', id)
);
```

`req.constraint` in memory source should have been replaced everywhere by the primary key constraint if/when it existed. This moves the primary key constraint definition up to the top of the function and replaces all references to `req.constraint`

---
# User Reports

- https://discord.com/channels/830183651022471199/1374349312888147968/1374349316902359132
- https://discord.com/channels/830183651022471199/1334350942304206919/1374073823934156861

---

These query forms pass in `chinook.pg-test.ts`. Figuring that out separately.

### DIFF
--- a/packages/zql-integration-tests/src/chinook/chinook.pg-test.ts
+++ b/packages/zql-integration-tests/src/chinook/chinook.pg-test.ts
@@ -200,6 +200,69 @@ test.each(
             ])
             .limit(10),
       },
+      {
+        name: 'Related with where',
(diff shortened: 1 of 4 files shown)


arv (commented):
LGTM

PR #4405, opened 2025-05-20
feature(zbugs): emails according to spec

cesara:


apps/zbugs/server/email.ts
```diff
@@ -30,19 +33,31 @@ export async function sendEmail({
     return;
   }
 
+  const titleMessage = [title, message].filter(Boolean).join('\n');
+  // --- headers for threading ---
+  const threadId = `<issue-${issue.id}@bugs.rocicorp.dev>`;
```
aboodman, line 38: So cool! Do you have a screencap of this working??

apps/zbugs/server/notify.ts
```diff
@@ -202,20 +258,80 @@ function clip(s: string) {
...
   issueID: string,
+  isAssigneeChange = false,
+  previousAssigneeID?: string | null | undefined,
 ): Promise<string[]> {
   const sql = tx.dbTransaction.wrappedTransaction;
 
-  const recipientRows = await sql`SELECT DISTINCT "user".email
-  FROM (
-      SELECT DISTINCT "creatorID" AS "userID" FROM "comment" WHERE "issueID" = ${issueID}
+  // Get all recipient candidates
```
aboodman, line 266: eppppic. I love how powerful custom mutators turn out to be.

aboodman (approved):
did not review code in detail but looks right structurally and if it works it works

apps/zbugs/server/notify.ts
```diff
@@ -202,20 +258,80 @@ function clip(s: string) {
...
+    FROM recipient_candidates
+    WHERE email IS NOT NULL
+    AND (
+      -- If issue is public, include all candidates
+      EXISTS (
+        SELECT 1 FROM issue_info WHERE visibility = 'public'
+      )
+      OR
+      -- If issue is not public, only include crew members
+      role = 'crew'
```
tantaman, line 333: I'd write a test for this query just to be sure we don't end up emailing people on internal issues.

apps/zbugs/server/notify.test.ts
```diff
@@ -0,0 +1,210 @@
+import {describe, it, expect} from 'vitest';
+import {type Schema} from '../shared/schema.ts';
+import {gatherRecipients} from './notify.ts';
+import type {ServerTransaction} from '@rocicorp/zero';
+import type postgres from 'postgres';
+
+// Mock database state
+const mockDB = {
```
tantaman, line 8: I was thinking something to test the SQL statement itself since that's where the core logic is that gathers recipients.

We have some `.pg-test.ts` files which will spin up and seed a postgres DB that you can hit.

So I'd:

1. Create a `email.pg-test.ts`
2. Seed with some data
3. Check that the big SQL statement `const recipientRows = await sql` returns what you expect for the `internal` issue case.

tantaman: example .pg-test.ts: https://github.com/rocicorp/mono/blob/51995101d0657519207f1c4695a8765b9016e07c/packages/zero-cache/src/services/change-source/pg/change-source.pg-test.ts#L56-L68

apps/zbugs/server/notify.pg-test.ts
```diff
@@ -0,0 +1,159 @@
...
+      -- Insert test data
+      INSERT INTO "user" (id, email, role) VALUES
+        ('user1', 'user1@example.com', 'user'),
+        ('user2', 'user2@example.com', 'user'),
+        ('user3', 'user3@example.com', 'crew'),
+        ('user4', 'user4@example.com', 'user');
+
+      INSERT INTO "issue" (id, "creatorID", "assigneeID", visibility) VALUES
+        ('issue-123', 'user1', 'user2', 'public'),
+        ('issue-456', 'user3', 'user4', 'private');
```
tantaman, line 57: I think we use `internal` and not `private` for visibility

tantaman (approved):

PR #4406, opened 2025-05-20
chore(zero-cache): revert split-health check behavior

darkgnotic:
Revert the behavior of announcing health at `/` early.

This was previously needed to allow the `replication-manager` to have an arbitrarily long startup time even if not monitored by an elastic load balancer.

The AWS docs for [`healthCheckGracePeriodSeconds`](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service_definition_parameters.html#sd-networkconfiguration) have since been updated to indicate that it works for container health checks as well as ELB health checks.

> **healthCheckGracePeriodSeconds**
>
> The period of time, in seconds, that the Amazon ECS service scheduler ignores unhealthy Elastic Load Balancing, VPC Lattice, and container health checks after a task has first started. If you do not specify a health check grace period value, the default value of 0 is used. If you do not use any of the health checks, then healthCheckGracePeriodSeconds is unused.
>
> If your service's tasks take a while to start and respond, you can specify a health check grace period of up to 2,147,483,647 seconds (about 69 years). During that time, the Amazon ECS service scheduler ignores the health check status. This grace period can prevent the service scheduler from marking tasks as unhealthy and stopping them before they have time to come up.

Thus it is no longer necessary to special-case container health checks.

### DIFF
--- a/packages/zero-cache/src/server/change-streamer.ts
+++ b/packages/zero-cache/src/server/change-streamer.ts
@@ -51,24 +51,6 @@ export default async function runWorker(
   });
   void warmupConnections(lc, changeDB, 'change');
 
-  // Start the web server immediately to respond to container health checks
-  // on 4849. This prevents the container from being considered unhealthy if
-  // initial sync takes a long time (the max container grace period is only
-  // 5 minutes), and since the server explicitly advertises readiness via the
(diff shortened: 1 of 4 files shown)


arv (commented):
LGTM

PR #4412, opened 2025-05-21
feat(zql)!: throw on order-by and limit application to junction relationships

tantaman:
ZQL cannot do:
```ts
z.issue.related('labels', q => q.limit(n))
// or 
z.issue.related('labels', q => q.orderBy('name', 'asc'))
```

So throw if someone tries to do either one of those.

Since there is a junction table in the middle items will come out in the order of the junction table rather than the order of the final table. The limit is currently applied by ZQL to the final join which is incorrect in the label case. The limit would need to be in the middle join. To be correct in all cases, the limit needs to be after both joins.

I believe @grgbkr has a plan to implement these.

cc @aboodman since this is more loudly broken than before.


aboodman:
thanks

PR #4417, opened 2025-05-21
fix(z2s): handle limit = 0

tantaman:
Found by the fuzzer. Seed: -681315740

### DIFF
--- a/packages/z2s/src/compiler.ts
+++ b/packages/z2s/src/compiler.ts
@@ -123,7 +123,7 @@ function select(
 }
 
 export function limit(limit: number | undefined): SQLQuery {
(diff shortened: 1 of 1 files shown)


aboodman:
yay!

packages/z2s/src/compiler.ts
```diff
@@ -123,7 +123,7 @@ function select(
 }
 
 export function limit(limit: number | undefined): SQLQuery {
-  if (!limit) {
+  if (limit === undefined) {
```
aboodman, line 126: honest question: do we care about unit testing every corner with this fuzzing technology?

PR #4421, opened 2025-05-22
fix(zql)!: throw if a relationship name matches a column name

tantaman:
If a relationship name matches a column name and we try to sort by that column, we end up trying to sort on the relationship iterable.

E.g.,

```ts
employee.related('reportsTo').orderBy('reportsTo', 'asc');
```

would fail

Found by the fuzzer.


arv (commented):
LGTM

Very nice. That fuzzer is really doing a good job. We should hire the person that wrote that!

PR #4422, opened 2025-05-22
fix(z2s): handle `limit(0)` against `singular` relationships

tantaman:
the compiler defaulted `singular` to `limit(1)` but `limit(0)` should take precedence here.

Found by the fuzzer.


arv (commented):
LGTM

PR #4434, opened 2025-05-23
chore(zero-cache): improve COPY performance by reducing string copying

darkgnotic:
Since the input is UTF-8, the TSV delimiters can be identified by their ascii codes from the Buffer's byte octets, so skip the initial string encoding.

### DIFF
--- a/packages/zero-cache/src/db/pg-copy.ts
+++ b/packages/zero-cache/src/db/pg-copy.ts
@@ -21,17 +21,15 @@ export class TextTransform extends Transform {
   // eslint-disable-next-line @typescript-eslint/naming-convention
   _transform(
     chunk: Buffer,
-    encoding: BufferEncoding,
+    _encoding: BufferEncoding,
     callback: (e?: Error) => void,
   ) {
     try {
-      const text = chunk.toString(encoding);
-
       let l = 0;
       let r = 0;
 
-      for (; r < text.length; r++) {
-        const ch = text.charCodeAt(r);
+      for (; r < chunk.length; r++) {
+        const ch = chunk[r];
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/db/pg-copy.ts
```diff
@@ -21,17 +21,15 @@ export class TextTransform extends Transform {
   // eslint-disable-next-line @typescript-eslint/naming-convention
   _transform(
     chunk: Buffer,
-    encoding: BufferEncoding,
+    _encoding: BufferEncoding,
     callback: (e?: Error) => void,
   ) {
     try {
-      const text = chunk.toString(encoding);
```
arv, line 28: nice

packages/zero-cache/src/db/pg-copy.ts
```diff
@@ -47,15 +45,15 @@ export class TextTransform extends Transform {
         switch (ch) {
           case 0x5c: // '\'
             // flush segment
-            l < r && (this.#currVal += text.substring(l, r));
+            l < r && (this.#currVal += chunk.subarray(l, r).toString('utf8'));
```
arv, line 48: Idea: Maybe we can use a streaming TextDecoder?

arv (commented):
LGTM

But I do think there is room for improvements here.

packages/zero-cache/src/db/pg-copy.ts
aboodman, line 48: These repeated `toString('utf8')` calls look sus to me. You could try pushing the string segments onto an array and joining. But otoh they aren't stressing the mac.

PR #4439, opened 2025-05-29
chore(zbugs): no forever ttl

tantaman:
This is an anti-pattern. Remove our use of it to discourage use.


tantaman:
can't do this until we switch from deleting queries on connection close

PR #4441, opened 2025-05-29
fix(debug): do not load the prettier config

tantaman:
Formatting the zql with the user's preferred prettier config in the debug tool is not important.

This was causing import errors when the prettier config was cjs.

https://discord.com/channels/830183651022471199/1376159734255714336/1376160190017306735


arv:
Ok. I think it is fine to remove this. It was just an extra polish.

I was hoping we could maybe keep it under some circumstances but I don't think we need to go out of our way to make this work.

PR #4446, opened 2025-06-03
chore(zero-cache): Add /heapz endpoint for getting a heapdump from node.

aboodman:
Is protected by same admin password as statz.


aboodman:
Argh this is probably useless because it will only capture the snapshot for the main isolate. Right @darkgnotic ?

PR #4447, opened 2025-06-03
chore(zero-cache): Refactor observability instruments

arv:
Structuring things as export default object prevents tree shaking

And with this refactor it becomes clear that we don't use the up-down counters. I'm leaving it in there for now.

### DIFF
--- a/packages/zero-cache/src/observability/counters.ts
+++ b/packages/zero-cache/src/observability/counters.ts
@@ -1,0 +1,55 @@
+import {getMeter} from './view-syncer-instruments.ts';
+
+function createCounter(name: string, options: {description: string}) {
+  return getMeter().createCounter(name, {
+    description: options.description,
+  });
+}
+
+export function replicationEvents() {
+  return createCounter('replication-events', {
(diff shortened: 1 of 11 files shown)


packages/zero-cache/src/observability/counters.ts
```diff
@@ -0,0 +1,55 @@
...
+    description: 'Number of replication events processed',
+  });
+}
+
+export function crudMutations() {
+  return createCounter('crud-mutations', {
+    description: 'Number of CRUD mutations processed',
+  });
+}
+
```
tantaman, line 20: I think this was a mistake I introduced but the counter shouldn't be re-created each time the function is invoked.

Before I made it lazy it was like this:
https://github.com/rocicorp/mono/blob/458349f46fabc7ae8e149f404f2d7a25454227a3/packages/zero-cache/src/observability/view-syncer-instruments.ts#L5-L8

tantaman:
> And with this refactor it becomes clear that we don't use the up-down counters. I'm leaving it in there for now.

I'll add the code to call those

PR #4449, opened 2025-06-03
feat(zero)!: drop `Schema` information from mutator return

tantaman:
Solves "type too deep" errors.

https://discord.com/channels/830183651022471199/1375464099252666460/1375464102620565637

The type of the `tx` argument for `Mutators` does not need to be exposed to callers (it is only called internally by Zero). Drop it.

This does now require custom mutator definitions to explicitly type their transaction argument. Done here via `MutatorTx` in `zbugs`.

### DIFF
--- a/apps/zbugs/server/server-mutators.ts
+++ b/apps/zbugs/server/server-mutators.ts
@@ -3,8 +3,9 @@ import {
   type CreateIssueArgs,
   type AddEmojiArgs,
   type AddCommentArgs,
+  type MutatorTx,
 } from '../shared/mutators.ts';
-import {type CustomMutatorDefs, type UpdateValue} from '@rocicorp/zero';
+import {type UpdateValue} from '@rocicorp/zero';
 import {schema} from '../shared/schema.ts';
 import {notify} from './notify.ts';
 import {assert} from '../../../packages/shared/src/asserts.ts';
@@ -24,7 +25,7 @@ export function createServerMutators(
     issue: {
       ...mutators.issue,
 
-      async create(tx, {id, title, description}: CreateIssueArgs) {
+      async create(tx: MutatorTx, {id, title, description}: CreateIssueArgs) {
         await mutators.issue.create(tx, {
           id,
           title,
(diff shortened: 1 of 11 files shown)


apps/zbugs/server/server-mutators.ts
```diff
@@ -194,5 +195,5 @@ export function createServerMutators(
         );
       },
     },
-  } as const satisfies CustomMutatorDefs<typeof schema>;
```
tantaman, line 197: looks like we can remove all the `satisfies` calls now.

tantaman:
will document and add a helper as a workaround

amosbastian:
where is this helper? I'm getting constant problems with this

apps/zbugs/server/server-mutators.ts
arv: The satisfies was never needed. It was there as an extra guard (afaict)

packages/zero-client/src/client/custom.ts
```diff
@@ -32,12 +32,14 @@ import type {ZeroLogContext} from './zero-log-context.ts';
 /**
  * The shape which a user's custom mutator definitions must conform to.
  */
-export type CustomMutatorDefs<S extends Schema> = {
+export type CustomMutatorDefs = {
   [namespaceOrKey: string]:
     | {
-        [key: string]: CustomMutatorImpl<S>;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        [key: string]: CustomMutatorImpl<any>;
```
arv, line 39: can we get rid of the type param now?

arv (commented):
What happened to this PR?

tantaman:
> where is this helper? I'm getting constant problems with this

We were going to wait to break the mutator API once when we updated it to match custom queries but it's been so long we probably should have just shipped this.

tantaman:
but this is finally coming to the next release with:

- #4699 
- #4700 

and future PRs

tantaman:
I do think our hesitance to break APIs while in alpha is a net negative for users.

tantaman:
After exploring the other API shape (linked in comment above) I think the hole it opens to allow users to make mistakes is too great. Resurrecting this PR.

PR #4454, opened 2025-06-04
perf(zql): avoid initiating fetch of permission relationships in pipeline driver

grgbkr:
Existing code avoided iterating the resulting Iterable, but this optimization avoids sending the query to sqlite entirely.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
+++ b/packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
@@ -571,7 +571,9 @@ class Streamer {
       switch (type) {
         case 'add':
         case 'remove': {
-          yield* this.#streamNodes(queryHash, schema, type, [change.node]);
+          yield* this.#streamNodes(queryHash, schema, type, () => [
+            change.node,
+          ]);
           break;
         }
         case 'child': {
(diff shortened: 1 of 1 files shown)


packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
```diff
@@ -571,7 +571,9 @@ class Streamer {
       switch (type) {
         case 'add':
         case 'remove': {
-          yield* this.#streamNodes(queryHash, schema, type, [change.node]);
```
aboodman, line 574: Are you sure this is even necessary? I haven't looked at impl of `streamNodes` but in similar code I remember realizing/proving to myself that the sql execute never even happens if nobody ever iterates the first time.

aboodman: That's why, for example, we just went ahead and setup the child relationships for each row "eagerly". Because we knew the sql statement to actually get them would never happen unless some downstream operator needed them.

tantaman: This might avoid the work of creating the query plan since we do not prepare the statement.

@grgbkr - was there a measurable difference with this change?

tantaman: I checked in some example "notebook style" analyses that explore perf if you need to gather metrics on this.

https://github.com/rocicorp/mono/blob/57a9aa52674e6e12aeb3f84007bf1e0fa0fbf997/packages/zql-benchmarks/src/frontend-analysis.pg-test.ts#L31-L40

PR #4456, opened 2025-06-04
chore(zero-cache): cache query transformation requests made to the API server

tantaman:


packages/shared/src/cache.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export class TimedCache<T> {
+  readonly #cache: Map<string, {value: T; expiresAt: number}>;
+  readonly #ttlMs: number;
+  readonly #intervalHandle: NodeJS.Timeout;
+
+  constructor(ttlMs: number) {
+    this.#cache = new Map();
+    this.#ttlMs = ttlMs;
+    this.#intervalHandle = setInterval(() => {
```
arv, line 16: Why are we doing this at an interval? We know when there is an entry to expire and we know when that entry should be collected.

You can use a single setTimeout to the smallest time.

For zero-cache I ended up overshooting slightly to reduce DB churn but I'm not sure that is a better design here.

PR #4457, opened 2025-06-04
chore(zero): add custom queries to the `queries-patch` protocol

tantaman:
@darkgnotic - please see the question in https://github.com/rocicorp/mono/commit/0f27ba8ddd71a92fb5ea843650355548ea20e43a

I'm getting test failures (for the new `oneCustomHash` case) because we sort `desiredQueryIDs` in the CvrConfigDriveUpdater but we do not sort them when loading from the DB.


tantaman:
Since custom queries are not yet enabled and I've got a growing stack of changes on top of this, I changed the custom query hash (in the cvr test) to sort correctly on load and update.

We'll need to come back to this question before turning on custom queries.

grgbkr:
> Since custom queries are not yet enabled and I've got a growing stack of changes on top of this, I changed the custom query hash (in the cvr test) to sort correctly on load and update.
> 
> We'll need to come back to this question before turning on custom queries.

I can't see any reason they need to be sorted.  But @darkgnotic or @arv may know something I don't.

grgbkr:
LGTM

packages/zero-cache/src/services/view-syncer/schema/cvr.ts
```diff
@@ -98,7 +99,7 @@ export type QueriesRow = {
   // This is the client AST _AFTER_ applying server name transformations.
   clientAST: JSONValue | null;
   queryName: string | null;
-  queryArgs: JSONValue | null;
+  queryArgs: readonly ReadonlyJSONValue[] | null;
```
arv, line 102: 🎂

packages/zero-cache/src/services/view-syncer/cvr.ts
```diff
@@ -270,18 +280,18 @@ export class CVRConfigDrivenUpdater extends CVRUpdater {
       return patches;
     }
     const newVersion = this._ensureNewVersion();
+    // why do we sort here?
```
arv, line 283: I cannot recall why this was done?

packages/zero-cache/src/services/view-syncer/cvr.ts
```diff
@@ -943,3 +953,35 @@ export function nextEvictionTime(cvr: CVR): number | undefined {
 function normalizeTTL(ttl: number): number {
   return ttl < 0 ? -1 : ttl;
 }
+
+function newQueryRecord(
+  id: string,
+  ast: AST | undefined,
+  name: string | undefined,
+  args: readonly ReadonlyJSONValue[] | undefined,
+): ClientQueryRecord | CustomQueryRecord {
```
arv, line 962: Now this is polymorphic. Maybe keep the properties and set them to undefined?

packages/zero-cache/src/services/view-syncer/cvr.ts
tantaman: oops, did not mean to commit the question. But yeah, still waiting on @darkgnotic to shed some light here.

packages/zero-cache/src/services/view-syncer/cvr.ts
tantaman: As in keep the properties in both types? So `CustomQueryRecord` will have an `ast` that can only be undefined?

arv: That was what I was thinking.

I'm worried about polymorphic code in hot code paths. Maybe it is a bit premature.

PR #4459, opened 2025-06-04
chore(zero-client): update `query-manager` to support named/custom queries

tantaman:


packages/zero-protocol/src/query-hash.ts
```diff
@@ -13,3 +13,11 @@ export function hashOfAST(ast: AST): string {
   hashCache.set(normalized, hash);
   return hash;
 }
+
+export function hashOfNameAndArgs(
+  name: string,
+  args: readonly unknown[],
```
grgbkr, line 19: can we not type args more specifically?

packages/zero-client/src/client/query-manager.ts
```diff
@@ -17,14 +20,19 @@ import {
...
+import type {ReadonlyJSONValue} from '../../../shared/src/json.ts';
 import {desiredQueriesPrefixForClient, GOT_QUERIES_KEY_PREFIX} from './keys.ts';
 import type {MutationTracker} from './mutation-tracker.ts';
 import type {ReadTransaction} from './replicache-types.ts';
 
 type QueryHash = string;
 
 type Entry = {
-  normalized: AST;
+  // May be undefined if the entry is a custom (named) query.
```
grgbkr, line 31: I think a discriminated union type would be better here (then you wouldn't need the comments in various places about fields being undefined for custom vs adhoc, it would be enforced by the type), but don't feel strongly about it since the 'ad-hoc' query type is going to go away.

packages/zero-client/src/client/query-manager.test.ts
```diff
@@ -66,10 +66,52 @@ test('add', () => {
...
+  const rm2 = queryManager.addCustom('customQuery', [1], 'forever');
+  expect(send).toBeCalledTimes(1);
+
+  rm2();
+  const rm3 = queryManager.addCustom('customQuery', [1], 'forever');
+  expect(send).toBeCalledTimes(1);
+  rm1();
+  rm3();
+  queryManager.addCustom('customQuery', [1], 'forever');
+  // once for del, another for put
```
grgbkr, line 111: Might be a bit better to assert the send for del before calling addCustom again and asserting the contents of what was sent.

packages/zero-client/src/client/query-manager.test.ts
```diff
@@ -66,10 +66,52 @@ test('add', () => {
     },
   ]);
 
-  queryManager.add(ast, 'forever');
+  queryManager.addLegacy(ast, 'forever');
   expect(send).toBeCalledTimes(1);
 });
 
+test('add and remove a custom query', () => {
```
grgbkr, line 73: We now have a lot more test coverage for addLegacy than for addCustom (covering behavior like gotCallback, recent query cache, etc). addLegacy and addCustom share logic for most of this stuff.  But in the long run addLegacy is going away. Is the thought to update these tests to use addCustom at the time we remove addLegacy?

grgbkr (commented):
LGTM

packages/zero-protocol/src/query-hash.ts
tantaman: good catch. Yes, it is `readonly ReadOnlyJSONValue[]` everywhere else.

packages/zero-client/src/client/query-manager.test.ts
tantaman: Let me see about sharing the same test logic for both paths.

PR #4462, opened 2025-06-05
chore(zql): named/custom query wrapper

tantaman:
```ts
const q = query.bindTo(schema);

q('issue', (tx, id: string) => tx.issue.where('id', id));

// or

query(schema, 'issue', (tx, id: string) => tx.issue.where('id', id));
```

The API is expected to evolve as I start using it in zbugs and other places.

### DIFF
--- a/packages/zql/src/query/named.test.ts
+++ b/packages/zql/src/query/named.test.ts
@@ -1,0 +1,52 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import {expect, expectTypeOf, test} from 'vitest';
+import {schema} from './test/test-schemas.ts';
+import {query, type NamedQuery} from './named.ts';
+import {StaticQuery} from './static-query.ts';
+import {ast, defaultFormat} from './query-impl.ts';
+
+test('defining a named query', () => {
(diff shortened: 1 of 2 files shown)


packages/zql/src/query/named.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  const named = bound('issue', (tx, id: string) => tx.issue.where('id', id));
+
+  expectTypeOf(named).toEqualTypeOf<NamedQuery<typeof schema, [string]>>();
+  check(named);
+});
+
+function check(named: NamedQuery<typeof schema, [string]>) {
+  const r = named(
+    {
+      issue: new StaticQuery(schema, 'issue', {table: 'issue'}, defaultFormat),
```
arv, line 28: The name here ('issue') can be anything?

I tested it myself and figured it out.

arv (commented):
LGTM

packages/zql/src/query/named.test.ts
tantaman: It can't. I think I should remove the parameter since it will always match the name of the table at the root of the ast.

PR #4464, opened 2025-06-05
chore(zql): add query name and args to builder interface

tantaman:
A custom query will have `name` and `args` filled in in addition to the AST.

- the AST is used for client execution
- name and args are sent to the server

The first commit in this PR is only kept around for historical interest / documentation of another approach.


grgbkr (commented):
LGTM

arv (commented):
LGTM

PR #4470, opened 2025-06-06
chore(zero-cache): optionally profile initial copy

darkgnotic:
Add an option to capture a cpu profile during the initial copy. Enable it for zbugs.


arv (commented):
LGTM

PR #4474, opened 2025-06-07
refactor(zero-solid): SolidView reconciles large changes to data

devagrawal09:
I noticed that the zero-solid integration was recreating the store with every change to the query, and directly setting new data when transaction completes.
Both these operations cause everything depending on the store to be disposed and recreated, instead of receiving granular updates.
These changes ensure that every `createQuery` call only creates a store the first time and passes on the setter to the view, and the view is responsible for reconciling data when there's large changes.

**Update:**
Additional changes to support the missing hash based caching

- Removed `createZero` in favor of `ZeroProvider` and `useZero` so that the zero instance can be accessed through context
- `useZero` return a signal of zero instance since the options passed to `new Zero()` can change which will recreate the zero client
- Un-deprecate `useQuery` since it now requires context. We should consider deprecating `createQuery` instead. `useX` is a good convention to communicate that this should be called inside a component and not outside.
- Added `createUseZero` to prevent redundant annotations, and so the api mirror's React's apis much more closely.

Docs PR for these changes coming soon

### DIFF
--- a/.github/workflows/sst-gigabugs-deploy.yml
+++ b/.github/workflows/sst-gigabugs-deploy.yml
@@ -2,7 +2,7 @@ name: Build and Deploy Gigabugs Zero-Cache AWS SST
 
 on:
   push:
-    branches: [darkgnotic/multipart-table-sync] # [gigabugs]
+    branches: [gigabugs]
 env:
   # Setting an environment variable with the value of a configuration variable
   ECR_IMAGE_ZERO_CACHE: zero-zgigabugs
--- a/apps/zbugs/api/index.ts
+++ b/apps/zbugs/api/index.ts
@@ -18,7 +18,8 @@ import {must} from '../../../packages/shared/src/must.ts';
 import * as v from '../../../packages/shared/src/valita.ts';
 import {handlePush} from '../server/push-handler.ts';
 import {authDataSchema, type AuthData} from '../shared/auth.ts';
-import {queries} from '../shared/schema.ts';
+import {getQuery} from '../server/get-query.ts';
+import type {ServerContext} from '../server/server-queries.ts';
 
 declare module 'fastify' {
   interface FastifyInstance {
@@ -142,13 +143,13 @@ fastify.post<{
   reply.send(response);
 });
 
-const emptyQuery = queries.issue.where(({or}) => or());
 fastify.post<{
   Querystring: Record<string, string>;
   Body: ReadonlyJSONValue;
 }>('/api/pull', async (request, reply) => {
+  let authData;
   try {
-    await maybeVerifyAuth(request.headers);
+    authData = await maybeVerifyAuth(request.headers);
   } catch (e) {
     if (e instanceof Error) {
       reply.status(401).send(e.message);
@@ -157,14 +158,23 @@ fastify.post<{
     throw e;
   }
 
+  const context: ServerContext = {
+    role: authData?.role,
+  };
   const transformRequest = v.parse(request.body, transformRequestMessageSchema);
   const response: TransformResponseMessage = [
     'transformed',
-    transformRequest[1].map(req => ({
-      id: req.id,
-      name: req.name,
-      ast: emptyQuery.ast,
-    })),
+    transformRequest[1].map(req => {
+      const query = getQuery(context, req.name, req.args);
+
+      const ret = {
+        id: req.id,
+        name: req.name,
+        ast: query.ast,
+      };
+
+      return ret;
+    }),
   ];
   reply.send(response);
 });
--- a/apps/zbugs/package.json
+++ b/apps/zbugs/package.json
@@ -19,7 +19,7 @@
     "run-query": "tsx ../../packages/zero-cache/src/scripts/run-query.ts",
     "run-query-brk": "tsx --inspect-brk ../../packages/zero-cache/src/scripts/run-query.ts",
     "create-keys": "tsx ../../packages/zero-cache/src/scripts/create-jwk.ts",
-    "test": "vitest run --config vitest.config.ts --project zbugs/no-pg",
+    "test": "cd ../../packages/zero && npm run build && cd ../../apps/zbugs && vitest run --config vitest.config.ts --project zbugs/no-pg",
     "test:watch": "vitest --config vitest.config.ts --project zbugs/no-pg",
     "analyze": "analyze -c vite.config.ts",
     "test:pg": "vitest run --config vitest.config.ts --project zbugs/pg-16",
--- a/apps/zbugs/server/get-query-test.ts
+++ b/apps/zbugs/server/get-query-test.ts
@@ -1,0 +1,74 @@
+import {describe, expect, vi, test} from 'vitest';
+import * as serverQueries from '../server/server-queries.ts';
+import * as sharedQueries from '../shared/queries.ts';
+import {getQuery} from './get-query.ts';
+import * as readPermissions from './read-permissions.ts';
+import type {NamedQueryImpl} from '@rocicorp/zero';
+
+vi.mock('./read-permissions', () => ({
+  applyIssuePermissions: vi.fn(),
+}));
+
+const context = {} as serverQueries.ServerContext;
+
+describe('permissions are applied for our sensitive queries', () => {
+  const queries = Object.keys(serverQueries).map(name => {
+    const args = mockArgs[name as keyof typeof mockArgs];
+    if (!args) {
+      throw new Error(`No mock args for query: ${name}`);
+    }
+    return [name, args] as const;
+  });
+
+  test.each(queries)('calls applyPermissions for %s', (name, args) => {
+    const permissionsSpy = vi.spyOn(readPermissions, 'applyIssuePermissions');
+    getQuery(context, name, args);
+    expect(permissionsSpy).toHaveBeenCalledOnce();
+  });
+});
+
+test('export names match between server and shared queries', () => {
+  const serverQueryNames = Object.keys(serverQueries);
+  const sharedQueryNames = Object.keys(sharedQueries);
+
+  const sharedNames = new Set(sharedQueryNames);
+  for (const name of serverQueryNames) {
+    expect(sharedNames.has(name)).toBe(true);
+  }
+});
+
+test('query name matches const name', () => {
+  expect(Object.keys(sharedQueries)).toEqual(Object.keys(mockArgs));
+  for (const [key, value] of Object.entries(mockArgs)) {
+    const queryFn = sharedQueries[key as keyof typeof sharedQueries];
+    const query = (queryFn as NamedQueryImpl)(...value);
+    // query names and their exported names should match in zbugs
+    expect(query.customQueryID?.name).toBe(key);
+    expect(query.customQueryID?.args).toEqual(value);
+  }
+});
+
+const mockArgs = {
+  allLabels: [],
+  allUsers: [],
+  issuePreload: ['userID'],
+  user: ['userID'],
+  userPref: ['key', 'userID'],
+  userPicker: [false, null, 'creators'],
+  issueDetail: ['id', '1', 'userID'],
+  prevNext: [null, null, 'next'],
+  issueList: [
+    {
+      open: null,
+      assignee: null,
+      creator: null,
+      labels: null,
+      textFilter: null,
+      sortField: 'created',
+      sortDirection: 'asc',
+    },
+    'userID',
+    10,
+  ],
+  emojiChange: ['subjectID'],
+};
--- a/apps/zbugs/server/get-query.ts
+++ b/apps/zbugs/server/get-query.ts
@@ -1,0 +1,31 @@
+import type {NamedQueryImpl, ReadonlyJSONValue} from '@rocicorp/zero';
+import * as serverQueries from '../server/server-queries.ts';
+import * as sharedQueries from '../shared/queries.ts';
+
+export function getQuery(
+  context: serverQueries.ServerContext,
+  name: string,
+  args: readonly ReadonlyJSONValue[],
+) {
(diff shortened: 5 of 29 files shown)


thdxr:
i tested this on my app with the profiler and can confirm it works

aboodman:
@grgbkr can you review this? @devagrawal09 - very happy to have you hear. Greg spent a large amount of time trying to get the solid integration to work and IIRC compromises were made.

devagrawal09:
Pushed another update that adds hash based caching and context dependent apis

packages/zero-solid/src/use-zero.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export function createUseZero<
+  S extends Schema,
+  MD extends CustomMutatorDefs<S> | undefined = undefined,
+>() {
+  return () => useZero<S, MD>();
+}
+
+export function ZeroProvider<
```
arv, line 46: The React one takes a Zero object as a prop. I can see how this is better because it deals with not having to use createZero.

packages/zero-solid/src/mod.ts
```diff
@@ -4,4 +4,4 @@ export {
   type CreateQueryOptions,
   type UseQueryOptions,
 } from './create-query.ts';
-export {createZero} from './create-zero.ts';
+export {useZero, ZeroProvider} from './use-zero.tsx';
```
arv: Unfortunately we need to keep createZero for at least one release

packages/zero-solid/src/create-query.ts
```diff
@@ -38,25 +38,25 @@ export function createQuery<
...
+  const [state, setState] = createStore<State>([
+    {
+      '': undefined,
+    },
+    unknown,
+  ]);
+
+  const z = useZero();
+
   // Wrap in in createMemo to ensure a new view is created if the querySignal changes.
```
arv: update comment

packages/zero-solid/src/create-query.ts
```diff
@@ -68,9 +68,9 @@ export function useQuery<
   return createQuery(querySignal, options);
 }
 
-type UnknownSolidView = SolidView<HumanReadable<unknown>>;
+type UnknownSolidView = SolidView;
```
arv: This can be removed now

packages/zero-solid/src/create-query.ts
```diff
@@ -38,25 +38,25 @@ export function createQuery<
...
+  createComputed(() => {
     const query = querySignal();
     const ttl = normalize(options)?.ttl ?? DEFAULT_TTL;
-    const view = getView(query, ttl);
-
-    // Use queueMicrotask to allow cleanup/create in the current microtask to
-    // reuse the view.
-    onCleanup(() => queueMicrotask(() => releaseView(query, view)));
-    return view;
+    getView(z().clientID, query, ttl, setState);
```
arv: The name getView is kind of misleading now since we no longer care about the return value. It would be better to rename getView and make it return void... I'm not sure what to name it... maybe useView?

packages/zero-solid/src/solid-view.ts
```diff
@@ -73,29 +67,23 @@ export class SolidView<V> implements Output {
...
+    );
+
     if (isEmptyRoot(initialRoot)) {
       this.#builderRoot = this.#createEmptyRoot();
     }
 
     if (queryComplete !== true) {
       void queryComplete.then(() => {
-        this.#setState(oldState => [oldState[0], complete]);
+        this.#setState(1, complete);
```
arv, line 80: nice

arv (commented):
Very nice

devagrawal09:
thanks @arv, will make those changes today

packages/zero-solid/src/solid-view.ts
```diff
@@ -104,7 +92,7 @@ export class SolidView<V> implements Output {
     const builderRoot = this.#builderRoot;
     if (builderRoot) {
       if (!isEmptyRoot(builderRoot)) {
-        this.#setState(oldState => [builderRoot, oldState[1]]);
+        this.#setState(0, reconcile(builderRoot));
```
grgbkr: `reconcile` is a challenging to use with Zero's view data model.  I don't think this is currently achieving what you want.

**Issue 1 keys**
`reconcile` wants all objects in arrays to be key-able by the same key (by default this key is 'id').  Zero's view data model is a tree of rows, which can have compound primary keys, and different primary keys at different levels of the tree (since the rows are from different tables).  

To address this either:
1. `null` should be passed for [`ReconcileOptions.key`](https://github.com/solidjs/solid/blob/3d3207dd3aeb84c2a38377cf9f3b895995c2d969/packages/solid/store/src/modifiers.ts#L6).  However, this will cause `reconcile`'s diffing mechanism to not work well in any case where rows change indexes.  
2. Create synthetic keys on all rows.  The combination of the primary key(s) for a rows could be combined into a single key value set a consistent key, say `__solidID`, which would be passed for [`ReconcileOptions.key`].  There is some overhead to creating these synthetic keys but the reconcile behavior would be much better.

**Issue 2 merge**
`reconcile`'s default behavior is [`merge: false`, which uses referential equality to determine if an object should be replaced](https://docs.solidjs.com/reference/store-utilities/reconcile#options).  Since builderRoot is not going to be referentially equal to the previous view's root Entry, this is always going to replace it, which will essentially have the same behavior as the old setState with no reconcile.  

`ReconcileOptions.merge` `true` seems to be what we want.

devagrawal09: Thanks for pointing these issues out.

> Create synthetic keys on all rows.

I think this is fair, and while this is something you don't need to do in the React version, it's something the developer needs to do themselves when consuming data in components. React also needs a key for reconciling the virtual DOM, and I'm curious how you approach this in zbugs. It's possible that stable synthetic keys would not only benefit the non-React bindings, but also benefit the React developers so that you don't have to come up with these synthetic keys for React yourself, you could just use the Zero provided keys.

> ReconcileOptions.merge true seems to be what we want.

You're right, I should make that change, but this really only works well if there's a good key to match with.

devagrawal09: > There is some overhead to creating these synthetic keys

What kind of overhead? If it's purely to compute the keys, you can do that lazily, i.e., inside a property getter so that they're only computed when asked for

aboodman: > Create synthetic keys on all rows. The combination of the primary key(s) for a rows could be combined into a single key value set a consistent key, say __solidID, which would be passed for [ReconcileOptions.key]. There is some overhead to creating these synthetic keys but the reconcile behavior would be much better.

Couple questions:

- What determines "large"? Is it just when the query changes or is there some other heuristic for determining when to call reconcile?
- Changes to a query can easily produce completely different data with no shared entities at all. Does it really make sense to reconcile whenever query changes? will it be efficient when there is nothing in common between the two trees?
- Does SolidJS require that every entity throughout the tree have a unique ID, even across levels (i.e., to detect moves) or only at the same level
- Could SolidJS be changed to accept an array of keys? This would be very convenient since this if the form that Zero keeps its concept of unique ID in.

packages/zero-solid/src/solid-view.ts
```diff
@@ -73,29 +67,26 @@ export class SolidView<V> implements Output {
...
-    [this.#state, this.#setState] = createStore<State>([
-      initialRoot,
-      queryComplete === true ? complete : unknown,
-    ]);
+
+    this.#setState = setState;
+    this.#setState(
+      reconcile([initialRoot, queryComplete === true ? complete : unknown], {
+        merge: true,
+        key: '__uniqueID',
```
arv: Can we use a symbol here?

arv: Where is __uniqeID added to the rows?

packages/zero-solid/src/solid-view.ts
devagrawal09: - There's really only two ways the view updates. Either the data changed or the query. If the data changed Zero provides the exact change that we can `produce`. If the query changes the only option is to `reconcile`.
- Reconcile falls back to just replacing the objects when the keys don't match so there's no downside of using it. Worst case it's the same as not using reconcile. Best case it moves around some data.
- Only at the same level.
- So an array of all possible properties that could be keys? I'm guessing the array would be in order of preference?

packages/zero-solid/src/solid-view.ts
devagrawal09: Yes a symbol should work as well, as long as its exported by Zero. Typescript might not like it though. 
We are expecting the unique id to be added to the data by Zero.

grgbkr: I think a Symbol is preferable here as it can't possibly collide with user data ( __uniqueID is a valid name for a postgres column).  Solid's typescript types expect a 'string', but the code looks like it will work just fine with a Symbol key, so I think its fine to just do a cast to get it through the TypeScript check (perhaps Solid's typescript type could be adjusted here).  

@devagrawal09 I also don't see where '__uniqueID' is populated.  Is your thinking that Zero should populate this field (rather than solid-view populating it)?  I know in the other thread you mentioned that it could be useful in React as a `key` when rendering lists.  I think this could be reasonable we could add it as a lazy memoized getter at a well known Symbol in the same place we add ref counts in view-apply-change https://github.com/rocicorp/mono/blob/55132ee1b4d0449cd0220350ac9cdffa2cda57cf/packages/zql/src/ivm/view-apply-change.ts#L415.
```
function makeNewEntryWithRefCountAndKey(row: Row, rc: number, primaryKey: PrimaryKey): RCEntry {
  const entry = {...row, [refCountSymbol]: rc};
  Object.defineProperty(entry, keyGetterSymbol, {
    get() {
      if (this[keySymbol] === undefined) {
        this[keySymbol] = computeKey(this, primaryKey);
      }
      return this[keySymbol];
    },
  });
  return entry;
}
```
@arv what do you think of that?

arv: @grgbkr I was thinking the same, except for...

@devagrawal09 An other option would be to use a function that computes the key that reconcile uses. Having to put things at a known property works for us as well but we would do this unconditionally on all row objects I think. Installing a getter is not cheaper than adding a data property, but the main problem is that we do not want to lazily add anything to the row objects because then they shape (as in V8 hidden class) would change which will lead to polymorhic or megamorphic code.

An array like @aboodman suggested would not work since different levels of the tree may have different primary keys.

arv:
@devagrawal09 How do you feel about the status of this?

To make progress I suggest a few things.

1. Add back `useQuery` and deprecate `createQuery`. Sorry.
2. Let's add the `ZeroProvider` and `useZero` and deprecate `createZero`
3. Add `createUseZero`
4. Change `createMemo` to `computed`
5. Add the hash memoization to view. I think this might need some care so that old code works for at least one major release.
6. Last thing would be to do the reconciliation in a different PR so we can discuss that in isolation.

Splitting things up in multiple PRs would make things easier but it does not have to be 6 PRs.

Let me know if you want us to take over any of this work?

@grgbkr We could also just make this a breaking change for solid and remove all the deprecated stuff. WDYT?

devagrawal09:
@arv i like that direction, we can deal with keyed reconciliation logic in a separate PR. For now I can get rid of the reconcile and simply replace the store data with the new query data, and recommend keeping the queries static if people run into performance issues due to the lack of proper reconciliation (this is the solution @thdxr ended up going with).

packages/zero-solid/src/use-zero.tsx
```diff
@@ -0,0 +1,60 @@
...
+  S extends Schema,
+  MD extends CustomMutatorDefs<S> | undefined = undefined,
+>(): () => Zero<S, MD> | undefined {
+  const zero = useContext(ZeroContext);
+
+  // TODO: Remove this once we have a way to ensure that useZero is used within a ZeroProvider.
+  // if (zero === undefined) {
+  //   throw new Error('useZero must be used within a ZeroProvider');
+  // }
+  return zero ?? (() => undefined);
```
devagrawal09, line 36: temporarily removing the undefined check so that `createQuery` works without the provider

packages/zero-solid/src/create-query.ts
```diff
@@ -38,25 +38,25 @@ export function createQuery<
   querySignal: () => Query<TSchema, TTable, TReturn>,
   options?: CreateQueryOptions | Accessor<CreateQueryOptions>,
 ): QueryResult<TReturn> {
-  // Wrap in in createMemo to ensure a new view is created if the querySignal changes.
-  const view = createMemo(() => {
+  const [state, setState] = createStore<State>([
```
arv, line 41: Just to be clear. The old implementation cached the store between different calls to createQuery and shared as long as the query object was the same.

I think this new semantics is fine but I wanted to know the main motivation for this change.

arv (approved):

packages/zero-solid/src/create-query.ts
devagrawal09: Wouldn't a change in dependencies cause an entirely new query object to be generated?
Dax shared a performance profile where we also noticed `createStore` being called on a query change.
The main motivation behind this change was that the query object couldn't be the same since it gets recreated on every dependency change.

arv: The Query object should change on every dependency change since queries are immutable.

The question if the store should change is orthogonal. I do buy the argument that the store should not change because that allows the UI to do less updates.

grgbkr:
> Thanks. Can you update the tests too?

@devagrawal09 Just FYI, I'm working on updating the tests for this PR and getting it landed.  Thanks for getting this going in the right direction.

grgbkr:
See https://github.com/rocicorp/mono/pull/4559

PR #4477, opened 2025-06-09
chore(zero): error on no pull url, throw on missing table

tantaman:
Some extra error reporting and invariants.

- if custom queries are received by zero-cache but no pull url is configured, log an errr
- if zero-pg receives a query for a table that is not in the schema, fatal


arv (approved):
LGTM

PR #4479, opened 2025-06-09
chore(zql): add `makeSchemaQuery` for server side query implementations

tantaman:
A user's API server needs a way to create a query AST to send back to Zero. `StaticQuery` and `makeSchemaQuery` allow this.

The general use on the API server would be something like:

```ts

const builders = makeSchemaQuery(schema);
app.post('/api/query', (req) => {
  const respones = [];
  for (query of req.body[1]) {
    responses.push(mutators[query.name](builders, query.args));
  }
});

```


packages/zql/src/query/named.test.ts
```diff
@@ -83,3 +83,25 @@ function check(named: NamedQuery<typeof schema, [string], any>) {
     hashOfAST((r as StaticQuery<typeof schema, 'issue'>).ast),
   );
 }
+
+test('makeSchemaQuery', () => {
+  const builders = makeSchemaQuery(schema);
+  const q1 = builders.issue.where('id', '123').nameAndArgs('myName', ['123']);
```
arv, line 89: What does `nameAndArgs` do? Is that important for this test? Do you need the `'123'` and `'123'` to be same?

packages/zql/src/query/named.ts
```diff
@@ -47,3 +50,34 @@ query.bindTo =
     fn: NamedQueryImpl<S, TArg, TReturnQuery>,
   ): NamedQuery<S, TArg, TReturnQuery> =>
     query(s, name, fn);
+
+/**
+ * This produces the query builders for a given schema.
+ * For use in Zero on the server to process custom queries.
+ */
+export function makeSchemaQuery<S extends Schema>(schema: S): SchemaQuery<S> {
```
arv, line 58: This name is pretty confusing. I think we need to see it in action before coming up with a better name.

How about calling this `makeSchemaQueryNameTBD` for now?

packages/zql/src/query/named.test.ts
tantaman: It pins the query to a given "custom query." So the "name and args" is what is sent to the server rather than the AST. It can be completely divorced from what parameters are passed to the query builder.

Let me work on naming here.

PR #4481, opened 2025-06-10
fix(zero-cache): ref count and clean up `pusher` & `mutagen`

tantaman:
Increments the ref count on the pusher and mutagen services each time a connection is opened.
Decrements the ref count each time a connection is closed for that service.

Services stop themselves when their ref count hits 0 and mark themselves invalid.

ViewSyncer works similarly internally by keeping a map of currently served clients and destroying itself when that hits 0 (and various other conditions are met).

### DIFF
--- a/packages/zero-cache/src/workers/connection.ts
+++ b/packages/zero-cache/src/workers/connection.ts
@@ -250,6 +250,7 @@ export class Connection {
         this.#lc.warn?.(
           `failed to parse close message "${reason}": ${String(e)}`,
         );
+        this.close('WebSocket close event', {code, reason, wasClean});
         return;
       }
 
--- a/packages/zero-cache/src/workers/syncer.test.ts
+++ b/packages/zero-cache/src/workers/syncer.test.ts
@@ -1,0 +1,236 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import {
+  afterAll,
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  test,
+  vi,
+} from 'vitest';
+
+let receiver: WebSocketReceiver<any>;
+vi.mock('../types/websocket-handoff.ts', () => ({
+  installWebSocketReceiver: vi
+    .fn()
+    .mockImplementation((_lc, _server, receive, _sender) => {
+      receiver = receive;
+    }),
+}));
+
+import {Syncer} from './syncer.ts';
+import {createSilentLogContext} from '../../../shared/src/logging-test-utils.ts';
+import type {ZeroConfig} from '../config/zero-config.ts';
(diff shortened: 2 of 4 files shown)


packages/zero-cache/src/workers/connection.ts
```diff
@@ -250,6 +250,7 @@ export class Connection {
         this.#lc.warn?.(
           `failed to parse close message "${reason}": ${String(e)}`,
         );
+        this.close('WebSocket close event', {code, reason, wasClean});
```
darkgnotic, line 253: Suggestion: maybe slightly easier to follow if we instead put the final `this.close()` call in a `finally` clause.

packages/zero-cache/src/workers/syncer.test.ts
```diff
@@ -0,0 +1,236 @@
...
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  test,
+  vi,
+} from 'vitest';
+
+let receiver: WebSocketReceiver<any>;
+vi.mock('../types/websocket-handoff.ts', () => ({
```
darkgnotic, line 13: Fancy!  (and thank you for adding this test)

packages/zero-cache/src/workers/syncer.ts
```diff
@@ -75,9 +75,9 @@ export class Syncer implements SingletonService {
       id => viewSyncerFactory(id, notifier.subscribe(), this.#drainCoordinator),
       v => v.keepalive(),
     );
-    this.#mutagens = new ServiceRunner(lc, mutagenFactory);
+    this.#mutagens = new ServiceRunner(lc, mutagenFactory, m => m.hasRefs());
```
darkgnotic, line 78: This is a nice tie-in with the existing `isValid` logic in the ViewSyncer. I'm wondering whether the ViewSyncer can be simplified to use this as well, as opposed to the keepalive method currently used to mitigate the delay between when the object is looked up and when it starts bookkeeping the associated connection. It may be more complicated for that class because it hold so much state, both in terms of cached CVR data and queued lock-holding requests. 

Anyway, this is mostly rambling.  😁

darkgnotic (approved):
Nicely done!

packages/zero-cache/src/workers/syncer.ts
tantaman: yes, that would be nice. Let me add another task to my list to visit after custom mutators.

packages/zero-cache/src/workers/syncer.test.ts
tantaman: its a lot of mocking in the test 😬 but seems worth it to test the cleanup case.

PR #4482, opened 2025-06-10
chore(zql): allow static queries to be converted to runnable queries

tantaman:
This will allow building queries without first having access to a Zero instance.

E.g., a user can define queries by only referencing their schema and then run those queries later by binding them to a delegate on the client or server.

```ts
const q = schema.query;
const issues = q.issue;
const issueOne = q.issue.where('id', 1);

// inside the useQuery hook we would call `asRunnable`
issueOne.asRunnable(z).materialize();
```

### DIFF
--- a/packages/zql/src/query/query-impl.ts
+++ b/packages/zql/src/query/query-impl.ts
@@ -147,13 +147,13 @@ export abstract class AbstractQuery<
   TReturn = PullRow<TTable, TSchema>,
 > implements Query<TSchema, TTable, TReturn>
 {
-  readonly #schema: TSchema;
-  readonly #tableName: TTable;
-  readonly #ast: AST;
+  protected readonly _schema: TSchema;
+  protected readonly _tableName: TTable;
+  protected readonly _ast: AST;
   readonly format: Format;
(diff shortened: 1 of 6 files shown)


tantaman:
I think I'm going to update this to have `materialize`, `run`, `preload` take a `delegate` as an argument instead.

I find it confusing (in later PRs) that I can get a query that I can call `run` on but it throws because the delegate is missing.

tantaman:
this change has been superseded by https://github.com/rocicorp/mono/pull/4488

packages/zql/src/query/query-impl.ts
```diff
@@ -147,13 +147,13 @@ export abstract class AbstractQuery<
   TReturn = PullRow<TTable, TSchema>,
 > implements Query<TSchema, TTable, TReturn>
 {
-  readonly #schema: TSchema;
-  readonly #tableName: TTable;
-  readonly #ast: AST;
+  protected readonly _schema: TSchema;
```
arv, line 150: All these protected properties is a code smell... Maybe we should rethink how the query implementations are done?

PR #4483, opened 2025-06-10
chore(zql): rename `makeSchemaQuery` to `query`

tantaman:
to get a handle on the query builders one can now do:

```ts
const builders = query(schema);
builders.issue.where(...);
```

This is used for defining custom queries. Could also be used to remove the superfluous `useZero` in react components.

I did not stick the builder on schema mainly for practical considerations. It would increase the size of the schema type which we want to avoid given TypeScript's type size limits.

### DIFF
--- a/packages/zql/src/query/named.ts
+++ b/packages/zql/src/query/named.ts
@@ -23,15 +23,29 @@ type NamedQueryImpl<
   TReturnQuery extends Query<S, keyof S['tables'] & string>,
 > = (tx: SchemaQuery<S>, ...arg: TArg) => TReturnQuery;
 
+export function query<S extends Schema>(s: S): SchemaQuery<S>;
 export function query<
   S extends Schema,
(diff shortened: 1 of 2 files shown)


packages/zql/src/query/named.ts
```diff
@@ -23,15 +23,29 @@ type NamedQueryImpl<
...
   TArg extends ReadonlyArray<ReadonlyJSONValue>,
   TReturnQuery extends Query<S, keyof S['tables'] & string>,
 >(
-  _s: S,
+  s: S,
   name: string,
   fn: NamedQueryImpl<S, TArg, TReturnQuery>,
-): NamedQuery<S, TArg, TReturnQuery> {
+): NamedQuery<S, TArg, TReturnQuery>;
+export function query<
```
arv, line 36: This looks suspect to me. Only the first one defines the actual type. So it is a type error to call query with name as undefined.

PR #4484, opened 2025-06-10
chore(zbugs): stub handler for custom queries

tantaman:


packages/zql/src/query/query.test.ts
```diff
@@ -856,16 +856,6 @@ test('custom materialize factory', () => {
   >();
 });
 
-test('Make sure that QueryInternal does not expose the ast', () => {
```
tantaman, line 859: API servers must pull the AST in order to send it back to Zero-cache.

packages/zql/src/query/query.ts
```diff
@@ -164,6 +164,7 @@ export interface Query<
    * the server which tracks queries by `name:args`.
    */
   hash(): string;
+  readonly ast: AST;
```
arv, line 167: It is back!

I guess we cannot prevent this from escaping... Now we need to export all the types that the AST references.

packages/zql/src/query/query.test.ts
arv: Can we make this an opaque type?

apps/zbugs/api/index.ts
```diff
@@ -147,6 +144,56 @@ fastify.post<{
   reply.send(response);
 });
 
+const queryBuilders = query(schema);
+const emptyQuery = queryBuilders.issue.where(({or}) => or());
+
+fastify.post<{
+  Querystring: Record<string, string>;
+  Body: ReadonlyJSONValue;
+}>('/api/pull', async (request, reply) => {
```
arv, line 153: Is this the recommended endpoint URL?

@aboodman

apps/zbugs/api/index.ts
```diff
@@ -147,6 +144,56 @@ fastify.post<{
...
+  const response: TransformResponseMessage = [
+    'transformed',
+    transformRequest[1].map(req => ({
+      id: req.id,
+      name: req.name,
+      ast: emptyQuery.ast,
+    })),
+  ];
+  reply.send(response);
+});
```
arv, line 174: I wonder if we should provide a higher level API for this.

PR #4486, opened 2025-06-10
feat(zbugs): no need to call `useZero` to build queries in React

tantaman:
trying this on for size before converting zbugs to custom queries.

Feels nice.
- less generics (<Schema, Mutators> being passed around
- slightly less to type (queries.issue vs z.query.issue)
- no `const z = useZero();` just to then `z.query.issue`

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -79,7 +79,11 @@ import {
 } from '../../../zero-schema/src/name-mapper.ts';
 import {customMutatorKey} from '../../../zql/src/mutate/custom.ts';
 import {newQuery} from '../../../zql/src/query/query-impl.ts';
-import {type Query, type RunOptions} from '../../../zql/src/query/query.ts';
+import {
+  type PreloadOptions,
+  type Query,
+  type RunOptions,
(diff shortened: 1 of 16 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -736,6 +743,14 @@ export class Zero<
     return createLogOptions(options);
   }
 
+  preload(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    query: Query<S, keyof S['tables'] & string, any>,
+    options: PreloadOptions,
+  ): void {
```
tantaman, line 750: nit: should return cleanup fn

packages/zql/src/query/query.ts
```diff
@@ -172,6 +172,7 @@ export interface Query<
     name: string,
     args: ReadonlyArray<ReadonlyJSONValue>,
   ): Query<TSchema, TTable, TReturn>;
+  delegate(delegate: QueryDelegate): Query<TSchema, TTable, TReturn>;
```
arv, line 175: The delegate is tied to the Zero instance so I don't think it is needed on the Query interface.

arv (commented):
This is nice

PR #4488, opened 2025-06-11
chore(zql): make `delegate` optional to `QueryImpl`

tantaman:
This is untying queries from a concrete implementation.
so you can do:

```ts
const builders = query(schema);
const issueQ = builders.issue.where(...);
```

and later execute the queries in the appropriate context

```ts
issueQ.delegate(zqlite / zql / pg).run();
```

users will  generally never need to call delegate themselves as the useQuery hook or the custom mutator transaction will have already done it for them. Custom queries are never executed in the function where they're defined so no need to delegate there either.

Preload is the one place where a query delegate would not be set automatically for the user.

E.g.,

```ts
issueQ.preload();
```

So a delegate parameter has been added here for the time being. I think the better route would be to move `preload` to `Zero` since preloading only makes sense in the context of having a zero client. That is a breaking API change, however, so we'll discuss it first.

Example of the "better route":

```ts
const z = new Zero();
z.preload(issueQ);
```

instead of 

```ts
issueQ.preload(z);
```

- **edit:** https://github.com/rocicorp/mono/pull/4486 moves `preload` to Zero without breaking the API.

### DIFF
--- a/packages/zql/src/query/query-impl.ts
+++ b/packages/zql/src/query/query-impl.ts
@@ -60,7 +60,7 @@ export function newQuery<
   TSchema extends Schema,
   TTable extends keyof TSchema['tables'] & string,
 >(
-  delegate: QueryDelegate,
+  delegate: QueryDelegate | undefined,
   schema: TSchema,
   table: TTable,
 ): Query<TSchema, TTable> {
@@ -817,7 +817,7 @@ export class QueryImpl<
     complete: Promise<void>;
   } {
     const delegate = must(
-      this._delegate,
+      pullDelegate(options?.delegate ?? this._delegate),
       'preload requires a query delegate to be set',
     );
     const {resolve, promise: complete} = resolver<void>();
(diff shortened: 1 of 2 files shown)


packages/zql-integration-tests/src/helpers/static.ts
```diff
@@ -9,14 +9,26 @@ export function staticToRunnable<TSchema extends Schema>({
...
-  query: AnyStaticQuery;
+  query: AnyQuery;
   schema: TSchema;
   harness: Awaited<ReturnType<typeof bootstrap>>;
 }) {
   // reconstruct the generated query
   // for zql, zqlite and pg
-  const zql = query.asRunnable(harness.delegates.memory);
-  const zqlite = query.asRunnable(harness.delegates.sqlite);
+  const zql = new QueryImpl(
```
tantaman, line 18: zql-integration-tests do not need to use static queries anymore.

arv (commented):
LGTM

I like where this is going but I think we need a design doc for the API soon.

PR #4489, opened 2025-06-11
fix(zero-cache): fix cvr migration to remove column from the right table

darkgnotic:
Fixes typo in #4478:  https://github.com/rocicorp/mono/pull/4478/files#diff-49db3c39da5be2abfbd9adecc2d59f0c067af2c7928dad605b6f4c641bd0d216


arv (commented):
LGTM

PR #4490, opened 2025-06-11
fix(zero-cache): fix handling of invalid websocket requests

darkgnotic:
Fixes a bug introduced in the removal of multi-tenant dispatch (#4325) in which invalid websocket requests were not properly being handled.

User report:
* https://discord.com/channels/830183651022471199/1288232858795769917/1382142719404806174

### DIFF
--- a/packages/zero-cache/src/integration/integration.pg-test.ts
+++ b/packages/zero-cache/src/integration/integration.pg-test.ts
@@ -40,6 +40,7 @@ import {DbFile} from '../test/lite.ts';
 import type {PostgresDB} from '../types/pg.ts';
 import {childWorker, type Worker} from '../types/processes.ts';
 import {stream, type Sink} from '../types/streams.ts';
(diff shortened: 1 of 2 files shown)


arv (commented):
LGTM

It is not clear to me how this fixes the problem? The old code seems to have checked the path as well?

darkgnotic:
> LGTM
> 
> It is not clear to me how this fixes the problem? The old code seems to have checked the path as well?

Sorry, it's about two levels deep. `installWebSocketReceiver()` calls `createWebSocketHandoffHandler()` which is what handles Errors thrown by the `handoff` callback:

https://github.com/rocicorp/mono/blob/21449d1b2b75095af6b9599fae9a9fac19444005/packages/zero-cache/src/types/websocket-handoff.ts#L91

I can't recall why I had unrolled that logic when I did this refactoring. Maybe it had something to do with the type parameters. Anyway, I figured it out this time.  😅

PR #4491, opened 2025-06-11
feat(zbugs): migrate to custom queries

tantaman:
This is a fancy trick that the current API can preserve --

CleanShot 2025-06-11 at 14 15 16

# TODO:
- null vs undefined shenanigans (ReadOnlyJSONValue)
- eslint rule to ensure name matches const (for us anyway)
- optional arg problems... fix in custom queries by returning empty queries? Create an empty query sentinel.
- priority on preload
- allow backend to return many queries in response to 1 custom query request
-


apps/zbugs/shared/queries.ts
```diff
@@ -0,0 +1,169 @@
+import {escapeLike, query, type Row} from '@rocicorp/zero';
+import {builders, type Schema} from './schema.ts';
+import {INITIAL_COMMENT_LIMIT} from './consts.ts';
+const {named} = query;
```
arv: This is a smell to me. If we expect them to use named as a bare function we should export it as a bare function.

apps/zbugs/shared/queries.ts
```diff
@@ -0,0 +1,169 @@
...
+
+export const allLabels = named('allLabels', () => builders.label);
```
arv: maybe this should be called `namedQuery`

Other options are:

```ts
query({name: 'allLabels', fn: () => builders.label})
```

apps/zbugs/shared/queries.ts
```diff
@@ -0,0 +1,169 @@
...
+export const allUsers = named('allUsers', () => builders.user);
+
+export const issuePreload = named('issuePreload', (userID: string) =>
+  builders.issue
```
arv: builders -> builder

packages/zero-client/src/client/zero.ts
```diff
@@ -562,6 +568,7 @@ export class Zero<
       slowMaterializeThreshold,
       assertValidRunOptions,
     );
+    this.queryDelegate = this.#zeroContext;
```
arv: I'm a little bit concerned about exposing things that we do not expect people to use.

packages/zero-client/src/client/zero.ts
```diff
@@ -736,6 +743,14 @@ export class Zero<
     return createLogOptions(options);
   }
 
+  preload(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    query: Query<S, keyof S['tables'] & string, any>,
+    options: PreloadOptions,
+  ): void {
+    query.delegate(this.#zeroContext).preload(options);
```
arv: I think zero.preload makes some sense but do we want query.preload and query.run then?

We should either put things on zero or on query and it seems like we are moving towards having queries not be tied to a zero instance.

arv (commented):
I think we need to design our API a bit.

Does this mean that all queries are now custom queries? If so, do we need to call them "named"?

apps/zbugs/shared/queries.ts
tantaman: changed to `namedQuery`

packages/zero-client/src/client/zero.ts
tantaman: I'm not sure yet. Still trying this out. I think we may want to keep `query.run / then` for use in custom mutators. Rather than:

```ts
await query.run(tx);
```

apps/zbugs/shared/queries.ts
tantaman: fixed

tantaman:
> I think we need to design our API a bit.
> Does this mean that all queries are now custom queries? If so, do we need to call them "named"?

Queries can either be:
- ad-hoc / what they are today
- custom


A query switches to being custom only when a name is assigned to it.

Let me put a high level overview together. The API has some really cool features, unlocked by how nearly identical custom and normal queries are, that are hard to get across without examples.

packages/zero-client/src/client/zero.ts
arv: The alternative could be:

```js
await tx.run(query);
```

PR #4492, opened 2025-06-12
fix(zero-cache): lazy load otel auto instrumentation

cesara:


packages/zero-cache/src/server/otel-start.ts
```diff
@@ -61,10 +63,19 @@ class OtelManager {
     }
 
     const logger = logs.getLogger('zero-cache');
+
+    // Lazy load the auto-instrumentations module
+    // avoid MODULE_NOT_FOUND errors in environments where it's not being used
+    if (!this.#autoInstrumentations) {
+      this.#autoInstrumentations =
+        autoInstrumentationsModule.getNodeAutoInstrumentations();
```
tantaman, line 71: My reading of the case was that it was lazy:

https://github.com/rocicorp/mono/blob/25c5f437d46780d7c8b6b784e7947b058bbb339d/packages/zero-cache/src/server/otel-start.ts#L40-L42

So I don't know why this fixes things. 🤷‍♂️

PR #4493, opened 2025-06-12
feat(inspector): Add Logarithmic Histogram Support for Query Metrics

arv:
### Summary

This PR introduces support for measuring query performance metrics using logarithmic histograms. The following metrics are now tracked:

- **Materialization Time**: Measures the time taken to materialize each query, grouped by query hash.
- **Advance Time**: Measures the time taken to process changes during query updates.

### Testing

- Updated inspector tests to verify histogram data is correctly exposed for queries.

### DIFF
--- a/packages/zero-client/src/client/context.ts
+++ b/packages/zero-client/src/client/context.ts
@@ -1,6 +1,7 @@
 import type {NoIndexDiff} from '../../../replicache/src/btree/node.ts';
 import type {Hash} from '../../../replicache/src/hash.ts';
 import {assert} from '../../../shared/src/asserts.ts';
+import {LogarithmicHistogram} from '../../../shared/src/logarithmic-histogram.ts';
 import type {AST} from '../../../zero-protocol/src/ast.ts';
 import {ErrorKind} from '../../../zero-protocol/src/error-kind.ts';
 import type {FilterInput} from '../../../zql/src/ivm/filter-operators.ts';
@@ -15,6 +16,7 @@ import type {
 } from '../../../zql/src/query/query-delegate.ts';
 import type {RunOptions} from '../../../zql/src/query/query.ts';
 import type {TTL} from '../../../zql/src/query/ttl.ts';
+import type {HistogramsDelegate, HistogramsMap} from './inspector/types.ts';
 import {type IVMSourceBranch} from './ivm-branch.ts';
 import type {QueryManager} from './query-manager.ts';
 import type {ZeroLogContext} from './zero-log-context.ts';
@@ -30,7 +32,7 @@ export type UpdateCustomQuery = QueryManager['updateCustom'];
  * Replicache data and pushes them into IVM and on tells the server about new
  * queries.
  */
-export class ZeroContext implements QueryDelegate {
+export class ZeroContext implements QueryDelegate, HistogramsDelegate {
   // It is a bummer to have to maintain separate MemorySources here and copy the
   // data in from the Replicache db. But we want the data to be accessible via
   // pipelines *synchronously* and the core Replicache infra is all async. So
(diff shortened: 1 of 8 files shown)


arv:
@tantaman @grgbkr Let me know what other metrics we want to expose to the debug panel?

packages/zero-client/src/client/context.ts
```diff
@@ -169,7 +189,9 @@ export class ZeroContext implements QueryDelegate {
   ) {
     this.batchViewUpdates(() => {
       try {
+        const t0 = performance.now();
         this.#mainSources.advance(expectedHead, newHead, changes);
```
tantaman, line 193: I wonder if this includes time to update the UI for some frameworks. Maybe not in that they would schedule UI changes in a microtask?

packages/zero-client/src/client/context.ts
```diff
@@ -192,4 +214,16 @@ export class ZeroContext implements QueryDelegate {
       }
     }
   }
+
+  queryHistograms(hash: string): HistogramsMap {
+    const histogram = this.#materializeHistograms.get(hash);
+    if (histogram === undefined) {
+      return new Map();
+    }
+    return new Map([['materialize', histogram]]);
```
tantaman, line 223: why is this a map? 🤔 Does it only have 2 entries? `materialize` and `advance`?

packages/zql/src/query/query-impl.ts
```diff
@@ -745,8 +745,7 @@ export class QueryImpl<
     let queryComplete = delegate.defaultQueryComplete;
     const gotCallback: GotCallback = got => {
       if (got) {
-        const t1 = Date.now();
-        delegate.onQueryMaterialized(this.hash(), ast, t1 - t0);
+        delegate.onQueryMaterialized(this.hash(), ast, performance.now() - t0);
```
tantaman, line 748: this is a very misleading metric in that it will include the time it took the server to hydrate all queries sent in a `changeDesiredQueries` message.

wdyt about removing it while you're here and exposing a different metric like a "changeDesiredQueries histogram"?

packages/zero-client/src/client/context.ts
```diff
@@ -123,6 +135,14 @@ export class ZeroContext implements QueryDelegate {
         duration,
       );
     }
+
+    // Add a histogram for the materialization duration.
+    let histogram = this.#materializeHistograms.get(hash);
+    if (histogram === undefined) {
+      histogram = new LogarithmicHistogram();
+      this.#materializeHistograms.set(hash, histogram);
+    }
```
tantaman, line 144: where do the histograms get cleaned up?

tantaman:
> @tantaman @grgbkr Let me know what other metrics we want to expose to the debug panel?

- time between mutation send and mutation confirmed
- time between sending `changeDesiredQueries` and the corresponding poke response

Looks like this currently captures the time to advance all active queries. Might be useful to get the time to advance each individual query as well. Idk though. Depends on the effort. They could also dive into the profiler once they know there is a slow `advance`.

packages/zero-client/src/client/context.ts
arv: I was expecting there to be more histograms per query. I think removing it now would be fine.

packages/zero-client/src/client/context.ts
arv: They do not get cleaned up. They are in kept memory. One of the benefits of histograms is that they have a very small memory and runtime footprint.

tantaman: but when someone is using a typeahead there will be 1 histogram per keystroke. Feels like there should be some bound on how many we keep.

arv:
Abandoning this PR. Instead I'm using TDigests and the same metrics have been added.

PR #4495, opened 2025-06-12
chore(zero): nits

tantaman:
- missing type exports
- extraneous comment
- pass the correct config to the view-syncer


arv (commented):
LGTM

PR #4498, opened 2025-06-12
chore(zero-cache): handle json args for custom queries

tantaman:
custom queries which contained a boolean in their argument list would fail to insert.

https://github.com/porsager/postgres/issues/386

https://github.com/porsager/postgres/blob/9b92b65da6a5121545581a6dd5de859c2a70177f/src/types.js#L227

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/cvr-store.ts
+++ b/packages/zero-cache/src/services/view-syncer/cvr-store.ts
@@ -421,11 +421,45 @@ export class CVRStore {
 
   putQuery(query: QueryRecord): void {
     const change: QueriesRow = queryRecordToQueryRow(this.#id, query);
+    // ${JSON.stringify(change.queryArgs)}::text::json is used because postgres.js
+    // gets confused if the input is `[boolean]` and throws an error saying a bool
+    // cannot be converted to json.
+    // https://github.com/porsager/postgres/issues/386
     this.#writes.add({
       stats: {queries: 1},
-      write: tx => tx`INSERT INTO ${this.#cvr('queries')} ${tx(change)}
+      write: tx => tx`INSERT INTO ${this.#cvr('queries')} (
+        "clientGroupID",
+        "queryHash",
+        "clientAST",
+        "queryName",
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/services/view-syncer/cvr-store.ts
```diff
@@ -421,11 +421,45 @@ export class CVRStore {
...
+        ${change.queryArgs === undefined ? null : JSON.stringify(change.queryArgs)}::text::json,
+        ${change.patchVersion},
+        ${change.transformationHash ?? null},
+        ${change.transformationVersion ?? null},
+        ${change.internal},
+        ${change.deleted ?? false}
+      )
       ON CONFLICT ("clientGroupID", "queryHash")
-      DO UPDATE SET ${tx(change)}`,
+      DO UPDATE SET
```
darkgnotic, line 454: FYI, if you want to avoid the duplication and keep the query string like it was before, I believe you can achieve this by creating a new object to pass into for substitution:

```ts
const row = {
  ...change,
  queryArgs: ...
  transformationHash: ...
  transformationVersion: ...
  deleted: ...
};

write: tx => tx`INSERT INTO ${this.#cvr('queries')} ${tx(row)} ... DO UPDATE SET ${tx(row)} ...`
```

darkgnotic (approved):

packages/zero-cache/src/services/view-syncer/cvr-store.ts
tantaman: how will it handle the casting then?

darkgnotic: ah, i totally missed that. sorry.  😅 

i guess you have to expand it all out to cast, unless this works:

```ts
const row = {
  ...change,
  queryArgs: tx.json(change.queryArgs === undefined ? null : change.queryArgs),
  transformationHash: ...
  transformationVersion: ...
  deleted: ...
};
```

but i remember you saying that the `json()` helper wasn't working for some reason.

PR #4499, opened 2025-06-12
chore(zero): bump zero-sqlite3 1.0.5

cesara:


tantaman:
thanks for catching this.

PR #4500, opened 2025-06-13
feat(zero): Limit TTL to 10 minutes

arv:
If the TTL is set to "never" or a value greater than 10 minutes, we now
default it to 10 minutes. This is to prevent that queries are active for
too long, which can lead to performance issues and resource exhaustion.

This is dealt with on both the client and server. On the client we log
a warning and change the TTL. On the server we treat `-1` as 10 minutes so that old
existing queries will also expire.


tantaman:
We'll need to change how client deletion works too, correct? As in, do not mark queries inactive on disconnect but rather on next connect.

E.g., zbugs does a normal load then a preload. That preload query will have been marked as inactive on client disconnect and, I assume, will be cleaned up if the page is visited > 10 minutes later.

arv:
Reopened now that we have `ttlClock` and inactive queries in `initConnection`.

PR #4501, opened 2025-06-13
feat(zero-react): Add placeholder for ZeroInspector

arv:
Right now it only contains the Zero version number but the infrastructure is in place to add more features later.


arv:
https://github.com/user-attachments/assets/25ffb94d-04c9-4075-81bb-2c2dc0932be9

PR #4509, opened 2025-06-18
chore(zero-client): Better alive clients mock

arv:
When we do not have Web Locks API support, we keep track of the clients in memory.

### DIFF
--- a/packages/zero-client/src/client/alive-clients-manager.ts
+++ b/packages/zero-client/src/client/alive-clients-manager.ts
@@ -23,7 +23,9 @@ function fromLockKey(
   };
 }
 
-const lockManager = getBrowserGlobal('navigator')?.locks;
+// When we do not have the `navigator.locks` API available, we will keep track
+// of the "locks" in memory.
+const allMockLocks = new Set<{name: string}>();
 
 /**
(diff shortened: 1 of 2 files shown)


packages/zero-client/src/client/alive-clients-manager.ts
```diff
@@ -42,26 +44,44 @@ export class AliveClientsManager {
...
 
-    void lockManager?.request(
-      toLockKey(clientGroupID, clientID),
-      () => this.#resolver.promise,
+    const name = toLockKey(clientGroupID, clientID);
+    let mockLock: {name: string};
+
+    if (this.#lockManager) {
+      this.#lockManager
+        .request(name, {signal}, () => this.#resolver.promise)
```
arv, line 58: I ended up with a signal because it simplifies things when used from the Zero class.

grgbkr:
What was the motivation for a better mock here?  Does this matter in non-browser environments, or does this help simplify testing?

arv:
> What was the motivation for a better mock here? Does this matter in non-browser environments, or does this help simplify testing?

Not for testing but for the case where there are more than one Zero instance in the same global context and that context does not have web locks.

PR #4510, opened 2025-06-18
chore(zer-client): Send active clients on connect

arv:
This sends the active clients on connect.

The active clients will later be used to determine what desired queries can be inactivated

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -86,7 +86,7 @@ import {
 } from '../../../zql/src/query/query.ts';
 import {nanoid} from '../util/nanoid.ts';
 import {send} from '../util/socket.ts';
-import {AliveClientsManager} from './alive-clients-manager.ts';
+import {ActiveClientsManager} from './active-clients-manager.ts';
 import * as ConnectionState from './connection-state-enum.ts';
 import {ZeroContext} from './context.ts';
 import {
@@ -371,7 +371,7 @@ export class Zero<
 
   // We use an accessor pair to allow the subclass to override the setter.
   #connectionState: ConnectionState = ConnectionState.Disconnected;
-  #aliveClientsManager: AliveClientsManager | undefined;
+  #aliveClientsManager: ActiveClientsManager | undefined;
 
   #setConnectionState(state: ConnectionState) {
     if (state === this.#connectionState) {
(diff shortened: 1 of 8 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -1960,6 +1974,8 @@ export async function createSocket(
   let queriesPatch: Map<string, UpQueriesPatchOp> | undefined =
     await queriesPatchP;
 
+  const activeClients = await aliveClientsManager?.getActiveClients();
```
grgbkr, line 1977: is there a distinction between active and alive?

packages/zero-client/src/client/zero.ts
```diff
@@ -1217,6 +1227,7 @@ export class Zero<
       this.#options.push,
       this.#options.maxHeaderLength,
       additionalConnectParams,
+      this.#aliveClientsManager,
```
grgbkr, line 1230: maybe we should assert this.#aliveClientsManager is not undefined here...

it shouldn't be possible for it to be undefined because we `await this.clientGroupID` above.  

Alternatively we could make this.#aliveClientsManager  a Promise<ActiveClientsManager> and await it here.

packages/zero-client/src/client/zero.ts
arv: No, I stopped using alive and forgot to rename this

packages/zero-client/src/client/zero.ts
arv: I tried making the ActiveClientsMananger sync. It lead to a race condition with the "event" when closing.

In general I try to make construction sync and hide the async details in the implementation but in this case it cleaned things up to make the "constructor" async.

grgbkr: What I'm hoping to address is that with the current code it appears this.#aliveClientsManager could be undefined here (and thus we wouldn't send the active clients on connect).  

I think we can change the createSocket arg from 

```
 aliveClientsManager:
    | Pick<ActiveClientsManager, 'getActiveClients'>
    | undefined,
```

to just
```
 aliveClientsManager: Pick<ActiveClientsManager, 'getActiveClients'>
 ```
 
 If we made this.#aliveClientsManager a promise that resolved to activeClientsManager when it was constructred, and then here we would just pass
 
 `await this.#aliveClientsManager`

PR #4512, opened 2025-06-19
chore: Fix turbo deps a bit

arv:
No need to rebuild replicache* when zero* changes.


tantaman:
How does turbo track dependencies these days? Given our package.json files do not list internal packages as dependencies.

arv:
We list all the file deps in /turbo.json like before but since replicache* does not depend on zero they do not need to be rebuilt when zero changes.

arv:
We could also add the devDependencies and remove the internal ones in the prepack hook.

arv:
Actually, we no longer have the postinstall script so we should be able to remove all of these hacks.

PR #4515, opened 2025-06-19
chore(zero): get zql-integration-tests running in ci

tantaman:
Looks like tests with `.pg-VV` were not getting picked up unless the package was listed as "special"

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,7 +1,13 @@
 import {readdirSync} from 'node:fs';
 import {defineConfig} from 'vitest/config';
 
-const special = ['shared', 'zero-cache', 'z2s', 'zero-pg'];
+const special = [
+  'shared',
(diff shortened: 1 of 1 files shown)


arv (approved):
LGTM

vitest.config.ts
```diff
@@ -1,7 +1,13 @@
 import {readdirSync} from 'node:fs';
 import {defineConfig} from 'vitest/config';
 
-const special = ['shared', 'zero-cache', 'z2s', 'zero-pg'];
+const special = [
```
arv, line 4: We could probably find these by looking at the vitest.config.ts

I think that all these special ones have a `.test.workspace`.

arv: https://bugs.rocicorp.dev/issue/3925

PR #4517, opened 2025-06-19
chore(zero): rm `server` export

tantaman:
The server is kicked off via `cli.js` in the `zero` package. The export is not needed.


arv (commented):
LGTM

PR #4520, opened 2025-06-19
chore(zql): disable fuzzer for now..

tantaman:
Too many timeouts. This isn't great for running in CI. We'll need to come up with another plan.

Options:
a. limit the complexity of queries created in CI
b. disable in CI and run on a dedicated machine outside of CI
c. kick timeouts up to 30 seconds?


packages/zql-integration-tests/src/chinook/chinook-fuzz-hydration.pg-test.ts
```diff
@@ -27,11 +27,15 @@ const harness = await bootstrap({
   pgContent,
 });
 
-test.each(Array.from({length: 100}, () => createCase()))(
+test.each(Array.from({length: 0}, () => createCase()))(
```
arv, line 30: could you have used 1 so the code is excercised?

arv (commented):
LGTM

PR #4523, opened 2025-06-19
fix(zero): Remove red error text when starting tools w/o .env.

aboodman:
Also remove yellow info text about injecting from .env file.

Fixes https://bugs.rocicorp.dev/issue/3853


packages/shared/src/dotenv.ts
```diff
@@ -0,0 +1,8 @@
+import {config as dotenvxConfig} from '@dotenvx/dotenvx';
+
+// Import env vars from .env file if present but don't whine if not present.
+// Also no free marketing for dotenvx.
```
tantaman, line 4: 🤣 so mean

tantaman (approved):
lgtm

PR #4524, opened 2025-06-20
chore(zql): increase chinook timeouts

tantaman:
only change here is adding 

```ts
describe(
  'Chinook PG Tests',
  {
    timeout: 30_000,
  },
```

putting the timeout in the vitest config had no impact.


arv (commented):
LGTM

PR #4527, opened 2025-06-20
ZeroProvider can now instantiate Zero too.

aboodman:


packages/zero-react/src/zero-provider.tsx
```diff
@@ -0,0 +1,77 @@
...
+  init?: (zero: Zero<S, MD>) => void;
+  children: ReactNode;
+};
+
+export function ZeroProvider<
+  S extends Schema,
+  MD extends CustomMutatorDefs<S> | undefined = undefined,
+>({children, init, ...props}: ZeroProviderProps<S, MD>) {
+  const [zero, setZero] = useState<Zero<S, MD> | undefined>(
+    'zero' in props ? props.zero : undefined,
```
arv, line 47: why not?

```suggestion
    props.zero,
```

packages/zero-react/src/zero-provider.tsx
```diff
@@ -0,0 +1,77 @@
...
+  );
+
+  // If Zero is not passed in, we construct it, but only client-side.
+  // Zero doesn't really work SSR today so this is usually the right thing.
+  // When we support Zero SSR this will either become a breaking change or
+  // more likely server support will be opt-in with a new prop on this
+  // component.
+  useEffect(() => {
+    if ('zero' in props) {
```
arv, line 56: Why not?

```suggestion
    if (props.zero) {
```

packages/zero-react/src/zero-provider.tsx
```diff
@@ -0,0 +1,77 @@
...
+      setZero(props.zero);
+      return;
+    }
+
+    const z = new Zero(props);
+    if (init) {
+      init(z);
+    }
```
arv:
```suggestion
    init?.(z);
```

arv (approved):
LGTM

This looks cleaner.

The whole Atom story was to allow us to start Zero before we loaded and initialized React. IIRC, Matt said the benefit was not very large so this is good to me.

packages/zero-react/src/zero-provider.tsx
aboodman: TypeScript complains about that. I don't understand why it does in this case. It has to do with the `|` type.

packages/zero-react/src/zero-provider.tsx
aboodman: TS complains.

packages/zero-react/src/zero-provider.tsx
aboodman: done

aboodman:
hm, little worried about the delayed startup. will see what greg ends up
doing when optimizing zero-records.

PR #4528, opened 2025-06-20
chore: Simplify turbo config

arv:
We no longer have a postinstall script that runs so we can Simplify the turbo config


tantaman (commented):
lgtm. We'll just need to be diligent about listing dependencies in package.json.

PR #4529, opened 2025-06-20
fix(custom-mutators): handle errors thrown in errorMode, only break on ooo errors

tantaman:
Two bugs

# Errors thrown in Error Mode

There was a race where a mutator could fail for some reason then fail again with OOO or AlreadyProcessed in error mode. If it failed with OOO or AlreadyProcessed, in error mode, we would not catch and handle this.

E.g.,

```
Step 1: runMutator(foo);
Step 2: foo throws new Error();
Step 3: pushProcessor catches error
Step 4: push processor re-runs in error mode
Step 5: error mode throws OOO or AlreadyProcessed
```

The error from step (5) would not be caught. This means the ooo error would not reach the client or, if it was an already processed error, the rest of the mutations would not be processed.

# Not Skipping Mutations

If a mutation returned an error we would break the loop rather then continuing to process the remaining mutations in the array 🤯

### DIFF
--- a/packages/zero-pg/src/push-processor.pg-test.ts
+++ b/packages/zero-pg/src/push-processor.pg-test.ts
@@ -3,11 +3,13 @@ import {beforeEach, describe, expect, test} from 'vitest';
 import type {PostgresDB} from '../../zero-cache/src/types/pg.ts';
 import {getClientsTableDefinition} from '../../zero-cache/src/services/change-source/pg/schema/shard.ts';
 
-import {PushProcessor} from './push-processor.ts';
+import {OutOfOrderMutation, PushProcessor} from './push-processor.ts';
 import {PostgresJSConnection} from './postgresjs-connection.ts';
 import type {PushBody} from '../../zero-protocol/src/push.ts';
 import {customMutatorKey} from '../../zql/src/mutate/custom.ts';
 import {ZQLDatabase} from './zql-database.ts';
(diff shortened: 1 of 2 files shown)


packages/zero-pg/src/push-processor.pg-test.ts
```diff
@@ -210,6 +212,223 @@ test('lmid still moves forward if the mutator implementation throws', async () =
...
+      relationships: {},
+      version: 1,
+    }),
+  );
+
+  expect(
+    await processor.process(
+      mutators,
+      params,
+      makePush([1, 2, 3, 4], Array(4).fill(customMutatorKey('foo', 'baz'))),
```
tantaman, line 228: the baz mutator always throws

packages/zero-pg/src/push-processor.ts
```diff
@@ -304,3 +308,17 @@ class OutOfOrderMutation extends Error {
...
+function makeAppErrorResponse(m: Mutation, e: unknown): MutationResponse {
+  return {
+    id: {
+      clientID: m.clientID,
+      id: m.id,
+    },
+    result: {
+      error: 'app',
+      details:
+        e instanceof Error ? e.message : 'exception was not of type `Error`',
```
grgbkr, line 321: maybe
`e instanceof Error ? e.message : String(e)`

grgbkr:
LGTM

PR #4533, opened 2025-06-21
fix(zero-cache): Colorize log output according to log level.

aboodman:
Clean up logging code a bit to make it more consistent throughout.

Also prefix consistently with pid and log level.


aboodman:
sample output:

CleanShot 2025-06-21 at 00 02 48@2x

PR #4538, opened 2025-06-25
chore(zero-client): Add active clients detection

arv:
This uses a combination of web locks and broadcast channel.

Each client sets up an exclusive lock for its name, and then uses a broadcast channel to announce its presence.

Other clients listen for these announcements on the broadcast channel and sets up a shared lock for the client's name. This way the shared lock will be acquired when the client goes away.

In case web locks are not supported we use an in memory fallback which at least allows multiple clients in the same global scope to work together.


aboodman:
What happens if they miss the announcement, either because the receive tab is frozen or because the receive tab starts up after the source tab?

aboodman:
I think to make this robust source clients will have to write some state into IndexedDB announcing themselves that way. Then receive clients will have to read this state and try to open the lock to see if source client is still there.

arv:
They also read the "state" from navigator.locks.

The broadcast channel is used to allow old clients to know about new clients.

But... Maybe using the dag store and use experimental watch would be better?

aboodman:
I'm sorry I misunderstood this PR the first time. Please disregard my comments, I will review again.

packages/zero-client/src/client/active-clients-manager.ts
```diff
@@ -0,0 +1,243 @@
...
+  readonly #lockManager = getBrowserGlobal('navigator')?.locks;
+  readonly #signal: AbortSignal;
+  readonly #unlisteners: (() => void)[] = [];
+  #activeClients: Set<string> = new Set();
+
+  /**
+   * A callback that is called when the list of active clients changes. It
+   * receives a `Set` of client IDs that are currently active in the client group.
+   */
+  onChange: ((activeClients: ReadonlySet<string>) => void) | undefined;
```
aboodman, line 78: It seems like what users will want is to know when an active client is removed. As is this API will force all users to track this themselves. Perhaps this class should do that work and directly provide the added/removed client.

packages/zero-client/src/client/active-clients-manager.ts
```diff
@@ -0,0 +1,243 @@
...
+        const client = fromLockName(e.data);
+        if (client?.clientGroupID === this.clientGroupID) {
+          this.#addListener(client.clientID);
+          this.#notifyClientActivated(client.clientID);
+        }
+      },
+      {signal},
+    );
+
+    let mockLock: {name: string; mode: 'exclusive'};
```
aboodman: Surprising to see this kind of branching in the code. Why not implement a common interface for the mock/polyfill and the real one so there doesn't need to be branching?

packages/zero-client/src/client/active-clients-manager.ts
```diff
@@ -0,0 +1,243 @@
...
+  readonly #unlisteners: (() => void)[] = [];
+  #activeClients: Set<string> = new Set();
+
+  /**
+   * A callback that is called when the list of active clients changes. It
+   * receives a `Set` of client IDs that are currently active in the client group.
+   */
+  onChange: ((activeClients: ReadonlySet<string>) => void) | undefined;
+
+  constructor(clientGroupID: string, clientID: string, signal: AbortSignal) {
```
aboodman: I find the `AbortSignal` API terrible and really hard to read. Consider just having a `close()` or `shutdown()` method on this class which uses `AbortSignal` internally as necessary to interact with the browser APIs.

aboodman (commented):
OK this makes sense. Sorry for misunderstanding at first.

Did you consider decoupling the registering and listening features? I don't think they need to be coupled and it might make it easier to work with.

Imagine this alternate and more general API:

```ts
// can work with js contexts besides tabs, 'tab' just more evocative
import { registerTab, listenForTabs } from 'cross-tab-registry.ts';

// first param is subset of LockManager that this function needs. Can be polyfilled too.
const unregister = registerTab(window.locks, {any, arbitrary:data, here});

type ListenCallbackParams = {added: unknown} | {removed: unknown};
const unlisten = listenForTabs(window.locks, change: ListenCallbackParams => {
  if (change.added) {
    console.log("handle change.added data");
  } else {
    console.log("handle change.removed data");
  }
});

// call unregister() and unlisten() if you want to stop using before unload.
```

To make this work, each registration would allocate its own unique ID, different from the clientID. But it would make this little library more general purpose and I think maybe a little easier to understand.

Also I feel each of the two functions would be easier to test and understand separated from the other.

packages/zero-client/src/client/active-clients-manager.ts
arv: That is a fair point.

In the case of zero, we do want the active clients so if we change to `{add: string[]; remove: string[]}` we will have to do more work.

I will see what I end up with in zero and I can change this as needed later.

packages/zero-client/src/client/active-clients-manager.ts
arv: I mock LockManager is just more code... but I can see if there is some way to keep this abstracted.

packages/zero-client/src/client/active-clients-manager.ts
arv: We must use a signal, that is the only thing that can abort a lock. We can change it to hide that as an implementation detail but we already have a signal from zero that fits perfectly.

arv:
The lock name must encode the client group id and the client id. This is the way we can figure out what clients are running. 

The broadcast channel name could be anything as long as all tabs agree. The data can be managed in the postMessage.

packages/zero-client/src/client/active-clients-manager.ts
aboodman: OK don't do anything zero doesn't want/need.

packages/zero-client/src/client/active-clients-manager.ts
aboodman: You don't have to mock the whole LockManager just the big that this code actually needs. You can use `Pick` to subset the interface of `LockManager` actually needed.

Or alternately define your own more minimal lockmanager to rely on, then write an adapter for the native one.

packages/zero-client/src/client/active-clients-manager.ts
aboodman: Yeah that's what I meant - to hide it as an impl detail. But I guess if you already have a signal in Zero then I see why you did this.

arv:
> Did you consider decoupling the registering and listening features? I don't think they need to be coupled and it might make it easier to work with.

I've been thinking and I do not see how this works.

A few things to remember.
- We are working with clients which there can be more than one per tab but that doesn't change anything.
- To be able to listen to a client going away you need to know how it was registered and you can only do it after the client was registered.

packages/zero-client/src/client/active-clients-manager.ts
```diff
@@ -0,0 +1,301 @@
...
+    signal: AbortSignal,
+  ) {
+    this.clientGroupID = clientGroupID;
+    this.clientID = clientID;
+    this.#lockManager = getClientLockManager(signal);
+    this.#activeClients.add(clientID);
+  }
+
+  async #init(
+    clientGroupID: string,
```
grgbkr, line 108: why take clientGroupID and clientID as args rather than using the corresponding instance vars?

PR #4541, opened 2025-06-27
feature(zero-cache): anonymous telemetry

cesara:


surfaceflinger:
Maybe consider supporting env DO_NOT_TRACK instead of creating new ZERO_ENABLE_USAGE_ANALYTICS?

https://consoledonottrack.com/

aboodman:
DO_NOT_TRACK sgtm.

packages/zero-cache/src/config/zero-config.ts
```diff
@@ -611,6 +611,15 @@ export const zeroOptions = {
     type: v.string().optional(),
     desc: [`The version string outputted to logs when the server starts up.`],
   },
+
+  enableUsageAnalytics: {
+    type: v.boolean().default(true),
+    desc: [
+      `Enable anonymous usage analytics telemetry. Set to false to opt out of`,
+      `telemetry collection. This helps us improve Zero Cache by collecting`,
```
darkgnotic: Nit: Maybe reorder/rephrase this and put the "Set to false" sentence last.

packages/zero-cache/src/server/anonymous-otel-start.ts
```diff
@@ -0,0 +1,368 @@
...
+    if (
+      process.env.AWS_LAMBDA_FUNCTION_NAME ||
+      process.env.AWS_REGION ||
+      process.env.AWS_EXECUTION_ENV
+    )
+      return 'aws';
+    if (process.env.RAILWAY_ENV || process.env.RAILWAY_STATIC_URL)
+      return 'railway';
+    if (process.env.RENDER || process.env.RENDER_SERVICE_ID) return 'render';
+    return 'unknown';
```
darkgnotic, line 304: If possible, can you look for a similar GCP variable? This will be helpful for gauging demand for GCP cloud zero clusters.

According to https://firebase.google.com/docs/functions/config-env?gen=2nd, maybe:
* `GCP_PROJECT`
* `GCLOUD_PROJECT`
* `GOOGLE_CLOUD_PROJECT`

packages/zero-cache/src/server/anonymous-otel-start.ts
```diff
@@ -0,0 +1,368 @@
...
+        `Telemetry: cached attributes=${JSON.stringify(this.#cachedAttributes)}`,
+      );
+    }
+    return this.#cachedAttributes;
+  }
+
+  #getPlatform(): string {
+    if (process.env.FLY_APP_NAME || process.env.FLY_REGION) return 'fly.io';
+    if (
+      process.env.AWS_LAMBDA_FUNCTION_NAME ||
```
darkgnotic: I'm not sure when / how zero-cache would be run via a lambda function, but I guess this doesn't hurt to have.

Other useful ones would be `ECS_CONTAINER_METADATA_URI_V4` and `ECS_CONTAINER_METADATA_URI`, one of which should be present for the SST/Fargate setup that we advertise. (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html)

packages/zero-cache/src/services/view-syncer/row-record-cache.ts
```diff
@@ -205,6 +206,8 @@ export class RowRecordCache {
             Date.now() - start
           } ms)`,
         );
+        // Record telemetry for successfully synced rows
+        recordRowsSynced(rows);
```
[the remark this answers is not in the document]
darkgnotic: This looks like a valid bug. Nice, Copilot.

packages/zero-cache/src/services/view-syncer/cvr.ts
```diff
@@ -162,6 +167,12 @@ export class CVRUpdater {
     );
     counters.cvrRowsFlushed().add(flushed.rows);
     histograms.cvrFlushTime().record(elapsed);
+
+    // Update telemetry with estimated CVR size
+    // Use JSON string length as an approximation of CVR size
+    const cvrSizeBytes = JSON.stringify(this._cvr).length;
```
darkgnotic: I'm not sure if this is a good idea. We just saw that cvrs [can be pretty big](https://github.com/rocicorp/mono/commit/2c8ec035cb8b8e357584008c1f52ed5e595b5247), and stringifying the whole thing on every flush can very possibly slow things down.

Also, `this._cvr` only contains the metadata (clients and queries), and does not include rows. What metric are we trying to capture with "updateCvrSize()"? Is it the size of the metadata specifically, or are we interested in the row contents as well?

It's possible that we don't need this metric at all if we're already tracking a "rows synced" metric.

darkgnotic (commented):
Nice work @cesara!

packages/zero-cache/src/services/view-syncer/row-record-cache.ts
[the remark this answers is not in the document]
cesara: co-pilot trying to trick us :) rows is a number

PR #4545, opened 2025-06-27
chore(zero): Remove closeConnection and close beacon

arv:
This deprecates the `closeConnection` protocol message and removes the
`closeConnection` beacon that was sent by the client when the client was
closed.
    
The `closeConnection` message was used to inactivate queries when
closing the client. We are going to use a different approach to
inactivate queries and this does not involve inactivating queries when
the client is closed.


packages/zero-protocol/src/protocol-version.ts
```diff
@@ -24,7 +24,8 @@ import {assert} from '../../shared/src/asserts.ts';
 // -- Version 16 adds a new error type (alreadyProcessed) to mutation responses
 // -- Version 17 deprecates `AST` in downstream query puts. It was never used anyway.
 // -- Version 18 adds `name` and `args` to the `queries-patch` protocol
-export const PROTOCOL_VERSION = 18;
+// -- Version 19 removes `closeConnection`
+export const PROTOCOL_VERSION = 19;
```
arv: @cesara I'll have to wait to for the zbugs deployment is resolved again.

After that I might need to make this 21? We will see, there are a bunch of these piled up now.

packages/zero-protocol/src/ast.test.ts
```diff
@@ -607,5 +607,5 @@ test('protocol version', () => {
   // old code will not understand the new schema, bump the
   // PROTOCOL_VERSION and update the expected values.
   expect(hash).toEqual('1c228prd1v53r');
-  expect(PROTOCOL_VERSION).toEqual(18);
+  expect(PROTOCOL_VERSION).toEqual(19);
```
darkgnotic: Thinking about this ... if a v18 client talks to a v19 server, what will the v19 server do when it receives a legacy closeConnection message ... will it ignore it or will it fail in some undesirable way?

If it's the latter, we may need to do this in two steps: Remove the client-side code in v19, and then remove server-side code (and protocol) in v20.

(But I don't know which is the case ... leaving that to you!)

arv:
I'll just keep the protocol for now. It is not really hurting.

PR #4550, opened 2025-06-27
feat(zero): added insert/update default functions to the schema

0xcadams:
This is an expansion of the current schema to support default values. Specifically, it allows:

1. On insert/update (or upsert), specify a default function that is called separately on both server and client.
2. Optionally, allow the database to fill in the value.

```tsx
table('user')
  .columns({
    id: string(),
    name: string(),
    createdAt: number().onInsert(() => Date.now()),
    updatedAt: number()
      .onInsert(() => Date.now())
      .onUpdate(() => Date.now()),
    createdAtWithDbDefault: number()
      .onInsert(() => Date.now())
      .dbGenerated('insert'),
  })
  .primaryKey('id')
```

This lets custom mutators omit values for fields that have a default:

```tsx
tx.mutate.user.insert({
  id: '1',
  name: 'John Doe',
});
// {
//   id: '1',
//   name: 'John Doe',
//   createdAt: 1743018158777,
//   updatedAt: 1743018158777,
//   createdAtWithDbDefault: 1743018158888, // db-generated value
// }
```

This also renames `optional` to `nullable` to more accurately reflect the database type.

The different combinations are:

1. A user passes a default `onInsert` function, and also runs it on the server (e.g. `createdAt` timestamp).
2. A user passes a default `onInsert` function, and opts to not run it on the server (e.g. db-generated `createdAt` timestamp or serial id).
3. A user passes a default `onUpdate` function, and also runs it on the server (e.g. `updatedAt` timestamp).
4. A user passes a default `onUpdate` function, and opts to not run it on the server (e.g. db trigger-generated `updatedAt` timestamp).

In these cases, the types are:

**Select**:

- 1-4: If the field is nullable, the value is nullable.

**Insert**:

- 1 & 2: The value does not need to be specified, but the user can optionally override it (but if `dbGenerated` is passed and the database does not generate a value, and the field is NOT NULL, the insert will fail).
- 3 & 4: The value needs to be defined.

**Update**:

- 1-4: The value is optional. The user can optionally override it.

**Upsert**:

- 1 & 2: The value is optional. The user can optionally override it.
- 3 & 4: The user must pass the key with either a value or an explicit `undefined`. This is because we don't know whether this will be an insert or update.

The defaults do not handle IDs well, since the function runs twice on the client and server. So columns with defaults are not allowed to be primary keys.


aboodman:
This is fantastic @0xcadams and i think it's basically perfectly structurally. My comments are more API design level:

1. I find `onInsert` and `onUpdate` little unintuitive. These functions are providing default values. So what about something like this:

```ts
table('user')
  .columns({
    id: string(),
    name: string(),
    createdAt: number()
      .default({
        insert: {
          client: () => Date.now(),  // can only be a function
          server: 'db',  // can also be a function
        }
      }),
    updatedAt: number()
      .default({
        insert: {
          client: () => Date.now(),
          server: 'db',
        },
        update: {
          client: () => Date.now(),
          server: 'db',
        },
      }),
  })
  .primaryKey('id')
```

This is a little laborious, but it can easily be abstracted by users for common cases:

```ts
const created = {
  createdAt: number()
    .default({
      insert: {
        client: () => Date.now(),  // can only be a function
        server: 'db',  // can also be a function
      }
    }),
};

const updated = {
  updatedAt: number()
    .default({
      insert: {
        client: () => Date.now(),
        server: 'db',
      },
      update: {
        client: () => Date.now(),
        server: 'db',
      },
    }),
};

table('user')
  .columns({
    id: string(),
    name: string(),
    created,
    updated,
  })
  .primaryKey('id')
```

2. Regarding renaming `optional`. I agree that `nullable` is a better name. But I worry the rename is going to be a big breaking change. Can we keep supporting `optional` but deprecate it?

3. Regarding disallowing defaults for PK. I'm trying to be less prescriptive with Zero. People sometimes find surprising ways to use it. It's fine to default PKs for some apps depending on the details of the app. If the UI is reactive and the ID doesn't escape the app it works. Also, I can imagine future expansions of this system to allow generating IDs that are stable. We could have a `tx.id(key: string)` function that generates a stable ID for `key` per-mutation. Then we could expose `tx` as an argument to default functions and they could call this to generate IDs. Having access to `tx` might be useful for other reasons too. Long story short I don't think we should disallow this, just document it as perhaps unwise.

You've already put a lot of work into this. If you don't want to make these changes I understand and we can take it over. But if you want to get it in fast, you'll probably be faster than us.

0xcadams:
No worries at all - I will make these changes tomorrow!

0xcadams:
Just updated! To respond to those points:
1. This will allow some interesting use-cases where the client default fn isn't the same as server.
2. This is a good point - I changed it so `optional()` is marked as deprecated.
3. Yeah, passing in the `tx` was something I was thinking about as well. It would be nice to have that - e.g. basing a default value on a value currently seen in the DB. It might be a bit clunky with custom mutators vs CRUD. And stable ID generation would reduce a ton of boilerplate.

To that point, there may be some confusion w/ CRUD vs custom mutators. Probably something to include in the docs?

apps/zbugs/shared/schema.ts
```diff
@@ -26,13 +44,13 @@ const user = table('user')
 const issue = table('issue')
   .columns({
     id: string(),
-    shortID: number().optional(),
+    shortID: number().nullable(),
     title: string(),
     open: boolean(),
-    modified: number(),
-    created: number(),
+    modified,
```
aboodman, line 50: niiiice

packages/zero-client/src/client/crud.ts
```diff
@@ -245,16 +261,42 @@ export function makeCRUDMutator(schema: Schema): CRUDMutator {
...
+      let override: ReadonlyJSONValue | null = null;
+
+      const defaultConfig = schema.columns[name]?.defaultConfig;
+
+      if (
+        operation === 'insert' &&
+        defaultConfig?.insert?.client &&
+        typeof defaultConfig.insert.client === 'function'
+      ) {
+        override = defaultConfig.insert.client() as ReadonlyJSONValue;
```
aboodman: Is this cast safe? Ideally the types require `client()` to return `ReadonlyJSONValue` in which case this cast is unnecessary. But if that's not possible for some reason then we should check it and throw an error otherwise.

packages/zero-pg/src/custom.ts
```diff
@@ -156,29 +156,72 @@ export function makeSchemaCRUD<S extends Schema>(
...
+  const rv: Record<string, unknown> = {};
+
   for (const [key, val] of Object.entries(value)) {
     if (val !== undefined) {
-      valueWithoutUndefined[key] = val;
+      rv[key] = val;
+    }
+  }
+
+  if (type) {
```
aboodman: Why is this function duplicated between custom.ts and crud.ts. Can we share it?

Maybe in packages/zero-schema

packages/zero-client/src/client/crud.ts
```diff
@@ -245,16 +261,42 @@ export function makeCRUDMutator(schema: Schema): CRUDMutator {
...
+      } else if (
+        operation === 'update' &&
+        defaultConfig?.update?.client &&
+        typeof defaultConfig.update.client === 'function'
+      ) {
+        override = defaultConfig.update.client() as ReadonlyJSONValue;
```
aboodman: Same here.

packages/zero-client/src/client/crud.ts
```diff
@@ -324,7 +370,19 @@ export async function updateImpl(
     return;
   }
   const update = arg.value;
+  const defaults = addDefaultToOptionalFields({
```
aboodman, line 334: todo

packages/zero-pg/src/custom.ts
```diff
@@ -156,29 +156,72 @@ export function makeSchemaCRUD<S extends Schema>(
...
+  value,
+  schema,
+  type,
+}: {
+  value: T;
+  schema: TableSchema;
+  type?: 'insert' | 'update';
+}): T {
+  const rv: Record<string, unknown> = {};
+
```
aboodman: It seems like a simpler implementation will be to walk the columns of the schema and check if field in `value` is `undefined`. If it is then use the default? Can eliminate this first loop.

packages/zero-pg/src/custom.ts
```diff
@@ -221,28 +279,38 @@ function makeTableCRUD(schema: TableSchema): TableCRUD<TableSchema> {
...
           ', ',
-        )} WHERE ${primaryKeyClause(schema, serverTableSchema, value)}`,
+        )} WHERE ${primaryKeyClause(schema, serverTableSchema, updateWithDefaults)}`,
       );
       const tx = this[dbTxSymbol];
       await tx.query(stmt.text, stmt.values);
     },
     async delete(this: WithHiddenTxAndSchema, value) {
-      value = removeUndefined(value);
+      const deleteWithDefaults = addDefaultToOptionalFields({
```
aboodman: I do not believe that PK fields are allowed to be optional in Zero. Can you verify this and if so remove this part?

aboodman:
@tantaman can you review this too please.

packages/zero-pg/src/custom.ts
0xcadams: Yes - confirmed that this function wasn't doing anything (`primaryKeyClause` would behave the same regardless of whether undefined keys are filtered), so removed this.

tantaman:
Very nice.

I remember having problems with setting created/modified date within the mutator body which this change effectively does. I'm spending some time understanding why that problem has gone away or it is still a thing.

One high-level question I have: how does this interact with existing CRUD mutators? My assumption is that CRUD mutators will _not_ compute default values on the server and will just respect the default computed by the client.

I'll be glad when CRUD mutators are finally removed.

apps/zbugs/shared/schema.ts
```diff
@@ -12,12 +12,30 @@ import {
 } from '@rocicorp/zero';
 import type {Role} from './auth.ts';
 
+const created = number().default({
+  insert: {
```
tantaman, line 16: did you guys discuss a shorthand for both client & server? e.g., `insert: () => Date.now()`

tantaman (commented):
I'm assuming you did not use the default value mechanism in some places on the issue page because you ran into the infinite loop problem?

I was able to repro the i-loop problem by changing the `assignee` selector to not provide a modified time so it would have to use the default value feature (video attached).

https://github.com/user-attachments/assets/2b880d8a-bf14-4a86-98fe-fe2876aa0422


The reason this happens is because the `prev/next` queries depend on the `modified` time of the currently viewed issue. The currently viewed issue sets a new modified time on each rebase which changes the prev/next queries which cause a new rebase which change the modified time again.

I think you're change is still good though. Separately we should figure out how to resolve this sort of dependency. Ideally changing system metadata (what queries are got and desired) would not cause rebases of mutations. wdyt @grgbkr ?

tantaman:
Thinking through if we should merge the defaults change as-is or do something about the dependency that can arise between a value set in a mutator and a query that uses the value.

apps/zbugs/shared/schema.ts
0xcadams: We didn't - @aboodman what do you think about this?

I like this simplified API.

0xcadams:
Sorry for the delay here - I was traveling this weekend.

> One high-level question I have: how does this interact with existing CRUD mutators? My assumption is that CRUD mutators will not compute default values on the server and will just respect the default computed by the client.

Yeah exactly, the server configuration doesn't apply to CRUD mutators. So this will need to be called out in the docs.

> I'm assuming you did not use the default value mechanism in some places on the issue page because you ran into the infinite loop problem?

Honestly, no, I just missed those code changes on the issue page.

I was just testing this and IIUC the problem above arises from dynamic values computed in the mutation, regardless of this feature or not? (like just calling Date.now() inside the mutator?) I am not able to reproduce it and I think I am not understanding how to cause the issue (see the video below) but open to any fixes you're hoping to see here.

https://screen.studio/share/hEWuWlDY

0xcadams:
Here is the reason this PR was closed:

> aa:
> It's not really that the mutators need to be deterministic. It's just that they can't pick a new value everytime they run. If they do, there are cases where reactivity loops can be easy to accidentally create. 
> 
> As an example of what I mean, a mutator could consult some external datasource to decide what to do. But the decision should be relatively stable. Because everytime it changes it will cause a re-render. If that re-render itself causes Zero to create or change a query, it can cause another render which will re-run the pending mutator, etc.
> The defaults PR was a pathological example - everytime a mutator ran that had defaults, it would pick a new value for Date.now().
> The way this specifically caused a problem in zbugs is that the issue detail page does a query (for the next / prev feature) that includes the current issue row as a param to the query.
> You'd do a mutation to the issue, and this would update the updated field because of the default. This would cause a new prev/next query to get registered. Because of how queries work in Zero, this would cause the pending issue update to rebase (run again). It would then get a new updated value, which would cause a new query, etc.
> This loop continues until the issue update mutation is confirmed b y server.
> So it's not an infinite loop but it does run many times while waiting for server which is confusing.
> We want to solve this by stopping the new query from causing a rebase but that is a bigger change that we can take on right now.

PR #4553, opened 2025-06-30
Revert "Revert "chore(zero): Add activeClients to initConnection (#45…

arv:
…40)" (#4544)"

This reverts commit f996bbb77d078009c52716183dec7e7fa7d7a541.


arv:
Revert of the revert of the revert

The reason it was reverted was because the deployment of zbugs was not working (due to an sql error in a different pr).

PR #4555, opened 2025-06-30
Support https protocol for change streamer

tjenkinson:


packages/zero-cache/src/services/change-streamer/storer.pg-test.ts
tjenkinson, line 1: Recommend ignoring whitespace. Wrapped all the existing tests in another describe so can have one for http and the other for https

packages/zero-cache/src/services/change-streamer/change-streamer-http.ts
```diff
@@ -155,7 +155,12 @@ export class ChangeStreamerHttpClient implements ChangeStreamer {
     if (!address) {
       throw new Error(`no change-streamer is running`);
     }
-    const uri = new URL(path, `http://${address}/`);
+    const uri = new URL(
```
tjenkinson, line 158: not tested currently. Not sure on the best approach given the current tests rely running a local server. Maybe need to start a https server too and disable certificate validation?

packages/zero-cache/src/services/change-streamer/change-streamer-http.ts
```diff
@@ -155,7 +155,12 @@ export class ChangeStreamerHttpClient implements ChangeStreamer {
...
+      path,
+      address.startsWith('http://') || address.startsWith('https://')
```
darkgnotic: Maybe: `address.includes("://")` so that it takes any scheme/protocol into account? Not that we expect people to use it, but ws:// and wss:// would also be valid.

packages/zero-cache/src/services/change-streamer/storer.ts
```diff
@@ -132,8 +135,12 @@ export class Storer implements Service {
     const db = this.#db;
     const owner = this.#taskID;
     const ownerAddress = this.#discoveryAddress;
-    await db`UPDATE ${this.#cdc('replicationState')} SET ${db({owner, ownerAddress})}`;
-    this.#lc.info?.(`assumed ownership at ${ownerAddress}`);
+    const ownerProtocol = this.#discoveryProtocol;
+    // we omit `http://` so that old view syncer versions that are not expecting the protocol continue to not get it
```
darkgnotic: Well thought out. Thank you!

darkgnotic (changes requested):

packages/zero-cache/src/services/change-streamer/change-streamer-http.ts
tjenkinson: Hey @darkgnotic !

Sure can update tomorrow. Should I allow the others in the config too then?

tjenkinson: Actually is http/https even valid for a websocket? Maybe it should be ws/wss?

tjenkinson: Looks like really it should be `ws`/`wss` but TIL support for `http`/`https` was added to support relative urls https://github.com/whatwg/websockets/pull/45 https://github.com/websockets/ws/pull/2162

I also had ```/^[a-zA-Z0-9+\-.]+:/.test(address) ? `${address}/` : `ws://${address}/`,``` locally to accept any protocol here, but if the config only accepts `ws`/`wss` maybe it's better to be stricter?

packages/zero-cache/src/services/change-streamer/storer.ts
```diff
@@ -132,8 +135,12 @@ export class Storer implements Service {
     const db = this.#db;
     const owner = this.#taskID;
     const ownerAddress = this.#discoveryAddress;
-    await db`UPDATE ${this.#cdc('replicationState')} SET ${db({owner, ownerAddress})}`;
-    this.#lc.info?.(`assumed ownership at ${ownerAddress}`);
+    const ownerProtocol = this.#discoveryProtocol;
+    // we omit `ws://` so that old view syncer versions that are not expecting the protocol continue to not get it
+    const addressWithProtocol =
+      ownerProtocol === 'wss' ? `wss://${ownerAddress}` : ownerAddress;
```
darkgnotic: I think something along the lines of what you had before is more flexible, in case we need to bring `http` or `https` back.

```ts
ownerProtocol === 'ws' ? ownerAddress : `${ownerProtocol}://${ownerAddress}`
```

packages/zero-cache/src/services/change-streamer/change-streamer-http.ts
```diff
@@ -155,7 +155,12 @@ export class ChangeStreamerHttpClient implements ChangeStreamer {
     if (!address) {
       throw new Error(`no change-streamer is running`);
     }
-    const uri = new URL(path, `http://${address}/`);
+    const uri = new URL(
+      path,
+      address.startsWith('ws://') || address.startsWith('wss://')
```
darkgnotic: Limiting the `protocol` value to `ws` and `wss` in the option definition seems fine, but I would prefer leaving this part to be a scheme agnostic check of `includes('://')`. 

Juuust in case something out there depends on the scheme being http. Hopefully nothing does.  🤞

darkgnotic (approved):
Thank you @tjenkinson !

packages/zero-cache/src/services/change-streamer/storer.ts
tjenkinson: ah yep this is better. done

packages/zero-cache/src/services/change-streamer/change-streamer-http.ts
tjenkinson: Done! Although I'm not sure how something else could get into the db?

darkgnotic: The idea is that it would be easier to add more options if necessary (fewer places in the code to modify).

Thank you! I'll merge this!

PR #4556, opened 2025-06-30
chore(zero-cache): Inactive queries on connect

arv:
When a client connects it sends the active clients in the client group. Based on this we can figure out what clients are inactive and we can inactive the queries that belong to those clients.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/view-syncer.pg-test.ts
+++ b/packages/zero-cache/src/services/view-syncer/view-syncer.pg-test.ts
@@ -452,6 +452,33 @@ const permissionsAll = await definePermissions<AuthData, typeof schema>(
   }),
 );
 
+async function nextPoke(client: Queue<Downstream>): Promise<Downstream[]> {
+  const received: Downstream[] = [];
+  for (;;) {
+    const msg = await client.dequeue();
+    received.push(msg);
+    if (msg[0] === 'pokeEnd') {
+      break;
+    }
+  }
+  return received;
+}
+
+async function nextPokeParts(
+  client: Queue<Downstream>,
+): Promise<PokePartBody[]> {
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/services/view-syncer/view-syncer.pg-test.ts
```diff
@@ -1741,6 +1734,344 @@ describe('view-syncer/service', () => {
...
+    const {source: newStreamA, queue: newClientA} = connectWithQueueAndSource(
+      newCtxA,
+      [{op: 'put', hash: 'query-hashA', ast: ISSUES_QUERY, ttl}],
+      undefined,
+      ['clientA', 'clientB'],
+    );
+
+    // desired +A +B -C
+    // got +A +B +C
+    // rows +A +B +C -- why is it sending the rows again?
```
arv: @darkgnotic This does not seem correct to me. Maybe this test is not setting things up correctly.

darkgnotic: I think what's going on is that `connectWithQueueAndSource()` by default connects with `baseCookie: null`, indicating that the client is starting from scratch:

https://github.com/rocicorp/mono/blob/10ad9e2bba0defc7da6573979f29ab900188ba06/packages/zero-cache/src/services/view-syncer/view-syncer.pg-test.ts#L730

To simulate clientA reconnecting from its last state, you can set the `baseCookie` value to whatever the `cookie` value was returned in the last poke it received.

arv: Ah, that makes sense. I will try that.

arv: That was it!

Thanks

PR #4557, opened 2025-06-30
chore(zero): Change default TTL to 1s/5s

arv:
For normal queries the default TTL is now 1s (used to be 0).

For preload queries the default TTL is now 5s (used to be 0).


tantaman:
the expectation for typeahead search is now that the user should explicitly set a 0 ttl?

arv:
> the expectation for typeahead search is now that the user should explicitly set a 0 ttl?

Yes, @aboodman's solution to up the ttl when the search settles makes things work nicely.

Setting the default to 1s allows in app navigation without churn for the rows.

PR #4558, opened 2025-06-30
fix(zero): Inspector queries were broken

arv:
When we added custom queries we changed the type of the ast column to nullable but the inspector code was not updated to handle that. This commit updates the inspector code to handle the nullable ast column as well as the name and args columns that were added when custom queries were added.


arv:
@tantaman Do we keep the last AST somewhere? It would be nice to show what the client and server ASTs were.

tantaman:
> @tantaman Do we keep the last AST somewhere? It would be nice to show what the client and server ASTs were.

ASTs are not currently stored for custom queries.

PR #4559, opened 2025-06-30
refactor(zero-solid): reconcile changes to data due to query changes

grgbkr:
Starting from this PR contribution: https://github.com/rocicorp/mono/pull/4474

Prior to this change the zero-solid integration created a new store whenever a query changed.
This caused everything depending on the store to be disposed and recreated, instead of receiving granular updates.
This PR ensures that every `createQuery` call only creates a store the first time and passes on the setter to the view, and the view is responsible for reconciling data when there's large changes.

- Add a `ZeroProvider` and `useZero` so that the zero instance can be accessed through context, `ZeroProvider` can take a signal for a `ZeroOptions` or a `Zero` instance.
- `useZero` return a signal of `Zero` instance since the signal to ZeroProvider can change the `ZeroOptions` or `Zero` instance.
- Un-deprecate `useQuery` since it now requires context. Deprecates `createQuery` instead. `useX` is a good convention to communicate that this should be called inside a component and not outside.
- Added `createUseZero` to prevent redundant annotations, and so the api mirror's React's apis much more closely.

Next steps:
- Use solid's `reconcile` utility to achieve granular updates for they initial hydration of a changed query.  This will require adding synthetic ids to rows.
- Update docs


grgbkr:
FYI @devagrawal09

devagrawal09:
Thanks for creating this PR. What's the plan for adding ids?
We don't need every row inside the view tree to be identifiable, we only need objects inside the same array to be identifiable among its siblings. it's okay for nested stuff to have the same ids.

Also is it possible that when a new query is constructed, the data returned from zero has the same references? We might not need ids if the data is always kept in memory by zero and reused across queries, because `reconcile` understands references and is able to key them automatically.

arv:
> same references

We can definitely do same references. That might be simpler and more performant

packages/zero-solid/package.json
```diff
@@ -20,6 +20,8 @@
     "@rocicorp/eslint-config": "^0.7.0",
     "@rocicorp/prettier-config": "^0.3.0",
     "@solidjs/testing-library": "^0.8.10",
+    "jsdom": "^26.1.0",
```
arv: Why? We are running the tests in a real browser, no?

packages/zero-solid/src/solid-view.ts
```diff
@@ -26,17 +20,16 @@ export type QueryResultDetails = {
   readonly type: ResultType;
 };
 
-type State = [Entry, QueryResultDetails];
+export type State = [Entry, QueryResultDetails];
 
-const complete = {type: 'complete'} as const;
-const unknown = {type: 'unknown'} as const;
+export const COMPLETE: QueryResultDetails = {type: 'complete'} as const;
```
arv: Maybe we should freeze these to prevent more surprises

packages/zero-solid/src/solid-view.ts
```diff
@@ -73,29 +67,25 @@ export class SolidView<V> implements Output {
       node => ({type: 'add', node}),
       initialRoot,
     );
-    [this.#state, this.#setState] = createStore<State>([
+
+    this.#setState = setState;
+    const s = [
```
arv: inline?

packages/zql/src/query/query-impl.ts
arv, line 1: remove commented out logs

packages/zero-solid/vitest.config.ts
```diff
@@ -1 +1,5 @@
-export {default} from '../shared/src/tool/vitest-config.ts';
+import solid from 'vite-plugin-solid';
+import {defineConfig, mergeConfig} from 'vitest/config';
+import config from '../shared/src/tool/vitest-config.ts';
+
+export default mergeConfig(config, defineConfig({plugins: [solid()]}));
```
arv, line 5: Is this needed? Maybe we were not using jsx in the tests before?

packages/zero-solid/src/solid-view.test.ts
```diff
@@ -642,6 +687,14 @@ test('collapse', () => {
     commit = cb;
   };
 
+  const [state, setState] = createStore<State>([
+    {
+      '': undefined,
+    },
+    {type: 'unknown'},
+  ]);
+  const data = () => state[0][''];
```
arv, line 696: Maybe const {data, state, setState} = createTestStore()

packages/zero-solid/src/use-zero.ts
```diff
@@ -0,0 +1,77 @@
...
+>() {
+  return () => useZero<S, MD>();
+}
+
+export function ZeroProvider<
+  S extends Schema,
+  MD extends CustomMutatorDefs<S> | undefined = undefined,
+>(props: {
+  children: JSX.Element;
+  zeroSignal: () => ZeroOptions<S, MD> | {zero: Zero<S, MD>};
```
arv: I find this type a bit strange. I want the props to be called zero and options.

Is this how the React one is done?

packages/zero-solid/src/use-zero.ts
```diff
@@ -0,0 +1,77 @@
...
+}) {
+  const zero = createMemo(() => {
+    const z = props.zeroSignal();
```
arv: So zeroSignal must be present?

packages/zero-solid/src/use-zero.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+test('if zeroSignal change ZeroProvider closes previous instance if it created it', () => {
+  const [zeroOptions, setZeroOptions] = createSignal({
+    server: 'foo',
+    userID: 'u',
+    schema: {tables: {}, relationships: {}},
+  });
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  const MockZeroProvider = (props: {children: JSX.Element}) => (
+    <ZeroProvider zeroSignal={zeroOptions}>{props.children}</ZeroProvider>
```
arv:
```suggestion
    <ZeroProvider options={zeroOptions}>{props.children}</ZeroProvider>
```

I hope

packages/zero-solid/src/use-zero.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+test('if zeroSignal changes ZeroProvider closes previous instance if it created it', () => {
+  const fakeZero0 = new FakeZero() as unknown as Zero<Schema>;
+  const [zeroSignal, setZeroSignal] = createSignal<
+    ZeroOptions<Schema, undefined> | {zero: Zero<Schema, undefined>}
+  >({
+    zero: fakeZero0,
+  });
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  const MockZeroProvider = (props: {children: JSX.Element}) => (
+    <ZeroProvider zeroSignal={zeroSignal}>{props.children}</ZeroProvider>
```
arv:
```suggestion
    <ZeroProvider zero={zeroSignal}>{props.children}</ZeroProvider>
```

packages/zql/src/query/query-impl.ts
grgbkr: done

packages/zero-solid/package.json
grgbkr: removed

packages/zero-solid/src/solid-view.ts
grgbkr: done.  this actually modifies solid's behavior as it checks for frozen objects and copies them, but i think its best not to rely on it.

packages/zero-solid/src/solid-view.ts
grgbkr: done

packages/zero-solid/src/use-zero.test.tsx
grgbkr: i made this match the React version, so it would be something like:
```
<ZeroProvider server="https://foo.com" userID="u1" schema={schema}>
```

packages/zero-solid/src/use-zero.test.tsx
grgbkr: 👍

packages/zero-solid/src/use-zero.ts
grgbkr: I fixed this to more closely match the React API.  So its either:

```
<ZeroProvider zero={zero}>...</ZeroProvider>
```

or

```
<ZeroProvider server="https://foo.com" userID="u1" schema={schema}>...</ZeroProvider>
```

packages/zero-solid/vitest.config.ts
grgbkr: exactly, added because we now are using jsx in the tests.

packages/zero-solid/src/use-zero.ts
```diff
@@ -0,0 +1,77 @@
...
+      ...z,
+      batchViewUpdates: batch,
+    });
+    onCleanup(() => createdZero.close());
+    return createdZero;
+  });
+
+  return ZeroContext.Provider({
+    value: zero,
+    get children() {
```
grgbkr, line 79: so it seems this getter for children was the secret sauce... but why?

Why does the following not work
```
return ZeroContext.Provider({
  value: zero,
  children: props.children
  });
  
```

grgbkr: Should value also be a getter?  There is still something fundamental about solidjs that i dont grok.

packages/zero-solid/src/use-zero.ts
```diff
@@ -0,0 +1,77 @@
...
+      return props.children;
+    },
+  });
```
devagrawal09, line 82:
```suggestion
  return <ZeroContext.Provider value={zero()}>
    {props.children}
  </ZeroContext.Provider>
```
You should be able to do just this

devagrawal09: https://docs.solidjs.com/concepts/context#context

grgbkr: We need to avoid using JSX here as esbuild doesn't seem able to build a library that uses both React JSX and Solid JSX (at least we couldn't figure it out).

packages/zero-solid/src/use-zero.ts
devagrawal09: Ah you can't use JSX here

In that case the getter is correct.
value would also need to be a getter.

In solid when you access something from the props like `props.value` inside a reactive scope, it needs to behave similar to accessing a signal `value()` so that it can be tracked as a dependency. So the JSX transform changes `<A b={c()} />` to `A({ get b() { return c{} } })`.

This makes sure that every property passed through the component tree can be tracked independently.

grgbkr:
> Thanks for creating this PR. What's the plan for adding ids? We don't need every row inside the view tree to be identifiable, we only need objects inside the same array to be identifiable among its siblings. it's okay for nested stuff to have the same ids.
> 
> Also is it possible that when a new query is constructed, the data returned from zero has the same references? We might not need ids if the data is always kept in memory by zero and reused across queries, because `reconcile` understands references and is able to key them automatically.

We do return the same references for rows that haven't changed.  Let me do some testing and see if i can reconcile based on references working.

packages/zero-solid/src/use-zero.ts
arv: `children` needs to be a getter to get the state tracking.

In the case of `value`, `zero`  is already a signal so there is no need to use a getter.

arv:
> We do return the same references for rows that haven't changed. Let me do some testing and see if i can reconcile based on references working.

We should also be able to "merge" the rows that changed if the primary keys didn't change. This can be dealt with in `applyChange`.

packages/zero-solid/src/use-query.test.tsx
```diff
@@ -0,0 +1,456 @@
...
+    resultDetailsLog.length = 0;
+  };
+
+  runWithOwner(renderHookResult.owner, () => {
+    createEffect(() => {
+      rowLog.push(rows());
+    });
+    createEffect(() => {
+      const details = resultDetails();
+      details.type;
```
arv, line 169: no op?

arv: I guess it does set up the tracking?

packages/zero-solid/src/use-zero.test.tsx
arv, line 1: Very nice and extensive tests!

packages/zero-solid/src/use-zero.ts
```diff
@@ -0,0 +1,83 @@
...
+  const zero = useContext(ZeroContext);
+
+  // TODO: uncomment when we require ZeroProvider in a future release.
+  // if (zero === undefined) {
+  //   throw new Error('useZero must be used within a ZeroProvider');
+  // }
+  return zero ?? (() => undefined);
+}
+
+export function createUseZero<
```
arv, line 46: This could use some JSDoc.

grgbkr:
> > Thanks for creating this PR. What's the plan for adding ids? We don't need every row inside the view tree to be identifiable, we only need objects inside the same array to be identifiable among its siblings. it's okay for nested stuff to have the same ids.
> > Also is it possible that when a new query is constructed, the data returned from zero has the same references? We might not need ids if the data is always kept in memory by zero and reused across queries, because `reconcile` understands references and is able to key them automatically.
> 
> We do return the same references for rows that haven't changed. Let me do some testing and see if i can reconcile based on references working.

@devagrawal09 @arv 
The reference based approach isn't going to work due to some other work I'm doing to modify rows in place to achieve sub-row level precise reactivity.  I got the idSymbol w synthetic id working.  https://github.com/rocicorp/mono/pull/4575

packages/zero-solid/src/use-query.test.tsx
grgbkr: yeah necessary to set up the tracking

arv:
@devagrawal09 Should we file a bug/pr on Solid to allow the key to be a symbol?

PR #4561, opened 2025-07-01
chore(zero-client): Send deleteClient when client is deleted

arv:
This uses the `ActiveClientsManager` to get notified when a client becomes inactive (deleted) and sends a `deleteClient` message to the zero cache server.

This is in addition to the existing mechanism that Replicache uses to delete clients when they are garbage collected from IDB.


packages/zero-client/src/client/zero.test.ts
```diff
@@ -3599,3 +3600,38 @@ test('onError is called on error', async () => {
     ]
   `);
 });
+
+test('We should send a deleteClient when a Zero instance is closed', async () => {
```
grgbkr, line 3604: nice

grgbkr (commented):
LGTM

grgbkr:
Is the server handling of deleteClient resilient to receiving a delete for the same clientID multiple times (its going to get it from multiple tabs and potentially later from the replicache process)?

arv:
> Is the server handling of deleteClient resilient to receiving a delete for the same clientID multiple times (its going to get it from multiple tabs and potentially later from the replicache process)?

I think there is a bug there... I saw an assert flying by that seems related to this.

PR #4562, opened 2025-07-01
fix(zql): handle `null` compare in `table-source`

tantaman:
todo: adding `push` to the fuzzer as that probably would have caught this.

### DIFF
--- a/packages/zqlite/src/table-source.ts
+++ b/packages/zqlite/src/table-source.ts
@@ -490,7 +490,7 @@ export class TableSource implements Source {
     if (constraint) {
       for (const [key, value] of Object.entries(constraint)) {
         constraints.push(
-          sql`${sql.ident(key)} = ${toSQLiteType(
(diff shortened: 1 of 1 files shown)


aboodman:
Maybe there should be some tests that cover the filter cases too?

tantaman:
there's still a bug in the `limit 1` case 🤔

tantaman:
the `limit 1` case is a separate problem and also broken in memory-source zql.

packages/zql-integration-tests/src/chinook/chinook.pg-test.ts
```diff
@@ -535,6 +535,44 @@ describe(
               },
             };
           })(),
+          // ordering by nullable columns
+          {
+            name: 'Order by nullable column',
+            createQuery: q => q.customer.orderBy('company', 'desc').limit(10),
+          },
+          // TODO: case below is broken.
```
tantaman, line 543: this is breakage is unrelated to the current change and is broken on main as well.

PR #4563, opened 2025-07-01
fix(zql): error in `take` when bound node is edited

tantaman:
When the bound node was edited to no longer be inside the window, take would not push a remove for that node.

Example:

```ts
const employees = new MemorySource([{
  id: 1,
  name: 'April',
}, {
  id: 2,
  name: 'BApril',
}]);

const view = z.employees.limit(1).orderBy('name', 'asc').materialize();

// view is:
[{
  id: 1,
  name: 'April',
}];

// now edit april
employees.push({
  type: 'edit',
  newRow: {
    id: 1,
    name: 'ZApril',
  },
  oldRow: {
    id: 1,
    name: 'April',
  }
});

// the view (before this fix) would be:
[{
  id: 1,
  name: 'April',
}, {
  id:2,
  name: 'BApril'
}];

// because only the `add` change for the node that fell into the window
// would be pushed.
// The `remove` change for the node to be ejected was not pushed.
```

### DIFF
--- a/packages/zql/src/ivm/take.ts
+++ b/packages/zql/src/ivm/take.ts
@@ -498,6 +498,12 @@ export class Take implements Operator {
         newBoundNode.row,
         maxBound,
       );
+      this.#withRowHiddenFromFetch(newBoundNode.row, () => {
+        this.#output.push({
+          type: 'remove',
+          node: change.oldNode,
+        });
+      });
       this.#output.push({
         type: 'add',
         node: newBoundNode,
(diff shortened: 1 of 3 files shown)


packages/zql/src/ivm/take.ts
```diff
@@ -498,6 +498,12 @@ export class Take implements Operator {
         newBoundNode.row,
         maxBound,
       );
+      this.#withRowHiddenFromFetch(newBoundNode.row, () => {
```
grgbkr, line 501: Wow the comment above has it right, but the code doesn't do what the comment says.

Looks like we had a unit test for this, but the test was also wrong, see
`edit row at boundary, making it fall outside the window`

thats part of the 'take with partition' > 'edit' suite.  

It would be good to add a test for this to the 'take with no partition' > 'edit' suite as well.

PR #4564, opened 2025-07-02
chore(zero-cache): Fix a really massive log line.

aboodman:
I theorize this is causing an OOM for Erik Munson. The CVR can be massive.


arv (commented):
LGTM

aboodman:
To follow up this was the source of the OOM. Need to be careful with what
we log.

PR #4565, opened 2025-07-02
fix: added type param to custom mutators

0xcadams:
This is to enable passing types similar to @jokull's blog post here: https://www.solberg.is/zero-custom-db-connection

```tsx
{...} as const satisfies CustomMutatorDefs<
  Schema,
  PostgresJsZeroTransaction<typeof db>
>;
```

Server mutators:

```tsx
if (tx.location === 'server') {
  const user =
    await tx.dbTransaction.wrappedTransaction.query.User.findFirst({
      where: eq(User.id, auth.sub),
    });

  if (!user) {
    throw new Error('User not found');
  }
}
```

Screenshot 2025-07-02 at 11 17 35 AM


packages/zero-client/src/client/custom.ts
```diff
@@ -32,22 +32,29 @@ import type {ZeroLogContext} from './zero-log-context.ts';
 /**
  * The shape which a user's custom mutator definitions must conform to.
  */
-export type CustomMutatorDefs<S extends Schema> = {
+export type CustomMutatorDefs<
+  S extends Schema,
+  TWrappedTransaction = unknown,
```
tantaman, line 37: I do worry that this will explode the size of the type for mutators but that will be fixed soon by this refactor (https://github.com/rocicorp/mono/pull/4570) if everyone agrees on the new approach.

Otoh, people can drop back to `unknown` if it is a problem.

tantaman:
Thanks. This has been on my list of things to do for a long time.

PR #4566, opened 2025-07-02
chore: Update vitest to 5.2

arv:


packages/zql/src/ivm/fan-out-fan-in.test.ts
```diff
@@ -291,7 +291,7 @@ test('cleanup forwards too all branches', () => {
 
   const filterSpy1 = vi.spyOn(filter1, 'filter');
   const filterSpy2 = vi.spyOn(filter2, 'filter');
-  const filterSpy3 = vi.spyOn(filter2, 'filter');
+  const filterSpy3 = vi.spyOn(filter3, 'filter');
```
arv, line 294: @grgbkr This look a bug in the original test.

grgbkr: agree, did you just happen to notice, or did this upgrade expose it some how?

grgbkr:
LGTM

PR #4569, opened 2025-07-02
feat(zero): custom mutator api that matches custom queries

tantaman:
This is the backend (push-processor) API. Frontend is currently unchanged.

The old API still exists and is implemented by calling the new API.


tantaman:
- being superseded by https://github.com/rocicorp/mono/pull/4699 and related

PR #4570, opened 2025-07-02
feat(zbugs) - use custom mutator API that matches custom queries

tantaman:
Swaps to a mutator definiton format like:

```ts
export const editComment = (authData: AuthData | undefined) =>
  mutator(
    'editComment',
    async (tx: Transaction<Schema>, {id, body}: {id: string; body: string}) => {
      await assertIsCreatorOrAdmin(authData, tx.query.comment, id);
      await tx.mutate.comment.update({id, body});
    },
  );
```

rather than:

```ts
function createMutators(authData: AuthData | undefined) {
  return {
    comment: {
      edit(tx, {id, body}: {id: string; body: string}) => {
        await assertIsCreatorOrAdmin(authData, tx.query.comment, id);
        await tx.mutate.comment.update({id, body});
      }
    }
  }
}
```

## Pros:

1. The new format supports code splitting

The `mutators` arg on `zero.options` would technically need to change to:

```ts
type ZeroOptions = {
  ...,
  mutators: (name: string) => Promise<Mutator>
};
```

where the user provides a lookup function from name to mutator. Allowing them to import the associated module(s) as needed.

Also useful in serverless so only the relevant mutator code for the current request needs to be loaded.

2. Matches the CustomQuery API
3. Allows removal of the second generic param to Zero. E.g., `Zero<S, MD>` becomes `Zero<S>`.
4. Solves the typescript "type too big" problem as there is no longer a single bag of all mutators
5. Allows the user to organize their mutator implementations however they desire
6. Solves the "go to definition" problem

https://github.com/user-attachments/assets/c83fa893-8cc7-4369-8bab-cd17cf5208b5

## Cons:

1. Redundancy in name. E.g., "editComment" is repeated.

```ts
export const editComment = (authData: AuthData | undefined) =>
  mutator(
    'editComment',
```

2. The user must manually route to the correct mutator implementation on the server.

CleanShot 2025-07-02 at 13 48 27


tantaman:
- to be replaced by https://github.com/rocicorp/mono/pull/4699 and related

PR #4571, opened 2025-07-02
test(zql): update to check pushesWithFetch

grgbkr:
Follow up fixing some test changes in https://github.com/rocicorp/mono/pull/4563


grgbkr:
@tantaman Maybe the test helpers could be tweaked to make this less error prone...

PR #4572, opened 2025-07-02
feat(zero): add drizzle adapter, clean up postgresjs adapter

tantaman:
Simplify custom mutator setup by providing turn-key adapters.

Our existing postgresjs adapter (packages/zero-pg/src/postgresjs-connection.ts) was confusing so this cleans that up (deletes some code and adds example usage) in addition to adding a drizzle adapter.


apps/zbugs/server/push-handler.ts
```diff
@@ -1,19 +1,13 @@
...
 import {createServerMutators, type PostCommitTask} from './server-mutators.ts';
 import type {AuthData} from '../shared/auth.ts';
 import type {ReadonlyJSONValue} from '@rocicorp/zero';
 
 const processor = new PushProcessor(
-  new ZQLDatabase(
-    new PostgresJSConnection(postgres(process.env.ZERO_UPSTREAM_DB as string)),
-    schema,
-  ),
+  zeroPostgresJS(schema, postgres(process.env.ZERO_UPSTREAM_DB as string)),
```
arv: If we are making this tied to `postgres` npm package, can we put that inside `zeroPostgresJS`?

packages/zero-server/package.json
```diff
@@ -23,6 +23,8 @@
   "devDependencies": {
     "@rocicorp/prettier-config": "^0.3.0",
     "@vitest/runner": "3.2.4",
+    "drizzle-orm": "^0.44.0",
+    "postgres": "^3.4.0",
```
arv: Do we also want peer dependencies? And make sure that these are part of the `@rocicorp/zero` npm package too.

arv (approved):
LGTM

I like it

packages/zero-server/package.json
tantaman: I may need your help figuring out peer deps. When I stuck `drizzle` into peer dependencies here I ended up getting errors about React versions conflicting of all things 🤣

Turns out drizzle has a peer dependency on expo which has a peer dependency on react native which has a peer dependency on react. This conflicts with some docusaurus peer dependencies on react. This wouldn't be a problem in practice but something about the monorepo setup makes it a problem.

apps/zbugs/server/push-handler.ts
tantaman: Yeah, good idea.

tantaman: I think I'll support both options.

1. if the user has a connection already established they can pass it 
2. otherwise, they can just pass the connection url and we'll make the connection

tantaman:
Updated the Drizzle adapter to follow @jokull's approach: https://www.solberg.is/zero-custom-db-connection

0xcadams:
Funny, I just landed something this week very similar in drizzle-zero - but glad you're adding upstream!
https://github.com/0xcadams/drizzle-zero/blob/main/src/node-postgres/index.ts
https://github.com/0xcadams/drizzle-zero/blob/main/src/postgres-js/index.ts

One thing - drizzle supports multiple PG clients. So it might be good to namespace them.

packages/zero-server/package.json
arv: Let's keep it as a devDependency for now and see how that works out.

tantaman:
Good call @0xcadams. Does this otherwise look correct from a drizzle perspective? I got the idea for providing a set of adapters from better-auth as that made the onboarding experience for that pretty seamless.

packages/zero/package.json
```diff
@@ -113,6 +113,14 @@
       "types": "./out/zero-server/src/mod.d.ts",
       "default": "./out/zero/src/server.js"
     },
+    "./server/adapters/postgresjs": {
+      "types": "./out/zero-server/src/adapters/postgres.d.ts",
+      "default": "./out/zero-server/src/adapters/postgresjs.js"
+    },
+    "./server/adapters/drizzle": {
```
0xcadams: I think there should be multiple exports here - separating out `node-postgres`, drizzle `node-postgres`, plain `postgres.js`, drizzle `postgres.js`. This would map better onto Drizzle adapters since the underlying client/types are different between those drizzle adapters.

They also have `neon-postgres` and `vercel-serverless`, which users may want as well.

https://github.com/drizzle-team/drizzle-orm/tree/main/drizzle-orm/src

jokull:
I wouldn't be upset if I wasn't credited in the release notes, but I also would be pretty ecstatic if that happened. Just saying :) :) :)

PR #4575, opened 2025-07-03
perf(zero-solid): use reconcile to minimize reactive changes on query changes

grgbkr:
This is a follow up to: https://github.com/rocicorp/mono/pull/4559

To effectively use solidjs's `reconcile` utility it was necessary to add a consistent identity key to every row in the view.


packages/zql/src/ivm/view-apply-change.ts
```diff
@@ -12,6 +12,7 @@ import type {SourceSchema} from './schema.ts';
 import type {Entry, Format} from './view.ts';
 
 export const refCountSymbol = Symbol('rc');
+export const idSymbol = Symbol('id');
```
grgbkr, line 15: Or should this be 
`export const keySymbol = Symbol('key');`

packages/zero-solid/src/use-query.test.tsx
```diff
@@ -194,6 +198,401 @@ test('useQuery query deps change', async () => {
   expect(resultDetailsLog).toEqual([{type: 'complete'}]);
 });
 
+test('useQuery query deps change, reconcile minimizes reactive updates', async () => {
```
grgbkr, line 201: This and the next test cases are new and verify this is giving us the precise reactive updates we want.

packages/zero-solid/src/use-query.test.tsx
```diff
@@ -194,6 +198,401 @@ test('useQuery query deps change', async () => {
...
+
+  queryDelegate.gotCallbacks.forEach(cb => cb?.(true));
+  await 1;
+
+  expect(row0Log).toEqual([]);
+  expect(row1Log).toEqual([]);
+  expect(resultDetailsLog).toEqual([{type: 'complete'}]);
+  resetLogs();
+
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
grgbkr, line 250: I don't know why the linter is getting this wrong... this method does not return a promise.

grgbkr:
FYI @devagrawal09

packages/zql/src/ivm/view-apply-change.ts
```diff
@@ -412,6 +417,19 @@ function getSingularEntry(parentEntry: Entry, relationship: string): RCEntry {
...
-function makeNewEntryWithRefCount(row: Row, rc: number): RCEntry {
-  return {...row, [refCountSymbol]: rc};
+function makeNewEntryWithRefCount(
+  row: Row,
+  schema: SourceSchema,
+  withIDs: boolean,
+  rc: number,
+): RCEntry {
+  const id = withIDs ? makeID(row, schema) : '';
+  return {...row, [refCountSymbol]: rc, [idSymbol]: id};
```
grgbkr, line 427: @arv I know you were concerned with polymorphic object shape.  Does this look ok?  Would it be ok to make the [idSymbol] optional?

packages/zql/src/ivm/view-apply-change.ts
arv: I don't really care. This is fine

packages/zql/src/ivm/view-apply-change.ts
arv: I think we do not need to add the symbol if `withIDs` is `false`. In non SolidJS it will always be `false` and in SolidJS it will always be `true`.

arv (commented):
LGTM

I don't think we should add the idSymbol for react. It is pretty confusing to have that there and it always being an empty string.

grgbkr:
> LGTM
> 
> I don't think we should add the idSymbol for react. It is pretty confusing to have that there and it always being an empty string.

Great, I will send a change to clean this up.

PR #4576, opened 2025-07-04
perf(zql, zero-solid): edit entries rather than replacing them 

grgbkr:
Edit entries rather than replacing them, this has two benefits:
1. enables sub-row level fine grain reactivity in SolidJS.  Currently minimum granularity of update is a whole row, now SolidJS can correctly narrowly react to single column changes.
2. reduces Object churn and thus GC

We already edit the relationship arrays in place.  

This is safe because:

- in the React bindings we deeClone the view's data.
- in the SolidJS binding we use `produce`, which implements an immer like copy on write.

### DIFF
--- a/packages/zql/src/ivm/view-apply-change.ts
+++ b/packages/zql/src/ivm/view-apply-change.ts
@@ -14,7 +14,10 @@ import type {Entry, Format} from './view.ts';
 export const refCountSymbol = Symbol('rc');
 export const idSymbol = Symbol('id');
 
-type RCEntry = Writable<Entry> & {[refCountSymbol]: number};
+type RCEntry = Writable<Entry> & {
+  [refCountSymbol]: number;
+  [idSymbol]: string;
+};
 type RCEntryList = RCEntry[];
 
 /**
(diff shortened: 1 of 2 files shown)


grgbkr:
FYI @devagrawal09 @thdxr

packages/zero-solid/src/solid-view.test.ts
```diff
@@ -1375,6 +1377,151 @@ test('basic with edit pushes', () => {
...
+    commit = cb;
+  };
+
+  const [state, setState] = createStore<State>([
+    {
+      '': undefined,
+    },
+    {type: 'unknown'},
+  ]);
+  const data: () => EntryList = () => state[0][''] as EntryList;
```
arv:
```suggestion
  const data: () => EntryList = () => state[0][''];
```

or

```suggestion
  const data = () => state[0][''] as EntryList;
```

packages/zero-solid/src/solid-view.test.ts
```diff
@@ -1375,6 +1377,151 @@ test('basic with edit pushes', () => {
...
+  const data3 = [
+    {a: 0, b: 'b3', [refCountSymbol]: 1, [idSymbol]: '0'},
+    {a: 1, b: 'a', [refCountSymbol]: 1, [idSymbol]: '1'},
+  ];
+  expect(data()).toEqual(data3);
+  expect(row0Log).toEqual([
+    {a: 0, b: 'b3', [refCountSymbol]: 1, [idSymbol]: '0'},
+  ]);
+  expect(row1Log).toEqual([
+    {a: 1, b: 'a', [refCountSymbol]: 1, [idSymbol]: '1'},
```
arv, line 1517: Why did this change?

packages/zql/src/ivm/view-apply-change.ts
```diff
@@ -14,7 +14,10 @@ import type {Entry, Format} from './view.ts';
 export const refCountSymbol = Symbol('rc');
 export const idSymbol = Symbol('id');
 
-type RCEntry = Writable<Entry> & {[refCountSymbol]: number};
+type RCEntry = Writable<Entry> & {
```
arv: Maybe rename to MetaEntry?

packages/zql/src/ivm/view-apply-change.ts
```diff
@@ -233,55 +239,59 @@ export function applyChange(
...
-            schema,
-            withIDs,
-            rc,
+            schema.compareRows,
           );
-
-          view[pos] = newEntry;
+          if (
+            oldEntry[refCountSymbol] === 1 &&
+            (oldPos === pos || oldPos + 1 === pos)
```
arv: I'm not sure what `oldPos + 1` is for here. Care to add some comments?

packages/zql/src/ivm/view-apply-change.ts
```diff
@@ -233,55 +239,59 @@ export function applyChange(
...
-            oldEntry,
-            format.relationships,
-            schema,
-            withIDs,
-            rc,
+            schema.compareRows,
           );
-
-          view[pos] = newEntry;
+          if (
```
arv, line 265: I think this needs comments explaining what is going on here. I have this fresh in my head from last night but I do not think I will be able to understand this in 6 months.

arv (approved):
LGTM with a few nits/suggestions

packages/zql/src/ivm/view-apply-change.ts
grgbkr: done

packages/zql/src/ivm/view-apply-change.ts
grgbkr: added a comment 

         // A special case:
          // when refCount is 1 (so the row is being moved
          // without leaving a placeholder behind), and the new pos is
          // the same as the old, or directly adjacent to the old:
          // the row does not need to be moved, it can just be edited in place.

packages/zero-solid/src/solid-view.test.ts
grgbkr: Here I effectively swapped the two rows ({a: 3, b: 'b3'}, changed to {a: 0, b: 'b3'} and moved to index 0, and {a: 1, b: 'a'}, move to index1 ), so it makes sense that the effects reading these two indexes would fire.

PR #4585, opened 2025-07-08
fix(zero): speed limit on auth retries

tantaman:
See commentary on line 1538: `// Force a sleep to avoid a busy loop...`

This was not reported by a user but I did create this case when trying to repro this  user report: https://discord.com/channels/830183651022471199/1391264604046360626/1391264609322795129


packages/zero-client/src/client/zero.ts
```diff
@@ -1518,12 +1519,44 @@ export class Zero<
...
 
+      if (Date.now() - loopWindowStart > RUN_LOOP_INTERVAL_MS) {
+        loopWindowStart = Date.now();
+        loopsPerWindow = 1;
+      } else {
+        ++loopsPerWindow;
+      }
+
+      if (loopsPerWindow > MAX_LOOPS_PER_WINDOW) {
+        // Force a sleep to avoid a busy loop.
```
grgbkr, line 1538: The existing backoff logic almost covers this case.  When Connected we await the this.#rejectMessageError, which throws if an error message is received.  The error handling for that throw applies the backOff.  

The part that is problematic is

```
        if (isAuthError(ex)) {
          if (!needsReauth) {
            needsReauth = true;
            // First auth error, try right away without waiting.
            continue;
          }
          needsReauth = true;
        }
```

The continue skips the backoffMs sleep.

We set `needsReauth = false` on successful connect, so in this case we are constantly skipping the backoffMs sleep.

I think we should adjust that logic rather than introduce another backoff.

tantaman: The issue here is that we never know when to set `needsReauth` back to false. 

We can do it on successful mutation, but that does not cover the query auth failure case.

Working on restructuring auth handling altogether though -- https://www.notion.so/replicache/Remove-401-Handling-2683bed8954580769bf2f942ac13fb26

PR #4587, opened 2025-07-08
chore(tests): fix `worker-test.ts`

tantaman:
https://github.com/rocicorp/mono/pull/4576 updated to editing nodes in place in the view. See why that is safe in that pr.


grgbkr:
LGTM!

Thanks.

arv:
One question is why this was not picked up as a dep. I think maybe Vite is not picking up the dep... Fix is here https://github.com/rocicorp/mono/pull/4592

PR #4588, opened 2025-07-08
zero-cache(fix): reduce OTEL cardinality

cesara:


packages/zero-cache/src/observability/histograms.ts
```diff
@@ -3,54 +3,82 @@ import {cache, getMeter} from './view-syncer-instruments.ts';
 
 const getOrCreate = cache<Histogram>();
 
-function getOrCreateHistogram(name: string, description: string) {
-  return getOrCreate(name, name =>
-    getMeter().createHistogram(name, {description, unit: 'milliseconds'}),
-  );
+// Custom bucket boundaries optimized for millisecond timing
+const TIMING_BUCKETS = [0, 1, 5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000];
```
tantaman, line 7: I don't think we need this change given we removed all the labels. What happens if values end up in the > 5000 range with these custom buckets? Do we lose precision? As in, 10,000 ms gets represented as 5,000ms?

PR #4594, opened 2025-07-09
chore(zero-cache): less loud error when websocket is closed

arv:
Use ErrorWithLevel instead of Error for websocket closed errors. This allows the error to be logged at the 'info' level instead of the default 'error' level, making it less disruptive in the logs.


arv:
I realize this test is pretty lame... I should have tested it at a higher level to make sure we log at info. 🤷🏼

PR #4597, opened 2025-07-09
chore(zero-client): fix active clients test flake

arv:
Instead of using a hard coded `sleep(5)` we use `vi.waitFor` which iterates until the condition is met or times out. This should make the test more robust against timing issues, especially in CI environments where the timing can be less predictable.


arv:
Of course I cannot repro the test flakiness locally so I will have to wait for the CI to run.

tantaman:
Did not know about `waitFor`. Neat. lgtm.

PR #4598, opened 2025-07-09
chore(zero-cache): emit deprecation warnings for deprecated options

darkgnotic:
Adds a `deprecated` field to Options definitions that emits a warning when the value is present:

Screenshot 2025-07-09 at 11 29 25

Also (finally) changed the `parseOptions()` function to take an options object instead of a list of params.

### DIFF
--- a/packages/shared/src/options.ts
+++ b/packages/shared/src/options.ts
@@ -37,6 +37,9 @@ export type WrappedOptionType = {
   /** Description lines to be displayed in --help. */
   desc?: string[];
 
+  /** Logged as a warning when parsed. */
+  deprecated?: string[];
+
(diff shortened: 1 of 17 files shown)


packages/shared/src/options.ts
```diff
@@ -276,42 +279,58 @@ function getRequiredOrDefault(type: OptionType) {
   };
 }
 
+export type ParseOptions = {
+  /** Defaults to process.argv.slice(2) */
+  argv?: string[];
```
arv, line 284: In general we usually do:

```ts
argv?: string[] | undefined;
```

to allow explicitly passing undefined.

PR #4599, opened 2025-07-09
fix(zero-protocol)!: update min protocol version to 18 (release 0.21.2025070200)

grgbkr:
Protocol compatibility was accidentally broken at some point thus min-protocol version needs to be updated https://discord.com/channels/830183651022471199/1288232858795769917/1392200724825505902.

I didn't find exactly where it broke, but the min-protocol version is currently claiming support for really old builds (back to 0.16).  Our policy is just to maintain support for the previous major release.  Updated min to protocol version 18, corresponding to the last major release  0.21.2025070200.  This allowed clean up of the deprecated `AST` in downstream query puts.   

Verified using hello-zero that a zero client on 0.21.2025070200 works with a server built at head including this change.

BREAKING CHANGE=To update zero-cache to version 0.22, client needs to be updated to at least version 0.21.2025070200.

### DIFF
--- a/packages/zero-protocol/src/ast.test.ts
+++ b/packages/zero-protocol/src/ast.test.ts
@@ -607,5 +607,5 @@ test('protocol version', () => {
   // old code will not understand the new schema, bump the
   // PROTOCOL_VERSION and update the expected values.
   expect(hash).toEqual('1c228prd1v53r');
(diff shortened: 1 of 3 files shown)


darkgnotic (approved):
LGTM

PR #4602, opened 2025-07-10
feat(zero-client): Expose client materialize time in inspector

arv:
# Add metrics collection for query materialization performance

This PR introduces metrics collection to track query materialization performance in Zero, accessible through the Zero Inspector.

## What's New

- Automatically tracks query materialization time with zero performance overhead
- Collects client-side and end-to-end performance metrics using T-Digest for percentile calculations
- Adds `metrics` property to Inspector and Query objects for performance analysis

### Global Performance Metrics
```typescript
const inspector = await zero.inspect();
const clientMetrics = inspector.metrics['query-materialization-client'];

console.log(`Total queries: ${clientMetrics.count()}`);
console.log(`P50 latency: ${clientMetrics.quantile(0.5)}ms`);
console.log(`P90 latency: ${clientMetrics.quantile(0.9)}ms`);
```

### Per-Query Analysis
```typescript
const queries = await inspector.client.queries();
const slowQueries = queries.filter(q => 
  q.metrics?.['query-materialization-client'].quantile(0.9) > 100
);
console.log(`Found ${slowQueries.length} queries with >100ms P90 latency`);
```

## Interface Changes

```typescript
interface Inspector {
  readonly metrics: Metrics;
  // ... existing properties
}

interface Query {
  readonly metrics: Metrics | null;
  // ... existing properties
}

type Metrics = {
  readonly 'query-materialization-client': ReadonlyTDigest;
  readonly 'query-materialization-end-to-end': ReadonlyTDigest;
};
```

Enables developers to identify performance bottlenecks and optimize slow queries in Zero applications.

### DIFF
--- a/packages/zero-client/src/client/query-manager.ts
+++ b/packages/zero-client/src/client/query-manager.ts
@@ -396,9 +396,6 @@ export class QueryManager implements InspectorMetricsDelegate {
   ): void {
     // We track all materializations of queries as well as per
     // query materializations.
-    metric satisfies
-      | 'query-materialization-client'
-      | 'query-materialization-end-to-end';
     this.#metrics[metric].add(value);
 
     const queryID = args[0];
--- a/packages/zero-client/src/client/query-manager.ts
+++ b/packages/zero-client/src/client/query-manager.ts
@@ -71,6 +71,7 @@ export class QueryManager implements InspectorMetricsDelegate {
     'query-materialization-end-to-end': new TDigest(),
   };
   readonly #queryMetrics: Map<string, Metric> = new Map();
+  readonly #slowMaterializeThreshold: number;
 
   constructor(
     lc: ZeroLogContext,
@@ -81,6 +82,7 @@ export class QueryManager implements InspectorMetricsDelegate {
     experimentalWatch: ReplicacheImpl['experimentalWatch'],
     recentQueriesMaxSize: number,
     queryChangeThrottleMs: number,
+    slowMaterializeThreshold: number,
   ) {
     this.#lc = lc.withContext('QueryManager');
     this.#clientID = clientID;
@@ -89,6 +91,7 @@ export class QueryManager implements InspectorMetricsDelegate {
     this.#send = send;
     this.#mutationTracker = mutationTracker;
     this.#queryChangeThrottleMs = queryChangeThrottleMs;
+    this.#slowMaterializeThreshold = slowMaterializeThreshold;
 
     this.#mutationTracker.onAllMutationsApplied(() => {
       if (this.#pendingRemovals.length === 0) {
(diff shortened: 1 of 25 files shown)


packages/zero-client/src/client/query-manager.ts
```diff
@@ -358,12 +369,39 @@ export class QueryManager {
...
       }
     }
   }
+
+  get metrics() {
+    return this.#metrics;
+  }
+
+  addMetric(metric: keyof MetricMap, value: number, queryID: string): void {
+    // We track of all materializations of queries as well as per
```
tantaman:
```suggestion
    // We track all materializations of queries as well as per
```

packages/zero-client/src/client/query-manager.ts
```diff
@@ -358,12 +369,39 @@ export class QueryManager {
...
+        assert(lruQueryID);
+        this.#queries.delete(lruQueryID);
+        this.#recentQueries.delete(lruQueryID);
+        this.#queryMetrics.delete(lruQueryID);
+        this.#queueQueryChange({op: 'del', hash: lruQueryID});
       }
     }
   }
+
+  get metrics() {
```
tantaman: a comment on this method about what metrics to expect to be returned would be useful

packages/zql/src/query/query-impl.ts
```diff
@@ -733,43 +736,47 @@ export class QueryImpl<
...
       input.destroy();
       removeCommitObserver?.();
       removeServerQuery();
     };
 
+    const t0 = performance.now();
+
+    const removeServerQuery = this.customQueryID
+      ? delegate.addCustomQuery(this.customQueryID, ttl, gotCallback)
+      : delegate.addServerQuery(ast, ttl, gotCallback);
```
tantaman: I know it was like this before but why does `removeServerQuery` call `addServerQuery`?

packages/zql/src/query/query-impl.ts
```diff
@@ -733,43 +736,47 @@ export class QueryImpl<
...
+      this.customQueryID
+        ? delegate.updateCustomQuery(this.customQueryID, newTTL)
+        : delegate.updateServerQuery(ast, newTTL);
+    };
+
     const gotCallback: GotCallback = got => {
       if (got) {
-        const t1 = Date.now();
-        delegate.onQueryMaterialized(this.hash(), ast, t1 - t0);
+        delegate.onQueryMaterialized(this.hash(), ast, performance.now() - t0);
```
tantaman: this represent the full server -> client materialization, correct? How is that metric distinguished from the purely client side materialization? It looks like `delegate.onQueryMaterialized` is called for both cases.

tantaman (commented):
.

packages/zql/src/query/query-impl.ts
arv: This is using the return-unsubscribe-function pattern. I'll rename to make it cleaner

packages/zql/src/query/query-impl.ts
arv: I changed this to a new metric instead called query-materialization-end-to-end

PR #4604, opened 2025-07-11
chore(zero): Add metrics for client side update to query

arv:
# Add Client-Side Query Update Time Metrics

## Summary
This PR adds metrics tracking for query update performance on the client side by measuring how long individual query updates take to propagate through the IVM pipeline.

## Problem
We could measure query materialization time but lacked visibility into query update performance, making it difficult to identify bottlenecks in the update pipeline.

## Solution
- **New `MeasurePushOperator`**: Wraps push operations to measure execution time and record `query-update-client` metrics
- **Enhanced `QueryDelegate`**: Added `decorateSourceInput` method to enable measurement decoration
- **Updated `buildPipeline`**: Now requires `queryID` parameter for proper query identification

## Inspector Integration
Metrics are exposed through existing inspector infrastructure via `getMetrics()` and `getQueryMetrics(queryID)`.

### DIFF
--- a/packages/analyze-query/src/bin-analyze.ts
+++ b/packages/analyze-query/src/bin-analyze.ts
@@ -39,9 +39,7 @@ import {
   serverToClient,
 } from '../../zero-schema/src/name-mapper.ts';
 import {buildPipeline} from '../../zql/src/builder/builder.ts';
-import type {FilterInput} from '../../zql/src/ivm/filter-operators.ts';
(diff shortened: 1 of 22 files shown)


arv:
@tantaman Second commit only.

tantaman (commented):
`decorateSourceInput` and `MeasurePushOperator` look good to me.

PR #4607, opened 2025-07-11
fix(zero-client): Queries containing non-latin queries break startup.

aboodman:
Fixes https://bugs.rocicorp.dev/issue/4008

We encode queries registered early in startup in the sec protocols and they have to be URL-encoded. To save space we also base64 encode, but btoa doesn't support non-latin chars.

Queries can contain any arbitrary string so that means we need to encode these non-latin chars. So add another layer of encoding!

### DIFF
--- a/packages/zero-protocol/src/connect.ts
+++ b/packages/zero-protocol/src/connect.ts
@@ -60,13 +60,19 @@ export function encodeSecProtocols(
     initConnectionMessage,
     authToken,
   };
-  // base64 encoding the JSON before URI encoding it results in a smaller payload.
-  return encodeURIComponent(btoa(JSON.stringify(protocols)));
+  // WS sec protocols needs to be URI encoded. To save space, we base64 encode
+  // the JSON before URI encoding it. But InitConnectionMessage can contain
+  // arbitrary unicode strings, so we need to encode the JSON as UTF-8 first.
+  // Phew!
+  const bytes = new TextEncoder().encode(JSON.stringify(protocols));
+  return encodeURIComponent(btoa(String.fromCharCode(...bytes)));
 }
 
 export function decodeSecProtocols(secProtocol: string): {
   initConnectionMessage: InitConnectionMessage | undefined;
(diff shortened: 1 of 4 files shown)


aboodman:
Also changed issue-page to only send the required fields. Issues can be arbitrarily big so it seems worth doing this.

packages/zero-protocol/src/connect.ts
```diff
@@ -60,13 +60,19 @@ export function encodeSecProtocols(
...
     authToken,
   };
-  // base64 encoding the JSON before URI encoding it results in a smaller payload.
-  return encodeURIComponent(btoa(JSON.stringify(protocols)));
+  // WS sec protocols needs to be URI encoded. To save space, we base64 encode
+  // the JSON before URI encoding it. But InitConnectionMessage can contain
+  // arbitrary unicode strings, so we need to encode the JSON as UTF-8 first.
+  // Phew!
+  const bytes = new TextEncoder().encode(JSON.stringify(protocols));
+  return encodeURIComponent(btoa(String.fromCharCode(...bytes)));
```
arv, line 68: Normally base64 makes things about 33% larger. The reason why it makes things better in our case is because JSON has a lot of characters that need to be URL encoded.

I wish there was a simpler way than this dance but nothing comes to mind at the moment.

aboodman:
The PR already adds a test: https://github.com/rocicorp/mono/pull/4607/files#diff-ee4794749457202fff8ffe3c316fd319e238783b265d4272d74c604dd8f4c707

packages/zero-protocol/src/connect.ts
aboodman: I mean we could compress it?

https://developer.mozilla.org/en-US/docs/Web/API/CompressionStream

arv: Interesting. It is even available in Node.

But the end result is actually larger because compressing leads to lots of non url safe characters.

PR #4608, opened 2025-07-11
fix(zero): fix race with mutation resolution

tantaman:
The big picture is that we should have gone with writing mutation responses to the database in the first place.

The specific problem --

A user's API server could return an error but have applied the mutations. In this case, the mutation tracker was assuming that the mutations were not applied and would wait for a retry to apply them.

The retry would never happen since the mutations were applied with the original request.

The fix is to mark error responses to pushes as being in an unknown or limbo state. When the lmid is advanced, we resolve limbo mutations whose lmid is less than or equal to the current lmid.

We do not resolve outstanding mutations on every lmid advance since we may receive the push response for those mutations. If we resolved on every lmid bump then we can miss application errors being returned in response to mutations.

The most correct fix is to write mutation responses to the DB and send them down with `poke`. This will make mutation promise resolution transactional with the syncing of the side-effects of the mutation.

### DIFF
--- a/packages/zero-client/src/client/mutation-tracker.ts
+++ b/packages/zero-client/src/client/mutation-tracker.ts
@@ -21,10 +21,7 @@ type ErrorType =
   | Error
   | unknown;
 
-const transientPushErrorTypes: PushError['error'][] = [
-  'zeroPusher',
-  'http',
-
+const completeFailureTypes: PushError['error'][] = [
   // These should never actually be received as they cause the websocket
   // connection to be closed.
   'unsupportedPushVersion',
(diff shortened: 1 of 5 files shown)


aboodman:
I think we should do the correct thing. This feels like it is going to keep causing problems. Doing the correct thing doesn't seem that hard. We already sync a `clients` table, we can also sync an `errors` table. Zero client can delete from this table once it has resolved the promise. zero server can delete all errors for a client when it deletes the client.

packages/zero-client/src/client/mutation-tracker.ts
```diff
@@ -48,15 +45,21 @@ export class MutationTracker {
...
 
   constructor(lc: ZeroLogContext) {
     this.#lc = lc.withContext('MutationTracker');
     this.#outstandingMutations = new Map();
     this.#ephemeralIDsByMutationID = new Map();
-    this.#allMutationsConfirmedListeners = new Set();
+    this.#allMutationsAppliedListeners = new Set();
+    this.#limboMutations = new Set();
+    this.#largestOutstandingMutationID = 0;
+    this.#currentMutationID = 0;
```
arv, line 62: FYI, all of these can be set outside the constructor.

PR #4610, opened 2025-07-11
fix(solid): export createUseZero helper

asterikx:
Export `createUseZero` helper.


grgbkr:
@asterikx Thanks for the submission.  Can you fix the prettier error by running `npm run format` in the `packages/zero-solid` directory.  Thanks!

PR #4611, opened 2025-07-11
chore(zero-cache): drop inactive replication slots when unable to create a new one

darkgnotic:
When a replication slot cannot be created due to reaching the maximum configured value, drop inactive replication slots and retry.

This address a pathological scenario in which inactive replication slots fill the slot quota due to repeated initial sync failures. 

User reports:
* https://discord.com/channels/830183651022471199/1389157062881050686/1389197578033303674
* https://rocicorp.slack.com/archives/C08V7L0DB9B/p1751817231088029

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/initial-sync.ts
+++ b/packages/zero-cache/src/services/change-source/pg/initial-sync.ts
@@ -1,4 +1,7 @@
-import {PG_INSUFFICIENT_PRIVILEGE} from '@drdgvhbh/postgres-error-codes';
+import {
+  PG_CONFIGURATION_LIMIT_EXCEEDED,
+  PG_INSUFFICIENT_PRIVILEGE,
+} from '@drdgvhbh/postgres-error-codes';
 import type {LogContext} from '@rocicorp/logger';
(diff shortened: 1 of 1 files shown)


tantaman:
> When a replication slot cannot be created due to reaching the maximum configured limit, drop inactive replication slots and retry.

Should we drop sooner? My understanding is that each replication slot will hold onto WAL entries so we should not keep any more slots around than are being used by zero-cache.

darkgnotic:
> > When a replication slot cannot be created due to reaching the maximum configured limit, drop inactive replication slots and retry.
> 
> Should we drop sooner? My understanding is that each replication slot will hold onto WAL entries so we should not keep any more slots around than are being used by zero-cache.

Good question and good point. There already is obsolete replication slot cleanup when a replication session begins:

https://github.com/rocicorp/mono/blob/main/packages/zero-cache/src/services/change-source/pg/change-source.ts#L349

We wait until this then to do the clean up in order to allow non-disruptive resyncs. The pathological case is that we don't get there if we're unable to create a replication slot to begin with, which is what this PR fixes.

PR #4614, opened 2025-07-12
chore(zero-cache): TTL was using the wrong clock

arv:
`markDesiredQueriesAsInactive` and `deleteClient` were using the wall clock time instead of the ttlClock. This led to the timeout being scheduled far in the future and usually not being hit.


darkgnotic (approved):

arv:
I added an opaque type for TTLClock so there is no more accidental usage of wallclock when a ttlclock is needed.

PR #4617, opened 2025-07-13
Notification preferences

aboodman:

PR #4618, opened 2025-07-13
Aa/bug 3877

aboodman:


KATT:
One thing that I know I missed is a link in the email to unsubscribed quickly, without needing to login and click unsubscribe

Probably something like

1. `/api/issues/unsubscribe?issueId=x&email=y` - unsubscribed users 
2. Redirect to `/issues/{id}?toast=unsubscribed`
3. Show toast and replaceState `?toast` param


---

If there was db migrations setup I'd also make it so that any user who interacted with an issue was added to the issue subscription table and make the subscription to an issue more explicit and skip the checks if the user has interacted

aboodman:
Yeah we don't have a migration thing. We built this before we had figured out how to integrate w/ drizzle / use those and now we keep kicking the can down the road. I agree the explicit design is better.

KATT:
Another thing I noticed with zbugs, just as an FYI, is that the monorepo isn't used properly, I started on #4620 but didn't finish it as TS wasn't passing due to a "cannot be used without a named reference to X"-bug. 

- It'll makes it a bit easier to work with and more grokkable how it'd translate to a standalone app outside of the monorepo.
- It will alert you about issues in how the packages are built  (like the "cannot be named issue") before you publish pakcages

aboodman:
Superceded by https://github.com/rocicorp/mono/pull/4617

PR #4624, opened 2025-07-14
feat(zbugs): Image Upload

alexhking:


apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,218 @@
...
+  className = '',
+}: ImageUploadAreaProps) {
+  const [isDragOver, setIsDragOver] = useState(false);
+  const [textareaRect, setTextareaRect] = useState<DOMRect | null>(null);
+  const dragCounterRef = useRef(0);
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  const fileInputRef = useRef<HTMLInputElement>(null);
+  const {isUploading, uploadFiles} = useImageUpload({onUpload});
+
+  const handleFileSelect = async (e: React.ChangeEvent<HTMLInputElement>) => {
```
aboodman: Nit: but we would generally import `ChangeEvent` directly so no need to say `React.ChangeEvent`.

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,218 @@
...
+    const files = Array.from(e.target.files || []);
```
aboodman: Nit: `??` is preferred over `||` these days.

aboodman: I believe you can also say:

```
const files = [....e.target.files ?? []];
```

And that is slightly more idiomatic.

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,218 @@
...
+            width: textarea.offsetWidth,
+            height: textarea.offsetHeight,
+          } as DOMRect);
+        }
+      }
+    };
+
+    // Update rect on mount and when dragging
+    updateRect();
+    if (isDragOver) {
```
aboodman: This makes no  sense to me. First line updates rect, and then second line does again if some boolean is true. Why need to do it twice in that case?

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,218 @@
...
+      updateRect();
+    }
+  }, [isDragOver]);
+
+  const handleDragEnter = (e: React.DragEvent) => {
+    e.preventDefault();
+    e.stopPropagation();
+    dragCounterRef.current++;
```
aboodman: There is a way to do this better, I can never remember how.  You want to listen to enter/exit changes on just the root node of the drag area. I think it has to do with looking at either the `target` of the event or the `currentTarget`.

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,218 @@
...
+    e.stopPropagation();
+  };
+
+  const handleDrop = async (e: React.DragEvent) => {
+    e.preventDefault();
+    e.stopPropagation();
+    setIsDragOver(false);
+    dragCounterRef.current = 0;
+
+    const files = Array.from(e.dataTransfer.files).filter(file =>
```
aboodman: `Array.from() -> [...]` here and elsewhere.

apps/zbugs/vite.config.ts
```diff
@@ -1,5 +1,5 @@
 import react from '@vitejs/plugin-react';
-import {defineConfig, type ViteDevServer} from 'vite';
+import {defineConfig, loadEnv, type ViteDevServer} from 'vite';
```
aboodman, line 2: It should not be necessary to load env vars here. See how `DISCORD_WEBHOOK_URL` works? It's not loaded here but is still accessible from `discord.ts`. Just using `process.env` inside the `upload.ts` file should be sufficient.

Also please add these env vars to `.env.example`.

packages/shared/src/dotenv.ts
```diff
@@ -1,8 +1,14 @@
 import {config as dotenvxConfig} from '@dotenvx/dotenvx';
+import path from 'path';
```
aboodman, line 2: This file should not need to change.

packages/shared/src/dotenv-cjs.cjs
```diff
@@ -0,0 +1,8 @@
+const {config} = require('@dotenvx/dotenvx');
```
aboodman, line 1: We should not need to add this file. Did ChatGPT maybe suggest this? We don't use cjs at all in zbugs.

apps/zbugs/src/pages/issue/issue-composer.tsx
```diff
@@ -82,6 +84,26 @@ export function IssueComposer({isOpen, onDismiss}: Props) {
     }
   };
 
+  const handleImageUpload = (markdown: string) => {
```
aboodman: This code seems to be repeated in issue-composer and issue-detail and comment-composer. Can we factor it into a helper component that is shared?

apps/zbugs/src/components/image-upload-area.tsx
alexhking: I can look into this more. I tried looking at currentTarget and ran into some problems with the box displaying, but I will try again.

apps/zbugs/src/pages/issue/issue-composer.tsx
alexhking: Yes we can. Will add

apps/zbugs/src/components/image-upload-area.tsx
alexhking: Updated. Let me know if this works

alexhking:
The only test failing appears to be a viewsyncer test. Unsure if this is related to my changes or something else.

aboodman:
It's not you.

apps/zbugs/api/index.ts
```diff
@@ -163,6 +163,36 @@ fastify.post<{
...
+    }
+    throw e;
+  }
+
+  if (!authData) {
+    reply.status(401).send('Authentication required');
+    return;
+  }
+
+  const {getPresignedUrl} = await import('../src/server/upload.ts');
```
aboodman: Not sure what's going on here? Should not need a dynamic import. Can this be imported the same way as other code in this file?

apps/zbugs/api/index.ts
```diff
@@ -163,6 +163,36 @@ fastify.post<{
...
+
+  try {
+    const result = await getPresignedUrl(request.body.contentType);
+    reply.send(result);
+  } catch (error) {
+    console.error('Error generating presigned URL:', error);
```
aboodman: There are no other usages of console.error in this file. Match what existing code does which appears to be to return the error message in the response.

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,215 @@
...
+        }
+      }
+
+      if (files.length > 0) {
+        await uploadFiles(files);
+      }
+    }
+  };
+
+  const dropZoneClasses = `
```
aboodman: Use `classnames` - look elsewhere in zbugs for examples.

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,215 @@
...
+      onDrop={handleDrop}
+      onPaste={handlePaste}
+      style={{
+        position: 'relative',
+      }}
+    >
+      {children}
+      {isDragOver && textareaRect && (
+        <div
+          style={{
```
aboodman: Maybe pull out the constant styles in either a module css or css file so that it's easier to see the logic?

apps/zbugs/src/pages/issue/comment-composer.tsx
```diff
@@ -78,22 +83,27 @@ export function CommentComposer({
...
-      />
-      <Button
-        className="secondary-button"
-        eventName={id ? 'Save comment edits' : 'Add new comment'}
-        onAction={save}
-        disabled={currentBody.trim().length === 0}
-      >
-        {id ? 'Save' : 'Add comment'}
-      </Button>{' '}
+      <ImageUploadArea onUpload={handleImageUpload}>
```
aboodman: Can we make the usage look more like:

```tsx
<ImageUploadArea>
  <textarea ...></textarea>
</ImageUploadArea>
```

The handleImageUpload could should be able to be composed inside `<ImageUploadArea>` without having to be repeated at each callsite. Then we don't need the two separate new hooks either as all that code can be inside `<ImageUploadArea>`.

apps/zbugs/.env.example
```diff
@@ -46,3 +46,8 @@ VITE_PUBLIC_JWK=''
 # Discord webhook to send notifications to. Not required. Notifications won't
 # be sent if absent.
 DISCORD_WEBHOOK_URL=''
+
+# AWS credentials for image uploads. Required for image upload functionality.
```
aboodman: @cesara how will this work in prod? Is this the right way to get access to s3 for us? If so do we have to add these env vars?

alexhking: @cesara When I set this up I created a specific user for zbugs image uploads and have a key id / access key pair generated.

apps/zbugs/api/index.ts
```diff
@@ -163,6 +164,37 @@ fastify.post<{
   );
 });
 
+fastify.post<{
+  Body: {contentType: string};
+}>('/api/upload/presigned-url', async (request, reply) => {
+  let authData: JWTData | undefined;
+  try {
```
aboodman: This pattern for calling `maybeVerifyAuth` is duplicated in many places in this file. Can you clean it up @0xcadams ?

zbugs has gotten a bit bitrotted over the months as we have explored different ideas and learned but we really do need to keep it as tight and beautiful as we can at the code level, as people often refer to it as an example of more advanced usage of Zero.

apps/zbugs/src/components/image-upload-area.module.css
```diff
@@ -0,0 +1,41 @@
+.dragOverlay {
```
aboodman: This is good but I noticed that @0xcadams is still using tailwindisms. You guys need to fight it out and figure out if we're using tailwind or not and then convert it one way or other so it's consistent.

I personally have a minor preference for no tailwind as I'm a grumpy old man set in my ways. But as long as somebody else figures out the right tailwind setup I'm happy to copy/paste.

I think maybe you should choose whichever is used more in current codebase which I believe is css. Just because less work to convert. Unless there is some big benefit ongoing to adding features in the future using tailwind.

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,330 @@
...
+  useCallback,
+  type ChangeEvent,
+} from 'react';
+import type {ReactNode} from 'react';
+import classNames from 'classnames';
+import {useLogin} from '../hooks/use-login.tsx';
+import {Button} from './button.tsx';
+import styles from './image-upload-area.module.css';
+
+interface ImageUploadAreaProps {
```
aboodman: We tend to prefer using `type` over `interface`.

apps/zbugs/src/hooks/use-image-upload.ts
```diff
@@ -0,0 +1,94 @@
+import {useState} from 'react';
```
aboodman: I think you forgot to remove this file.

apps/zbugs/src/hooks/use-textarea-image-insert.ts
```diff
@@ -0,0 +1,32 @@
+import {useCallback, type RefObject} from 'react';
+
+export function useTextareaImageInsert(
```
aboodman: I think you forgot to remove this file.

apps/zbugs/src/pages/issue/comment-composer.tsx
```diff
@@ -78,22 +81,27 @@ export function CommentComposer({
...
-      />
-      <Button
-        className="secondary-button"
-        eventName={id ? 'Save comment edits' : 'Add new comment'}
-        onAction={save}
-        disabled={currentBody.trim().length === 0}
-      >
-        {id ? 'Save' : 'Add comment'}
-      </Button>{' '}
+      <ImageUploadArea>
```
aboodman: aaah, very nice.

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,330 @@
...
+}: ImageUploadAreaProps) {
+  const [isDragOver, setIsDragOver] = useState(false);
+  const [textareaRect, setTextareaRect] = useState<DOMRect | null>(null);
+  const [isUploading, setIsUploading] = useState(false);
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  const fileInputRef = useRef<HTMLInputElement>(null);
+  const {loginState} = useLogin();
+
+  // Image upload logic (from use-image-upload.ts)
+  const validateFile = (file: File): string | null => {
```
aboodman: These functions that do not rely on any variables from within the component should be moved out to bare functions in the same file (`function validateFile(...) { .... }`) so that it's clear they are static. (Also just to reduce the size of this function).

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+    if (!response.ok) {
+      throw new Error(`Failed to get presigned URL: ${response.statusText}`);
+    }
+
+    return response.json();
+  };
+
+  // Textarea image insert logic (from use-textarea-image-insert.ts)
+  const insertMarkdown = useCallback((markdown: string) => {
```
aboodman: This one could arguably stay inline since it is relying on `wrapperRef`. But you could also move it out into its own function and pass `wrapperRef.current` in. I would probably do that, again to make the react component smaller and easier to read.

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,330 @@
...
+    if (wrapperRef.current) {
+      const textarea = wrapperRef.current.querySelector('textarea');
+      if (textarea) {
+        const start = textarea.selectionStart;
+        const end = textarea.selectionEnd;
+        const text = textarea.value;
+        const newText =
+          text.substring(0, start) + markdown + text.substring(end);
+
+        // Need to simulate user input here
```
aboodman: Spurious comment?

apps/zbugs/src/components/image-upload-area.tsx
```diff
@@ -0,0 +1,330 @@
...
+          'value',
+        )?.set;
+
+        if (nativeInputValueSetter) {
+          nativeInputValueSetter.call(textarea, newText);
+        } else {
+          textarea.value = newText;
+        }
+
+        // Create a synthetic event that React will recognize as a real user input
```
aboodman: Hm, I can see reusing this `<ImageUploadArea>` is a little complicated by the fact that sometimes the inner component is an HTML element and sometimes a React component.

I did not like using events to interact with React at first, but I'm coming around to it after thinking through alternatives. But in this function it looks like we try three different ways to set the value:

1. Use the native setter
2. Use `textarea.value=...`
3. Send an event

It seems like just (3) would work for all cases. Does it not? I'm not sure we need 1 and 2.

Big picture I'm not sure what the right way is to factor this component w/o playing with a few options. @0xcadams can you look and see? Obviously we want to share as much code as possible but we shouldn't do anything unnatural or brittle in react either.

apps/zbugs/.env.example
cesara: yes this is good. we will need to add it to vercel. ill look at the key permissions

apps/zbugs/src/components/image-upload-area.tsx
alexhking: No. It is in fact there on purpose. The code below is there to mimic how the browser would handle native user input, because after some changes the auto resize function was not working properly when it generated the markdown for embedded images.

apps/zbugs/src/components/image-upload-area.module.css
alexhking: Ultimately, I can use both, but I dislike Tailwind for a variety of reasons. With Tailwind, you're sacrificing maintainability and readability for short-term convenience, and to me that tradeoff doesn't make sense for any project we're working on, especially when we have a high-level of expertise with CSS. Some other gripes I have with Tailwind:

* Separation of concerns is violated. Embedding dozens of utility classes into the templates clutters markup and makes it harder to scan for structure, spot bugs, and understand the intention behind styles without reading class by class. This especially inpacts long-lived, complex projects like zbugs where clarity and maintainability are more important than rapid prototyping.
* It's harder to scale and maintain. As this project grows we have a tailwind class soup that eventually becomes overwhelming. Reusability suffers (unless you manually extract components or use @apply, which defeats Tailwind's "utility first" mantra. Refactoring becomes harder because styles are baked into markup and not centralized.
* As I said before it's really a poor fit for teams that know CSS well. I find it's often slower and more frustrating to work against the Tailwind mindset. You end up customizing config files to reintroduce semantics (example: renaming text-slate-900 to text-heading), fighting the naming conventions, and losing the expressive power that plain CSS provides.

I could go on, but I'm curious to hear a counter-opinion. Once again, if maybe CSS is not a strong point for everyone and this would help others, I could totally work with Tailwind. It's just not my preference.

apps/zbugs/src/components/image-upload-area.tsx
0xcadams: At a high level - do we want to build everything in zbugs without libraries? If I were building this as an app dev, I would use something like:
1. https://react-dropzone.js.org/
2. https://tiptap.dev/

A WYSIWYG editor feels to me like something that I wouldn't want to have to maintain. Most teams don't build custom editors like this because it's too much work - showing them how to integrate with popular libraries makes more sense to me.

apps/zbugs/src/components/image-upload-area.module.css
0xcadams: Honestly this is just a personal preference thing. It's just CSS in the end. I will make the counter-argument here for the sake of it, but strong opinion loosely held:

I personally think tailwind or CSS-in-JS is far easier to maintain (no dealing with CSS scopes, reusable variables are handed to you, styling is clearly defined next to components).

I think the "React way" is building components that abstract styles out. I've built multiple large projects in Tailwind and I've had the opposite experience - it's just how you componentize.

For instance in the following lines in this PR, there's some hardcoded colors, border radius, and font size. It's really hard to keep those aligned as projects grow, and it feels like swimming upstream to continuously keep CSS consistent across the project. As opposed to that falling out of your styling solution and it's swimming upstream to add one-off styling.

I would say a lot of teams don't use scoped CSS anymore - it's far more common to use CSS-in-JS, Tailwind, some solution that helps with code reuse. e.g. shadcn has a ton of adoption because it's super easy to add a new component with styles. It's not a super important point but again it's probably what your users are using.

I'm not gonna come in and flip the styling patterns your team has on their head - so going forward I will happily write CSS. Tailwind is just what I'm used to (and it's used throughout the project(s) so I'm just adding on that).

aboodman: To be clear, by "CSS-in-JS" do you mean css modules or something else?

apps/zbugs/src/components/image-upload-area.tsx
aboodman: It's kind of a case-by-case value judgement. i would prefer to use libraries if they are small, lightweight, and very high quality.

For drag/drop it makes sense to me to use react-dropzone if it is lightweight and works well. It's also not that hard to do from scratch so if there's not a good library that matches I'm fine doing it manually too.

@alexhking did you consider any libraries? (a fine answer would be I forgot to – I often forget how high level react librariers have gotten)

For the editor I specifically **don't** want the experience to be rich text. I want it to be just plain markdown. I feel like rich text editors never work very well and technical users prefer to just write markdown.

For the drag/drop image experience i wanted it to do what GitHub does where it injects markdown for the image.

Within those constraints I am very happy to use libraries that work well.

apps/zbugs/src/components/image-upload-area.module.css
0xcadams: No there are a lot of styling libraries which let you colocate styles with components. styled-components, tamagui (nate built this), tokenami (jjenzz built this), emotion. Similar in devex to tailwind but usually generate styles at runtime. Tailwind is somewhere in-between.

apps/zbugs/src/components/image-upload-area.tsx
0xcadams: Yeah that makes sense. My thinking was that Linear uses tiptap, and zbugs is directly inspired by that.

Sounds good to me.

apps/zbugs/src/components/image-upload-area.module.css
aboodman: OK I am so out of date. How do people keep up with all of this.

apps/zbugs/src/components/image-upload-area.tsx
aboodman: Linear is used by normals 😂.

alexhking: I did not consider any libraries, mainly because I was trying to keep this as lightweight as possible, but I'm open to it if it helps in any way.

0xcadams:
@aboodman this has been reworked to use `react-dropzone` and tested locally. The AWS secrets have been added to Vercel.

aboodman:
Seems like there is a lint error or something. Let me know if you need help landing.

0xcadams:
OK, I need help with merging this. I've rerun the benchmark and JS tests a number of times and they fail every time.

PR #4626, opened 2025-07-15
chore(view-syncer): remove maxRowCount and related eviction logic

arv:
- Eliminated the maxRowCount property and its associated eviction mechanism for inactive queries.
- Removed the getInactiveQueries function call and the #evictInactiveQueries method.
- Cleaned up related comments and documentation to reflect the removal of row count constraints.


packages/zero-cache/src/config/zero-config.test.ts
```diff
@@ -359,11 +359,6 @@ test('zero-cache --help', () => {
                                                                                                                                                                    
      --target-client-row-count number                            default: 20000                                                                                    
        ZERO_TARGET_CLIENT_ROW_COUNT env                                                                                                                            
-                                                                 The target number of rows to keep per client in the client side cache.
```
arv, line 362: @darkgnotic Is it not supposed to show anything if deprecated? Maybe I misunderstood how to use the deprecated config option?

darkgnotic: Hmmm ... maybe I should have implied `hide: true` with there is a deprecation. Maybe you can add `hide: true` for now and we can change the library to do this by default. 

wdyt?

arv: Is it warning when using the deprecated option?

darkgnotic: Yes

PR #4629, opened 2025-07-15
feat(zero-client): allow the client to provide a mutate/query url

tantaman:
This URL will be required to match one of the pre-configured URLs in `ZERO_MUTATE/QUERY_URL`


packages/zero-client/src/client/options.ts
```diff
@@ -105,8 +108,12 @@ export interface ZeroOptions<
    * push.queryParams can be used to augment the URL
    * used to connect to your API server so it includes
    * variables in the query string.
+   *
+   * DEPRECATED: Use `userMutateParams` instead.
```
arv, line 112: Use `@deprecated` JSDoc. VSCode and others will warn on usage.

packages/zero-protocol/src/connect.ts
```diff
@@ -21,15 +21,30 @@ export const connectedMessageSchema = v.tuple([
...
+  /**
+   * A client driven URL to send queries or mutations to.
+   * This URL must match one of the URLs set in the zero config.
+   *
+   * E.g., Given the following environment variable:
+   * ZERO_QUERY_URL=[https://*.example.com/query]
+   *
+   * Then this URL could be:
+   * https://myapp.example.com/query
+   */
```
arv, line 34: I don't see where this is enforced?

arv (commented):
LGTM

packages/zero-client/src/client/options.ts
aboodman: I actually prefer we don't officially deprecate until query api is done.

aboodman: we should probably remove queryparams and just have the bare URLs right?

aboodman: no need for query params complexity?

packages/zero-protocol/src/connect.ts
tantaman: enforced on the backend here: https://github.com/rocicorp/mono/blob/08f69494eaca1d53f2a9153ea392e1e91edef76a/packages/zero-cache/src/custom/fetch.ts#L86-L98

packages/zero-client/src/client/options.ts
tantaman: The one upside to the query params is that it keeps things a bit more DRY. The URL is only specified in one place (ZERO_CONFIG) and doesn't need to be kept in sync.

aboodman:
This change does not actually allow the user to change the push/mutate URL . It always passes the first entry from the env var list:

https://github.com/rocicorp/mono/blob/b6e8ad4c3f5e81e64afb14a915ef3b1a9365ee88/packages/zero-cache/src/services/mutagen/pusher.ts#L379

PR #4638, opened 2025-07-16
fix(zero-cache): anon otel - readd sessionid, and taskid, delay otel start, and container detect

cesara:


packages/zero-cache/src/server/anonymous-otel-start.ts
```diff
@@ -279,12 +303,54 @@ class AnonymousTelemetryManager {
...
+      if (existsSync('/usr/local/bin/docker-entrypoint.sh')) {
+        return true;
+      }
+
+      if (process.env.KUBERNETES_SERVICE_HOST) {
+        return true;
+      }
+
+      if (
+        process.env.DOCKER_CONTAINER_ID ||
```
darkgnotic, line 330: Note that we can also add an ENV variable in our own Dockerfile image as a signal, e.g. adding

```Dockerfile
ENV ZERO_IN_CONTAINER=1
```

in https://github.com/rocicorp/mono/blob/9e4c6932f10e1830536b06eb5343cc117e14c4d2/packages/zero/Dockerfile#L45

packages/zero-cache/src/server/anonymous-otel-start.ts
```diff
@@ -55,30 +61,43 @@ class AnonymousTelemetryManager {
       return;
     }
 
-    if (this.#started || !config.enableUsageAnalytics) {
+    if (this.#started || this.#shouldStart || !config.enableUsageAnalytics) {
```
darkgnotic: Technically I think this can result in setting multiple timeouts if `start()` is called multiple times.

Here's a suggestion:

```ts
start(...) {
  if (this.#starting || !config.enableUsageAnalytics) {
    return;
  }
  
  this.#starting = true;
  setTimeout(() => this.#run());
}

#run() {
  if (this.#stopped) {
    return;
  }
  ...
}

shutdown() {
  this.#stopped = true;
}

File: packages/zero-cache/src/server/syncer.ts [hunk 3]
```diff
@@ -149,6 +146,9 @@ export default function runWorker(
     parent,
   );
 
+  // Start telemetry with view-syncer count
+  startAnonymousTelemetry(lc, config, syncer.viewSyncerCount);
```
darkgnotic: viewSyncerCount is not correct. You want to use the value in `config.numSyncWorkers`. However, this may be undefined, in which case it defaults to availableParallelism() - 1:

https://github.com/rocicorp/mono/blob/9e4c6932f10e1830536b06eb5343cc117e14c4d2/packages/zero-cache/src/server/main.ts#L48

Let me check something in so that the config is set for child processes ...

darkgnotic: Okay with https://github.com/rocicorp/mono/pull/4639 you can reference `config.numSyncWorkers` to determine how many are running.

packages/zero-cache/src/server/anonymous-otel-start.ts
darkgnotic: You'd need to add the check for that here too then ...

```ts
  process.env.ZERO_IN_CONTAINER ||
```

PR #4639, opened 2025-07-16
chore(zero-cache): pass the computed `numSyncWorkers` to subprocesses

darkgnotic:
This allows subprocesses know now many sync workers are running (for analytics logic).


cesara:
🙏 thanks

PR #4641, opened 2025-07-16
feat(zbugs): notification preferences

0xcadams:
Built on notification prefs. Specifically updates `issueNotifications` with:

1. Subscribed when the user creates the issue
2. Subscribed when the user is assigned to the issue
3. Unsubscribed when the user is unassigned from the issue?
4. Adding an emoji on a comment doesn't do anything anymore?
5. Emoji'ing the issue itself subscribes you to the issue
6. Adding labels does nothing
7. Adding a comment to the issue subscribes you

And then anyone can subscribe/unsubscribe to issues that they can see.

**Before deployment:**
1. We need to update Loops to include the `unsubscribe` link in email notification content.
2. We need to migrate the DB to add `issueNotifications` and migrate current data into this.


apps/zbugs/api/index.ts
```diff
@@ -163,6 +163,37 @@ fastify.post<{
...
+
+  const existingUser = existingUserResult[0];
+
+  if (!existingUser) {
+    reply.status(401).send('Unauthorized');
+    return;
+  }
+
+  // Look up the actual issue ID from the shortID
+  const issueResult =
```
aboodman, line 192: Check the result of `parseInt` to ensure it's actually an integer.

apps/zbugs/api/index.ts
```diff
@@ -163,6 +163,37 @@ fastify.post<{
   );
 });
 
+fastify.get<{
+  Querystring: {id: string; email: string};
+}>('/api/unsubscribe', async (request, reply) => {
+  const existingUserResult =
```
aboodman, line 182: Check that email param is present.

apps/zbugs/docker/share_sql/base.sql
```diff
@@ -105,6 +105,16 @@ CREATE TABLE "userPref" (
     PRIMARY KEY ("userID", "key")
 );
 
+-- issueNotifications
+
+CREATE TABLE "issueNotifications" (
+    "userID" VARCHAR REFERENCES "user"(id) ON DELETE CASCADE,
+    "issueID" VARCHAR REFERENCES issue(id) ON DELETE CASCADE,
+    "subscribed" BOOLEAN DEFAULT true,
```
aboodman, line 115: My first reaction was that `subscribed` column not necessary since presence can mean subscribed and non-presence, not-subscribed.

But I realize you are probably trying to prevent user from getting resub'd if they continue interacting w/ bug. In that case, I agre but can you add a comment explaining the schema choice since it might not be immediately obvious.

apps/zbugs/src/pages/issue/issue-page.tsx
```diff
@@ -352,6 +357,28 @@ export function IssuePage({onReady}: {onReady: () => void}) {
   useEmojiDataSourcePreload();
   useShowToastForNewComment(comments, virtualizer, highlightComment);
 
+  useEffect(() => {
```
aboodman: The toast is very cute idea but I think I would prefer to save the complexity budget on this page for other things. How about we just print out "OK! You are unsubscribed to <issue link>" in the api response as plain text.

apps/zbugs/server/notify.ts
```diff
@@ -71,35 +64,44 @@ export async function notify(
...
         message: [issue.title, clip((await issue.description) ?? '')]
           .filter(Boolean)
           .join('\n'),
-        link: `https://bugs.rocicorp.dev/issue/${issue.shortID}`,
+        link: issueLink,
       };
       for (const recipient of recipients) {
         postCommitTasks.push(async () => {
-          await sendEmail({tx, email: recipient, ...payload, issue});
+          await sendEmail({
```
aboodman: There is a lot of duplicated code in here. I know you didn't do it but can you clean it up?

apps/zbugs/server/notify.ts
```diff
@@ -71,35 +64,44 @@ export async function notify(
...
   assert(modifierUser);
 
-  const recipients = await gatherRecipients(
-    // TODO: we shouldn't have to assert and cast this given this is called from `server-mutators`.
-    tx as ServerTransaction<Schema, postgres.TransactionSql>,
-    issueID,
-    isAssigneeChange,
-    previousAssigneeID,
-  );
+  const recipients = await gatherRecipients(tx, issueID, modifierUserID);
```
aboodman: Maybe call this `emailRecipients` to make more clear since it's not used for discord.

apps/zbugs/server/server-mutators.ts
```diff
@@ -32,79 +33,75 @@ export function createServerMutators(
           created: Date.now(),
           modified: Date.now(),
         });
-        await notify(
-          tx,
-          authData,
-          {kind: 'create-issue', issueID: id},
-          postCommitTasks,
-        );
+        if (tx.location === 'server') {
```
aboodman: Why is this necessary? i thought this file was only used on server (hence name).

Is it basically casting `tx`? If so maybe this instead:

```
assert(tx.location === 'server');
```

apps/zbugs/shared/mutators.ts
```diff
@@ -179,6 +246,49 @@ export function createMutators(authData: AuthData | undefined) {
...
+      created: number;
+    },
+  ) {
+    await assertUserCanSeeIssue(tx, userID, issueID);
+
+    if (subscribed === 'subscribe') {
+      await tx.mutate.issueNotifications.upsert({
+        userID,
+        issueID,
+        subscribed: true,
```
aboodman, line 291: This will update the user to subscribed even if they have already previously unsubscribed. I don't think that's what we want right? I think this function should only insert a record and set it to true if there isn't already a record set to false. For the unsubscribe case, upsert is correct.

aboodman: For the case where a user specifically sets their notification pref, using the UI then of course what it's doing now is the right thing.

apps/zbugs/src/pages/issue/issue-page.tsx
aboodman: (this page just already has a *lot* going on and I want to make sure we're using our complexity for things that are important/unique to Zero)

apps/zbugs/docker/share_sql/base.sql
0xcadams: Honestly I just built on Alex/your work. I previously had something different. But this is a good point and I will add the resub logic

apps/zbugs/src/pages/issue/issue-page.tsx
0xcadams: Sounds good

apps/zbugs/server/server-mutators.ts
0xcadams: Yes it's only used on the server - it's a type helper to remove a downstream cast that was marked TODO. Changing to use asserts instead

apps/zbugs/shared/mutators.ts
0xcadams: Yeah this was a good point and this is now updated for this functionality

apps/zbugs/server/notify.ts
```diff
@@ -260,82 +228,30 @@ function clip(s: string) {
...
-  isAssigneeChange = false,
-  previousAssigneeID?: string | null | undefined,
+  actorID: string,
 ): Promise<string[]> {
   const sql = tx.dbTransaction.wrappedTransaction;
 
-  // Get all recipient candidates
+  // we filter out the actor to not send them notifications on their own actions
+  // and filter by issue visibility - only crew members get notifications for internal issues
   const recipientRows = await sql`
```
tantaman, line 236: 🎉

apps/zbugs/server/server-mutators.ts
```diff
@@ -32,6 +33,8 @@ export function createServerMutators(
           created: Date.now(),
           modified: Date.now(),
         });
+
+        assert(tx.location === 'server');
```
tantaman, line 37: hmm, I wonder if we can fix the typing to not require the assert. For another day.

0xcadams: Spitballing here - would it make sense to change the type of the `CustomMutatorDefs` so that if a `TWrappedTransaction` is passed, then it's a server mutator definition?

```tsx
// before
export type Transaction<S extends Schema, TWrappedTransaction = unknown> =
  | ServerTransaction<S, TWrappedTransaction>
  | ClientTransaction<S>;

// after
export type Transaction<
  S extends Schema,
  TWrappedTransaction = unknown
> = [TWrappedTransaction] extends [unknown]
  ? ServerTransaction<S, TWrappedTransaction> | ClientTransaction<S>
  : ServerTransaction<S, TWrappedTransaction>;
```

apps/zbugs/README.md
```diff
@@ -31,8 +31,7 @@ npm install
 ### 2. Run the "upstream" Postgres database
 
 ```bash
-cd docker
-docker compose up
+npm run db-up
```
aboodman: I think the README also refers to these commands. Can you remove this from PR and we'll do it separately.

apps/zbugs/api/index.ts
```diff
@@ -163,6 +163,51 @@ fastify.post<{
...
+    reply.status(404).send('Issue not found');
+    return;
+  }
+
+  await sql`INSERT INTO "issueNotifications" ("userID", "issueID", "subscribed") 
+    VALUES (${existingUser.id}, ${issue.id}, false)
+    ON CONFLICT ("userID", "issueID") 
+    DO UPDATE SET "subscribed" = false`;
+  reply.send(
+    `OK! You are unsubscribed from issue #${request.query.id}: ${issue.title}.`,
```
aboodman: Sorry little nit can you make it into a URL that way it will appear clickable in browsers.

apps/zbugs/server/notify.ts
```diff
@@ -71,40 +64,70 @@ export async function notify(
...
 
   // If no recipients, skip notification
-  if (recipients.length === 0) {
+  if (recipientEmails.length === 0) {
     console.log('No recipients for notification', issueID);
     return;
   }
 
+  if (issue.shortID === null) {
+    console.log('No short ID for issue', issueID);
```
aboodman: This should not be possible. By the time we get to this code it should have been filled in. Maybe `assertNotNull()` instead.

aboodman (approved):
Good work. Please work w/ @grgbkr or @darkgnotic or @tantaman to do the migration.

aboodman:
😍

a (phone)

PR #4642, opened 2025-07-16
fix(zero-cache): anon-otel fix race condition for getFsId

cesara:


packages/zero-cache/src/server/anonymous-otel-start.ts
```diff
@@ -290,18 +290,20 @@ class AnonymousTelemetryManager {
...
-      return newId;
+      try {
+        writeFileSync(fsidPath, newId, {encoding: 'utf8', flag: 'wx'});
+        return newId;
+      } catch (writeError) {
+        if ((writeError as NodeJS.ErrnoException).code === 'EEXIST') {
+          const existingId = readFileSync(fsidPath, 'utf8').trim();
+          return existingId;
+        }
+        throw writeError;
```
darkgnotic, line 305: Unrelated to this fix but a relevant question: is it ever possible that we don't have write permissions to the home dir? Like in a container or something? 

I don't expect this to be common, but would that prevent zero-cache from starting up?

cesara: ya, it is possible but we catch the throw and return "unknown" as the fsid

PR #4644, opened 2025-07-17
chore(zero-cache): Fix exception when deleting client before TTL clock is read

arv:
The reading of the TTL clock from the CVR is async so it may not be
available when the client is deleted. If that is the case we do not try
to write the TTL clock back to the CVR. This avoids an exception that
can occur when the client is deleted before the TTL clock is read.


arv:
You can ignore the first commit. It just splits the test file (it is still huge)

PR #4651, opened 2025-07-18
chore(zero): wire client and server together for mutation results  [2/n]

tantaman:
Add to the `poke` protocol so updates to the `mutations` table can be poked down to the client.

On poke end, the client uses the merged poke result to resolve mutation promises. This happens after Replicache has been updated so a user can issue a read after their promise resolves and get the latest state.

### DIFF
--- a/packages/zero-client/src/client/custom.test.ts
+++ b/packages/zero-client/src/client/custom.test.ts
@@ -24,8 +24,27 @@ import type {WriteTransaction} from './replicache-types.ts';
 import {MockSocket, zeroForTest} from './test-utils.ts';
 import {createDb} from './test/create-db.ts';
 import {getInternalReplicacheImplForTesting} from './zero.ts';
+import type {Row} from '../../../zql/src/query/query.ts';
+import {refCountSymbol} from '../../../zql/src/ivm/view-apply-change.ts';
 
 type Schema = typeof schema;
+
+// beforeEach(() => {
+//   let frameId = 0;
+
+//   vi.stubGlobal(
+//     'requestAnimationFrame',
+//     vi.fn(callback => {
+//       frameId++;
+//       setTimeout(() => callback(performance.now()), 0);
+//       return frameId;
+//     }),
+//   );
+// });
+
+// afterEach(() => {
+//   vi.unstubAllGlobals();
+// });
 
 test('argument types are preserved on the generated mutator interface', () => {
   const mutators = {
@@ -492,6 +511,7 @@ describe('server results and keeping read queries', () => {
     await z.triggerConnected();
     await z.waitForConnectionState(ConnectionState.Connected);
 
+    const q = z.query.issue.limit(1).materialize();
     const create = z.mutate.issue.create({
       id: '1',
       title: 'foo',
@@ -502,7 +522,6 @@ describe('server results and keeping read queries', () => {
     });
     await create.client;
 
-    const q = z.query.issue.limit(1).materialize();
     q.destroy();
 
     z.queryDelegate.flushQueryChanges();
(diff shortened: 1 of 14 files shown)


packages/zero-client/src/client/zero-poke-handler.ts
```diff
@@ -262,16 +285,27 @@ export function mergePokes(
...
       lastMutationIDChanges: mergedLastMutationIDChanges,
       patch: mergedPatch,
       cookie,
     },
   };
+
+  // For backwards compatibility. Because we're strict on our validation,
+  // zero-client must be able to parse pokes with this field before we introduce it.
+  // So users can update their clients and then start using custom mutators that write responses to the db.
+  if (mutationResults.length > 0) {
```
grgbkr, line 305: Seems this could be simplified to just return the empty array instead of undefined, since the good that uses this return value just converts the undefined back into [].

packages/zero-client/src/client/zero-poke-handler.test.ts
```diff
@@ -842,6 +849,7 @@ test('replicachePoke throwing error calls onPokeError and clears', async () => {
     clientID,
     schema,
     logContext,
+    new MutationTracker(logContext),
```
grgbkr: Would be good to add a basic test of the interaction with MutationTracker, probably using a mock.

packages/zero-protocol/src/mutations-patch.ts
```diff
@@ -0,0 +1,11 @@
+import * as v from '../../shared/src/valita.ts';
+import {mutationResponseSchema} from './push.ts';
+
+export const putOpSchema = v.object({
```
grgbkr, line 11: Maybe a comment as to why there is only put.

packages/zero-client/src/client/custom.test.ts
```diff
@@ -24,9 +24,27 @@ import type {WriteTransaction} from './replicache-types.ts';
 import {MockSocket, zeroForTest} from './test-utils.ts';
 import {createDb} from './test/create-db.ts';
 import {getInternalReplicacheImplForTesting} from './zero.ts';
+import type {Row} from '../../../zql/src/query/query.ts';
 
 type Schema = typeof schema;
 
+// beforeEach(() => {
```
grgbkr: remove?

packages/zero-client/src/client/custom.test.ts
```diff
@@ -537,12 +556,16 @@ describe('server results and keeping read queries', () => {
...
+    // lmid advancement is not in a RAF callback
+    // so tick a few times
+
     // mutation is no longer outstanding, query is removed.
-    expect(filter(messages)).toMatchInlineSnapshot(`
-      [
-        "["changeDesiredQueries",{"desiredQueriesPatch":[{"op":"del","hash":"1vsd9vcx6ynd4"}]}]",
-      ]
-    `);
+    await vi.waitFor(() => {
```
grgbkr, line 547: vi.waitFor is new to me, pretty nice test helper.

packages/zero-client/src/client/custom.test.ts
```diff
@@ -584,15 +608,107 @@ describe('server results and keeping read queries', () => {
...
+      ]);
+    });
+
     messages.length = 0;
 
     await z.close();
-    await expect(close.server).rejects.toEqual({error: 'app'});
+  });
+
+  test.only('after the server promise resolves (via poke), reads from the store return the data from the server', async () => {
```
grgbkr: remove test.only.

grgbkr (approved):

packages/zero-client/src/client/zero-poke-handler.ts
tantaman: I'll have to confirm again, but I believe we use `strict` when receiving poke messages on the client. If a server sends `mutationResults` to a client that is not expecting it, the client will fatal.

I'd like everyone to be able to deploy the next zero release, get their clients up to date, then turn on the "mutations write to db" feature.

grgbkr: We should update to not use strict parse.  I'll send a PR for this.  We should be able to introduce new fields on the server and deploy the server before the client is deployed.  We expect server updates to be deployed before the client update, and adding new fields is perfectly backwards compatible.  

I'm still not understanding how that relates to this logic and the comment above about backwards compatibility.   This code is after we've parsed the poke message.

PR #4652, opened 2025-07-18
feat(zero): offline

0xcadams:
There was a usability issue with the online status with Zero. React doesn't rerender when Zero's online property updates. This proposes an API, `onOnline`, to make it easier to listen for updates to this. And an accompanying hook in React/SolidJS, `useZeroOnline`.

### DIFF
--- a/apps/zbugs/src/zero-init.tsx
+++ b/apps/zbugs/src/zero-init.tsx
@@ -1,9 +1,8 @@
 import {ZeroProvider} from '@rocicorp/zero/react';
 import {useLogin} from './hooks/use-login.tsx';
-import {createMutators, type Mutators} from '../shared/mutators.ts';
+import {createMutators} from '../shared/mutators.ts';
 import {useMemo, type ReactNode} from 'react';
-import {schema, type Schema} from '../shared/schema.ts';
-import type {ZeroOptions} from '@rocicorp/zero';
+import {schema} from '../shared/schema.ts';
 
 export function ZeroInit({children}: {children: ReactNode}) {
   const login = useLogin();
@@ -22,7 +21,7 @@ export function ZeroInit({children}: {children: ReactNode}) {
         }
         return login.loginState?.encoded;
       },
-    } as const satisfies ZeroOptions<Schema, Mutators>;
+    };
   }, [login]);
 
   return <ZeroProvider {...props}>{children}</ZeroProvider>;


apps/zbugs/src/pages/issue/issue-composer.tsx
```diff
@@ -49,18 +51,25 @@ export function IssueComposer({isOpen, onDismiss}: Props) {
...
       created: Date.now(),
       modified: Date.now(),
     });
+
     reset();
     onDismiss(id);
+
+    const raceResult = await promiseRace([sleep(5000), result.server]);
+    if (raceResult === 0) {
+      // TODO show toast
```
0xcadams: Do we want this to be a toast?

packages/zero-client/src/client/zero.ts
```diff
@@ -897,6 +903,9 @@ export class Zero<
 
     lc.debug?.('Closing Zero instance. Stack:', new Error().stack);
 
+    // should we unsubscribe all listeners here?
+    // this.#onlineManager.cleanup();
```
0xcadams: TODO calling this out for feedback

aboodman:
I'm confused - we already have `onOnlineChange`. Why not use that at the basis for `useOnline`?

aboodman:
@alexhking can you design an online indicator for zbugs? Note that people will probably copy whatever we do into their apps.  

Here are my product thoughts:

1. I do ___not___ (cannot stress enough how much I do not want this) want it to animate or throb or do anything when syncing. This just becomes a spinner and destroys the feeling of instant.
2. I think when online, it should be invisible. Linear doesn't have an 'online' state. You assume it is online.
3. I am not sure where it should go visually.
4. As always I think it should be simple, minimal, tasteful.
5. Maybe we should consider animations and transitions when going offline and returning online.

0xcadams:
> I'm confused - we already have `onOnlineChange`. Why not use that at the basis for `useOnline`?

Yeah I can do that - it just means managing that state in two places and adds a new Context. Will change.

aboodman:
> > I'm confused - we already have `onOnlineChange`. Why not use that at the basis for `useOnline`?
> 
> Yeah I can do that - it just means managing that state in two places and adds a new Context. Will change.

I am missing something.... it seems trivial to wrap a hook around the existing feature. I don't get why any change is needed in zero.ts.

aboodman:
Oooh it is because there can be many hooks, but there can only be one `onOnlineChange` event. Yeah that was super ghetto, let's fix that. Good call. Will review PR with that frame.

apps/zbugs/src/pages/issue/issue-composer.tsx
aboodman: I think that we should rely on a new global `onError` handler that Matt is constantly wanting to add. It is half-implemented in the codebase right now, but we got distracted by API design questions and never finished it.

I don't want every mutator to have to do this kind of error handling manually.

apps/zbugs/src/zero-init.tsx
```diff
@@ -21,7 +22,10 @@ export function ZeroInit({children}: {children: ReactNode}) {
         }
         return login.loginState?.encoded;
       },
-    };
+    } as const satisfies ZeroOptions<
+      typeof schema,
+      ReturnType<typeof createMutators>
```
aboodman: The `Mutators` type is already exported.

packages/shared/src/subscribable.ts
```diff
@@ -0,0 +1,38 @@
+export class Subscribable<
```
aboodman, line 1: I was going to suggest nanoevent but this is so simple maybe it makes sense to have it in our codebase in case we want to change it.

@arv @tantaman ?

aboodman: Also @arv just confirming there's nothing like this already in Replicache so we don't duplicate code?

packages/zero-client/src/client/zero.ts
aboodman: I would say yes to be safe, but technically it should not be necessary since if Zero/Replicache is not buggy no  more events should be sent.

packages/zero-client/src/client/zero.ts
```diff
@@ -1787,22 +1796,32 @@ export class Zero<
...
-    this.#online = online;
-    this.#onOnlineChange?.(online);
+    this.#onlineManager.setOnline(online);
   }
 
   /**
    * A rough heuristic for whether the client is currently online and
    * authenticated.
+   *
+   * This does *not* update in React when the online status changes.
```
aboodman: I don't think we should have this comment since it standard that any simple field is not reactive. We don't want to add this warning to every field in the Zero API.

packages/zero-client/src/client/zero.ts
```diff
@@ -1787,22 +1796,32 @@ export class Zero<
...
   }
 
+  /**
+   * Subscribe to online status changes.
+   *
+   * This is useful when you want to update state based on the online status.
+   * @param listener - The listener to subscribe to.
+   * @returns A function to unsubscribe the listener.
+   */
+  addOnlineChangeListener = (
```
aboodman: How about `onOnline` and then we deprecate `onOnlineChange`.

packages/zero-client/src/client/zero.ts
```diff
@@ -454,7 +455,12 @@ export class Zero<
       );
     }
 
-    this.#onOnlineChange = onOnlineChange;
```
aboodman, line 458: This is a breaking change but it is so simple to update I am supportive.

apps/zbugs/src/pages/issue/issue-composer.tsx
0xcadams: This makes sense - agreed it would be much better to have this in `onError`. So I can add this toast later and only have this PR include online indicator, once the onError handler lands.

0xcadams:
@alexhking I changed the design based on the above comments. Open to feedback. See the description for the current design. I can add tooltip, etc.

tantaman:
@aboodman  - should we just merge rather than having this open/blocked and fix forward? Seems annoying and unproductive to have to have PRs open for such a long time (having to rebase and resolve merge conflicts to revisit).

alexhking:
> @alexhking I changed the design based on the above comments. Open to feedback. See the description for the current design. I can add tooltip, etc.

Hey Chase, for some reason I didn't get notified when you sent this last week. Checking out now and will give feedback.

alexhking:
@0xcadams @aboodman 

Ok, after reviewing I have several questions and comments:

* Do we want to completely remove the state of the visual representation of the user being logged in when offline?
* If so, why are we still showing the log out control?
* The “Offline” badge has a completely different style than the logged in user state, and I’m not sure why. Right now it seems to closely resemble what we use for labels, but with different padding.
* It may be nice to add a tooltip to explain a bit more about what offline means and what limitations there are.

I would like to take a stab at redesigning this just a bit at least visually.

aboodman:
> @aboodman - should we just merge rather than having this open/blocked and fix forward? Seems annoying and unproductive to have to have PRs open for such a long time (having to rebase and resolve merge conflicts to revisit).

Typically we would but just because Chase was new to the team I suggested he block on reviews and start conservatively. Will re-evaluate.

apps/zbugs/src/zero-init.tsx
```diff
@@ -21,7 +22,7 @@ export function ZeroInit({children}: {children: ReactNode}) {
         }
         return login.loginState?.encoded;
       },
-    };
```
aboodman: I don't understand why this change is needed.

packages/zero-client/src/client/options.ts
```diff
@@ -118,6 +118,12 @@ export interface ZeroOptions<
 
   /**
    * `onOnlineChange` is called when the Zero instance's online status changes.
+   *
+   * @deprecated Use `onOnline` on the Zero instance instead. e.g.
```
aboodman, line 122: Nuts. @grgbkr had gone to a lot of trouble to pass in all the configuration to `Zero`. But I think didn't consider this problem/use-case. So maybe passing in all configuration cannot work.

0xcadams:
> * Do we want to completely remove the state of the visual representation of the user being logged in when offline?
> * If so, why are we still showing the log out control?
> * The “Offline” badge has a completely different style than the logged in user state, and I’m not sure why. Right now it seems to closely resemble what we use for labels, but with different padding.
> * It may be nice to add a tooltip to explain a bit more about what offline means and what limitations there are.
> 
> I would like to take a stab at redesigning this just a bit at least visually.

On these - I am completely open to the design you want to go with. I am still getting used to what design styles the team likes, so I did not put a lot of time into the design (this was a very loose copy of how Linear does it). Please let me know if you want me to add tooltip, etc.

apps/zbugs/src/zero-init.tsx
0xcadams: It's not - an artifact of other changes made and then kept because it added type safety. Can remove

packages/zero-client/src/client/options.ts
0xcadams: We can support both here. The adding/removal of listeners doesn't need to be all up-front. And this doesn't need to be deprecated.

Or, if it's better, we can move this all to the React/Solid layers. And continue to use `onOnlineChange` and track the online state with React Context.

aboodman:
> @0xcadams @aboodman
> 
> Ok, after reviewing I have several questions and comments:
> 
> * Do we want to completely remove the state of the visual representation of the user being logged in when offline?

I don't think so... is there some reason to do @0xcadams ? Prefer to keep login state looking the same while offline.

> * It may be nice to add a tooltip to explain a bit more about what offline means and what limitations there are.
> I would like to take a stab at redesigning this just a bit at least visually.

Please! Sounds great!

packages/zero-client/src/client/options.ts
grgbkr: I think it will be very convoluted to track the listeners state in the React/Solid layers.  I'm ok with this change.  

The other to event listener type configs we currently have are `onUpdateNeeded` and `onClientStateNotFound`.  Possibly these should move on to the Zero instance as well, and our general pattern would be event listeners on the Zero instance.  I think these are unlikely to need multiple listeners like onOnline might, but it may just be more ergonomic to have them on Zero given React hooks.  I think it would be good to build a UI for `onUpdateNeeded` to figure out what is most ergonomic.

apps/zbugs/src/zero-init.tsx
aboodman: What does it improve about type safety? Currently for me I do not see any type errors in my env w/o this.

The reason I'm asking is because zbugs is a testbed for how we expect Zero to be used and I wasn't expecting this to be required of users.

0xcadams:
I have removed all changes to zbugs so we can update the styling/display of offline in a future PR.

aboodman:
I think it is fine to land the zbugs changes and let Alex style them
separately. It wasn’t bad enough to delay. And having the usage of the api
somewhere is great for dogfooding purposes.

a (phone)

0xcadams:
Sounds good - added back and merging and will note to follow up w/ @alexhking to get better styling here.

PR #4653, opened 2025-07-18
chore(zero-cache): Rename (swap) Sender and Receiver process interfaces

darkgnotic:
The previous naming was intended to represent the destination process, i.e. you call "send()" on the receiving process. 

But this is kind of confusing.

The new naming is more intuitive: Sender.send() and Receiver.onMessage().


arv:
LGTM

PR #4660, opened 2025-07-20
chore(zero-cache): Fix assertion related to TTL clock

arv:
The main issue was that we were using `#getTTLClock` in `handleConfigUpdate` before we had initialized `#ttlClock`. By always reading the ttlClock from the CVR we can simplify the logic and avoid the assertion failure.


arv:
@cesara FYI

arv:
@darkgnotic I'm not very happy with how the ttlClock ended up. I know you are on vacation, but when you get back I would really appreciate it if you could take a closer look and see if you can think of something cleaner and more reliable? Thanks

PR #4663, opened 2025-07-23
chore(zero): expand protocol to support acking mutation responses

tantaman:
https://rocicorp.slack.com/archives/C013XFG80JC/p1753285835544609

note: The `ack` message will not be sent until `zero-cache` begins including mutation results in `pokes`.

I'll be updating the `push protocol` to `version 2` in a future PR. When the API server sees a push at that version it will write mutation results to the `mutations` table which will cause `zero-cache` to poke them down.

### DIFF
--- a/packages/zero-cache/src/config/zero-config.test.ts
+++ b/packages/zero-cache/src/config/zero-config.test.ts
@@ -27,370 +27,371 @@ test('zero-cache --help', () => {
   expect(logger.info).toHaveBeenCalled();
   expect(stripAnsi(logger.info.mock.calls[0][0])).toMatchInlineSnapshot(`
     "
-     --upstream-db string                                        required                                                                                                                              
-       ZERO_UPSTREAM_DB env                                                                                                                                                                            
-                                                                 The "upstream" authoritative postgres database.                                                                                       
-                                                                 In the future we will support other types of upstream besides PG.                                                                     
-                                                                                                                                                                                                       
-     --upstream-max-conns number                                 default: 20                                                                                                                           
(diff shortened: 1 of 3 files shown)


packages/zero-protocol/src/protocol-version.ts
```diff
@@ -29,7 +29,8 @@ import {assert} from '../../shared/src/asserts.ts';
 // -- Version 21 removes `AST` in downstream query puts which was deprecated in Version 17, removes support for versions < 18 (0.22)
 // -- Version 22 adds an optional 'userQueryParams' field to `initConnection` (0.22)
 // -- Version 23 add `mutationResults` to poke (0.22)
-export const PROTOCOL_VERSION = 23;
+// -- Version 24 adds `ackMutationResults` to poke (0.22).
```
grgbkr, line 32: `to poke` -> `to upstream protocol`

packages/zero-client/src/client/mutation-tracker.ts
```diff
@@ -104,15 +111,23 @@ export class MutationTracker {
...
+        }
 
-      if ('error' in patch.mutation.result) {
-        this.#processMutationError(patch.mutation.id, patch.mutation.result);
-      } else {
-        this.#processMutationOk(patch.mutation.id, patch.mutation.result);
+        if ('error' in patch.mutation.result) {
+          this.#processMutationError(patch.mutation.id, patch.mutation.result);
+        } else {
+          this.#processMutationOk(patch.mutation.id, patch.mutation.result);
```
grgbkr, line 123: how would we adapt this if we decided to only write responses for errors and mutations with a return value?

PR #4667, opened 2025-07-25
fix(ttl): fix incorrect calculation of expired queries

tantaman:
There are two paths the check if queries are expired:

1. `hasExpiredQueries` in view-syncer.ts
2. `getInactiveQueries` in cvr.ts

(1) only returns queries that are expired for all clients. (2) was returning queries if they were expired for 1 or more clients.

On reading, it appears that both paths should caluclate expiration the same way. The second path should not return a query as inactive if one of its client has not inactivated it.

It looks like this was addressed before here: https://github.com/rocicorp/mono/pull/3843 but it did not take into account the `undefined` case. Or I'm missing something.

before (re-schedules eviction every 50ms because it thinks a query is expired but it is not):

https://github.com/user-attachments/assets/42459a6a-2716-4a94-b843-53aa37475206

after (stable):

https://github.com/user-attachments/assets/e8893bd4-be06-4290-9016-7b46eb33ad65

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/cvr.ts
+++ b/packages/zero-cache/src/services/view-syncer/cvr.ts
@@ -917,6 +917,14 @@ function mergeRefCounts(
   return Object.values(merged).some(v => v > 0) ? merged : null;
 }
 
+/**
+ * The query must be inactive for all clients to be considered inactive.
+ * This is because expiration is defined that way: a query is expired for a client group
+ * only if it is expired for all clients in the group.
+ *
+ * If all clients have inactivated the query, we return
+ * the one with the expiration furthest in the future.
+ */
 export function getInactiveQueries(cvr: CVR): {
   hash: string;
   inactivatedAt: TTLClock;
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/services/view-syncer/cvr.ts
```diff
@@ -936,29 +944,36 @@ export function getInactiveQueries(cvr: CVR): {
...
         }
+        break;
+      }
+
+      const clampedTTL = clampTTL(ttl);
+      if (existing) {
+        // The stored one might be too large because from a previous version of
+        // zero
+        const existingTTL = clampTTL(existing.ttl);
+        // Use the last eviction time.
```
grgbkr, line 963: Would 
// Use the expiration furthest in the future.

be a more accurate comment?

grgbkr:
Fix looks correct to me but probably a good idea for @arv to review as well.

arv:
The PR description looks correct. I haven't looked at the actual code yet.

Feel free to land I'll take a look later

packages/zero-cache/src/services/view-syncer/cvr.ts
```diff
@@ -917,6 +917,14 @@ function mergeRefCounts(
   return Object.values(merged).some(v => v > 0) ? merged : null;
 }
 
+/**
+ * The query must be inactive for all clients to be considered inactive.
+ * This is because expiration is defined that way: a query is expired for a client group
```
arv, line 922: This sentence reads a bit odd.

arv (commented):
LGTM

I looked at the code in detail and it looks good to me.

Thank you

PR #4669, opened 2025-07-25
feat(zero-client): expo-sqlite storage

0xcadams:
Adds a new `@rocicorp/zero/expo` package to add a SQLite `StoreProvider` for Expo/React Native.

```tsx
import { schema } from "@/schema";
import { expoSQLiteStoreProvider } from "@rocicorp/zero/expo";
import { ZeroProvider } from "@rocicorp/zero/react";
import { Stack } from "expo-router";

const storeProvider = expoSQLiteStoreProvider();

export default function RootLayout() {
  return (
    <ZeroProvider
      kvStore={storeProvider}
      server="http://localhost:4848"
      userID="anon"
      schema={schema}
    >
      <Stack />
    </ZeroProvider>
  );
}
```

Builds on @austinm911's work.

- Adds a core `SQLiteDatabaseManager` which handles schema setup, PRAGMAs, and connection handling.
- Adds a generic `SQLiteStore` which works with bare SQLite and uses prepared statements with `RWLock`.

**Note:**
iOS and Android both had issues with WAL journal mode - they hang on COMMIT indefinitely. [According to Expo](https://docs.expo.dev/versions/latest/sdk/sqlite/#executing-pragma-queries), this is recommended, but it doesn't seem to work, so it is disabled by default, with a performance hit. Here is the performance difference with `better-sqlite3`:

Screenshot 2025-07-28 at 10 23 53 AM

### Expo Demo

https://github.com/user-attachments/assets/99274ac3-4b46-4a60-ada8-e4fcfd74d094


arv:
Exciting! I'm on vacation until beginning of August so I will not be able to provide any good feedback on this until then.

packages/replicache/src/kv/sqlite-store.bench.ts
```diff
@@ -0,0 +1,168 @@
...
+  walSQLite3DatabaseManager.clearAllStoresForTesting();
+  defaultSQLite3DatabaseManager.clearAllStoresForTesting();
+});
+
+describe('zero-sqlite3 WAL', () => {
+  bench(
+    `write`,
+    async () => {
+      await withWrite(walStore, async wt => {
+        await wt.put('foo1', 'bar1');
```
tantaman: what is the bench attempting to measure? Just raw _transactions_ per second? Or do you want more information on _writes_ per second?

Writes per second should be roughly on the order of `transactions_per_second * writes_per_transaction` so 1 write per transaction will give you the same writes per second as transactions per second.

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  async read(): Promise<Read> {
+    const release = await this.#rwLock.read();
+    try {
+      const read = new SQLiteStoreRead(this.#preparedStatements, release);
+
+      return Promise.resolve(read);
+    } catch (e) {
+      return Promise.reject(e);
```
tantaman: do you not need to release the lock here if the above threw?

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  async write(): Promise<Write> {
+    const release = await this.#rwLock.write();
+    try {
+      const write = new SQLiteStoreWrite(this.#preparedStatements, release);
+
+      return Promise.resolve(write);
+    } catch (e) {
+      return Promise.reject(e);
```
tantaman: same question as with read.

It does feel weird to be doing locking given the interface we expose to SQLite is technically synchronous.

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,340 @@
...
+
+    return promiseVoid;
+  }
+
+  get closed(): boolean {
+    return this.#closed;
+  }
+}
+
+export class SQLiteStoreRead implements Read {
```
tantaman: Curious why the reads are not run in a sqlite transaction.

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,340 @@
...
+    this.#preparedStatements = preparedStatements;
+  }
+
+  commit(): Promise<void> {
+    if (this._pending.size === 0) return promiseVoid;
+
+    // we bundle writes together and run the commit in a synchronous block
+    const next = new Promise<void>(resolve => {
+      // BEGIN IMMEDIATE grabs an RESERVED lock
+      this.#preparedStatements.beginImmediate.run();
```
tantaman: why do we only begin the sqlite transaction on commit?

Yes, the read-write locks should be ensuring proper concurrency control (so long as no other _process_ is interacting with the db) but, for performance, it seems that we would want to scope the transaction around all operations in `withWrite` and `withRead` as each `read` operation will open an implicit transaction if one is not already open.

tantaman (commented):
Thanks for figuring this out

packages/replicache/src/kv/sqlite-store.ts
0xcadams: Yeah this is good point. I will continue to look into scoping the transaction higher - originally there were issues with sharing a single SQLite connection across the StoreProvider. The async reads/writes were causing interleaving of statements ("cannot start a transaction inside a transaction" type errors). This was before adding locking though.

packages/replicache/src/kv/sqlite-store.ts
0xcadams: This should now be fixed

packages/replicache/src/kv/sqlite-store.bench.ts
0xcadams: Yeah this is a good question. I was using this benchmark to improve performance of the SQLite store w/ better-sqlite3, as a way to test performance of the store as I was making changes to schema/PRAGMAs. I updated this to be simpler and compare WAL modes. Let me know if this doesn't fit into existing benchmarks or any ways you see improving this.

packages/replicache/src/kv/sqlite-store.ts
tantaman: Yeah, the locks should fix all of that.

0xcadams: This is done now and the transactions are opened/closed around `withWrite` and `withRead`. With explicit transactions in both.

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,376 @@
...
+  isInTransaction(): boolean;
+}
+
+type SQLitePreparedStatements = {
+  begin: PreparedStatement;
+  beginImmediate: PreparedStatement;
+  commit: PreparedStatement;
+  rollback: PreparedStatement;
+
+  savepoint: PreparedStatement;
```
aboodman: Surprised to see savepoint/release here. I didn't expect anything needed in the impl other than normal begin/commit/rollback.

savepoint/release seem to be used to allow nested transactions. We don't want nested transactions.

Either me / @arv / or @grgbkr need to review the semantics of this I think.

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,376 @@
...
+  // Similar to https://github.com/WiseLibs/better-sqlite3/blob/674ce6be68a26742d9e24f8672da7888cea0aebb/lib/methods/transaction.js#L35-L39
+  savepoint: db.prepare('SAVEPOINT `\t_rc.\t`'),
+  release: db.prepare('RELEASE `\t_rc.\t`'),
+
+  get: db.prepare('SELECT value FROM entry WHERE key = ?'),
+  put: db.prepare('INSERT OR REPLACE INTO entry (key, value) VALUES (?, ?)'),
+  del: db.prepare('DELETE FROM entry WHERE key = ?'),
+});
+
+const rwLocks = new Map<string, RWLock>();
```
aboodman: It's unfortunate to need multiple levels of locking but I guess it is needed so we dont' have to create a connection per read/write.

aboodman: I do not think we need a global map of locks though. Each SQLiteStore can own a connection to SQLite which uses SQLite's built-in file based locking. We don't expect lots of instances of SQLiteStore for the same name (or really any?) in the same js context. Each SQLiteStore can then also own a RWLock for in-memory queuing.

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,376 @@
...
+
+  constructor(
+    preparedStatements: SQLitePreparedStatements,
+    isInTransaction: boolean,
+    release: () => void,
+  ) {
+    super(preparedStatements, release);
+    this.#isInTransaction = isInTransaction;
+
+    if (!this.#isInTransaction) {
```
aboodman: We should just use BEGIN here and let it queue I think.

packages/replicache/src/kv/sqlite-store.ts
0xcadams: Yeah that makes sense. This was for safety but I will remove if there's that guarantee. Yeah, exactly, the alternative was spinning up many connections.

packages/replicache/src/kv/sqlite-store.ts
0xcadams: Yeah actually this is a condition of opening multiple async read transactions. If you look at `multiple reads at the same time` it will fail with `cannot start a transaction within a transaction` if it's not a nested transaction.

Basically two `withRead` promises executed without awaiting can encounter a situation. My understanding:

1. The first read runs which starts a transaction with `BEGIN`
2. This yields to the event loop and the second read starts and tries to start its own tx with `BEGIN`
3. There is already a transaction open so this fails

So it seems there are two options:
1. Don't open explicit transactions for read.
2. Use savepoints and allow nested transactions for read.

tantaman: > Don't open explicit transactions for read.

We want explicit transactions for read so if many reads are done in `withRead` they are all consistent. One could argue that the `rwLock` will guarantee this but a separate process  _could_ modify the SQLite file.

> Use savepoints and allow nested transactions for read.

that or, if you're already in a read transaction, don't do anything. E.g.,

```ts
if (inTransaction) {
  // run the callback
} else {
  // begin a transaction
  // then run the callback
}
```

0xcadams: Another option would be to use separate sqlite connections for reads. E.g. a pool of readers with N=CPU cores. What do you think?

tantaman: You know the other problem with `withRead` is that you don't know when to commit.

The transactions may not be nested but rather run as siblings

```ts
await Promise.all([withRead(), withRead(), withRead()]);
```

I think the current code doesn't handle this as the first withRead to run would try to commit the outer transaction when it finishes, the other `withReads` would not have yet closed their savepoints.

Separate connections or serializing reads seem like the only options.

0xcadams:
Changed this to have a read connection pool and a single writer connection.

There is also still a global RWLock. This is because in this test:

https://github.com/rocicorp/mono/blob/891d60a092ddda9eba0be228b51b84bc880fc145/packages/replicache/src/kv/store-test-util.ts#L238-L287

r1, r2 start → w1 runs/commits → w2 calls write() (now waiting) → r3 sneaks in, starts a
new read transaction and observes the database state that still contains “w1” →
w2 finally gets the write lock, overwrites the value with “w2”, commits

0xcadams:
I benchmarked performance (with `@rocicorp/zero-sqlite3`) of using a read connection pool versus opening a new connection on every read.

TLDR; the read pooling is roughly 8-14x faster.

The last test, `plain read`, is probably the clearest comparison, since it's just running 5 reads in parallel.

```tsx
bench(
  `plain read`,
  async () => {
    const readP1 = withRead(store, async rt => {
      expect(await rt.get('foo')).equal('bar');
    });
    const readP2 = withRead(store, async rt => {
      expect(await rt.get('foo')).equal('bar');
    });
    const readP3 = withRead(store, async rt => {
      expect(await rt.get('foo')).equal('bar');
    });
    const readP4 = withRead(store, async rt => {
      expect(await rt.get('foo')).equal('bar');
    });
    const readP5 = withRead(store, async rt => {
      expect(await rt.get('foo')).equal('bar');
    });

    await Promise.all([readP1, readP2, readP3, readP4, readP5]);
  },
  {
    throws: true,
    setup: async () => {
      await withWrite(store, async wt => {
        await wt.put('foo', 'bar');
      });
    },
  },
);
```

**With a read connection pool of 2:**

Screenshot 2025-08-06 at 11 25 25 AM


**With a new connection on every read:**

Screenshot 2025-08-06 at 11 25 09 AM

packages/replicache/src/kv/sqlite-store-test-util.ts
```diff
@@ -0,0 +1,54 @@
...
+            fs.unlinkSync(filename);
+          }
+        },
+        prepare: (sql: string) => {
+          const stmt = db.prepare(sql);
+          if (logging) {
+            console.log('prepare', sql);
+          }
+          return {
+            all: <T>(params?: unknown[]): T[] => {
```
arv: This can be `all(params: unknown[] = []) {}` and you can remove the ternary.

Same in `run`

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,528 @@
...
+      };
+    };
+
+    const release = await lock();
+
+    return {
+      preparedStatements: entry.preparedStatements,
+      release,
+    };
+  }
```
arv: I don't think you need the `lock` function here. You can inline it.

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,528 @@
...
+
+  constructor(
+    name: string,
+    dbm: SQLiteDatabaseManager,
+    opts: SQLiteDatabaseManagerOptions,
+  ) {
+    this.#name = name;
+    this.#dbm = dbm;
+
+    this.#rwLock = new RWLock();
```
arv: move to declaration

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,528 @@
...
+    if (unsafeValue === undefined) return promiseUndefined;
+    const parsedValue = JSON.parse(unsafeValue) as ReadonlyJSONValue;
+    const frozenValue = deepFreeze(parsedValue);
+    return Promise.resolve(frozenValue);
+  }
+
+  #getSql(key: string): string | undefined {
+    const rows = this._preparedStatements.get.all<{value: string}>([key]);
+    if (rows.length === 0) return undefined;
+    return rows?.[0]?.value;
```
arv:
```suggestion
    return rows[0].value;
```

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,528 @@
...
+    release: () => void,
+  ) {
+    super(preparedStatements, release);
+
+    // BEGIN IMMEDIATE grabs a RESERVED lock
+    this._preparedStatements.beginImmediate.run();
+  }
+
+  put(key: string, value: ReadonlyJSONValue): Promise<void> {
+    this._preparedStatements.put.run([key, JSON.stringify(value)]);
```
arv: is there a reason why the type of run takes rest args but the callers always wrap in a single array? Consider removing one array allocation.

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,528 @@
...
+      this.#ensureSchema(newDb);
+    }
+
+    if (opts.busyTimeout !== undefined) {
+      // we set a busy timeout to wait for write locks to be released
+      const busyTimeout = newDb.prepare(
+        `PRAGMA busy_timeout = ${opts.busyTimeout}`,
+      );
+      busyTimeout.run();
+      busyTimeout.finalize();
```
arv: Nit

```js
const exec = sql => {
  const statement = newDb.prepare(sql);
  statement.run();
  statement.finalize();
}
```

packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -0,0 +1,528 @@
...
+      db.close();
+    }
+    for (const db of dbInstance.dbs) {
+      db.destroy();
+    }
+    this.#dbInstances.delete(name);
+  }
+
+  #ensureSchema(db: SQLiteDatabase): void {
+    const begin = db.prepare('BEGIN IMMEDIATE');
```
arv: didn't you have this prepared already?

arv (approved):
LGTM

I know Matt is reviewing this already but I was very curious.

I looked at the locking structure and it seems to have the same semantics as the IDBStore.

packages/replicache/src/kv/sqlite-store.ts
0xcadams: Good catch - moved to rest args

packages/replicache/src/kv/sqlite-store-test-util.ts
```diff
@@ -0,0 +1,54 @@
...
+import sqlite3 from '@rocicorp/zero-sqlite3';
+import fs from 'fs';
+import path from 'path';
+import {SQLiteDatabaseManager} from './sqlite-store.ts';
+
+export const getTestSQLiteDatabaseManager = (logging: boolean = false) =>
+  new SQLiteDatabaseManager({
+    open: name => {
+      const filename = path.resolve(__dirname, `${name}.db`);
+      // this cannot be :memory: because multiple read connections must access
```
tantaman, line 11: it's technically doable via memory with some fancy options. But yeah, better to just keep it file based.

tantaman (approved):
lgtm

aboodman:
OK on more thought I understand the reason for the design here.

I was wondering why we cannot rely on RWLock within the context and thereby share a single connection across the context. But the problem is that RWLock wants to allow read tx to overlap. And we cannot represent overlapped, separate read tx with a single connection (at least not cleanly?).

LGTM too.

0xcadams:
Unfortunately the tests keep failing and I don't have permissions to override the PR requirements

PR #4671, opened 2025-07-27
feat(zero): add support for 'time' datatype

xfh:
Hi

I have some existing database schema that uses postgres' `time` datatype.
Zero doesn't allow me to sync these tables. 

This is an attempt to add support for `time`, relying on postgres' parser for string formats.
https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-INPUT-TIMES

I don't know how I can build a version of zero that I can test locally. Some test are failing, but unrelated to my changes.
Any guidance is welcome.

### DIFF
--- a/packages/zero-schema/src/permissions.ts
+++ b/packages/zero-schema/src/permissions.ts
@@ -19,6 +19,10 @@ import {clientToServer, NameMapper} from './name-mapper.ts';
 export const ANYONE_CAN = [
   (_: unknown, eb: ExpressionBuilder<Schema, never>) => eb.and(),
 ];
+
+/**
+ * @deprecated Use {@link ANYONE_CAN} instead.
+ */
 export const ANYONE_CAN_DO_ANYTHING = {
   row: {
     select: ANYONE_CAN,
@@ -30,6 +34,7 @@ export const ANYONE_CAN_DO_ANYTHING = {
     delete: ANYONE_CAN,
   },
 };
+
 export const NOBODY_CAN = [];
 
 export type Anchor = 'authData' | 'preMutationRow';
@@ -54,13 +59,25 @@ export type AssetPermissions<
 > = {
   // Why an array of rules?: https://github.com/rocicorp/mono/pull/3184/files#r1869680716
   select?: PermissionRule<TAuthDataShape, TSchema, TTable>[] | undefined;
+  /**
+   * @deprecated Use Mutators instead.
+   * @see {@link https://zero.rocicorp.dev/docs/writing-data}
+   */
   insert?: PermissionRule<TAuthDataShape, TSchema, TTable>[] | undefined;
+  /**
+   * @deprecated Use Mutators instead.
+   * @see {@link https://zero.rocicorp.dev/docs/writing-data}
+   */
   update?:
     | {
         preMutation?: PermissionRule<TAuthDataShape, TSchema, TTable>[];
         postMutation?: PermissionRule<TAuthDataShape, TSchema, TTable>[];
       }
     | undefined;
+  /**
+   * @deprecated Use Mutators instead.
+   * @see {@link https://zero.rocicorp.dev/docs/writing-data}
+   */
   delete?: PermissionRule<TAuthDataShape, TSchema, TTable>[] | undefined;
 };
 


tantaman:
Curious why you chose `string` rather than `bigint` and store `nanosecond` precision in zero-cache for `time`?

Seems easier to deal with time as "number of nanoseconds since the beginning of the day" rather than as a string.

xfh:
I think it is a lot more natural to work with strings, because of the browser support:
The [HTML time input](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/time) element supports strings.
The recent [Temporal PlainTime](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainTime) supports parsing an RFC 9557 string as well (using the `from`  static method).

Numeric values are a bit odd for `time`, because you will always have to convert it to something useful (user facing), whereas the strings can often be used directly. Also, you need to know the resolution of the numeric value, but you can't tell just from the value, which I always found harder to work with.
By the way, postgres supports only microseconds resolution for time.

xfh:
Hi @tantaman, please let me know what the idea of the pipeline-driver test is, regarding this column definition: https://github.com/rocicorp/mono/blob/ab93b206982e3c8bc52bfdcd58b343dfef262f2f/packages/zero-cache/src/services/view-syncer/pipeline-driver.test.ts#L61

Should the test reflect how it deals with unsupported datatypes? In that case, the "ignored" type can simply change to `BYTEA` to make the test pass.

Based on the values that are set later on and the git history, it looks like typical timestamp values.
https://github.com/rocicorp/mono/blob/ab93b206982e3c8bc52bfdcd58b343dfef262f2f/packages/zero-cache/src/services/view-syncer/pipeline-driver.test.ts#L84-L85

I can change the type to `timestamp` and update the expectations, or set some `time` values as strings if you prefer. Let me know what you prefer.

tantaman:
Pulling this down to check

tantaman:
`microseconds since 00` just seems less likely to get munged up and much easier to use anytime you need to do some sort of computation on a date.

To convert to a string it should be:

```ts
function microsecondsToTimeString(microseconds: number): string {
  // Convert microseconds to seconds
  const totalSeconds = Math.floor(microseconds / 1_000_000);
  
  // Get the fractional microseconds (remainder that does not fit into a whole second)
  const fractionalMicroseconds = microseconds % 1_000_000;
  
  // Calculate hours, minutes, and seconds
  const hours = Math.floor(totalSeconds / 3600);
  const minutes = Math.floor((totalSeconds % 3600) / 60);
  const seconds = totalSeconds % 60;
  
  // Format with leading zeros
  const hoursStr = hours.toString().padStart(2, '0');
  const minutesStr = minutes.toString().padStart(2, '0');
  const secondsStr = seconds.toString().padStart(2, '0');
  const microsecondsStr = fractionalMicroseconds.toString().padStart(6, '0');
  
  return `${hoursStr}:${minutesStr}:${secondsStr}.${microsecondsStr}`;
}
```

tantaman:
the problem with the tests was that all `time` columns were previously expected to be ignored during replication. Updating those. Also going with `milliseconds since 00` since that matches how we handle `date`.

xfh:
Thanks for taking this on. I'm out camping for a few days, but I am happy to help with whatever is left to do.

tantaman:
ok, this should be good to go now. Waiting on tests then will merge today.

packages/zero-cache/src/types/pg.test.ts
```diff
@@ -46,6 +371,7 @@ describe('dataTypeToZqlValueType', () => {
     ['float4', 'number'],
     ['float8', 'number'],
     ['date', 'number'],
+    ['time', 'string'],
```
xfh, line 374: I think this should change to ` number`, because you changed zero's format to ` milliseconds since 00`.

packages/zero-cache/src/types/pg.ts
```diff
@@ -257,6 +362,7 @@ export const pgToZqlTypeMap = Object.freeze({
 
   // Date/Time types
   'date': 'number',
+  'time': 'string',
```
xfh, line 365: same here. This should now be type ` number`

tantaman:
Good catch. Thanks

xfh:
fixed in https://github.com/rocicorp/mono/pull/4709

aboodman:
This PR didn't work, at least with custom mutators. When a mutator runs I get:

```
Error: Schema incompatibility detected between your zero schema definition and the database:

  - Type mismatch for column "t" in table "message": time without time zone is currently unsupported in Zero. Please file a bug at https://bugs.rocicorp.dev/

Please update your schema definition to match the database or migrate your database to match the schema.
    at assert (file:///Users/aa/work/hello-zero-solid/node_modules/@rocicorp/zero/out/shared/src/asserts.js:3:15)
    at getServerSchema (file:///Users/aa/work/hello-zero-solid/node_modules/@rocicorp/zero/out/zero-server/src/schema.js:88:5)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async makeServerTransaction (file:///Users/aa/work/hello-zero-solid/node_modules/@rocicorp/zero/out/zero-server/src/custom.js:23:26)
```

This appears to be happening because this `create table` statement:

```sql
CREATE TABLE "message" (
  "id" VARCHAR PRIMARY KEY,
  "sender_id" VARCHAR REFERENCES "user"(id),
  "medium_id" VARCHAR REFERENCES "medium"(id),
  "body" VARCHAR NOT NULL,
  "timestamp" TIMESTAMP not null,
  "t" TIME not null
);
```

leads to this datatype:

```sql
user@127.0.0.1:postgres> describe message;
+-----------+-----------------------------+-----------+
| Column    | Type                        | Modifiers |
|-----------+-----------------------------+-----------|
| id        | character varying           |  not null |
| sender_id | character varying           |           |
| medium_id | character varying           |           |
| body      | character varying           |  not null |
| timestamp | timestamp without time zone |  not null |
| t         | time without time zone      |  not null |
+-----------+-----------------------------+-----------+
```

which is not listed in `pg.ts`:

```ts
export const pgToZqlTypeMap = Object.freeze({
  // Numeric types
  ...pgToZqlNumericTypeMap,

  // Date/Time types
  'date': 'number',
  'time': 'number',
  'timestamp': 'number',
  'timestamptz': 'number',
  'timestamp with time zone': 'number',
  'timestamp without time zone': 'number',
```

aboodman:
I am not sure what all tests need to be fixed so leaving this for one of you two.

aboodman:
Also, I don't know if you came out at microseconds since 00:00 or milliseconds, but can you please make sure it matches whatever we do for `datetime`. I assume that we do millis since epoch there to match JS standard.

aboodman:
Here is a PR against one of our sample apps that you can use to replicate the problem: https://github.com/rocicorp/hello-zero-solid/pull/25

xfh:
I reproduce, thanks for the setup!

Adding `time without time zone` to the pgToZqlTypeMap does help with the "unsupported error", but it is not enough. I am afraid I don't understand the  architectural design yet.

There is some conversion of ZQL types to Postgres types happening here in `sql.ts`:
https://github.com/rocicorp/mono/blob/d981ea6721a1247d1a8f73839e61bb0b33b6327a/packages/z2s/src/sql.ts#L182-L232

But there is also the serialization & parsing logic in `pg.ts`:
https://github.com/rocicorp/mono/blob/d981ea6721a1247d1a8f73839e61bb0b33b6327a/packages/zero-cache/src/types/pg.ts#L193-L247

I don't understand, why there are two different approaches used, to convert to native postgres types. The placeholder approach in `sql.ts` is assuming that the ZQL value can be easily converted to postgres, mostly through casting. The `postgresTypeConfig` in `pg.ts` makes of typescript to serialize or parse values.

I've tested a `time` column in the `hello-zero` project (without custom mutators) - there the `time` values are saved and synced as they should. With custom mutators it doesn't work, because the value that is passed to postgres is still `milliseconds since midnight`. It seems a bit fragile, to implement the conversion twice. Wouldn't it be saver convert the value that is given to the query as an argument using the serializer functions implemented in `pg.ts`? I am probably not seeing the big picture here, sorry.

@tantaman, maybe you could reconsider using a `string` representation of TIME? It would remove all the complexity of converting to and from milliseconds since midnight. Note also @aboodman's assumption that it was millis since epoch. Since a time is independent of a date, it cannot be since epoch. In my opinion millis are rather confusing for TIME values. Having a string representation instead of a numeric one could indicate that time and date values are fundamentally different.

In the meantime, I have found a way to add support for time in `sql.ts` with the current `millis since midnight` approach:
```ts   
case 'time':
case 'time without time zone': 
    return `(INTERVAL '1 millisecond' * ${valuePlaceholder})::TIME`;
```

Let me know which direction you want to go and I'll implement tests and a PR.

aboodman:
Sorry I meant that I would have assumed millis since midnight. (matt and I
spoke about the string/number thing and I agree with him -- I don't like
the idea of transmitting time as a string)

Matt is on vacation due to the US holiday and will be back next week and
I'm sure he can sort this out.

PR #4676, opened 2025-07-29
fix(zero-cache): fix upgrade path for already replicated array types

tantaman:
Prior to the v0.21 we wrote arrays into the SQLite schema as `type[]`. Post v0.21 we write them as `type|TEXT_ARRAY`.

This updates `isArray` to treat either `[]` or `TEXT_ARRAY` as an array type so we can handle arrays for users coming from either version.

OTOH, someone stuck in the bad state could have wiped their replica.

for ref, pr that added array support: https://github.com/rocicorp/mono/pull/4390


packages/zero-cache/src/types/lite.test.ts
```diff
@@ -209,7 +212,7 @@ describe('types/lite', () => {
       '[12313214123432,12313214123432]',
     ],
     ['float[]', [123.456, 987.654], '[123.456,987.654]'],
-    ['bool[]', [true, false], '[1,0]'],
+    ['bool[]', [true, false], '[true,false]'],
```
tantaman, line 215: old code was converting this to a stringified array of `1` | `0` 🤔.

packages/zero-cache/src/types/lite.ts
```diff
@@ -202,7 +202,7 @@ export function dataTypeToZqlValueType(
   return upstreamDataTypeToZqlValueType(
     upstreamDataType(liteTypeString).toLowerCase(),
     liteTypeString.includes(TEXT_ENUM_ATTRIBUTE),
```
arv, line 204: this one should have used `isEnum` to centralize the check

packages/zero-cache/src/types/lite.test.ts
arv: very odd

arv (commented):
LGTM

My goal was to not write TEXT_ARRAY any more but 🤷🏼

PR #4677, opened 2025-07-29
chore(zero): persist mutator responses into replicache

tantaman:
We need to persist mutator responses since:

1. Two clients can connect
2. Client A disconnects
3. Client B syncs Client A's mutations
4. Server pokes down results
5. Client B syncs to IBD
6. Client A refreshes from IDB
7. Client A comes online
8. Client A will not receive pokes for its mutations since they were already synced for the client group

Client A can only get its response if the mutation results are persisted to replicache.

### DIFF
--- a/packages/zero-client/src/client/mutation-tracker.ts
+++ b/packages/zero-client/src/client/mutation-tracker.ts
@@ -5,16 +5,21 @@ import type {
 } from '../../../replicache/src/replicache-options.ts';
 import {assert} from '../../../shared/src/asserts.ts';
 import {emptyObject} from '../../../shared/src/sentinels.ts';
-import type {
-  MutationError,
-  MutationID,
-  MutationOk,
-  PushError,
-  PushOk,
-  PushResponse,
+import {
+  mutationResultSchema,
+  type MutationError,
+  type MutationID,
+  type MutationOk,
+  type PushError,
+  type PushOk,
+  type PushResponse,
 } from '../../../zero-protocol/src/push.ts';
 import type {ZeroLogContext} from './zero-log-context.ts';
-import type {MutationPatch} from '../../../zero-protocol/src/mutations-patch.ts';
+import type {ReplicacheImpl} from '../../../replicache/src/impl.ts';
+import {MUTATIONS_KEY_PREFIX} from './keys.ts';
+import type {NoIndexDiff} from '../../../replicache/src/btree/node.ts';
+import {must} from '../../../shared/src/must.ts';
+import * as v from '../../../shared/src/valita.ts';
 
 type ErrorType =
   | MutationError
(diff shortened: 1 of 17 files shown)


packages/zero-client/src/client/mutation-tracker.ts
```diff
@@ -110,25 +115,40 @@ export class MutationTracker {
...
+          } else {
+            this.#processMutationOk(clientID, mutationID, result);
+          }
+          break;
         }
+        case 'del':
+          break;
+        case 'change':
+          throw new Error('MutationTracker does not expect change operations');
       }
```
grgbkr, line 144: default:
      unreachable(diff.op);

packages/zero-client/src/client/mutation-tracker.ts
```diff
@@ -110,25 +115,40 @@ export class MutationTracker {
...
+    for (const diff of diffs) {
+      const mutationID = Number(
+        diff.key.slice(MUTATIONS_KEY_PREFIX.length + clientID.length + 1),
+      );
+      assert(
+        !isNaN(mutationID),
+        `MutationTracker received a diff with an invalid mutation ID: ${diff.key}`,
+      );
+      largestLmid = Math.max(largestLmid, mutationID);
+      switch (diff.op) {
```
grgbkr, line 130: I think it would be ideal if mutations resolved in order.  Seems like a nice invariant that will make debugging issues easier. 

So we'd need to sort these by lmid, and then resolve gaps between them as we loop through them.

packages/zero-client/src/client/mutation-tracker.ts
```diff
@@ -5,30 +5,28 @@ import type {
...
   | unknown;
 
-const completeFailureTypes: PushError['error'][] = [
-  // These should never actually be received as they cause the websocket
-  // connection to be closed.
-  'unsupportedPushVersion',
-  'unsupportedSchemaVersion',
-];
-
 let currentEphemeralID = 0;
```
grgbkr, line 30: Do we still need the ephemeral id system?  If so why?

packages/zero-client/src/client/mutation-tracker.ts
```diff
@@ -64,14 +57,26 @@ export class MutationTracker {
...
   }
 
-  set clientID(clientID: string) {
+  setClientIDAndWatch(
+    clientID: string,
+    experimentalWatch: ReplicacheImpl['experimentalWatch'],
+  ) {
+    assert(this.#clientID === undefined, 'clientID already set');
     this.#clientID = clientID;
+    experimentalWatch(
```
grgbkr, line 71: If I understand this correctly its essential for correctness that for a poke this watch callback must run before zero-poke-handler calls lmidAdvanced for that poke.

I assume this order works out because ZeroPokeHandler awaits

await this.#replicachePoke(merged); 

which must not resolve until the subscribers/watchers are run

before ZeroPokeHandler calls

lmidAdvanced.

It would be good to document this ordering constraint.

With the current impl of MutationTracker I think if this ever happened out of order and a poke had responses, we would resolve the mutations for those responses with success during the lmidAdvanced call, and then drop the responses when they came in through watch.  It would be good if we could at least assert that the order is as expected.

tantaman: Yep, that's right. There are some tests that check this but adding an invariant is a good idea. We can check that the response LMID is always > than the last LMID processed by mutation tracker.

PR #4680, opened 2025-07-30
feat(zero-server): write mutation error responses into upstream

tantaman:
- if a mutation returns an error that is written into the `mutations` table on upstream
- if it succeeds, we can skip writing to mutations as success can be implicitly resolved by looking at the `clients` table's lmid

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/pipeline-driver.test.ts
+++ b/packages/zero-cache/src/services/view-syncer/pipeline-driver.test.ts
@@ -1261,7 +1261,7 @@ describe('view-syncer/pipeline-driver', () => {
         clientGroupID: 'cg1',
         clientID: 'c1',
         mutationID: 1,
-        result: 'success',
+        result: {},
       }),
     );
 
@@ -1273,7 +1273,13 @@ describe('view-syncer/pipeline-driver', () => {
       ),
     ];
 
-    // TODO: get row?
+    expect(
+      pipelines.getRow('zeroz.mutations', {
+        clientGroupID: 'cg1',
+        clientID: 'c1',
+        mutationID: 1,
+      }),
+    ).toMatchInlineSnapshot(`undefined`);
   });
 
   test('schemaVersions change and insert', () => {
--- a/packages/zero-client/src/client/mutation-tracker.ts
+++ b/packages/zero-client/src/client/mutation-tracker.ts
@@ -258,7 +258,6 @@ export class MutationTracker {
       clientID === this.#clientID,
       'received mutation for the wrong client',
     );
-    this.#lc.error?.(`Mutation ${mid} returned an error`, error);
 
     const ephemeralID = this.#ephemeralIDsByMutationID.get(mid);
     if (!ephemeralID && error.error === 'alreadyProcessed') {
--- a/packages/zero-cache/src/server/syncer.ts
+++ b/packages/zero-cache/src/server/syncer.ts
@@ -50,8 +50,8 @@ export default function runWorker(
   const fileMode = v.parse(args[0], replicaFileModeSchema);
 
   const {cvr, upstream} = config;
-  assert(cvr.maxConnsPerWorker);
-  assert(upstream.maxConnsPerWorker);
+  assert(cvr.maxConnsPerWorker, 'cvr.maxConnsPerWorker must be set');
+  assert(upstream.maxConnsPerWorker, 'upstream.maxConnsPerWorker must be set');
 
   const replicaFile = replicaFileName(config.replica.file, fileMode);
   lc.debug?.(`running view-syncer on ${replicaFile}`);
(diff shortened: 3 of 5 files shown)


packages/zero-cache/src/services/mutagen/pusher.test.ts
```diff
@@ -668,102 +668,6 @@ describe('pusher streaming', () => {
     expect(result.type).toBe('ok');
   });
 
-  test('streams successful push response to correct client', async () => {
```
tantaman, line 671: we no longer use the websocket interface to send mutation successes to the client. These will come through poke.

packages/zero-cache/src/services/mutagen/pusher.ts
```diff
@@ -324,6 +321,7 @@ class PushWorker {
         }
       }
     } else {
+      // Look for mutations results that should cause us to terminate the connection
       const groupedMutations = groupBy(response.mutations, m => m.id.clientID);
```
tantaman, line 325: we still loop through the API server's mutation responses as an `ooo mutation` should cause us to fatal the connection.

packages/zero-cache/src/services/mutagen/pusher.ts
```diff
@@ -270,18 +270,15 @@ class PushWorker {
...
+   * 1. If the entire `push` fails, we send the error to relevant clients.
+   * 2. If the push succeeds, we look for any mutation failure that should cause the connection to terminate
+   *  and terminate the connection for those clients.
    */
-  async #fanOutResponses(response: PushResponse | Fatal) {
+  #fanOutResponses(response: PushResponse | Fatal) {
     const responses: Promise<Result>[] = [];
     const connectionTerminations: (() => void)[] = [];
+
+    // if the entire push failed, send that to the client.
```
tantaman, line 281: if the entire `push` request fails, we send this information to the client. This is unused on the client currently but could be surfaced in `onError`. Seems useful to surface given the whole push could fail due to configuration problems (bad push url, lacking ssl certs, api server is down, whatever).

packages/zero-cache/src/services/view-syncer/cvr.pg-test.ts
```diff
@@ -1278,18 +1266,6 @@ describe('view-syncer/cvr', () => {
                   },
                   type: 'simple',
                 },
-                {
```
tantaman, line 1306: the `mutationResults` query was changed to grab all mutation results for the entire client group rather than hydrating and cleaning up a query per client. This is how the lmids query works as well.

packages/zero-cache/src/services/view-syncer/pipeline-driver.test.ts
```diff
@@ -1241,6 +1251,31 @@ describe('view-syncer/pipeline-driver', () => {
...
+
+    [
+      ...pipelines.addQuery(
+        mutationResultsQuery.id,
+        mutationResultsQuery.ast,
+        startTimer(),
+      ),
+    ];
+
+    // TODO: get row?
```
tantaman: oops, forgot an expect

packages/zero-client/src/client/mutation-tracker.ts
```diff
@@ -185,20 +185,6 @@ export class MutationTracker {
    * to the DB while writing the LMID.
    */
   onConnected(lastMutationID: number) {
-    for (const [id, entry] of this.#outstandingMutations) {
```
tantaman, line 188: `this.lmidAdvanced(lastMutationID)` was already doing this logic. Removed.

`onConnected` is kept for now in case the user is running a new client with an outdated zero-cache.

packages/zero-server/src/push-processor.ts
```diff
@@ -204,7 +205,7 @@ export class PushProcessor<
     params: Params,
     req: PushBody,
     m: Mutation,
-    errorMode: boolean,
+    caughtError: unknown,
   ): Promise<MutationResponse> {
```
grgbkr, line 209: Should this still return a MutationResponse given that is now communicated via upstream?

packages/zero-server/src/zql-database.ts
```diff
@@ -80,6 +80,17 @@ export class ZQLDatabase<S extends Schema, WrappedTransaction>
 
           return {lastMutationID};
         },
+
+        async writeMutationResult(result) {
+          const formatted = formatPg(
+            sql`INSERT INTO ${sql.ident(transactionInput.upstreamSchema)}.mutations
```
grgbkr, line 86: Where does this mutations table get created?

grgbkr:
LGTM

packages/zero-server/src/zql-database.ts
tantaman: https://github.com/rocicorp/mono/blob/5c91a6372a78ef70db3d913e8fa2764dd11b604b/packages/zero-cache/src/services/change-source/pg/schema/shard.ts#L129-L138

and 

https://github.com/rocicorp/mono/blob/5c91a6372a78ef70db3d913e8fa2764dd11b604b/packages/zero-cache/src/services/change-source/pg/schema/init.ts#L186-L193

packages/zero-server/src/push-processor.ts
tantaman: Yeah, we still need to check if the mutation response contains an `ooo mutation` error as that will not be written to upstream in the current design.

PR #4681, opened 2025-07-31
chore(zero): clean up `mutations` table on client and client group delete

tantaman:
It is possible that a client disappears before acking mutations so we need a way to drop them when the client or client groups are later deleted.


packages/zero-cache/src/services/view-syncer/cvr.pg-test.ts
```diff
@@ -6847,8 +6913,145 @@ describe('view-syncer/cvr', () => {
     const cvr3 = await cvrStore.load(lc, t2);
     expect(cvr3).toEqual(cvr2);
   });
+
+  test('delete a client', async () => {
```
tantaman, line 6921: these are the main tests for the new code

tantaman:
@darkgnotic - on  second thought, maybe we should send the cleanup request through the API server to deal with the whole custom backend situation.

PR #4684, opened 2025-07-31
chore(mutators): workaround for custom upstream

tantaman:
The `zero_0.mutations` table is written by the API server in the mutator's transaction, with the exception of cleanup. Cleanup of mutations that the client has acknowledged receiving is done by zero-cache.

This is a problem since if the upstream is custom we cannot delete mutation results from it.

This PR is a temporary workaround to not break users with custom upstreams.

Alternatives:
1. Send a message to the API server when a mutation is acked, client is deleted or client group is deleted. 
2. Add more to the custom upstream interface and poke that. 
3. Have the API server clean the table on some interval given zero is not expected to run for prolonged periods offline.

### DIFF
--- a/packages/zero-cache/src/server/syncer.ts
+++ b/packages/zero-cache/src/server/syncer.ts
@@ -96,7 +96,7 @@ export default function runWorker(
       config.taskID,
       id,
       cvrDB,
-      upstreamDB,
+      config.upstream.type === 'pg' ? upstreamDB : undefined,
       new PipelineDriver(
         logger,
         config.log,
--- a/packages/zero-cache/src/services/view-syncer/cvr-store.ts
+++ b/packages/zero-cache/src/services/view-syncer/cvr-store.ts
@@ -126,7 +126,7 @@ export class CVRStore {
   readonly #taskID: string;
   readonly #id: string;
   readonly #db: PostgresDB;
-  readonly #upstreamDb: PostgresDB;
+  readonly #upstreamDb: PostgresDB | undefined;
   readonly #writes: Set<{
     stats: Partial<CVRFlushStats>;
     write: (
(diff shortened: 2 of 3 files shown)


packages/zero-cache/src/server/syncer.ts
```diff
@@ -96,7 +96,7 @@ export default function runWorker(
       config.taskID,
       id,
       cvrDB,
-      upstreamDB,
+      config.upstream.type === 'pg' ? upstreamDB : undefined,
```
tantaman, line 99: I'm so confused. I would not have expected a connection to `upstreamDB` to have been established here given `upstream` could be custom.

But the main thing https://github.com/rocicorp/mono/pull/4681 did was to thread this `upstreamDB` variable down (https://github.com/rocicorp/mono/pull/4681/files#diff-b658908c87dc22639c571586705bdb443d8f25cda12ab0d011745665596d9af1).

Introducing this check fixes `integation.pg-test.ts`

```
cd packages/zero-cache
npm run test -- integration.pg-test.ts --project=zero-cache/pg-17 --silent=false
```

tantaman: awaiting @darkgnotic 's wisdom

darkgnotic: Sorry for being way out of the loop.  😓 

Overall, my opinion is that we shouldn't be doing any PG-specific upstream logic here in view-syncer land. Upstream has thus far been tightly cordoned off in the `change-source` package, with the change stream protocol being the full interface for both pg and custom upstreams. Adding logic here introduces complexity here in terms of both managing CVRs and (current and future) upstream implementations. 

I almost think it's not worth the feature? I know a lot of people request this, but I don't know if it's worth the complexity. Thinking about it more (than I did a couple of weeks ago), having an upstream table that keeps the entire history of mutation responses seems like a dangerous/hard thing to manage.

The only semi-elegant alternative I can think of is to store the _last_ mutation-server response in the existing `clients` table. Clients would get updates for each mutation if they are syncing, but they'd only get the last one on catchup. I'm betting this won't satisfy everyone's use case, but it does make the data problem a lot more manageable.

But yeah, I'm also inclined to just push back on this feature and have the users manage it.  😞

tantaman: > having an upstream table that keeps the entire history of mutation responses seems like a dangerous/hard thing to manage.

It only keeps error responses and no others.

We can also move all of this code out of the view-syncer and send acks to the user's API server.

PR #4689, opened 2025-08-01
fix(zero-cache)!: fully support env-based OTEL client configuration

darkgnotic:
User report: https://discord.com/channels/830183651022471199/1400094101399015537/1400444793133338694

Remove the hard-coded configuration of logs, traces, and metrics exporters (previously set to the http/json implementations) in order to allow the otel client to be fully configurable via environment variables. 

This correctly supports configurations such as:
* setting an exporter to `none` or `console`
* choosing other protocols such as `grpc`
* choosing different resource detectors

For backwards compatibility, the protocol is set to `http/json` if not explicitly set.

Breaking change:
* The previous code never enabled logs export. 
* With this change, logs export correctly defaults to `otlp` when the otel endpoint is enabled. 
* To achieve the previous (disabled) logs export behavior, the `OTEL_LOGS_EXPORTER` env variable should be set to `none`.

### DIFF
--- a/packages/zero-cache/src/server/otel-start.ts
+++ b/packages/zero-cache/src/server/otel-start.ts
@@ -2,14 +2,7 @@ import {diag} from '@opentelemetry/api';
 import {logs} from '@opentelemetry/api-logs';
 import * as autoInstrumentationsModule from '@opentelemetry/auto-instrumentations-node';
 import type {Instrumentation} from '@opentelemetry/instrumentation';
-import {
-  defaultResource,
-  detectResources,
-  envDetector,
-  hostDetector,
-  processDetector,
-  resourceFromAttributes,
-} from '@opentelemetry/resources';
+import {resourceFromAttributes} from '@opentelemetry/resources';
 import {NodeSDK} from '@opentelemetry/sdk-node';
 import {ATTR_SERVICE_VERSION} from '@opentelemetry/semantic-conventions';
 import {LogContext} from '@rocicorp/logger';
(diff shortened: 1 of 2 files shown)


darkgnotic:
@tantaman with the exception of enabling logs export (per the PR description), I believe this behaves equivalently to the old hard-coded values unless overridden by environment variables, as per the functions:
* getResourceDetectorsFromEnv()
* configureMetricProviderFromEnv()
* getSpanProcessorsFromEnv(),
* getLogProviderFromEnv()

in https://github.com/open-telemetry/opentelemetry-js/blob/48538b46add710483af3ff8147572d5bd4b273e0/experimental/packages/opentelemetry-sdk-node/src/sdk.ts

Let me know if I missed something!

packages/zero-cache/src/server/otel-start.ts
```diff
@@ -62,60 +46,39 @@ class OtelManager {
...
       [ATTR_SERVICE_VERSION]: process.env.ZERO_SERVER_VERSION ?? 'unknown',
     });
 
-    const resource = defaultResource().merge(envResource).merge(customResource);
-
-    // Initialize logger provider if not already set
-    if (!logs.getLoggerProvider()) {
-      const provider = new LoggerProvider({resource});
-      if (otelLogsEnabled()) {
-        const processor = new BatchLogRecordProcessor(new OTLPLogExporter());
```
tantaman, line 83: so the env var just takes care of everything? Does it default to batching records?

darkgnotic: Yes, I checked that the end result is the same as what we originally had hard-coded.

Here is the path for logs:

https://github.com/open-telemetry/opentelemetry-js/blob/48538b46add710483af3ff8147572d5bd4b273e0/experimental/packages/opentelemetry-sdk-node/src/sdk.ts#L502

traces:

https://github.com/open-telemetry/opentelemetry-js/blob/48538b46add710483af3ff8147572d5bd4b273e0/experimental/packages/opentelemetry-sdk-node/src/utils.ts#L105

and metrics:

https://github.com/open-telemetry/opentelemetry-js/blob/48538b46add710483af3ff8147572d5bd4b273e0/experimental/packages/opentelemetry-sdk-node/src/sdk.ts#L159

darkgnotic:
Looking at the implementation history of the current auto-otel code, I realize that I should be sending these to @cesara. Sorry for the churn @tantaman!

darkgnotic:
Looking at the implementation history of the current auto-otel code, I realize that I should be sending these to @cesara. Sorry for the churn @tantaman!

PR #4692, opened 2025-08-02
fix(zero-solid): Require ZeroProvider in solid, just like in React.

aboodman:
This came up because the new `useZeroOnline` didn't work if `ZeroProvider` not used. But I think we should just require it, like in React.


0xcadams (approved):
This looks good to me.

PR #4693, opened 2025-08-02
fix(zero-cache): workaround the COPY hang bug in Windows

darkgnotic:
The COPY hang bug that is fixed in Node v22+ for Linux/MacOS is still present in the latest version of Node on Windows.

User report: https://discord.com/channels/830183651022471199/1400629013105475634/1401009565931147267

Work around this for `win32` only, by running running enough copy workers (i.e. connections) for each table to have its own connection, and avoiding waiting for connections to close.


arv:
LGTM

PR #4695, opened 2025-08-02
chore(zero-cache): fix otel duplicate registration errors

darkgnotic:
The [fix to surface otel diagnostic messages](https://github.com/rocicorp/mono/pull/4670) to the Logger revealed existing errors of the form:

```
Error: @opentelemetry/api: Attempted duplicate registration of API: {trace,context,propagation}
```

Screenshot 2025-08-02 at 15 25 09

This is due to the duplicate registration of the otel libraries:
* first by`--require=@opentelemetry/auto-instrumentations-node/register`
* then explicitly in [`otel-start.ts`](https://github.com/rocicorp/mono/blob/a1a46d6c0b793ad21978cf3360d2dcea9a418bc1/packages/zero-cache/src/server/otel-start.ts#L71) 

Remove the former and rely on the latter, which affords additional options such as setting the `service.version` attribute.

### DIFF
--- a/packages/zero-cache/src/server/otel-start.ts
+++ b/packages/zero-cache/src/server/otel-start.ts
@@ -1,6 +1,6 @@
 import {diag} from '@opentelemetry/api';
 import {logs} from '@opentelemetry/api-logs';
-import * as autoInstrumentationsModule from '@opentelemetry/auto-instrumentations-node';
+import {getNodeAutoInstrumentations} from '@opentelemetry/auto-instrumentations-node';
 import type {Instrumentation} from '@opentelemetry/instrumentation';
 import {resourceFromAttributes} from '@opentelemetry/resources';
 import {NodeSDK} from '@opentelemetry/sdk-node';
@@ -56,8 +56,7 @@ class OtelManager {
     // Lazy load the auto-instrumentations module
     // avoid MODULE_NOT_FOUND errors in environments where it's not being used
     if (!this.#autoInstrumentations) {
-      this.#autoInstrumentations =
-        autoInstrumentationsModule.getNodeAutoInstrumentations();
+      this.#autoInstrumentations = getNodeAutoInstrumentations();
     }
     // Set defaults to be backwards compatible with the previously
     // hard-coded exporters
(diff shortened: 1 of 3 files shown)


darkgnotic:
@cesara let me know if I'm mistaken in that we don't need the `--require=@opentelemetry/...` flags since we're explicitly starting the NodeSdk() in the code.

packages/zero-cache/src/server/otel-start.ts
```diff
@@ -56,8 +56,7 @@ class OtelManager {
     // Lazy load the auto-instrumentations module
     // avoid MODULE_NOT_FOUND errors in environments where it's not being used
```
darkgnotic, line 57: @cesara I didn't understand this comment. When does MODULE_NOT_FOUND happen? 

I'm wondering why we can't just do:

```ts
    const sdk = new NodeSDK({
      resource,
      autoDetectResources: true,
      instrumentations: [getNodeAutoInstrumentations()],
    });
```

as per the example in https://opentelemetry.io/docs/languages/js/libraries/#registration. What would break if we did that?

cesara: there was an issue with pnpm users that this lazy loading helped avoid duplicated otel library issues. Id need to experiment a bit and maybe this avoids it all together

cesara:
> @cesara let me know if I'm mistaken in that we don't need the `--require=@opentelemetry/...` flags since we're explicitly starting the NodeSdk() in the code.

I think this is correct. I am actually wondering if we want to node instrumentation data at all ?

darkgnotic:
> > @cesara let me know if I'm mistaken in that we don't need the `--require=@opentelemetry/...` flags since we're explicitly starting the NodeSdk() in the code.
> 
> I think this is correct. I am actually wondering if we want to node instrumentation data at all ?

Good question. From our Grafana I see:
* nodejs event loop timings
* http client / server timings
* v8js gc
* v8js heap usage

Maybe we can leave the code as is and set a default `OTEL_NODE_ENABLED_INSTRUMENTATIONS` environment variable to only enable the instrumentations that are useful.

https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/auto-instrumentations-node#usage-auto-instrumentation

packages/zero-cache/src/server/otel-start.ts
darkgnotic: I'm guessing that removing the lazy loading fixes the issue, but I'll wait for you to confirm.  🙏

PR #4698, opened 2025-08-05
fix: rework offline design

0xcadams:
Reworked based on new design:

Screenshot 2025-08-05 at 1 19 31 PM

Screen Recording:

https://screen.studio/share/wvSN9PHU


aboodman:
I think we should go ahead and do 'offline mode' as part of this change. I don't think it will be much work. Do:

- Somewhere down in mutation path check whether offline (as determined by `offline` state in class).
- If offline throw an error.
- Add a field to `ReplicacheOptions` to override offline behavior (and return to current behavior). Perhaps `allowOfflineWrites: boolean`, defaulting to `false`.

aboodman:
This would close this bug: https://bugs.rocicorp.dev/issue/4019

aboodman:
Despite me saying there won't be a flag here: https://bugs.rocicorp.dev/issue/4019#comment-2FKglJ651B8KStUlv9EHV, writing this suggestion out made me realize there will be a lot of bummed people if we turn this off suddenly and it's really cheap to maintain.

On next release we should have a big red warning box telling people the current downsides of using it.

0xcadams:
Closing and will reopen a future PR

PR #4699, opened 2025-08-06
feat(zero): `mutators` api to match `queries` api [1/n]

tantaman:
custom queries are defined like:

```ts
const q = queries({ name: fn, name: fn, ... });
```

add function to allow mutators to be defined the same way:

```ts
const m = mutators({ name: fn, name: fn, ... });
```

This PR defines the base `mutators` function.

Next PR:

- Allow using mutators defined in this way on the client

### DIFF
--- a/packages/zql/src/mutate/custom.ts
+++ b/packages/zql/src/mutate/custom.ts
@@ -1,5 +1,6 @@
 import {assert} from '../../../shared/src/asserts.ts';
 import type {Expand} from '../../../shared/src/expand.ts';
+import type {ReadonlyJSONValue} from '../../../shared/src/json.ts';
 import type {Schema} from '../../../zero-schema/src/builder/schema-builder.ts';
 import type {
   SchemaValueToTSType,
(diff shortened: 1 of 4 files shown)


packages/zql/src/mutate/custom.ts
```diff
@@ -148,3 +149,42 @@ export function customMutatorKey(namespace: string, name: string) {
...
+  string,
+  (...args: ReadonlyArray<ReadonlyJSONValue>) => Promise<void>
+>;
+
+/**
+ * Define mutators. Unlike queries, you must currently
+ * define _all_ mutators at once.
+ *
+ * The reason is that all mutators are currently
+ * passed to zero-client on construction.
```
tantaman, line 173: I think I can resolve this in the next pr

tantaman:
maybe going a different direction after considering the fact that not having mutators explicitly attached to zero introduces more chances for them to diverge.

E.g., someone may `createMutators` once to attach to zero then `createMutators` again (capturing a new environment) when calling a mutator.

https://rocicorp.slack.com/archives/C013XFG80JC/p1754575112973199

PR #4700, opened 2025-08-06
chore(mutators): wire new mutator API into zero-client [2/n]

tantaman:
Custom queries can be called like:

```ts
const issueList = issues(zero, args);
```

custom mutators can be called the same way with this change.

```ts
createIssue(zero, args);
```

Annoyance:
Custom mutators (like queries) need unique names. Given that, you can't have mutators defined like: 

```ts
const issueMutators = mutators({
  create() {},
  update() {},
});

const commentMutators = mutators({
  create() {},
  update() {},
});
```

since the names collide.

We detect this when mutators are added to the zero instance at runtime.


tantaman:
see https://github.com/rocicorp/mono/pull/4699#issuecomment-3164202219

PR #4701, opened 2025-08-06
docs: deprecation for write permissions

0xcadams:
Opening this for discussion.

### DIFF
--- a/packages/zero-schema/src/permissions.ts
+++ b/packages/zero-schema/src/permissions.ts
@@ -19,6 +19,10 @@ import {clientToServer, NameMapper} from './name-mapper.ts';
 export const ANYONE_CAN = [
   (_: unknown, eb: ExpressionBuilder<Schema, never>) => eb.and(),
 ];
+
+/**
+ * @deprecated Use ANYONE_CAN instead.
+ */
 export const ANYONE_CAN_DO_ANYTHING = {
   row: {
     select: ANYONE_CAN,
@@ -30,6 +34,7 @@ export const ANYONE_CAN_DO_ANYTHING = {
     delete: ANYONE_CAN,
   },
 };
+
 export const NOBODY_CAN = [];
 
 export type Anchor = 'authData' | 'preMutationRow';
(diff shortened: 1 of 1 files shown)


packages/zero-schema/src/permissions.ts
```diff
@@ -54,13 +59,22 @@ export type AssetPermissions<
 > = {
   // Why an array of rules?: https://github.com/rocicorp/mono/pull/3184/files#r1869680716
   select?: PermissionRule<TAuthDataShape, TSchema, TTable>[] | undefined;
+  /**
+   * @deprecated Use Mutators instead.
```
tantaman, line 63: should we provide a link to the docs?

0xcadams: Just added this, based on https://github.com/rocicorp/zero-docs/pull/158

packages/zero-schema/src/permissions.ts
```diff
@@ -19,6 +19,10 @@ import {clientToServer, NameMapper} from './name-mapper.ts';
 export const ANYONE_CAN = [
   (_: unknown, eb: ExpressionBuilder<Schema, never>) => eb.and(),
 ];
+
+/**
+ * @deprecated Use {@link ANYONE_CAN} instead.
```
tantaman, line 24: I'm confused by this one. Where is `ANYONE_CAN` supposed to be used?

0xcadams: I guess this is more of a timing question - are CRUD mutators and reads getting deprecated at the same time? If read permissions are still around for a bit, then this could be used there. The current docs change to deprecate CRUD mutators keeps read permissions around for now. cc @aboodman

aboodman: I definitely wanted to remove CRUD mutators from docs because I want to stop recommending them to new users.

I hadn't thought too hard about marking the API `@deprecated`.

I guess we do really want people to move to CM though, so let's do it. @tantaman I believe you have argued for deprecating/removing CRUD mutators sooner. If we are deprecating CM then ANYONE_CAN_DO_ANYTHING should be deprecated too since the purpose of it is to apply ANYONE_CAN to multiple permissions. But we only have one permission after deprecating CM.

aboodman:
LGTM

PR #4703, opened 2025-08-07
feat(zero): drop `schema` from `tx` type

tantaman:
resurrects https://github.com/rocicorp/mono/pull/4449 to solve type depth issues.

users must now explicitly type their `Transaction` argument in their mutator definitions.

### DIFF
--- a/packages/zero-client/src/client/custom.ts
+++ b/packages/zero-client/src/client/custom.ts
@@ -32,15 +32,14 @@ import type {ZeroLogContext} from './zero-log-context.ts';
 /**
  * The shape which a user's custom mutator definitions must conform to.
  */
-export type CustomMutatorDefs<
-  S extends Schema,
-  TWrappedTransaction = unknown,
-> = {
+export type CustomMutatorDefs = {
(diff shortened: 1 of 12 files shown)


packages/zero-client/src/client/custom.ts
```diff
@@ -32,15 +32,14 @@ import type {ZeroLogContext} from './zero-log-context.ts';
...
   [namespaceOrKey: string]:
     | {
-        [key: string]: CustomMutatorImpl<S, TWrappedTransaction>;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        [key: string]: CustomMutatorImpl<any>;
       }
-    | CustomMutatorImpl<S, TWrappedTransaction>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    | CustomMutatorImpl<any>;
 };
```
arv, line 43: I'm still not sure why these needs to be `any`? Can they not be `Schema` instead?

arv (commented):
LGTM

PR #4706, opened 2025-08-07
feat(zero): symmetric ⌯ server side APIs for mutators and queries

tantaman:
For processing mutation and query requests on the server, the user will invoke:

1. mapQueryRequest
2. mapMutationRequest

Each of these functions takes in a request object and a callback.

The callback is invoked for each query or mutation.

The user then transforms the query or runs the mutator.

The fact that we can namespace mutations means that we need to provide a `dispatchMutation` function since a mutator is not as trivial to look up as a query.

```ts
// mutate
reply.send(await mapMutationRequest(
  async transact => {
    // pre transaction code can go here
    
    await transact(db, (tx, name, args) =>
      // mutator is actually run here
      dispatchMutation(mutators, tx, name, args),
    );
    
    // post transaction code can go here
  },
  request.query,
  request.body,
  'info',
));

// query
reply.send(
  await mapQueryRequest(
    async (name, args) => ({query: getQuery(authData, name, args)}),
    schema,
    request.body,
  ),
);
```


grgbkr:
> The fact that we can namespace mutations means that we need to provide a dispatchMutation function since a mutator is not as trivial to look up as a query.

What does it mean to "namespace" mutations?  How does this make mutator lookup less trivial?

tantaman:
> What does it mean to "namespace" mutations? How does this make mutator lookup less trivial?

You can define mutators like:

```ts
new Zero({
  mutators: {
    issue: {
      create(tx, args) { ... }
    },
  }
});
```

where `create` is namespaced to `issue`

We encode the name when sending it to the server as: `issue|create`.

The user would need to split and walk down the mutator map to find the right implementation on the server. Not that hard but not as easy as custom queries which are not namespaced.

custom queries don't need namespace support since the user can do:

```ts
const issueQueres = queries({ ... });
const userQueries = queries({ ... });
```

apps/zbugs/api/index.ts
```diff
@@ -11,12 +11,17 @@ import {jwtVerify, SignJWT, type JWK} from 'jose';
...
 import postgres from 'postgres';
 import {must} from '../../../packages/shared/src/must.ts';
-import {handlePush} from '../server/push-handler.ts';
 import {jwtDataSchema, type JWTData} from '../shared/auth.ts';
 import {getQuery} from '../server/get-query.ts';
-import {getQueries} from '@rocicorp/zero/server';
+import {
+  mapQueryRequest,
+  mapMutationRequest,
+  getMutation,
```
tantaman, line 19: eh, should be `getMutator`

PR #4711, opened 2025-08-11
chore(zero-cache): Use vi.waitFor

arv:
Which does pretty much the same thing as the old code.


darkgnotic (commented):
This is really cool. Thank you for the tip!

PR #4713, opened 2025-08-11
chore: Bencher runner should use Node 22 too

arv:


arv:
I ended up ssh into ec2 to make sure apt-get update worked.

PR #4715, opened 2025-08-12
chore(zbugs): fix format, lint and check-types

arv:
9cb5af28eb78073b8a5421744f54f4dd129d1650 introduced a bunch of errors.


arv:
Landing this even though the change-streamer is still failing. At least it fixes, format, lint and ts

PR #4718, opened 2025-08-12
fix(zero-cache): destroy cgStorage when cleaning up the write authorizer

tantaman:


aboodman:
What happens in the case zero-cache crashes and this cleanup code doesn't happen?

I fully support cleaning up the space used, but it seems like name used should be random so that there cannot be cross-run collisions.

PR #4719, opened 2025-08-12
fix(statz): remove verbose stats

tantaman:
not that useful and can easily blow up the endpoint.


arv:
How does this blow up the endpoint? Too many queries?

We could stream the result but the question is how useful these are.

tantaman:
> How does this blow up the endpoint? Too many queries?

Yeah, too many queries. The AST can be sizable as well.

PR #4721, opened 2025-08-12
chore(zero-cache): Fix flake test

arv:
It seems to work locally. I ran the test 10 times without failure with the fix.
It fails after about 5 tries without the fix.


arv:
Argh... It failed after 13 attempts.

PR #4727, opened 2025-08-12
feat: Add support for ignoring tables during Zero Cache replication

Karavil:
## Summary

Adds support for ignoring specific tables from PostgreSQL publications during replication. Tables in `IGNORED_PUBLICATION_TABLES` get created in SQLite but stay empty - all changes are dropped. Useful for excluding audit logs or other high-volume tables.

Stored the ignored tables in the database with publications in `InternalShardConfig`. They define the replication boundary together - publications control what Postgres sends, ignored tables control what Zero Cache accepts.

**Important**: Changing ignored tables triggers a full resync (like changing publications). This prevents stale data from remaining in SQLite when a table becomes ignored.

## Design Decision

Ignored tables are like publications - both define what data replicates. Publications say what Postgres sends, ignored tables say what Zero Cache keeps. If nodes disagree on either during deployment, you get inconsistent data (some nodes have audit_logs, others don't).

Wrestled with where to store the config:

```
Option 1: In-memory from env vars (problematic)
================================================

PostgreSQL publishes: [users, logs, temp]
                           ↓

Rolling deployment:
┌─────────────┐    ┌─────────────┐    ┌─────────────┐
│   Node A    │    │   Node B    │    │   Node C    │
│   (v1.0)    │    │   (v1.0)    │    │   (v1.1)    │ ← new version\!
│             │    │             │    │             │
│ ENV: logs   │    │ ENV: logs   │    │ ENV: logs,  │
│              │    │              │    │      temp   │
└─────────────┘    └─────────────┘    └─────────────┘
      ↓                   ↓                   ↓
Replicates:          Replicates:          Replicates:
[users, temp]        [users, temp]        [users]      ← INCONSISTENT\!


Option 2: Database storage (what I did)
========================================

PostgreSQL publishes: [users, logs, temp]
                           ↓
                    ┌──────────────┐
                    │  shardConfig │
                    │  ignored:     │
                    │  [logs]       │
                    └──────────────┘
                           ↓

Rolling deployment (all nodes read same config):
┌─────────────┐    ┌─────────────┐    ┌─────────────┐
│   Node A    │    │   Node B    │    │   Node C    │
│   (v1.0)    │    │   (v1.0)    │    │   (v1.1)    │
└─────────────┘    └─────────────┘    └─────────────┘
      ↓                   ↓                   ↓
Replicates:          Replicates:          Replicates:
[users, temp]        [users, temp]        [users, temp] ← CONSISTENT\!
```

The database approach ensures all nodes agree on what to replicate, even mid-deployment. This matters because ignored tables (like publications) are replication config, not app config. App config affects how a node runs (ports, log levels) - fine to vary. Replication config affects what data exists - must be consistent or you get different data on different nodes.

First PR here, might be missing context. Happy to refactor if storing in database seems overkill - just felt like ignored tables belong with publications since they both define the replication boundary.

## Testing

```bash
export ZERO_UPSTREAM_DB="postgresql://user:pass@localhost:5432/mydb"
export ZERO_IGNORED_PUBLICATION_TABLES='["public.audit_logs"]'
```

Check:
- Logs show "Skipping initial sync for ignored table"
- Tables exist but empty
- Changes don't replicate to ignored tables
- Changing ignored tables triggers resync: "Dropping shard to change ignored tables"

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/change-source.pg-test.ts
+++ b/packages/zero-cache/src/services/change-source/pg/change-source.pg-test.ts
@@ -16,6 +16,7 @@ import {
 import {DbFile} from '../../../test/lite.ts';
 import {versionFromLexi, versionToLexi} from '../../../types/lexi-version.ts';
 import {type PostgresDB} from '../../../types/pg.ts';
+import type {ShardConfig} from '../../../types/shards.ts';
 import type {Source} from '../../../types/streams.ts';
 import type {
   ChangeSource,
@@ -828,7 +829,7 @@ describe('change-source/pg', {timeout: 30000, retry: 3}, () => {
         appID: APP_ID,
         publications: ['zero_foo', 'zero_zero'],
         shardNum: SHARD_NUM,
-        ignoredTables: [] as string[],
+        ignoredTables: [],
       },
       replicaFile2.path,
       {tableCopyWorkers: 5},
@@ -873,7 +874,7 @@ describe('change-source/pg', {timeout: 30000, retry: 3}, () => {
         appID: APP_ID,
         publications: ['zero_foo', 'zero_zero'],
         shardNum: SHARD_NUM,
-        ignoredTables: [] as string[],
+        ignoredTables: [],
       },
       replicaFile3.path,
       {tableCopyWorkers: 5},
@@ -948,7 +949,7 @@ describe('change-source/pg', {timeout: 30000, retry: 3}, () => {
           appID: APP_ID,
           shardNum: SHARD_NUM,
           publications: ['zero_different_publication'],
-          ignoredTables: [] as string[],
+          ignoredTables: [],
         },
         replicaDbFile.path,
         {tableCopyWorkers: 5},
--- a/packages/zero-cache/src/services/change-source/pg/initial-sync.pg-test.ts
+++ b/packages/zero-cache/src/services/change-source/pg/initial-sync.pg-test.ts
@@ -16,6 +16,7 @@ import {
   initDB as initLiteDB,
 } from '../../../test/lite.ts';
 import type {PostgresDB} from '../../../types/pg.ts';
+import type {ShardConfig} from '../../../types/shards.ts';
 import {ZERO_VERSION_COLUMN_NAME} from '../../replicator/schema/replication-state.ts';
 import {initialSync, INSERT_BATCH_SIZE} from './initial-sync.ts';
 import {fromLexiVersion} from './lsn.ts';
@@ -2164,7 +2165,7 @@ describe('change-source/pg/initial-sync', {timeout: 10000}, () => {
             appID: APP_ID,
             shardNum: SHARD_NUM,
             publications: c.requestedPublications ?? [],
-            ignoredTables: [] as string[],
+            ignoredTables: [],
           },
           replica,
           getConnectionURI(upstream),
@@ -2255,7 +2256,7 @@ describe('change-source/pg/initial-sync', {timeout: 10000}, () => {
   test('resume initial sync with invalid table', async () => {
     const lc = createSilentLogContext();
     const replica = new Database(lc, ':memory:');
-    const shardConfig = {appID: APP_ID, shardNum: SHARD_NUM, publications: [] as string[], ignoredTables: [] as string[]};
+    const shardConfig: ShardConfig = {appID: APP_ID, shardNum: SHARD_NUM, publications: [], ignoredTables: []};
 
     await ensureShardSchema(lc, upstream, shardConfig);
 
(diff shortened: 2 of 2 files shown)


packages/zero-cache/src/services/change-source/pg/change-source.pg-test.ts
```diff
@@ -35,7 +35,7 @@ import {dropEventTriggerStatements} from './schema/ddl.ts';
 const APP_ID = '23';
 const SHARD_NUM = 1;
 
-describe.skip('change-source/pg', {timeout: 30000, retry: 3}, () => {
+describe('change-source/pg', {timeout: 30000, retry: 3}, () => {
```
Karavil, line 39: ah, whoops? not sure why this was being skipped... I can undo it? but it seemed useful to run these tests.

aboodman:
Curious about this design decision:

> Tables in IGNORED_PUBLICATION_TABLES get created in SQLite but stay empty - all changes are dropped.

Why create the table in SQLite?

aboodman:
Thank you very much for the contribution. Exciting!

@darkgnotic is best person to review this but he is on vacation right now. That is why this hasn't been reviewed so far.

From a product pov can you explain a bit more why the publication approach is hard to use? I believe you that it is, I'm just curious...

Is it because you don't want to have to remember to update the PG publication when you add a table? Why is it not possible to maintain the publication as part of the rest of your schema management?

Karavil:
> Curious about this design decision:
> 
> > Tables in IGNORED_PUBLICATION_TABLES get created in SQLite but stay empty - all changes are dropped.
> 
> Why create the table in SQLite?

I was mostly worried about:

1. Client schema defined, table A exists
2. Zero deploy, table A added to ignore list, table no longer exists in sqlite
3. Client breaks?

Not super committed to this behavior though—can make changes!

@aboodman

Karavil:
> Thank you very much for the contribution. Exciting!
> 
> @darkgnotic is best person to review this but he is on vacation right now. That is why this hasn't been reviewed so far.
> 
> From a product pov can you explain a bit more why the publication approach is hard to use? I believe you that it is, I'm just curious...
> 
> Is it because you don't want to have to remember to update the PG publication when you add a table? Why is it not possible to maintain the publication as part of the rest of your schema management?

Honestly, it's just more to think about! We _have_ to use custom publications right now because of the sheer amount of data we have on file (initial syncs take _ages_ if we don't do this). And I like that Zero is simple! I'd rather not create and maintain a publication for a few tables I want to ignore.

I _do_ agree that a better abstraction around generating publications could work for this (we use Drizzle Zero; it would plug in nicely there). But it's just another point of friction for experimenting with Zero. Now I have to migrate my database before I can deploy Zero!

@aboodman

Karavil:
> > Thank you very much for the contribution. Exciting!
> > @darkgnotic is best person to review this but he is on vacation right now. That is why this hasn't been reviewed so far.
> > From a product pov can you explain a bit more why the publication approach is hard to use? I believe you that it is, I'm just curious...
> > Is it because you don't want to have to remember to update the PG publication when you add a table? Why is it not possible to maintain the publication as part of the rest of your schema management?
> 
> Honestly, it's just more to think about! We _have_ to use custom publications right now because of the sheer amount of data we have on file (initial syncs take _ages_ if we don't do this). And I like that Zero is simple! I'd rather not create and maintain a publication for a few tables I want to ignore.
> 
> I _do_ agree that a better abstraction around generating publications could work for this (we use Drizzle Zero; it would plug in nicely there). But it's just another point of friction for experimenting with Zero. Now I have to migrate my database before I can deploy Zero!
> 
> @aboodman

I could also see this playing nicely with your cloud offering in the future? It'd be pretty much plug and play:

-> Create a publication for all tables
-> Select tables to ignore (don't worry, you can edit this later!)
-> Zero instance ready

Karavil:
The naming here is also a bit weird. Should it be `ZERO_APP_IGNORED_TABLES` instead of `ZERO_APP_PUBLICATION_IGNORED_TABLES`? Naming it that way (without the Postgres context) would imply that this has to be supported for _all_ databases in the future, so I was wary of it. Happy to change it though.

darkgnotic:
Hi @Karavil. Thank you for a great proposal and well thought out implementation. I appreciate (and agree with) the design decisions you detailed in the PR description. 

I also agree that this would be a useful feature to provide in a cloud offering, and went through the exercise of what that might look like at a high level, were we to implement it.

At the end of the day, the problem boils down to a deficiency in the Postgres API (e.g. for `CREATE PUBLICATION`). However, Postgres does provide a way to achieve this: you can create an [`EVENT TRIGGER`](https://www.postgresql.org/docs/current/sql-createeventtrigger.html) on the `CREATE TABLE` event, and add the table to your publication (when desired) in the triggered function.

The advantage of implementing it at the Postgres layer is that:
* There would be less code and logic to maintain. It would a matter of adding a table to a publication, rather than intercepting multiple replication points / commands.
* It would be more efficient, by avoiding the bandwidth and serialization cost of the data of ignored tables (which, as you point out, can be large).

Would you be up for trying this approach? And if you're willing, we'd love to figure out how to make it available for other users, whether it be through documented setup examples, a cli, or something that zero-cache does under the covers during the initial setup.

PR #4730, opened 2025-08-13
chore(zero): Switch perf to percentage

arv:
It was previously using a t-test, which lead to delayed alerts which was very misleading.


tantaman:
is this because a t-test uses the standard deviation to determine statistical difference?

arv:
t-test also lags behind a lot which has been very confusing in the past.

[[](url)](https://bencher.dev/perf/zero?branches=8e414274-cbba-40eb-b007-d206483a2c51&heads=6dd337c7-8c67-4900-8c40-79193b37edfe&testbeds=7274290c-8365-4158-ba73-19904b7ef4ae&benchmarks=ed1c4f7e-ef3f-4184-9f6d-61c76f3bd686&measures=fc802276-c2ed-462d-9fa5-0c1fcf2deb8e&start_time=1752415139000&end_time=1755007139000&report=b876016c-fcb0-4f84-a0a8-79a60412394d&lower_boundary=true)

In this case the perf regression was me switching the runner to Node 22 but the alert came like 5 commits later.

But since we always flag perf it gets ignored.

The current situation is not ok

arv:
I'm keeping t-test but changing the max-sample-size.

PR #4731, opened 2025-08-13
chore: Try to use codspeed instead of bencher.dev

arv:


tantaman:
What's the advantage of codspeed?

arv:
Just wanted to see if something else was easier to use. Bencher.dev has not been easy.

arv:
Another thing that made me want to try codspeed is that instrumented code which should give much less flaky results.

PR #4733, opened 2025-08-13
feat(zql-viz): debug app for ZQL

tantaman:
Thinking through how to add `merge` is easier with tooling.

CleanShot 2025-08-13 at 14 44 00


apps/zql-viz/reports/api/.gitkeep
arv, line 1: what is this?

apps/zql-viz/eslint.config.js
arv, line 1: use shared eslint config to reduce code duplication

apps/zql-viz/tsconfig.node.json
arv, line 1: can we use the shared one?

arv (commented):
Can you add a README?

PR #4736, opened 2025-08-14
chore(zero-cache): introduce ZeroEvents for surfacing real-time events for ops/observability

darkgnotic:
Introduce ZeroEvents, a channel for surfacing realtime "events" for the purposes of production ops and observability. 

As a monitoring signal, events differ from (1) metrics, which are periodic and limited to numeric values, and (2) logs, which are textual and generally have a multi-minute processing delay. Events, on the other hand, are designed for low latency delivery / processing with structured data payloads.

ZeroEvents are published using the [Cloud Events specification](https://github.com/cloudevents/spec), with configuration designed to facilitate integration with [knative sink bindings](https://github.com/knative/eventing/blob/main/docs/spec/sources.md#sinkbinding). However, any CloudEvent routing framework can be used to route and handle the Cloud Events.

The first specific ZeroEvent type is used for publishing the status of replication. ReplicationStatusEvents are emitted:
* periodically during initial sync
* when replication begins (i.e. indicating that initial sync has finished)
* whenever a schema change is processed

The events themselves contain the replicated schema data, which facilitates verification by the operator that the expected tables and columns are being replicated.


packages/zero-events/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "zero-events",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "module": "true",
```
arv, line 6: Not sure what this is?

packages/zero-events/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "format": "prettier --write .",
+    "check-format": "prettier --check .",
+    "lint": "eslint --ext .ts,.tsx,.js,.jsx src/"
+  },
+  "dependencies": {
+    "shared": "0.0.0"
+  },
+  "devDependencies": {
+    "@rocicorp/eslint-config": "^0.7.0",
+    "@rocicorp/prettier-config": "^0.3.0",
```
arv, line 21: vitest?

It does look like this package has no test so you should remove the test scripts

packages/zero-events/src/status.ts
```diff
@@ -0,0 +1,104 @@
...
+export type ReplicationStage = v.Infer<typeof replicationStageSchema>;
+
+/**
+ * A ReplicationStatusEvent is a StatusEvent event subtype for the
+ * "replication" component.
+ */
+export const replicationStatusEventSchema = statusEventSchema.extend({
+  type: v
+    .literal('zero/events/status/replication/v1')
+    .assert(v => v.startsWith(STATUS_EVENT_PREFIX)),
```
arv, line 92: I'm not sure what this assert does here. Seems like it is more of a "safety check". You can probably achieve the same thing using typescript `satisfies`

packages/zero-cache/src/observability/events.ts
```diff
@@ -0,0 +1,123 @@
...
+    sink.push({...event, time: now.toISOString()});
+    return promiseVoid;
+  };
+}
+
+export function publishEvent<E extends ZeroEvent>(lc: LogContext, event: E) {
+  void publishFn(lc, event);
+}
+
+export async function publishCriticalEvent<E extends ZeroEvent>(
```
arv, line 99: is this just the same as publishEvent? What is the usecase?

packages/zero-cache/src/observability/events.ts
```diff
@@ -0,0 +1,123 @@
...
+  // Include any enumerable properties (e.g. of Error subtypes).
+  for (const [field, value] of Object.entries(err)) {
+    if (isJSONValue(value, pathUnused)) {
+      errorDetails[field] = value;
+    }
+  }
+  return errorDetails;
+}
+
+const pathUnused = {push: () => {}, pop: () => {}};
```
arv, line 123: Path unused can also be an array.

packages/zero-cache/src/services/change-source/pg/initial-sync.pg-test.ts
```diff
@@ -2247,6 +2252,29 @@ describe('change-source/pg/initial-sync', {timeout: 10000}, () => {
...
+            component: 'replication',
+            stage: 'Initializing',
+            status: 'OK',
+          },
+          {
+            type: 'zero/events/status/replication/v1',
+            component: 'replication',
+            stage: 'Initializing',
+            status: 'OK',
+            description: /Copying \d+ upstream tables at version \w+/,
```
arv, line 2268: TIL, you can use RegExps in these

packages/zero-cache/src/services/change-source/pg/initial-sync.pg-test.ts
```diff
@@ -2247,6 +2252,29 @@ describe('change-source/pg/initial-sync', {timeout: 10000}, () => {
           slotName: r.slot,
           lsn: fromLexiVersion(replicaState.stateVersion),
         });
+
+        expect(eventSink.slice(0, 2)).toMatchObject([
+          {
+            type: 'zero/events/status/replication/v1',
+            component: 'replication',
+            stage: 'Initializing',
+            status: 'OK',
```
arv, line 2261: Why doesn't this have the description?

packages/zero-cache/src/services/change-source/pg/initial-sync.ts
```diff
@@ -188,6 +199,12 @@ export async function initialSync(
         {rows: 0, flushTime: 0},
       );
 
+      statusPublisher.publish(
+        lc,
+        'Indexing',
+        `Creating ${indexes.length} indexes`,
+        5000,
```
arv, line 206: What is 5000? and how is it exposed/tested?

packages/zero-cache/src/services/replicator/replication-status.ts
```diff
@@ -0,0 +1,151 @@
...
+import type {Database} from '../../../../zqlite/src/db.ts';
+import {computeZqlSpecs, listIndexes} from '../../db/lite-tables.ts';
+import type {LiteTableSpec} from '../../db/specs.ts';
+import {
+  makeErrorDetails,
+  publishCriticalEvent,
+  publishEvent,
+} from '../../observability/events.ts';
+
+const byKeys = ([a]: [string, unknown], [b]: [string, unknown]) =>
```
arv, line 19: I'm not sure if this is hot but for hot code do not destructure arrays

packages/zero-cache/src/services/replicator/replication-status.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  publish(
+    lc: LogContext,
+    stage: ReplicationStage,
+    description?: string,
+    interval = 0,
+  ): this {
+    this.stop();
+    publishEvent(
```
arv, line 37: would it make more sense to make publishEvent a property that is passed in instead of overriding it for tests?

packages/zero-cache/src/observability/events.ts
darkgnotic: Most events are published fire-and-forget in the steady state.

But for errors, particularly when the server is going to be shut down, I expose this method that is intended to be `await`ed to make sure that the error event gets published before the server shuts down. This is to guarantee that the error makes it out of the server before it dies.

packages/zero-cache/src/observability/events.ts
darkgnotic: Hmmm ... can you clarify? I'm looking at this method: https://github.com/rocicorp/mono/blob/842c807fa1db90326bac5c56171eeb0e9a96d7f8/packages/shared/src/json.ts#L184

packages/zero-cache/src/services/change-source/pg/initial-sync.pg-test.ts
darkgnotic: This is the first status we publish when we start initial sync, before we've touched upstream at all. I couldn't think of any useful information to embellish here ... seems like "Initializing" is enough? It will be followed up with the "Copying ... " event pretty quickly.  🤷

packages/zero-cache/src/services/change-source/pg/initial-sync.ts
darkgnotic: It's the interval between continuous updates. So during initial sync, events will be published every 5 seconds to surface the latest state of initial sync. In the PG case, it's mainly the replica size that will change.

It isn't tested.  🤫

packages/zero-cache/src/services/replicator/replication-status.ts
darkgnotic: Not super hot ... events should be infrequent. But I'll remember this. Thank you!  🙏

packages/zero-cache/src/services/replicator/replication-status.ts
darkgnotic: that would be cleaner ... bit it would be a lot of plumbing to get to this object from the top level of the test. i chose the lazy / less invasive route. i think it's "okay" because tests are run serially per process ... but please correct me if i'm mistaken.

packages/zero-events/package.json
darkgnotic: sorry for the lack of context. i which we could have comments in package.json

eventually i intend to publish this package alone as something like `rocicorp/zero-events`, so that production / operational code (i.e. cloudzero) can import just the ZeroEvent interfaces and not full zero package.

arv: but `"module": "true",` is not part of package.json spec as far as I can tell.

packages/zero-cache/src/observability/events.ts
arv: The usage of isJSONValue usually takes an array. Array implements Path :-)

darkgnotic: Aah, I see. Roger.

packages/zero-events/package.json
darkgnotic: Oh, then it looks like I just cargo-culted it from another (mistaken) package.json  😅

packages/zero-events/src/status.ts
darkgnotic: Okay, I got something to work using template literals, so we can avoid the somewhat cryptic use of assert.

PR #4741, opened 2025-08-15
feat(zero-cache): garbage collection of inactive CVRs

darkgnotic:
### Feature

Adds a periodic process that incrementally purges cvrs that have not been accessed for more than `ZERO_CVR_GARBAGE_COLLECTION_INACTIVITY_THRESHOLD_HOURS`, which defaults to 48 hours.

### Mechanism

The `reaper` process purges up to 1000 cvrs per iteration, increasing the number if necessary to reach a steady state. Iterations happen every minute while there are remaining cvrs to purge, with the interval increasing exponentially up to 16 minutes when there are no cvrs to purge.

### Schema Migration

A backwards-compatible schema migration on the CVR db:
* creates an index over the existing `lastActive` column on the `instances` table
* configures all foreign key references to cascade deletes to simplify the purge implementation

### DIFF
--- a/packages/zero-cache/src/server/main.ts
+++ b/packages/zero-cache/src/server/main.ts
@@ -13,15 +13,12 @@ import {
   restoreReplica,
   startReplicaBackupProcess,
 } from '../services/litestream/commands.ts';
-import {initViewSyncerSchema} from '../services/view-syncer/schema/init.ts';
-import {pgClient} from '../types/pg.ts';
 import {
   childWorker,
   parentWorker,
   singleProcessMode,
   type Worker,
 } from '../types/processes.ts';
-import {getShardID} from '../types/shards.ts';
 import {
   createNotifierFrom,
   handleSubscriptionsFrom,
@@ -82,7 +79,6 @@ export default async function runWorker(
     return processes.addWorker(worker, type, name);
   }
 
-  const shard = getShardID(config);
   const {
     taskID,
     changeStreamer: {mode: changeStreamerMode, uri: changeStreamerURI},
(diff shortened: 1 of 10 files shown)


packages/zero-cache/src/server/main.ts
```diff
@@ -110,24 +106,25 @@ export default async function runWorker(
...
-    await initViewSyncerSchema(lc, cvrDB, shard);
-    void cvrDB.end();
+  const {promise: reaperReady, resolve: reaperStarted} = resolver();
+  if (numSyncers > 0) {
+    loadWorker('./server/reaper.ts', 'supporting').once(
+      'message',
+      reaperStarted,
+    );
+  } else {
+    reaperStarted();
```
arv, line 127: In this case there is no reaping?

packages/zero-cache/src/services/view-syncer/schema/init.ts
```diff
@@ -133,6 +133,31 @@ export async function initViewSyncerSchema(
...
+      await sql`
+        CREATE INDEX instances_last_active ON ${sql(schema)}.instances ("lastActive");
+      `;
+
+      // Update / add foreign key constraints to cascade deletes.
+      for (const [table, reference] of [
+        ['clients', 'instances'],
+        ['queries', 'instances'],
+        ['rows', 'rowsVersion'],
+      ] as [string, string][]) {
```
arv, line 147: `as const` is simpler but 🤷🏼

packages/zero-cache/src/server/main.ts
darkgnotic: Yeah, this is the `replication-manager` case. I think it's sufficient to run reapers on `view-syncers`. I prefer to not have any cvr logic (or cvr db connections) in the `replication-manager`.

packages/zero-cache/src/services/view-syncer/schema/init.ts
darkgnotic: Ah, thank you for the tip! I'll remember this next time.  🙏

packages/zero-cache/src/server/main.ts
arv: Maybe a comment would be useful.

PR #4743, opened 2025-08-15
chore(zero-cache): Disable more change source tests

arv:


arv:
Sorry @darkgnotic. I hope you can look at this as soon as you are back before we introduce regressions.

arv:
I'm going to try something else.

darkgnotic:
Is this to ameliorate the "invalid snapshot" issues?

arv:
Yes.

I was hoping vitest allowed sequential runs of projects which seems to be the reason for these failures but vitest does not seem to support that. 

We could coordinate using filesystem locks maybe?

darkgnotic:
> Yes.
> 
> I was hoping vitest allowed sequential runs of projects which seems to be the reason for these failures but vitest does not seem to support that.
> 
> We could coordinate using filesystem locks maybe?

tbh I don't know exactly what causes the invalid snapshot scenario. i think one thing we could try (though I don't know if this is possible with vitest), is to start up a pg container instance per worker, instead of a single one for the (e.g. pg-17) whole project. do you know if this is possible?

darkgnotic:
> > Yes.
> > I was hoping vitest allowed sequential runs of projects which seems to be the reason for these failures but vitest does not seem to support that.
> > We could coordinate using filesystem locks maybe?
> 
> tbh I don't know exactly what causes the invalid snapshot scenario. i think one thing we could try (though I don't know if this is possible with vitest), is to start up a pg container instance per worker, instead of a single one for the (e.g. pg-17) whole project. do you know if this is possible?

Giving this a shot:  https://github.com/rocicorp/mono/pull/4754

darkgnotic:
Maybe progress? The tests are no longer failing, but the test runner seems to get stuck (deterministically) after around 15 minutes, eventually timing out after 39 minutes.

https://github.com/rocicorp/mono/actions/runs/17019249632/job/48245971672

https://github.com/rocicorp/mono/actions/runs/17019607456/job/48246756547

Screenshot 2025-08-17 at 19 15 50

Is this a higher level issue, such as running out of CPU quota or something?

darkgnotic:
Definitely stuck ... 3 hours and counting ...

Screenshot 2025-08-17 at 22 21 16

Is it possible to try a larger runner to see if it fixes it?

Screenshot 2025-08-17 at 22 18 01

arv:
We could definitely try a larger runner.

darkgnotic:
> We could definitely try a larger runner.

👍  cesar said he will try this tomorrow am.

PR #4744, opened 2025-08-15
chore(zero): move global debug state into BuilderDelegate

tantaman:
Trying to make it easier for users to run analyze query against their production data. The easiest way to do this would be a new endpoint like `statz` but called `analyze_queryz` that will run analyze query and collect stats in process.

This will also greatly improve our lives as we can rapidly iterate on queries in the `zql-viz` tool

CleanShot 2025-08-15 at 11 23 26


packages/zql/src/builder/debug-delegate.ts
```diff
@@ -0,0 +1,84 @@
...
+  reset(): void;
+}
+
+export class Debug implements DebugDelegate {
+  readonly #rowCountsBySource: RowCountsBySource;
+  readonly #rowsBySource: RowsBySource;
+
+  constructor() {
+    this.#rowCountsBySource = new Map();
+    this.#rowsBySource = new Map();
```
arv, line 30: move out of constructor

packages/zql/src/builder/debug-delegate.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  getVendedRows(): RowsBySource {
+    return this.#rowsBySource;
+  }
+
+  initQuery(table: SourceName, query: SQL): void {
+    const {counts} = this.#getRowStats(table);
+    if (counts) {
+      if (!counts.has(query)) {
```
arv, line 44: merge

arv (commented):
LGTM

Is this testes somewhere?

PR #4746, opened 2025-08-15
chore(zero): split `runAst` out into its own function

tantaman:
So we can call this from other places. Specifically the `queryz` endpoint to debug a query on a production server.


arv:
LGTM

PR #4747, opened 2025-08-15
Add useSuspenseQuery hook

lewisl9029:
Hey Rocicorp team! 

Been enjoying Zero in production on https://reflame.app for a few months now, migrating from React Query v5, and the overall DX/UX has been significantly better in most aspects, but with one major exception w.r.t. DX: lack of React Suspense support. 

So far I've been working around this by preloading all of my users' readable data until completion before rendering the rest of my app, and then just ignoring the type === complete checks entirely. This works pretty well with only a few minor exceptions (more on this later), but is starting to break down as customer data sets grow larger (and it means giving up selective sync, which is one of the major advantages of Zero vs other sync engines).

So, I took a crack at implementing Suspense myself, and this PR is the result. I have it running in production right now at https://reflame.app (published at https://www.npmjs.com/package/@lewisl9029/zero in case anyone else wants to give it a try), and it seems to be working as expected for the most part.

I did discover one _very sharp edge_ that I'm not sure what to do about though. In some of my components I have queries that look like this involving `Date.now()`: 

```js
  console.log('before invites')
  // FIXME: seems to trigger infinite re-suspend here?
  const [invites] = useSuspenseQuery(
    z.query.invites
      .whereExists('email', (q) => q.where('userId', '=', user.id))
      .where('status', '=', 'pending')
      .where('validUntil', '>=', Date.now())
      .related('email')
      .related('organization')
      .related('userSender'),
  )
  console.log('invites', invites)
```

The `Date.now()` causes the query to re-suspend infinitely in a tight loop, presumably because every time the component is rendered, useSuspenseQuery sees a brand new query using a new Date.now() value, so has to start from `unknown` status, causing another suspension, and then the cycle repeats. 

Caching in useState or useRef wasn't enough to work around this because the component gets unmounted entirely when suspending, so local state/refs get wiped out. The workaround I'm using now is calling Date.now() outside of the render function, which is fine for this use case, but might not be for others (because it makes the timestamp stale for the duration of the user's session). 

Looking at the section on [query completeness](https://zero.rocicorp.dev/docs/reading-data#completeness), I think we'll need this change to land in order to address the root cause behind this issue:

> in the future, Zero will be able to return this result type when it knows that all possible data for this query is already available locally.

Curious where this stands on the roadmap? And whether y'all think we should release suspense support in this state with this edge case present? From my testing so far the hook has been working as expected outside of this one edge case, fwiw.


aboodman:
Thank you so much for doing this. I have been wanting this for awhile. Taking a look at PR now.

aboodman:
Are you sure that the sharp edge you experience is unique to `useSuspenseQuery` ? We have found something similar: https://bugs.rocicorp.dev/issue/3465#comment-jrBHt1uohlTRFf6Kxrqjg (see latest comment).

packages/zero-react/src/use-query.tsx
```diff
@@ -53,11 +55,48 @@ export function useQuery<
...
   );
   // https://react.dev/reference/react/useSyncExternalStore
-  return useSyncExternalStore(
+  const snapshot = useSyncExternalStore(
     view.subscribeReactInternals,
     view.getSnapshot,
     view.getSnapshot,
   );
+
+  if (suspense && snapshot[1].type === 'unknown') {
```
aboodman: So the thing here is that this will wait until complete which is kind of a bummer for some use cases. Sometimes you'll want ot suspend only until the first non-empty result.

What if `useQuery` took an extra `suspendUntil` option which had a few values:

```
// what is currently implemented - no suspends
none

// what you have implemented - suspend until complete
complete

// suspends until first non-empty result, regardless of type
non-empty
```

Then we would not need a new entrypoint, either.

lewisl9029:
> Are you sure that the sharp edge you experience is unique to `useSuspenseQuery` ? We have found something similar: https://bugs.rocicorp.dev/issue/3465#comment-jrBHt1uohlTRFf6Kxrqjg (see latest comment).

I'm unsure if this is the same issue as the linked comment, but I'm 90% sure the root cause isn't inherent to the useSuspenseQuery version of the hook. 

Even in the regular useQuery hook, a query with an inline Date.now() always starts with `unknown` before transitioning to `complete`, even if all the data is preloaded and locally available. 

That transition is pretty harmless when using the regular useQuery hook, but it's deadly for Suspense because it causes the fallback to render (and the query component to be unmounted), which triggers the re-suspend cycle endlessly since useSuspenseQuery never sees the exact same query when Date.now() is involved, so has to start with `unknown` again and again.

Though a better workaround just popped into my head: we can probably just floor Date.now() to have a bit less precision instead of declaring it outside the render function. As long as its imprecise enought to avoid the initial unknown -> complete -> unknown loop we should be fine.

That said I do still think this will bite quite a few folks if shipped in its current state though, since I'd expect Date.now() in queries might be pretty common in real world apps.

packages/zero-react/src/use-query.tsx
lewisl9029: Ah yep suspendUntil option sounds good to me. I'll try to give it a shot over the weekend. Or feel free to take over the PR if you can't wait. :)

I'm not so sure about unifying the suspense and non-suspense hooks though. Earlier react query v5 betas used to have just a single useQuery hook with a `suspense: boolean` flag, but they transitioned to a separate useSuspenseQuery hook by the time full release came out.

IIRC it was due to how flipping between suspense: true and false between renders resulted in some nasty bugs and how some args/outputs only make sense in a non-suspense context (e.g. the error and loading outputs, in suspense we'd just use suspense fallbacks for loading and error boundaries for error), so splitting the two made types simpler to model.

The implementation and input/output interfaces here look pretty straightforward so far, so I'm not sure how much of their learnings apply in our case, but I figured I'd err on the side of caution and start with a separate hook first. Happy to merge the two if that's the API you'd prefer though!

lewisl9029: Found some links with more context on the change in react query: https://github.com/TanStack/query/discussions/6348
https://github.com/TanStack/query/discussions/6206

lewisl9029: Hmm I just gave adding the `suspendUntil: non-empty` option a try but it wasn't immediately obvious how that could be implemented?

Right now we only have `unknown` and `complete` statuses upstream, so this would require implementing the new status everywhere, correct? Not sure if that should be in scope of this PR. @aboodman Wdyt?

aboodman: Ugh those TanStack links make me not want to support suspense. So. many. edge cases. I'll have to read through the rationale against a single entrypoint and think more carefully.

I get that it's weird to switch wanting to use suspense on and off within one component but i don't understand why that can't be interpreted as just re-mounting the query. Similar to if you'd changed the query itself.

Regarding implementing `non-empty`.

`non-empty` doesn't really make sense as a result type for `useQuery` since `useQuery` must always return synchronously. The `resultType` field for `useQuery` is saying "OK you got some result, what kind of result is it".

But `useSuspenseQuery` is different. It doesn't always have to return data. The flag is saying under what conditions *to* return data. So it's a little different meaning.

Implementing `suspendUntil: non-empty` is super easy. Just wait until "complete || result.length > 0".

aboodman:
> Even in the regular useQuery hook, a query with an inline Date.now() always starts with unknown before transitioning to complete, even if all the data is preloaded and locally available.

Yes this is inherent to Zero until we implement consistency, which is a ways out.

packages/zero-react/src/use-query.tsx
lewisl9029: Yeah Suspense support will definitely increase the maintenance burden. I can't say I fully appreciate all the nuances myself, but even just as a user of Suspense it's been difficult to wrap my head around some of the edge cases I've encountered (the infinite re-suspend thing here is a good example that took me a while to figure out). I do think the DX improvement with Suspense makes it all worthwhile though. Never having to branch on loading/error states in components has truly been a game changer.

Appreciate the extra pointers on `non-empty`! I added it in a new commit [here](https://github.com/rocicorp/mono/pull/4747/commits/9ba5d06c76df1778e43e41f1c7acf290ebadaf34) (had to introduce another promise to suspend on similar to the one for complete). Ended up also moving to 2 distinct implementations for the 2 hooks since they now have different input args as well.

Also added a few more tests that I managed to get passing, but I haven't had a chance to test it in my product yet (haven't had a use case for suspending on anything other than complete at the moment).

packages/zero-react/src/use-query.tsx
```diff
@@ -30,6 +31,21 @@ export type UseQueryOptions = {
   ttl?: TTL | undefined;
 };
 
+export type UseSuspenseQueryOptions = {
```
grgbkr:
```suggestion
export type UseSuspenseQueryOptions =
  UseQueryOptions & {
    /**
     * Whether to suspend until the query is complete or until the query has non-empty data.
     * Default is 'complete'.
     */
     suspendUntil?: 'complete' | 'non-empty';
  };
```

lewisl9029: Ah great call, applied it manually in this commit: https://github.com/rocicorp/mono/pull/4747/commits/96bcfab738d16c7f0400f6a8d2957fdeee745de0

Also updated the branch against latest main to get rid of the merge conflict. Let me know if there's anything else I can do to help move this along!

packages/zero-react/src/use-query.tsx
```diff
@@ -274,6 +351,10 @@ class ViewWrapper<
   #snapshot: QueryResult<TReturn>;
   #reactInternals: Set<() => void>;
   #ttl: TTL;
+  #completeResolver: Resolver<void> | undefined;
```
grgbkr, line 354: I think this can be simplified a little bit. 

```
#complete = false;  // set to true when you resolve #completeResolver
#completeResolver = resolver<void>();
#nonEmpty = false; // set to true when you resolve #nonEmptyResolver
#nonEmptyResolver = resolver<void>();

// use these when checking if useSuspenseQuery
// should suspend
get complete() {
  return this.#complete;
}
get nonEmpty() {
  return this.#nonEmpty;
}
```

Where you need the promise just do
`this.#completeResolver.promise;`

Finally, I think the most correct place to reset these resolvers is where we destroy the view / dematerialize (see the callback returned from `subscribeReactInternals`).  

We can do 
```
 // We already destroyed the view
if (this.#view === undefined) {
  return;
}
this.#view?.destroy();
this.#view = undefined;
this.#complete = false;
this.#completeResolver = resolver();
this.#nonEmpty = false;
this.#nonEmptyResolver = resolver();
```

This also means #completeResolver and #nonEmptyResolver are never undefined so some of the `?.`s can go away.

packages/zero-react/src/use-query.tsx
```diff
@@ -350,4 +447,28 @@ class ViewWrapper<
     this.#ttl = ttl;
     this.#view?.updateTTL(ttl);
   }
+
+  waitForComplete(): Promise<void> {
+    if (!this.#complete) {
```
grgbkr, line 452: I don't understand why `waitForComplete` and `waitForNonEmpty` reset these resolvers.  In the case that `useSuspenseQuery` is reusing a cached view that is already complete/non-empty, we don't want the hook to suspend until the view happens to have a change and invokes the this.#onData listener and re-resolve these promises.  The reused cached view is already complete/non-empty, and their is no need to suspend.

packages/zero-react/src/use-query.tsx
```diff
@@ -64,6 +73,70 @@ export function useQuery<
...
+  // https://react.dev/reference/react/useSyncExternalStore
+  const snapshot = useSyncExternalStore(
+    view.subscribeReactInternals,
+    view.getSnapshot,
+    view.getSnapshot,
+  );
+
+  // React 19 exposes use(), otherwise we throw the promise to suspend
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+  const useHook = (React as unknown as {use?: (p: Promise<void>) => void}).use;
```
grgbkr, line 112: Note React.use is technically not a hook, as it can be safely used in conditionals and loops.

```suggestion
// reactUse and suspect can be hoisted to the top of this module
// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
const reactUse = (React as unknown as {use?: (p: Promise<unknown>) => void})
  .use;
const suspend: (p: Promise<unknown>) => void = reactUse
  ? reactUse
  : p => {
      throw p;
    };

if (suspendUntil === 'complete' && !view.complete) {
  suspend(view.waitForComplete());
} 
if (suspendUntil === 'non-empty' && !view.nonEmpty) {
  suspend(view.waitForNonEmpty());
} 

```

I think it's best we avoid duplicating the complete and non-empty checks here.   To much room for an inconsistency bug.  For instance I think the 'non-empty'  check below should be 

```
suspendUntil === 'non-empty' &&
    // this complete check was missing
    snapshot[1].type !== 'complete' &&
    (query.format.singular
      ? snapshot[0] === undefined
      : (snapshot[0] as unknown[]).length === 0
```

It's tempting to simplify even more and just do

```
if (suspendUntil === 'complete') {
  suspend(view.waitForComplete());
} 
if (suspendUntil === 'non-empty') {
  suspend(view.waitForNonEmpty());
} 
```

But for React 18 I think its best to avoid `throwing` when we can, since Exception dispatch is a slow/heavy operation.

grgbkr (commented):
Thanks so much for the contribution.  I sent you a friend request on Discord, feel free to message me there if you'd like to discuss any of the feedback.

grgbkr:
> Thanks so much for the contribution. I sent you a friend request on Discord, feel free to message me there if you'd like to discuss any of the feedback.

@lewisl9029 I think it'll be easiest for me to just take this PR over and land it, see https://github.com/rocicorp/mono/pull/4826.  Thanks for the great start.

lewisl9029:
> > Thanks so much for the contribution. I sent you a friend request on Discord, feel free to message me there if you'd like to discuss any of the feedback.
> 
> @lewisl9029 I think it'll be easiest for me to just take this PR over and land it, see #4826. Thanks for the great start.

Sounds good, looking forward to seeing it in the next release! Cheers! :)

aboodman:
Hm, sadly this wasn't as epic as I dreamed. Feedback wanted: https://github.com/rocicorp/mono/pull/4830

PR #4750, opened 2025-08-16
feat(zero): Add inspector metrics for query-update-server

arv:
This PR adds inspector metrics for `query-update-server`, enabling measurement of update times per query (`queryID`) and globally across all updates. Metrics are collected via a `MeasurePush` operator in the IVM pipeline, similar to the client-side approach. Inspector responses now include these metrics. Also includes a test fix for metric integration.### Summary
Adds inspector metrics for `query-update-server`, allowing measurement of update times both per query (`queryID`) and globally.

### Changes
- Collects update time metrics via `MeasurePush` operator in the IVM pipeline (mirroring client-side).
- Inspector responses now include these new metrics.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/view-syncer.ts
+++ b/packages/zero-cache/src/services/view-syncer/view-syncer.ts
@@ -14,9 +14,7 @@ import {assert, unreachable} from '../../../../shared/src/asserts.ts';
 import {stringify} from '../../../../shared/src/bigint-json.ts';
 import {CustomKeyMap} from '../../../../shared/src/custom-key-map.ts';
 import {must} from '../../../../shared/src/must.ts';
-import {mapValues} from '../../../../shared/src/objects.ts';
 import {randInt} from '../../../../shared/src/rand.ts';
-import {TDigest} from '../../../../shared/src/tdigest.ts';
 import type {AST} from '../../../../zero-protocol/src/ast.ts';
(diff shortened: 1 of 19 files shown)


packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -1079,6 +1067,7 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
           span.setAttribute('table', transformedAst.table);
           for (const _ of this.#pipelines.addQuery(
             transformationHash,
+            hash,
```
arv, line 1070: @darkgnotic I think this is correct but I wanted to verify with you. I see in some other places we map from the transformationHash to multiple client side hashes (aka queryID).

What happens if two queries ends up with the same transformationHash? Do they share the IVM pipeline? If so, then I would need to map the ID similar to how we do it in other places.

darkgnotic: Yes, if two queries end up with the same transformationHash, they share the same pipeline. This is not expected to happen often but it does happen, for example, when two queries map to the "not-permitted" query and return an empty result.

So technically a one-to-many map would be needed here.

arv: OK. I will figure out how to find this mapping.

PR #4754, opened 2025-08-17
chore(zero-cache): run a pg (container) per vite worker

darkgnotic:
For the zero-cache PG tests, run a pg container per vite worker (i.e. subprocess) rather than per project (e.g. pg-15, pg-16, pg-17), using the [Per Scope Test Context](https://vitest.dev/guide/test-context.html#per-scope-context-3-2-0) feature of vite.

The theory is that isolating PG access to a single thread / process will reduce "Invalid snapshot identifier" flakiness in the PG tests.


arv:
LGTM

Was this the magic saude?

PR #4755, opened 2025-08-17
chore(zero-cache): increase vitest timeouts, disable isolation

darkgnotic:
Disable vitest isolation so that a single worker can run multiple files. Increase default timeouts.

### DIFF
--- a/packages/zero-cache/vitest.config.ts
+++ b/packages/zero-cache/vitest.config.ts
@@ -23,6 +23,13 @@ export function configForVersion(version: number, url: string) {
         reporter: [['html'], ['clover', {file: 'coverage.xml'}]],
         include: ['src/**'],
       },
+      poolOptions: {
+        forks: {
+          isolate: false,
+        },
+      },
+      testTimeout: 20000,
+      hookTimeout: 20000,
     },
   });
 }


packages/zero-cache/vitest.config.ts
```diff
@@ -23,6 +23,13 @@ export function configForVersion(version: number, url: string) {
         reporter: [['html'], ['clover', {file: 'coverage.xml'}]],
         include: ['src/**'],
       },
+      poolOptions: {
+        forks: {
+          isolate: false,
```
arv, line 28: Why does this help? Would it not cause me trouble if the tests are not isolated?

arv (approved):
If it works it works but I do not understand why this would work

packages/zero-cache/vitest.config.ts
darkgnotic: I'm attempting to reduce the number of pg containers by increasing the number of files that are used for a single worker, based on https://vitest.dev/guide/test-context.html#per-scope-context-3-2-0

> The worker scope will run the fixture once per worker. The number of running workers depends on various factors. By default, every file runs in a separate worker, so file and worker scopes work the same way.
>
> However, if you disable [isolation](https://vitest.dev/config/#isolate), then the number of workers is limited by the [maxWorkers](https://vitest.dev/config/#maxworkers) or [poolOptions](https://vitest.dev/config/#pooloptions) configuration.

I'm not really sure if it makes any difference.

PR #4758, opened 2025-08-18
chore: Run JS CI Tests on larger runner

arv:


arv:
Our self-hosted runner is only used by perf tests.

The Github provided larger runners is a github feature and we have to pay github instead of amazon.

I feel more comfortable to use GH here to manage our runner than having to manually manage this on EC2.

arv:
Summary of changes:

- Added to GH hosted runners. Both are ARM64 with "8-core · 32 GB RAM · 300 GB SSD".
  - The "test" one allows 50 concurrent jobs.
  - The "perf" one only allows 1 job. This is to reduce the flake.
- Change all the perf tests to use the perf runner (even the old replicache one)
  - This means we can retire the self hosted AWS/EC2 one.
    - But lets wait a little to see price and stability.

PR #4761, opened 2025-08-18
chore(zero): do not transform custom queries more than once

tantaman:
Calling the API server every single time we sync pipelines or add a custom query that has already been transformed can cause unexpected behavior on the client.

The issue is that once a client has `gotten` a query, a re-transform on the server could flip it into an `error` state due to the user's API server not responding or returning an error. We should not re-transform queries that are already running.

This PR updates `view-syncer` to not transform a custom query if the pipeline driver is already running that query.

The PR takes a minimally invasive approach by acting as if the transformation call was made when a query, with the same id, is already running.

I don't understand how often `syncQueryPipelineSet` is called (but it shouldn't matter with the approach taken here). Would appreciate some insight on that.

--
We'll eventually prefix the custom query id with client id so each client will cause a transform of its own query requests.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/pipeline-driver.test.ts
+++ b/packages/zero-cache/src/services/view-syncer/pipeline-driver.test.ts
@@ -683,7 +683,52 @@ describe('view-syncer/pipeline-driver', () => {
         startTimer(),
       ),
     ];
-    expect(pipelines.addedQueries()).toEqual(new Set(['hash1']));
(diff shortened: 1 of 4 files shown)


darkgnotic:
> I don't understand how often syncQueryPipelineSet is called (but it shouldn't matter with the approach taken here). Would appreciate some insight on that.

syncQueryPipelineSet is called for all incremental advancements, i.e. whenever a transaction is committed to the replica.

PR #4762, opened 2025-08-18
chore: Only run pg-17 tests on CI

arv:


arv:
If this work we can run the 15 and 16 afterwards  as a top level call to `npx vitest --project=zero-cache/pg-15`

PR #4763, opened 2025-08-18
chore(zero-cache): switch back to a per-vitest-project pg instance

darkgnotic:
The github actions test runner otherwise hangs when attempting to run a pg per-worker.

To reduce invalid-snapshot flakiness, limit the number table copy workers in some of the quick initial-sync tests. The theory is that the pool is being `setDone()`, and the replication session closed, before the pool workers start up, causing the pool workers to fail when attempting to start up (and initialize the snapshot).

### DIFF
--- a/packages/zero-cache/src/db/transaction-pool.pg-test.ts
+++ b/packages/zero-cache/src/db/transaction-pool.pg-test.ts
@@ -185,7 +185,8 @@ describe('db/transaction-pool', () => {
     });
   });
 
-  test('pool resizing before run', async () => {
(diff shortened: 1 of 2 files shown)


packages/zero-cache/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "test": "vitest run",
+    "test": "vitest run --maxWorkers 3",
```
arv: This needs to be in the root package json

arv:
LGTM

PR #4765, opened 2025-08-18
feat(zero-client)!: surface query transform errors to callers of `useQuery`

tantaman:
```ts
const [data, details] = useQuery(...);

if (details.type === 'error') {
  // do something with result.error
}
```

Coming next is a way to let the user `retry` failed queries. E.g.,

```ts
datails.retry();
```

Might go with a shape more like TanStack's --

https://tanstack.com/query/v4/docs/framework/react/reference/useQuery

which would look like:

```ts
const [data, {type, error, retry}] = useQuery(...);
```

### DIFF
--- a/packages/zero-client/src/client/query-manager.ts
+++ b/packages/zero-client/src/client/query-manager.ts
@@ -10,6 +10,7 @@ import {
   type AST,
 } from '../../../zero-protocol/src/ast.ts';
 import type {ChangeDesiredQueriesMessage} from '../../../zero-protocol/src/change-desired-queries.ts';
+import type {ErroredQuery} from '../../../zero-protocol/src/custom-queries.ts';
 import type {UpQueriesPatchOp} from '../../../zero-protocol/src/queries-patch.ts';
 import {
   hashOfAST,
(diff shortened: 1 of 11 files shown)


tantaman:
> This is a breaking change.

Is it? The error case is an additional option rather than a modification of an existing option.

tantaman:
Added tests. Meant to have this in draft beforehand.

tantaman:
I think we're going to have to mark the query as `error'ed` in the `useQuery` hook so it does not retry and cause i-loops.

Theoretical problem:
1. Send query
2. Query fails
3. Query retracted from manager
4. Component renders
5. Query re-added to manager
6. repeat

arv:
> > This is a breaking change.
> 
> Is it? The error case is an additional option rather than a modification of an existing option.

Yes. Adding a new enum option is a breaking change.


```ts
function f(v: V) {
  if (v === 1) ok();
  else if (v === 2) ok();
  else fail();
}

// in v1
type V = 1 | 2;

// in v2
type V = 1 | 2 | 3;
```

tantaman:
Needs tests for the new additions to ArrayView and useQuery

packages/zero-client/src/client/query-manager.ts
```diff
@@ -281,7 +292,7 @@ export class QueryManager implements InspectorDelegate {
     }
 
     let removed = false;
-    return () => {
+    const cleanupCb = () => {
```
arv, line 295: inline

arv: This was probably only changed for debugging purpose?

arv (commented):
LGTM

PR #4767, opened 2025-08-19
chore(zero-cache): wait for copiers to close before closing replication

darkgnotic:
This should correctly fix the initial snapshot issue by ensuring that all copy workers start up (and set their snapshots) before the replication session is closed.

Keep the no-wait workaround for win32 only (to address the COPY hang bug in Node win32).


arv:
LGTM

PR #4768, opened 2025-08-19
chore(zero-cache): reenable change-source.pg tests

darkgnotic:
Snapshot flakiness was fixed with https://github.com/rocicorp/mono/pull/4767


arv:
LGTM

PR #4769, opened 2025-08-19
optimize queries server side in get-query

grgbkr:


apps/zbugs/shared/queries.ts
```diff
@@ -112,24 +112,199 @@ export const queries = queriesWithContext({
...
+    buildBaseListQuery(
+      builder.issue,
+      listContext,
+      issue,
+      dir,
       auth?.role,
-    ),
+    ).one(),
+
+  prevNextForAssignee: (
```
tantaman: Realistically these wouldn't be in `shared`, correct? Since they are only needed on the server.

apps/zbugs/shared/queries.ts
```diff
@@ -112,24 +112,199 @@ export const queries = queriesWithContext({
...
+          listContext === null ? null : {...listContext, creator: null},
+          issue,
+          dir,
+          auth?.role,
+        ).one(),
+      )
+      .one();
+  },
+
+  prevNextForLabel: (
```
tantaman: this is unfortunate. I was hoping there'd be much less code but there is still quite a bit to try to keep in sync between all the variants of list queries. Seems like this would be an unending source of bugs.

Put another way, if I was developing a product with Zero I'd feel compelled to write tests for these queries to ensure they are always returning the same data and it doesn't seem that easy to write those tests and set up the data in such a way that each test would return matching results.

grgbkr: Agree.

I managed to refactor and get rid of a lot of redundancy.  I think its better now.  I'd still feel better with tests confirming the two things return the same thing.

apps/zbugs/server/get-query.ts
```diff
@@ -16,8 +29,116 @@ export function getQuery(
   args: readonly ReadonlyJSONValue[],
 ) {
   if (name in validated) {
+    if (name === 'issueList') {
+      assert(queries.issueList.parse);
```
aboodman: Hm, I think we can fix this so it has the correct (defined) type.

apps/zbugs/server/get-query.ts
```diff
@@ -16,8 +29,116 @@ export function getQuery(
...
+      const [listContext, userID, limit] = queries.issueList.parse([...args]);
```
aboodman: the approach I would have went for here is to build a separate server version of the queries. Kinda like how we have had `serverMutators` in the past.

Then run `withValidation` over that set instead.

grgbkr: took a stab at this, it is a bit cleaner.  nice.

PR #4770, opened 2025-08-19
narrow down which tests are hanging

darkgnotic:


darkgnotic:
@arv this PR doesn't seem to make the tests actually run, but I think the next step would be something like this, where we use process of elimination to identify if there's a particular test suite that causes the test runner to hang.

PR #4771, opened 2025-08-19
feat(zero): syncedQuery -- context, validation

tantaman:
https://www.notion.so/replicache/Custom-Query-Burndown-2543bed8954580b2af73e6b2d27e1c6b

# Basic usage

## A Simple Synced Query
```ts
const issueById = syncedQuery('issue', (id: string) =>
    builder.issue.where('id', id),
 );
```

## A Synced Query with Context
```ts
const viewerIssues = syncedQueryWithContext('viewersIssues', (ctx: AuthData) =>
    builder.issue.where('ownerId', ctx.userId)
 );

const query = viewerIssues({userID: 123});
```

Context arguments are not sent over the wire.

## A Synced Query with Validation
```ts
const issueById = syncedQuery(
  'issue',
  (...args: unknown[]) => {
    assert(typeof args[0] === 'string');
    return [args[0]] as const;
  },
  (id: string) => builder.issue.where('id', id),
);

// use the validated form:
const validated = withValidation(issueById);
validated(...unknownPayload);

// or the unvalidated form
issueById('123');
```

## A Synced Query with Validation and Context

```ts
const viewerIssues = syncedQuery(
  'viewerIssues',
  (...args: unknown[]) => {
    assert(args.length === 0 || typeof args[0] === number);
    return [args[0]] as const;
  },
  (ctx: AuthData, created?: number) => builder.issue.where('ownerId', ctx.userId).where('createdAt', '>', created),
);

// use the validated form:
const validated = withValidation(viewerIssues);
validated(ctx, ...unknownPayload);

// or the unvalidated form
viewerIssues(ctx, '123');
```

**Note:** Calling `withValidation` on a query that has no validator is an error.

### DIFF
--- a/apps/zbugs/shared/queries.ts
+++ b/apps/zbugs/shared/queries.ts
@@ -2,7 +2,7 @@ import {
   escapeLike,
   type Query,
   type Row,
-  queriesWithContext,
+  syncedQueryWithContext,
 } from '@rocicorp/zero';
 import {builder, type Schema} from './schema.ts';
 import {INITIAL_COMMENT_LIMIT} from './consts.ts';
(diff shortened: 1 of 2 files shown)


packages/zql/src/query/named.test.ts
```diff
@@ -27,64 +29,141 @@ test('defining a named query', () => {
...
+});
+
+test('defining a synced query with validation', () => {
+  const def = syncedQuery(
+    'myQuery',
+    (id: unknown) => {
+      assert(typeof id === 'string', 'id must be a string');
+      return [id] as const;
+    },
+    (id: string) => builder.issue.where('id', id),
```
aboodman: Can you add a test that the inference works? Should not need to provide type on `id` in query func.

packages/zql/src/query/named.test.ts
```diff
@@ -1,20 +1,22 @@
...
+const builder = createBuilder(schema);
 
-test('defining a named query', () => {
-  const queryBuilder = createBuilder(schema);
-  const x = queries({
-    myName: (id: string) => queryBuilder.issue.where('id', id),
-  });
-  const q = x.myName('123');
+test('defining a synced query', () => {
+  const def = syncedQuery('myQuery', (id: string) =>
```
aboodman, line 16: Can you add a test for no args case? Like for `allLabels`.

packages/zql/src/query/named.test.ts
```diff
@@ -27,64 +29,141 @@ test('defining a named query', () => {
...
-    hashOfAST((r as StaticQuery<typeof schema, 'issue'>).ast),
+test('defining a synced query with validation and context', () => {
+  const def = syncedQueryWithContext(
+    'myQuery',
+    (id: unknown, createdAt: unknown) => {
+      assert(typeof id === 'string', 'id must be a string');
+      assert(typeof createdAt === 'number', 'createdAt must be a number');
+      return [id, createdAt] as const;
+    },
+    (ctx: object, ownerId: string, createdAt: number) => {
```
aboodman: Can you add a test with the form: `syncedWithContext<string>("foo", validator, fn)` where the signature of fn doesn't need to include types for args or auth.

aboodman:
Can we make the `validator` argument accept an object that has a `parse` field which is the right signature? This is a little flourish that tanstack does which is nice.

```ts
// Note: `parse` implied
syncedQuery("foo", z.array(z.number(), z.string()), ... )
```

aboodman:
Actually my use of `z.array()` in all those examples is wrong. It must be `z.tuple()` to validate an argument list:

CleanShot 2025-08-19 at 14 09 16@2x

Also could we change the calling convention to allow this:

```ts
export const indexPage = syncedQuery(
  'indexPage', [z.string(), z.number()], (filter: string, count: number) => {
    return builder.artist
      .where('name', 'ILIKE', `%${filter}%`)
      .orderBy('popularity', 'desc')
      .limit(limit);
  },
);
```

Just reduces boilerplate of having to wrap in `tuple` and add parse.

PR #4773, opened 2025-08-19
feat(zero): `run` and `materialize` methods on the `zero` instance

tantaman:
needs some tests


aboodman:
LGTM

packages/zero-client/src/client/zero.ts
```diff
@@ -811,6 +817,42 @@ export class Zero<
...
+    query: Q,
+    factoryOrOptions?:
+      | ViewFactory<S, QueryTable<Q>, QueryReturn<Q>, T>
+      | MaterializeOptions
+      | undefined,
+    maybeOptions?: MaterializeOptions | undefined,
+  ) {
+    if (typeof factoryOrOptions === 'function') {
+      return (query as AnyQuery)
+        .delegate(this.#zeroContext)
```
arv, line 848: Can you remove `delegate` now?

arv: It should not be part of the public API unless we are willing to document its usage.

aboodman: I know it shouldn't but I told Matt it was OK to ship this w/ delegate if it was hard to remove (I don't really know if it is hard to remove).

It's not great but there are other instances of public undocumented APIs and I think it's just the reality of alpha software.

PR #4780, opened 2025-08-20
Aa/cq

aboodman:
Couldn't quite get this working. Something wrong with types.

The first commit is good and fixes a problem using valita. But the second one doesn't solve problem treating synced queries generically.

### DIFF
--- a/apps/zbugs/shared/queries.ts
+++ b/apps/zbugs/shared/queries.ts
@@ -2,11 +2,13 @@ import {
   escapeLike,
   type Query,
   type Row,
+  syncedQuery,
   syncedQueryWithContext,
 } from '@rocicorp/zero';
 import {builder, type Schema} from './schema.ts';
 import {INITIAL_COMMENT_LIMIT} from './consts.ts';
 import type {AuthData, Role} from './auth.ts';
+import * as v from '../../../packages/shared/src/valita.ts';
 
 function applyIssuePermissions<TQuery extends Query<Schema, 'issue', any>>(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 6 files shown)


apps/zbugs/shared/queries.ts
```diff
@@ -57,13 +71,15 @@ export const queries = {
 
   user: syncedQueryWithContext(
     'user',
-    (_auth: AuthData | undefined, userID: string) =>
+    v.tuple([v.string()]).parse,
+    (_auth: AuthData | undefined, userID) =>
```
florian-lefebvre, line 75: Since auth is always `AuthData | undefined`, maybe there could be an API where you define this once? Eg.
```ts
const { syncedQueryWithContext } = createQueriesWithContextFactory<AuthData>()

syncedQueryWithContext(
	'user',
	validator.parse,
	// Automatically typed as AuthData | undefined
	(auth, userID) => {}
)
```
This should be quite easy to do in TS

packages/zql/src/query/named.ts
```diff
@@ -256,13 +256,13 @@ export function withValidation<
...
           >
-        )(context, ...validator(...args))) as ValidatedSyncedQueryWithContext<
+        )(context, ...validator(args))) as ValidatedSyncedQueryWithContext<
         TContext,
         TReturnQuery
       >;
     }
     return ((...args) =>
-      fn(...validator(...args))) as ValidatedSyncedQuery<TReturnQuery>;
+      fn(...validator(args))) as ValidatedSyncedQuery<TReturnQuery>;
```
florian-lefebvre, line 265: For the validator, it would be great to support https://github.com/standard-schema/standard-schema

PR #4781, opened 2025-08-20
chore: Use playwright docker container

arv:


arv:
This will not work because we have a docker image for pg?

PR #4783, opened 2025-08-20
fix(zero): zod/valita/whatever validator functions take a single arg

tantaman:
```ts
const paramSchema = v.tuple([v.string(), v.number()]);
const def = syncedQueryWithContext(
  'myQuery',
  paramSchema.parse.bind(paramSchema),
  (ctx: object, ownerId, createdAt) => {
    expect(ctx).toEqual({});
    return builder.issue
      .where('ownerId', ownerId)
      .where('createdAt', '>', createdAt);
  },
);

let q = def({}, '123', 123);
expect(q.customQueryID).toEqual({
  name: 'myQuery',
  args: ['123', 123],
});

const validated = withValidation(def);
q = validated({}, '321', 321);
```

Some weird typing errors I need to work through when using withValidation(withContext( in zbugs -- https://github.com/rocicorp/mono/pull/4784

### DIFF
--- a/packages/zql/src/query/named.ts
+++ b/packages/zql/src/query/named.ts
@@ -57,7 +57,7 @@ export type CustomQueryID = {
 };
 
 export type Validator<T extends ReadonlyArray<ReadonlyJSONValue>> = (
-  ...args: readonly unknown[]
+  args: readonly unknown[],
 ) => T | readonly [...T];
 
 /**
@@ -247,6 +247,7 @@ export function withValidation<
   | ValidatedSyncedQuery<TReturnQuery>
   | ValidatedSyncedQueryWithContext<TContext, TReturnQuery> {
   const {validator, takesContext} = fn;
+
   if (validator) {
     if (takesContext) {
       return ((context, ...args) =>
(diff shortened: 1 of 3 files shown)


tantaman:
Some weird typing errors I need to work through when using `withValidation(withContext(` in zbugs -- https://github.com/rocicorp/mono/pull/4784

packages/zql/src/query/named.ts
```diff
@@ -57,7 +57,7 @@ export type CustomQueryID = {
 };
 
 export type Validator<T extends ReadonlyArray<ReadonlyJSONValue>> = (
-  ...args: readonly unknown[]
+  args: readonly unknown[],
 ) => T | readonly [...T];
```
aboodman, line 61: Do we really want the return value to possibly be `T` here? Wouldn't it always be `readonly [...T]` ?

Alternately the way I did it was just have the input be a single `unknown`, because that can represent an array too, and have the output be `T`:

```ts
export type Validator<T extends ReadonlyArray<ReadonlyJSONValue>> = (
  args: unknown,
) => T;
```

What is forcing it to be an array giving us?

tantaman: Without this we get errors when the user does `as const` in their validator

e.g.,

```
(args) => [args[0], args[1]] as const
```

I'll see if there is a workaround

aboodman: I dont think I care too much about the return type here, i was more curious. As long as the usage we have been working toward works, whatever type schenanigans are required seem OK to me.

PR #4784, opened 2025-08-20
validator, context on zbugs

tantaman:


apps/zbugs/server/get-query.ts
```diff
@@ -7,21 +12,21 @@ export function getQuery(
...
-    query = (
-      queries[name] as (
-        context: AuthData | undefined,
-        ...args: readonly ReadonlyJSONValue[]
-      ) => AnyQuery
-    )(context, ...args);
-  } else {
-    throw new Error(`Unknown query: ${name}`);
+  if (isQuery(name)) {
+    // The cast is required because, otherwise, TypeScript reduces `queries[name]` to the supertype of all
```
aboodman, line 16: I was thinking that we'd create a new map of all the wrapped `withContext(withValidation())` versions of the CQs local to this file. Then they will all have the same type (all take ReadonlyJSONValue[] as arg, and context) and no cast required.

PR #4786, opened 2025-08-20
feat(zero-client)!: Include client ZQL for custom queries

arv:
This is a breaking change for the inspector API.

This removes `ast` and `zql` in favor of `clientZQL` and `serverZQL`.

`serverZQL` is always `null` at this point.

To make this work `QueryManager.addCustomQuery` now takes the AST.


packages/zero-client/src/client/query-manager.ts
```diff
@@ -262,7 +262,7 @@ export class QueryManager implements InspectorClientMetricsDelegate {
       this.#queueQueryChange({
         op: 'put',
         hash: queryId,
-        ast: normalized,
+        ast: name === undefined ? normalized : undefined,
```
tantaman, line 265: 👍

PR #4793, opened 2025-08-21
feat(zero-cache): added passthrough of auth

0xcadams:
Changes zero-cache to allow pass-through of a token passed to `auth` without validation if no JWT configs are set. It will also print a warning to the user: `One of jwk, secret, or jwksUrl is not configured - the auth token must be manually verified by the user`.

On the client:

```tsx
import { expoClient } from "@better-auth/expo/client";
import { createAuthClient } from "better-auth/react";
import * as SecureStore from "expo-secure-store";

export const authClient = createAuthClient({
  baseURL: "http://localhost:3000",
  plugins: [
    expoClient({
      scheme: "hello-zero-expo",
      storagePrefix: "hello-zero-expo",
      storage: SecureStore,
    }),
  ],
});

new Zero({
  ...
  auth: authClient.getCookie(),
});
```

Server:

```tsx
app.use("*", async (c, next) => {
  const authHeader = c.req.raw.headers.get("Authorization");
  const cookie = authHeader?.split("Bearer ")[1];
  const newHeaders = new Headers(c.req.raw.headers);
  if (cookie) {
    newHeaders.set("Cookie", cookie);
  }
  
  const session = await auth.api.getSession({ headers: newHeaders });
  if (!session) {
    c.set("auth", null);
    return next();
  }
  c.set("auth", session);
  return next();
});
```

### DIFF
--- a/packages/zero-cache/src/workers/syncer.ts
+++ b/packages/zero-cache/src/workers/syncer.ts
@@ -167,10 +167,10 @@ export class Syncer implements SingletonService {
         new SyncerWsMessageHandler(
           this.#lc,
           params,
-          auth !== undefined && decodedToken !== undefined
+          auth !== undefined
             ? {
                 raw: auth,
-                decoded: decodedToken,
+                decoded: decodedToken ?? {},
               }
             : undefined,
           this.#viewSyncers.getService(clientGroupID),
--- a/packages/zero-cache/src/workers/syncer.ts
+++ b/packages/zero-cache/src/workers/syncer.ts
@@ -6,7 +6,7 @@ import {MessagePort} from 'node:worker_threads';
 import {WebSocketServer, type WebSocket} from 'ws';
 import {promiseVoid} from '../../../shared/src/resolved-promises.ts';
 import {ErrorKind} from '../../../zero-protocol/src/error-kind.ts';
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/workers/syncer.ts
```diff
@@ -112,25 +113,41 @@ export class Syncer implements SingletonService {
...
+              kind: ErrorKind.AuthInvalidated,
+              message: `Failed to decode auth token: ${String(e)}`,
+            },
+            e,
+          );
+          ws.close(3000, 'Failed to decode JWT');
+          return;
+        }
+      } else {
+        this.#lc.warn?.(
```
tantaman, line 153: It is technically still an error if they do not have both `ZERO_MUTATE_URL` and `ZERO_GET_QUERIES_URL` set as we need this information for crud and legacy queries.

I think it is worth handling that case given we'll have these features a few more weeks.

tantaman (approved):
lgtm with 1 suggestion.

packages/zero-cache/src/workers/syncer.ts
0xcadams: Makes sense - I will add a check for that and keep the error here in that case.

PR #4796, opened 2025-08-21
chore: Order the JS jobs

arv:
No need to waste CPU if format, lint or check-types fails.


arv:
It does increase the whole run by a minute but it seems reasonable.

PR #4798, opened 2025-08-21
chore: Use pnpm for lint, format and check-types

arv:


arv:
The good thing is that "npm install" step goes from 60s to 15s.

The bad thing is that we would have to change everywhere.

Not doing this at this point in time

PR #4804, opened 2025-08-21
feat(zero): Expose server AST of CQ to inspector

arv:
The server now keeps the returned AST from a custom query in memory so that it can be returned to the inspector as needed.


packages/zero-cache/src/server/inspector-delegate.ts
```diff
@@ -50,24 +53,35 @@ export class InspectMetricsDelegate implements MetricsDelegate {
...
+      ? this.#transformationASTs.get(transformationHash)
+      : undefined;
+  }
+
+  removeQuery(queryID: string): void {
     this.#perQueryServerMetrics.delete(queryID);
+    this.#queryIDToTransformationHash.delete(queryID);
     // Remove queryID from all hash-to-ID mappings
-    for (const [hash, idSet] of this.#hashToIDs.entries()) {
+    for (const [transformationHash, idSet] of this.#hashToIDs.entries()) {
```
tantaman, line 67: do we need to iterate the whole map? Wouldn't we have the specific transformation hash here from the line above:

```
this.#queryIDToTransformationHash.delete(queryID);
```

PR #4815, opened 2025-08-21
feat(zero-cache): added passthrough of auth (again)

0xcadams:
This is a reapply of https://github.com/rocicorp/mono/pull/4793, with fixes for zbugs.

The `auth` config defaults to an empty string in Replicache here:
https://github.com/rocicorp/mono/blob/842c0d5e6f7f588b3f5cfe7ca3fc6fcbc1879c51/packages/replicache/src/replicache-impl.ts#L444-L447

So `auth` is never `undefined`. This [commit](https://github.com/rocicorp/mono/commit/d5642c1131cb730ab340b826433ac49659ff2e7c) fixes handling this.

---

Changes zero-cache to allow pass-through of a token passed to `auth` without validation if no JWT configs are set, and a mutate & get queries URL are defined. It will also print a warning to the user: `One of jwk, secret, or jwksUrl is not configured - the auth token must be manually verified by the user`.

On the client:

```tsx
import { expoClient } from "@better-auth/expo/client";
import { createAuthClient } from "better-auth/react";
import * as SecureStore from "expo-secure-store";

export const authClient = createAuthClient({
  baseURL: "http://localhost:3000",
  plugins: [
    expoClient({
      scheme: "hello-zero-expo",
      storagePrefix: "hello-zero-expo",
      storage: SecureStore,
    }),
  ],
});

new Zero({
  ...
  auth: authClient.getCookie(),
});
```

Server:

```tsx
app.use("*", async (c, next) => {
  const authHeader = c.req.raw.headers.get("Authorization");
  const cookie = authHeader?.split("Bearer ")[1];
  const newHeaders = new Headers(c.req.raw.headers);
  if (cookie) {
    newHeaders.set("Cookie", cookie);
  }
  
  const session = await auth.api.getSession({ headers: newHeaders });
  if (!session) {
    c.set("auth", null);
    return next();
  }
  c.set("auth", session);
  return next();
});
```

### DIFF
--- a/packages/zero-cache/src/workers/syncer.ts
+++ b/packages/zero-cache/src/workers/syncer.ts
@@ -31,7 +31,6 @@ import {
   recordConnectionAttempted,
   setActiveClientGroupsGetter,
 } from '../server/anonymous-otel-start.ts';
-import {assert} from '../../../shared/src/asserts.ts';
 
 export type SyncerWorkerData = {
   replicatorPort: MessagePort;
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/workers/syncer.ts
```diff
@@ -154,10 +176,11 @@ export class Syncer implements SingletonService {
         new SyncerWsMessageHandler(
           this.#lc,
           params,
-          auth !== undefined && decodedToken !== undefined
+          // auth is an empty string if the user is not authenticated
+          auth
```
0xcadams, line 182: This was changed from `auth !== undefined` in the previous PR

aboodman (approved):
Neat. Nice work!

packages/zero-cache/src/workers/syncer.ts
```diff
@@ -115,25 +116,46 @@ export class Syncer implements SingletonService {
...
-          e,
+      const tokenOptions = tokenConfigOptions(this.#config.auth);
+
+      const hasPushOrMutate =
+        this.#config?.push?.url !== undefined ||
+        this.#config?.mutate?.url !== undefined;
+      const hasGetQueries = this.#config?.getQueries?.url !== undefined;
+
+      // must either have one of the token options set or have custom mutations & queries enabled
+      assert(
```
aboodman: It was already like this in prev code but in the future don't use assertions for this kind of user-level error. use normal `throw new Error...`. `assert` is for checking invariants - a failed assertion means that a Rocicorp dev screwed up, not the user.

aboodman: cc @tantaman

PR #4817, opened 2025-08-22
fix(zero-server): clean up adapters for drizzle and postgres

0xcadams:
Adds a few missing adapters for common use-cases, for both `pg` and `postgres`, and `drizzle`.

```tsx
import {
  zeroDrizzle,
  type DrizzleTransaction,
} from '@rocicorp/zero/server/adapters/drizzle';

import { drizzle } from 'drizzle-orm/node-postgres';
const drizzleDb = drizzle(pool, { schema: drizzleSchema });

// api.ts
const zql = zeroDrizzle(schema, drizzleDb);

// server-mutators.ts
// Use a type helper for the server transaction type
import type { ServerTransaction } from '@rocicorp/zero';

type ServerTx = ServerTransaction<
  Schema,
  DrizzleTransaction<typeof drizzleDb>
>;

async function createUser(
  tx: ServerTx,
  { id, name }: { id: string; name: string },
) {
  // this is then fully typed w/ drizzle
  await tx.dbTransaction.wrappedTransaction
    .insert(drizzleSchema.user)
    .values({ id, name });
}
```


0xcadams:
@aboodman this adds the `pg` adapter mentioned in the quickstart.

packages/zero-server/src/adapters/drizzle-pg.ts
```diff
@@ -1,74 +1,152 @@
...
+import type {Schema} from '../../../zero-schema/src/builder/schema-builder.ts';
 
-type DrizzleConnection<T extends Record<string, unknown>> =
-  NodePgDatabase<T> & {
+/**
+ * Helper type for the wrapped transaction used by drizzle-orm/node-postgres.
+ *
+ * @remarks Use with `ServerTransaction` as `ServerTransaction<Schema, NodePgDrizzleTransaction>`.
+ */
+export type NodePgDrizzleTransaction<
```
aboodman: It seems wrong to me that there is both a PG Drizzle adapter and also a Postgres Drizzle adapter. We should not have a combinatorial explosion here, but instead the postgres and drizzle adapters should be composable.I think maybe the Drizzle adapter should be able to take any underlying db adapter and expose it or something.

0xcadams: Yeah I agree with you. I took a stab at this and the types get very messy trying to combine them all into one Drizzle wrapper. It got so convoluted that I reverted it. But I changed the Drizzle adapters reuse more code from the base adapters.

FWIW Drizzle doesn't support many more PG adapters outside of these two. And most of this PR is just typing helpers so they get a nice `wrappedTransaction` type.

tantaman: Does Drizzle expose a higher level abstraction for running SQL that doesn't require hooking into the underlying pg/postgres?

my 2 cents:
1. we should support `pg` and `postgres` for drizzle only
2. if someone has something more exotic they use with drizzle, they can write the adapter themselves at that point

0xcadams: Thanks for this - I just reworked to merge the two drizzle adapters into one. So now any drizzle client can be used with:

```tsx
const drizzleDb = drizzle(pool, { schema: drizzleSchema });
const zql = zeroDrizzle(schema, drizzleDb);
```

This also supports:

```tsx
import { type DrizzleTransaction } from '@rocicorp/zero/server/adapters/drizzle';
import type { ServerTransaction } from '@rocicorp/zero';

type ServerTx = ServerTransaction<
  Schema,
  DrizzleTransaction<typeof drizzleDb>
>;

// then
await tx.dbTransaction.wrappedTransaction
    .insert(drizzleSchema.user)
    .values({ id, name });
```

packages/zero/package.json
```diff
@@ -114,12 +114,16 @@
       "default": "./out/zero/src/server.js"
     },
     "./server/adapters/postgresjs": {
-      "types": "./out/zero-server/src/adapters/postgres.d.ts",
+      "types": "./out/zero-server/src/adapters/postgresjs.d.ts",
       "default": "./out/zero/src/adapters/postgresjs.js"
     },
-    "./server/adapters/drizzle-pg": {
```
0xcadams, line 119: This is technically a breaking change, but I don't know how many people are using this entrypoint.

packages/zero-server/src/adapters/drizzle.ts
```diff
@@ -0,0 +1,195 @@
...
+    const stmt = fromDollarParams(sql, params);
+    const result = await this.wrappedTransaction.execute(stmt);
+    return toIterableRows(result);
+  }
+}
+
+/**
+ * Turn `$1, $2...` placeholders into a Drizzle SQL object with bound params.
+ */
+export function fromDollarParams(text: string, params: unknown[]): SQL {
```
tantaman, line 103: scary but seems reasonable. We're only ever sending keywords and identifiers so there _shouldn't_ ever be a $ coming in here except as a parameter.

tantaman (approved):

packages/zero-server/src/adapters/drizzle.ts
0xcadams: I know, I didn't love this approach either. But there's no way to do it in Drizzle, so it's either this, or multiple drizzle adapters.

austinm911:
i'm seeing some type errors @0xcadams 

`The inferred type of 'zql' cannot be named without a reference to '.bun/drizzle-orm@0.44.6/node_modules/drizzle-orm/pg-core'. This is likely not portable. A type annotation is necessary.ts(2742)`

Getting this on the client arg of `zeroDrizzle` and the generic of `DrizzleTransaction`
```ts 
 The types of '_.session' are incompatible between these types.
      Type 'PgSession<NodePgQueryResultHKT, typeof import("/Users/am/Coding/2025/repo/packages/core/src/db/schema/index"), ExtractTablesWithRelations<typeof import("/Users/am/Coding/2025/repo/packages/core/src/db/schema/index")>>' is not assignable to type 'PgSession<PgQueryResultHKT, Record<string, unknown>, ExtractTablesWithRelations<Record<string, unknown>>>'.
        Property 'dialect' is protected but type 'PgSession<TQueryResult, TFullSchema, TSchema>' is not a class derived from 'PgSession<TQueryResult, TFullSchema, TSchema>'.ts(2379)
```

image

0xcadams:
> i'm seeing some type errors @0xcadams
> 
> `The inferred type of 'zql' cannot be named without a reference to '.bun/drizzle-orm@0.44.6/node_modules/drizzle-orm/pg-core'. This is likely not portable. A type annotation is necessary.ts(2742)`
> 
> Getting this on the client arg of `zeroDrizzle` and the generic of `DrizzleTransaction`
> 
> ```ts
>  The types of '_.session' are incompatible between these types.
>       Type 'PgSession<NodePgQueryResultHKT, typeof import("/Users/am/Coding/2025/repo/packages/core/src/db/schema/index"), ExtractTablesWithRelations<typeof import("/Users/am/Coding/2025/repo/packages/core/src/db/schema/index")>>' is not assignable to type 'PgSession<PgQueryResultHKT, Record<string, unknown>, ExtractTablesWithRelations<Record<string, unknown>>>'.
>         Property 'dialect' is protected but type 'PgSession<TQueryResult, TFullSchema, TSchema>' is not a class derived from 'PgSession<TQueryResult, TFullSchema, TSchema>'.ts(2379)
> ```
> 
> image

Thanks for the report - I will take a look.

PR #4823, opened 2025-08-26
fix: changed the types for mutators to be discrete

0xcadams:
This changes the type helpers with Zero to not union both CRUD and custom mutators.

This encourages users to remove CRUD mutators if they have opted into custom mutators.

### DIFF
--- a/packages/zero-client/src/client/zero.test.ts
+++ b/packages/zero-client/src/client/zero.test.ts
@@ -1,5 +1,13 @@
 import {resolver} from '@rocicorp/resolver';
-import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest';
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  expectTypeOf,
+  test,
+  vi,
+} from 'vitest';
 import {setDeletedClients} from '../../../replicache/src/deleted-clients.ts';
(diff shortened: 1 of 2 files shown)


aboodman:
The current intended behavior is that you can use both crud and custom
mutators together. I think this changes that behavior.

a (phone)

0xcadams:
Understood - didn't realize we allowed this. I thought it was recommended to fully migrate.

aboodman:
It's recommended but we do allow a mix right now.

If you set the new `allowLegacyMutators: false` it does the same thing as
this proposed PR. And in a future zero version we will default
allowLegacyMutators to false.

0xcadams:
Yeah - maybe this should be a type fix for removing autocomplete when `allowLegacyMutators` is false?

Screenshot 2025-08-26 at 12 45 47 PM


Screenshot 2025-08-26 at 12 45 23 PM

aboodman:
argh, i assumed that was how it already worked.

it will throw a runtime error currently, but the type hints would be nice
to get rid of.

we will need an extra type param to control this on zero which will means
that useZero will also need it :-/.

maybe the two enable flags should move into `schema` so that they their
types travel naturally through to useZero and there is not opp for the type
and value to disagree...

aboodman:
I'm asking claude to do this. cross your fingers

PR #4825, opened 2025-08-27
feat(zero-client): Use admin password for inspector

arv:
Add support for admin password authentication in the inspector.

The server can now respond to any inspect rpc with `{op: 'authorized', value: false}`. If
the client receives this, it will prompt the user for the admin password
and send a `{op: 'authenticate', value: 'password'}` rpc request. When a successful auth response is
received, the client then retries the original inspect rpc.

On the server side, the `InspectorDelegate` keeps tracks of authenticated
client groups by keeping a set in memory.


aboodman:
We need to default the admin password to the upstream db password, or else the net effect is that you have to remember to setup an admin password before you can use this feature. The feature is super useful for debugging user problems so we need a default.

aboodman:
But I see your concern with using the Postgres password as the default. It feels icky. 

How about this as an alternative? If there is no admin password, at startup of zero-cache we generate one (a random strong password) and print it out to the logs. That way, if the user doesn't set an admin password, which will be common, and they want to use this feature, they can always go to the logs and find the current password. 

I think that this password that we generate should be the *admin* password, not the inspector password, meaning that it should work for any feature that's blocked by the admin password, such as the stats endpoint.

PR #4826, opened 2025-08-27
feat(zero-react): support for suspense via useSuspenseQuery

grgbkr:
`useSuspenseQuery` takes the same options as useQuery, plus a new `'suspendUntil'` option.

```
  /**
   * Whether to suspend until:
   * - 'non-empty': the query has non-empty results (non-empty array or defined
   *   value for singular results) which may be of result type 'unknown'
   *   or the query result type is 'complete' (in which case results may be
   *   empty).  This is useful for suspending until there are non-empty
   *   optimistic local results, or the query has completed loading from the
   * server.
   * - 'complete': the query result type is 'complete'.
   *
   * Default is 'non-empty'.
   */
 ```
 
Based on @lewisl9029's contribution https://github.com/rocicorp/mono/pull/4747

### DIFF
--- a/packages/zero-react/src/mod.ts
+++ b/packages/zero-react/src/mod.ts
@@ -11,6 +11,7 @@ export type {ResultType} from '../../zql/src/query/typed-view.ts';
 export {ZeroInspector} from './components/zero-inspector.tsx';
 export {
   useQuery,
+  useSuspenseQuery,
   type QueryResult,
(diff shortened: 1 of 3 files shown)


grgbkr:
@aboodman I'd appreciate feedback on the option name and documentation.  It kind of bothers me that with 'non-empty', the results may actually be empty (if the result type is complete).

PR #4830, opened 2025-08-28
nwip: I am not sure how to use suspense

aboodman:
Hey @lewisl9029 and @grgbkr I was playing with the new suspense support and it was not as glorious as I hoped. Maybe I just don't get the glory of Suspense though.

Here is how I had hoped to use it. Currently in zbugs, I have [manual logic](https://github.com/rocicorp/mono/pull/4830/files#diff-fe44d0a1725a1ac25dfedaf82dd420db12c23ecca73c7cb529c69341a8b51d06L83) to delay displaying any UI until the main content area query has completed. Then the entire UI comes up together. I had hoped suspense support would automate this in a cleaner way.

But it's surprisingly subtle!

Simply changing the content area queries to useSuspenseQuery and putting a `<Suspense>` around the content area doesn't do the trick. Because whenever a new query starts in the content area the entire page goes blank again, which is not what I want.

For example, if I let the app load, then type an obscure query in the search box, it causes a query to go to the server which causes a suspend. Because of suspense, the entire page goes blank again and flickers a loading state. But that's not what I want.

I only want the entire page to  go blank for the very *first* load. After that I don't actually want to use suspense at all. I just want the specific parts of the app waiting for load to go blank, like they do today. So like when you do a search in zbugs that cause a server load, the result table and result count UI should go blank but nothing else.

I realized don't actually know how to achieve this gracefully w/ suspense. I can blank the entire UI on first load by [conditionally wrapping](https://github.com/rocicorp/mono/pull/4830/files#diff-0efd385fff79d660eaddb6685e77217c217225a6c7c8322376ec5827fe02be6dR27) the entire content area in <Suspense>. But that means in the case when I don't wrap the content area I still need a suspense boundary somewhere or else I'll get errors.

My next thought was to wrap `<Suspense>` around, e.g., `<ListPage>`. But this doesn't work because `<ListPage>` also includes the search UI which shouldn't be disabled on second load. And I can't factor out the data table into a separate component that has a suspense around it because there is the matter of result count which uses the query result and is displayed outside the data table.

I [find myself wondering](https://github.com/rocicorp/mono/pull/4830/files#diff-9597c0281d1a51dac42b6c50666d911bc9fbdea3ae9ea97008a054f98926febcR88) what the purpose of this machinery is when what I had worked fine and was like 10 lines of code. The suspense code is much bigger and more complex and doesn't work yet!

In summary:

image

... or am I misunderstanding how to use Suspense?

### DIFF
--- a/apps/zbugs/src/components/maybe-suspend.tsx
+++ b/apps/zbugs/src/components/maybe-suspend.tsx
@@ -1,0 +1,26 @@
+import {Suspense} from 'react';
+
+export function MaybeSuspend({
+  children,
+  enabled,
+  onReveal,
+}: {
+  children: React.ReactNode;
+  enabled: boolean;
+  onReveal?: (() => void) | undefined;
+}) {
+  if (enabled) {
+    return (
+      <Suspense>
+        {children}
+        <Reveal on={onReveal} />
+      </Suspense>
+    );
+  }
+  return <>{children}</>;
+}
+
+function Reveal({on}: {on?: (() => void) | undefined}) {
+  on?.();
+  return null;
+}
--- a/apps/zbugs/src/pages/issue/issue-page.tsx
+++ b/apps/zbugs/src/pages/issue/issue-page.tsx
@@ -1,4 +1,4 @@
-import {useQuery} from '@rocicorp/zero/react';
+import {useQuery, useSuspenseQuery} from '@rocicorp/zero/react';
 import {useWindowVirtualizer, Virtualizer} from '@tanstack/react-virtual';
 import {nanoid} from 'nanoid';
 import {
@@ -80,7 +80,7 @@ function softNavigate(path: string, state?: ZbugsHistoryState) {
 
 const emojiToastShowDuration = 3_000;
 
-export function IssuePage({onReady}: {onReady: () => void}) {
+export function IssuePage() {
   const z = useZero();
   const params = useParams();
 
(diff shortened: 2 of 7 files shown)


grgbkr:
@aboodman Yeah so Suspense is a little bit complicated, and is particularly rough in React 18.  You have to use startTransition and useDeferredValue explicitly a lot to achieve what you are looking for https://18.react.dev/reference/react/Suspense#preventing-already-revealed-content-from-hiding.  The router we use in zbugs (wouter) is minimalist and does not have built in Suspense support (like auto wrapping navigations in transitions).

With React 19 they introduced [React.use](https://react.dev/reference/react/use) which avoids the need for startTransition in a lot of scenarios (in React 18 instead of React.use you threw promises to suspend).  We should update zbugs to React 19.  I think it will largely work as you want out of the box.

lewisl9029:
From your description, it sounds like the biggest issue here is search queries suspending the whole page, and that can be solved with useDeferredValue like @grgbkr mentioned: https://18.react.dev/reference/react/Suspense#showing-stale-content-while-fresh-content-is-loading

An auto transition wrapping router is definitely helpful if you want the kind of rails-like loading behavior where the old route continues to be displayed while the new route loads (although with suspense you can pretty easily offer an even cooler experience where you can show a loading indicator on the link itself instead of or in addition to a page level loading bar like rails sites tend to do),  but isn't necessary if you're ok with page navigations suspending immediately and rendering your loading fallbacks. 

In my experience with zero so far, suspends on page navigations are rare enough that either approach would probably be fine, but I do slightly prefer the rails like approach because users have something to look at and can preempt at any point during route transitions.

PR #4831, opened 2025-08-28
feat(zero-react): Rename 'non-empty' to 'partial'.

aboodman:
This is what ChatGPT autocompleted for the API and I think it is nicer.


grgbkr:
LGTM.  I like it.

PR #4833, opened 2025-08-28
Aa/solid use query

aboodman:


packages/zero-solid/package.json
```diff
@@ -20,8 +20,9 @@
     "@rocicorp/eslint-config": "^0.7.0",
     "@rocicorp/prettier-config": "^0.3.0",
     "@solidjs/testing-library": "^0.8.10",
-    "vite-plugin-solid": "^2.11.7",
-    "solid-js": "^1.9.4"
+    "jsdom": "^26.1.0",
```
arv, line 23: No jsdom. We use real browsers

aboodman:
This one wasn’t meant for review. It was a repro for a bug.

a (phone)

PR #4837, opened 2025-08-29
feat(replicache): Add support for React Native using Expo SQLite

arv:
This is a minimal change to Replicache to ensure less disruptions.

To use you have to get the `expoSQLiteStoreProvider` from the `replicache/expo` package.

Usage:

```js
import {Replicache} from 'replicache';
import {expoSQLiteStoreProvider} from 'replicache/expo';

const rep = new Replicache({
  name: 'user',
});
```


apps/zql-viz/package.json
```diff
@@ -5,7 +5,9 @@
   "type": "module",
   "scripts": {
     "dev": "npx api-extractor run --local && vite",
-    "preview": "vite preview"
+    "build": "npx api-extractor run --local && vite build",
```
aboodman, line 8: Did  you mean to re-enable this as part of the PR? It was disabled by cb7eb694be4c1350d62e4bbfe6d4ca1f0d26a655 because the canary script doesn't work with it.

arv: I wasn't aware that it was intentionally removed.

I thought it was bad that it wasn't being built. I wanted build failures to cause errors on the CI.

PR #4843, opened 2025-09-02
chore: Change bencher perf to 70%

arv:
This is to reduce alerts


arv:
This number is actually how confident "we" are that it is a regression. Closer to 1 means less warnings.

I changed this to 0.9 in #4844

PR #4846, opened 2025-09-02
feat(zero-cache): Generate admin password if not provided

arv:
In case adminPassword is not set we generate a cryptographically secure random password and log it.


arv:
I'm reverting this while we ponder on the right decision.

PR #4850, opened 2025-09-03
feat(replicache): Export sqlite kv store symbols

arv:
This allows creating a KV Store provider for different SQLite implementations. For example using `op-sqlite` in React Native.:

```ts
import { open, type Scalar } from "@op-engineering/op-sqlite";
import {
  SQLiteDatabaseManager,
  SQLiteStore,
  type KVStore,
  type KVStoreProvider,
} from "replicache/sqlite";

const dbManager = new SQLiteDatabaseManager({
  open(fileName: string) {
    const database = open({ name: fileName });
    return {
      close() {
        database.close();
      },
      destroy() {
        database.delete();
      },
      prepare(sql: string) {
        // OP-SQLite does not support sync execute on prepared statements.

        return {
          run(...params: unknown[]) {
            database.executeSync(sql, params as Scalar[]);
          },
          all<T>(...params: unknown[]): T[] {
            const result = database.executeSync(sql, params as Scalar[]);
            return result.rows as T[];
          },
          finalize() {},
        };
      },
    };
  },
});

export const provider: KVStoreProvider = {
  create(name): KVStore {
    return new SQLiteStore(name, dbManager, {
      readPoolSize: 2,
      busyTimeout: 5000,
      journalMode: "WAL",
      synchronous: "NORMAL",
    });
  },
  async drop(name): Promise<void> {
    dbManager.destroy(name);
  },
};
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,7 @@ Chinook_PostgreSql.sql
 WARP.md
 *.db-shm
 *.db-wal
+
+# We output files outside of out to make old React Native / Metro happy
+packages/replicache/expo
+packages/replicache/sqlite
--- a/packages/replicache/package.json
+++ b/packages/replicache/package.json
@@ -77,11 +77,17 @@
     },
     "./expo": {
       "types": "./out/replicache/src/expo.d.ts",
-      "default": "./out/expo.js"
+      "default": "./expo/index.js"
+    },
+    "./sqlite": {
+      "types": "./out/replicache/src/sqlite.d.ts",
+      "default": "./sqlite/index.js"
     }
   },
   "files": [
     "out",
+    "expo",
+    "sqlite",
     "!*.tsbuildinfo"
   ],
   "eslintConfig": {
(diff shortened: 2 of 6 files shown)


arv:
> OP-SQLite does not support sync execute on prepared statements.

This makes me wonder if we should have an async interface for op-sqlite. If so we would add an `AsyncSQLiteStore` which uses async open/delete/prepare/execute etc.

aboodman:
I think maybe the SQLite store should just be async. I don't think we win a huge amt by making it sync? Zero already does a lot of work batch writes.

aboodman:
And replicache lower layers all already async anyway so it would fit naturally.

aboodman:
I think the sync apis are nice when you're using sqlite super granularly, but we don't use it that way.

arv:
I agree.

https://www.powersync.com/blog/react-native-database-performance-comparison#:~:text=Only%20asynchronous/non%2Dblocking%20APIs%20are%20used.%20We%20think%20it%27s%20preferable%20to%20stay%20away%20from%20synchronous%20calls%20altogether%20because%20they%20have%20the%20potential%20to%20stutter%20or%20block%20the%20UI.

I will remove the sync apis

PR #4851, opened 2025-09-03
fix(expo-sqlite): do not close a db we're trying to open

tantaman:
The `SQLiteDatabaseManager` has a map of instances that it keys by database name. When someone tells the SQLiteDBManager to close a db by name, it looks up all connections that exist for that DB and closes them.

This was a problem when one Zero instance was being destroyed and another was being created. Both instances use the `replicache_dbs_v0` database so we could get into cases where a closing Zero would nuke the connections of an opening Zero. 

**Example:**

Zero Instance 1 would be closing.
Zero Instance 2 would be opening.

(note: zero.close is async but it is not / can not be awaited in useEffect in ZeroProvider)

The `replicache_dbs_v0` would be opened by Zero Instance 2. The close from Zero Instance 1 would come in.
The SQLiteDBManager would close all connections associated with `replicache_dbs_v0` which included Zero Instance 2's connections.

We would then fail when making any call to SQLite on that connection.

```
🟠 FunctionCallException: Calling the 'runSync' function has failed (at ExpoModulesCore/SyncFunctionDefinition.swift:137)
🟠 → Caused by: AccessClosedResourceException: Access to closed resource (at ExpoSQLite/SQLiteModule.swift:625)
```

This error does not show in the Expo logs but does show when:

a. running the app in xcode or
b. opening `Console.app` and selecting your simulator

CleanShot 2025-09-03 at 13 07 11

---

The fix is to not have a single `SQLiteDBManager` but to create one for each Store. When the store is closed it then only closes its own connections to SQLite rather than the connections of other stores.

I tested this using the zslack app. Previously, after logging in w/ WAL mode and restarting the app, my channel list would be empty. Now it all works fine.

CleanShot 2025-09-03 at 13 09 15

CleanShot 2025-09-03 at 13 09 26

https://rocicorp.slack.com/archives/C013XFG80JC/p1756914030938379

### DIFF
--- a/packages/replicache/src/expo/store.ts
+++ b/packages/replicache/src/expo/store.ts
@@ -4,65 +4,68 @@ import {
   type SQLiteBindParams,
 } from 'expo-sqlite';
 import {
+  safeFilename,
   SQLiteDatabaseManager,
   SQLiteStore,
   type SQLiteDatabaseManagerOptions,
 } from '../kv/sqlite-store.ts';
 import type {StoreProvider} from '../kv/store.ts';
 
-export const expoDbManagerInstance = new SQLiteDatabaseManager({
-  open: fileName => {
-    const db = openDatabaseSync(fileName);
-    let closed = false;
-
-    const close = () => {
-      if (!closed) {
-        db.closeSync();
-        closed = true;
-      }
-    };
-
-    return {
-      close,
-      destroy() {
-        close();
-        deleteDatabaseSync(fileName);
-      },
-      prepare(sql: string) {
-        const stmt = db.prepareSync(sql);
-        return {
-          run: (...params: unknown[]): void => {
-            stmt.executeSync(params as SQLiteBindParams);
-          },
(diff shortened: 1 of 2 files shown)


packages/replicache/src/expo/store.ts
```diff
@@ -4,65 +4,68 @@ import {
...
-        // we override the journal mode to undefined because
-        // setting it to WAL causes hanging COMMITs on Expo
-        journalMode: undefined,
-      }),
+      });
+    },
+
     drop: (name: string) => {
-      expoDbManagerInstance.destroy(name);
+      // Note that we cannot drop a database if it is open.
```
tantaman, line 66: this seems like a reasonable invariant to me and I actually prefer it to the old model. In that it prevents someone from mistakenly dropping a DB that is being used.

Although I'm not sure of all the ways Replicache internals may call this yet.

aboodman:
yay, makes sense. Thanks Matt.

PR #4852, opened 2025-09-04
fix(zero-cache): include and validate the client's primaryKey

darkgnotic:
Include the `primaryKey` from the client's schema in the `ClientSchema` message validated by the zero-cache (and used as part of the IDB hash).

The zero-cache will validate that the `primaryKey` corresponds to an upstream index over non-null columns (order agnostic).

Note that in order to be backwards compatible with old client versions, the `primaryKey` field is optional, and can be made non-optional when bumping the min-supported protocol version to 30+

### DIFF
--- a/packages/zero-protocol/src/client-schema.ts
+++ b/packages/zero-protocol/src/client-schema.ts
@@ -1,3 +1,4 @@
+import {must} from '../../shared/src/must.ts';
 import {mapAllEntries} from '../../shared/src/objects.ts';
 import * as v from '../../shared/src/valita.ts';
 
@@ -19,6 +20,9 @@ export type ColumnSchema = v.Infer<typeof columnSchemaSchema>;
 
 export const tableSchemaSchema = v.object({
   columns: v.record(columnSchemaSchema),
+  // TODO: Make this non-optional when bumping the
+  //       MIN_SERVER_SUPPORTED_SYNC_PROTOCOL to 30+.
+  primaryKey: v.array(v.string()).optional(),
 });
 
 export type TableSchema = v.Infer<typeof tableSchemaSchema>;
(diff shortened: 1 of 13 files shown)


packages/zero-protocol/src/client-schema.ts
```diff
@@ -39,12 +43,16 @@ const keyCmp = ([a]: [a: string, _: unknown], [b]: [b: string, _: unknown]) =>
...
-        .sort(keyCmp)
-        .map(([name, table]) => [
-          name,
-          {columns: mapAllEntries(table.columns, e => e.sort(keyCmp))},
-        ]),
+      tables.sort(keyCmp).map(([name, table]) => [
+        name,
+        {
+          columns: mapAllEntries(table.columns, e => e.sort(keyCmp)),
+          primaryKey: must(
```
grgbkr, line 50: I guess this is safe because its only called from the client code (the server still has to deal with this being possibly undefined).  Maybe we should move this into the zero-client package.

packages/zero-schema/src/builder/schema-builder.ts
```diff
@@ -143,16 +143,20 @@ export function clientSchemaFrom(schema: Schema): {
...
+    tables: mapEntries(
+      schema.tables,
+      (name, {serverName, columns, primaryKey}) => [
+        serverName ?? name,
+        {
+          columns: mapEntries(columns, (name, {serverName, type}) => [
+            serverName ?? name,
+            {type},
+          ]),
+          primaryKey: [...primaryKey],
```
grgbkr, line 155: map to server name?

grgbkr (commented):
LGTM

packages/zero-schema/src/builder/schema-builder.ts
darkgnotic: Thank you.

https://github.com/rocicorp/mono/pull/4857

PR #4853, opened 2025-09-04
feat(zero): Allow customization of mutate and get-queries URLs on the client

aboodman:
This PR allows users to customize the `mutate` and `get-queries` URLs on the client.

This functionality was always envisioned but it was broken in the current build and therefore not documented. 

The `mutate` version just wasn't wired up. But the `get-queries` version had a more fundamental problem: `get-queries` happens on behalf of a client-group, not a client. But we are configuring a client with the constructor param. So it was possible for `zero-cache` to get into a situation where it's doing a GET query fetch, but there are multiple clients having different values for the GET query customization URL. 

To fix this on the client, we incorporated the mutate URL, and the get-queries URL into the IDB name. The result is that changing these URLs gives a zero client a different client group ID. 

On the server, we simplified the communication and storage of these two URLs to be a single value in both the transform service, the pusher, and the view syncer because these services are all one-to-one with client groups. 

We also simplified the API and removed the ability to customize the query strings because by exposing the ability to customize the entire URL, we no longer need that and we removed all the supporting code that supported customizing the query strings. 

New unit tests were added where necessary, but I also tested manually that the functionality works in zbugs. 

Please review this PR carefully as it's a significant change and was implemented with the help of Claude Code.

### DIFF
--- a/packages/zero-cache/src/custom/fetch.ts
+++ b/packages/zero-cache/src/custom/fetch.ts
@@ -18,13 +18,11 @@ export async function fetchFromAPIServer(
   allowedUrls: string[],
   shard: ShardID,
   headerOptions: HeaderOptions,
-  queryParams: Record<string, string> | undefined,
   body: ReadonlyJSONValue,
 ) {
   lc.info?.('fetchFromAPIServer called', {
     url,
     allowedUrls,
-    queryParams,
   });
   
   if (!urlMatch(url, allowedUrls)) {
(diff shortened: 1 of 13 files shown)


packages/zero-cache/src/custom/fetch.ts
```diff
@@ -12,13 +13,18 @@ export type HeaderOptions = {
...
 export async function fetchFromAPIServer(
+  lc: LogContext,
   url: string,
   allowedUrls: string[],
   shard: ShardID,
   headerOptions: HeaderOptions,
-  queryParams: Record<string, string> | undefined,
   body: ReadonlyJSONValue,
 ) {
+  lc.info?.('fetchFromAPIServer called', {
```
aboodman, line 23: Added this because I feel it will be a very useful thing to see in the logs.

PR #4854, opened 2025-09-04
Abort watchdog timer when ConnectionLoop is closed

dpeek:
I'm using zero client from the command line as a simple, typed CRUD client.

When running in Bun the process currently waits for ~60s after calling `client.close()` before actually exiting, because the watchdog timer in `ConnectionLoop` never gets cancelled.

This PR adds an AbortController and aborts the watchdog `sleep` Promise on `ConnectionLoop.close()`


aboodman:
lgtm thank you for the contribution

PR #4855, opened 2025-09-04
perf(zql): implement FlippedJoin

grgbkr:
FlippedJoin implements an *inner* join which fetches nodes from its child input first and then fetches
their related nodes from its parent input.  Output nodes are the nodes from parent input (in parent input order), which have at least one related child.  These output nodes have a new relationship added to them,
which has the name `relationshipName`. The value of the relationship is a stream of related nodes from the child input (in child input order).

### DIFF
--- a/packages/zql/src/ivm/flipped-join.ts
+++ b/packages/zql/src/ivm/flipped-join.ts
@@ -106,6 +106,10 @@ export class FlippedJoin implements Input {
     return this.#schema;
   }
 
+  // TODO: When parentKey is the parent's primary key (or more
+  // generally when the parent cardinality is expected to be small) a different
+  // algorithm should be used:  For each child node, fetch all parent nodes
+  // eagerly and then sort using quicksort.
   *fetch(req: FetchRequest): Stream<Node> {
     const childNodes = [...this.#child.fetch({})];
     // FlippedJoin's split-push change overlay logic is largely
@@ -128,6 +132,8 @@ export class FlippedJoin implements Input {
     let threw = false;
     try {
       for (const childNode of childNodes) {
+        // TODO: consider adding the ability to pass a set of
+        // ids to fetch, and have them applied to sqlite using IN.
         // TODO: Can there be a conflict between req constraint and
         // the join constraint?
         const stream = this.#parent.fetch({


packages/zql/src/ivm/flipped-join.ts
```diff
@@ -0,0 +1,440 @@
...
+  type Input,
+  type Output,
+} from './operator.ts';
+import type {SourceSchema} from './schema.ts';
+import {first, type Stream} from './stream.ts';
+
+type Args = {
+  parent: Input;
+  child: Input;
+  // The order of the keys does not have to match but the length must match.
```
tantaman, line 24: "The order of the keys does not have to match"

This comment is strange. There is an order since values in corresponding slots are compared.

packages/zql/src/ivm/flipped-join.ts
```diff
@@ -0,0 +1,440 @@
...
+            type: 'edit',
+            oldNode: flip(change.oldNode),
+            node: flip(change.node),
+          });
+          break;
+        }
+        if (oldHasChild) {
+          this.#output.push({
+            type: 'remove',
+            node: flip(change.node),
```
tantaman, line 425: shouldn't we remove the `change.oldNode`?

packages/zql/src/ivm/flipped-join.ts
```diff
@@ -0,0 +1,440 @@
...
+              this.#parent
+                .getSchema()
+                .compareRows(
+                  minParentNode.row,
+                  this.#inprogressChildChange.position,
+                ) <= 0
+            ) {
+              overlaidRelatedChildNodes = relatedChildNodes.filter(
+                n => n !== this.#inprogressChildChange?.change.node,
+              );
```
tantaman, line 218: I'm a bit confused on how this relates to ` childNodes.splice(insertPos, 0, removedNode);` on line 129. Doesn't line 129 mean that `inProgressChildChange.change.node` is never in `relatedChildNodes`?

packages/zql/src/ivm/flipped-join.ts
```diff
@@ -0,0 +1,440 @@
...
+    // because if the removed node was the last related child, the
+    // related parents with position greater than change.position
+    // (which should not yet have the node removed), would not even
+    // be fetched here, and would be absent from the output all together.
+    if (this.#inprogressChildChange?.change.type === 'remove') {
+      const removedNode = this.#inprogressChildChange.change.node;
+      const compare = this.#child.getSchema().compareRows;
+      const insertPos = binarySearch(childNodes.length, i =>
+        compare(childNodes[i].row, removedNode.row),
+      );
```
tantaman, line 128: `binarySearch` looks like it expects `needle - haystack[i]` not `haystack[i] - needle`

https://github.com/rocicorp/mono/blob/345895c186e1f0eaa8b4a5f2d4b7e5e678f592e3/packages/shared/src/binary-search.test.ts#L8


```suggestion
      const insertPos = binarySearch(childNodes.length, i =>
        compare(removedNode.row, childNodes[i].row),
      );
```

packages/zql/src/ivm/flipped-join.ts
```diff
@@ -0,0 +1,440 @@
...
+          )
+        ) {
+          if (this.#inprogressChildChange.change.type === 'remove') {
+            if (
+              this.#parent
+                .getSchema()
+                .compareRows(
+                  minParentNode.row,
+                  this.#inprogressChildChange.position,
+                ) <= 0
```
tantaman, line 214: do we need to take into consideration `req.rev`?

packages/zql/src/ivm/flipped-join.ts
```diff
@@ -0,0 +1,440 @@
...
+                n => n !== this.#inprogressChildChange?.change.node,
+              );
+            }
+          } else if (
+            this.#parent
+              .getSchema()
+              .compareRows(
+                minParentNode.row,
+                this.#inprogressChildChange.position,
+              ) > 0
```
tantaman, line 226: same question about req.rev

packages/zql/src/ivm/flipped-join.ts
grgbkr: copy-pasta from join, agree, removed

packages/zql/src/ivm/flipped-join.ts
grgbkr: good catch, ill need to figure out a test to cover this

packages/zql/src/ivm/flipped-join.ts
grgbkr: Good call.  There was a test covering this but i missed the bad ordering in the snapshot.

packages/zql/src/ivm/flipped-join.ts
grgbkr: Line 21 puts the removed childNode back in the childNodes, so it may end up in relatedChild nodes.  This line then re-removes it from any parents for which the remove child change has already been pushed.

packages/zql/src/ivm/flipped-join.ts
grgbkr: No, we don't need to.  We push the child changes in default parent fetch order. this.#inprogressChildChange.position is the last parent for which the in progress child change was pushed.  This is testing if the inprogress child change has already been pushed for minParentNode.

packages/zql/src/ivm/flipped-join.ts
grgbkr: No, we don't need to.  We push the child changes in default parent fetch order. this.#inprogressChildChange.position is the last parent for which the in progress child change was pushed.  This is testing if the inprogress child change has not yet been pushed for minParentNode.

packages/zql/src/ivm/flipped-join.ts
grgbkr: Thinking about this a bit more, this case can't happen.  The edit is not allowed to change the relationship, so either both old and new have children, or neither have children.  Simplified.

PR #4858, opened 2025-09-05
fix(zero-cache): do not throw error on request timeout for anonymous OTEL

cesara:


tantaman:
any idea why these timeouts are happening? Something expected like firewall rules or something unexpected, like a bug in our implementation?

cesara:
> any idea why these timeouts are happening? Something expected like firewall rules or something unexpected, like a bug in our implementation?

it looks like grafana service at times returns timeouts
```
Forwarding to Grafana...
Grafana response: 504
Grafana JSON response: {"status":"error","errorType":"timeout","error":"rpc error: code = DeadlineExceeded desc = context deadline exceeded"}
```

PR #4859, opened 2025-09-05
chore(zbugs): Remove inspector icon since we're not using it anymore.

aboodman:
We should probably remove the supporting code too?


arv:
LGTM

Remove this too /packages/zero-react/src/components/zero-inspector.tsx

PR #4860, opened 2025-09-05
feat(replicache)!: Remove generic SQLite store and migrate to Expo-specific implementation

arv:
This removes the generic SQLite store implementation and replaces it with a specialized Expo SQLite store using `expo-sqlite` directly. This change simplifies the architecture and improves performance for Expo-based applications.

The generic SQLite store had become overly complex with multiple layers of abstraction - connection managers, read/write pools, and generic database interfaces. The new Expo store directly uses `expo-sqlite` APIs (`openDatabaseSync`, `prepareSync`, etc.) for better performance and simpler code paths.

**Breaking Changes:**
- Removed `SQLiteStore` and related infrastructure from `packages/replicache/src/kv/`
- SQLite store factory operations are now async to properly support database initialization

**Changes:**
- **Removed:** Generic SQLite store implementation (~1,600 lines across multiple files)
- **Added:** Expo-specific SQLite store with comprehensive test suite (~1,000 lines)
- **Simplified:** Resource management with reference counting for shared store instances
- **Improved:** Transaction handling using RWLock without connection pooling complexity

The Expo wrapper now properly manages database lifecycle, ensuring we don't call close on already closed databases. Resource cleanup is handled centrally by the database manager rather than being scattered across connection managers.

Added comprehensive tests covering concurrent operations, resource management, large data handling, and proper cleanup scenarios.

### DIFF
--- a/packages/replicache/src/expo/store.ts
+++ b/packages/replicache/src/expo/store.ts
@@ -9,16 +9,16 @@ import type {ReadonlyJSONValue} from '../../../shared/src/json.ts';
 import {deepFreeze} from '../frozen-json.ts';
 import type {Read, Store, StoreProvider, Write} from '../kv/store.ts';
 
-function safeFilename(name: string): string {
-  return name.replace(/[^a-zA-Z0-9]/g, '_');
-}
-
 export type ExpoSQLiteStoreOptions = {
   busyTimeout?: number | undefined;
   journalMode?: 'WAL' | 'DELETE' | undefined;
   synchronous?: 'NORMAL' | 'FULL' | undefined;
   readUncommitted?: boolean | undefined;
 };
+
+function safeFilename(name: string): string {
+  return name.replace(/[^a-zA-Z0-9]/g, '_');
+}
 
 type Statements = {
   has: SQLiteStatement;
--- a/packages/zero-react-native/src/mod.ts
+++ b/packages/zero-react-native/src/mod.ts
@@ -1,2 +1,4 @@
-export {expoSQLiteStoreProvider} from '../../replicache/src/expo/store.ts';
-export type {SQLiteDatabaseManagerOptions} from '../../replicache/src/kv/sqlite-store.ts';
+export {
+  expoSQLiteStoreProvider,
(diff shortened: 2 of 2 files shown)


packages/replicache/src/expo/store.ts
```diff
@@ -1,73 +1,352 @@
...
+    // Start shared read transaction if this is the first reader
+    // This ensures consistent reads across all concurrent readers
+    if (this.#activeReaders === 0 && !this.#sharedReadTransaction) {
+      this.#db.execSync('BEGIN');
+      this.#sharedReadTransaction = true;
+    }
+    this.#activeReaders++;
+
+    return new ExpoSQLiteRead(() => {
+      this.#activeReaders--;
```
tantaman, line 173: clever

packages/replicache/src/expo/store.ts
```diff
@@ -1,73 +1,352 @@
...
+    if (this.#closed) {
+      return;
+    }
+
+    const writeRelease = await this.#rwLock.write();
+
+    if (this.#sharedReadTransaction) {
+      this.#db.execSync('COMMIT');
+      this.#sharedReadTransaction = false;
+    }
```
tantaman, line 208: I don't understand this. Why would `sharedReadTransaction` be set if we've waiting for the write lock? Wouldn't all readers have committed by then?

packages/replicache/src/expo/store.ts
```diff
@@ -1,73 +1,352 @@
...
+  async read(): Promise<Read> {
+    if (this.#closed) {
+      throw new Error('Store is closed');
+    }
+
+    const release = await this.#rwLock.read();
+
+    // Start shared read transaction if this is the first reader
+    // This ensures consistent reads across all concurrent readers
+    if (this.#activeReaders === 0 && !this.#sharedReadTransaction) {
```
tantaman, line 166: This all looks good until I think about "shared connections between multiple store instances with the same name"

Wouldn't we need to track `sharedReadTransaction` across all stores of the same name if they're all using the same db connection?

PR #4864, opened 2025-09-05
fix(zero): close connection on 401 from upstream

tantaman:
Turns out we started swallowing this exception at some point


arv:
LGTM

PR #4865, opened 2025-09-05
feat(zero-react): refetch

tantaman:
add `refetch` to allow users to retry a query once it has failed. There is no good way to retry a query otherwise as if two components have a query mounted, unmounting and re-mounting one component will not retry the query. All users of the query must be unmounted and re-mounted.

Note that if a 401 is thrown the re-auth flow will run and queries will be retried automatically. This is a bit wonky given the 401 will also flow to the react components. Another argument for making auth refresh explicitly in the user's control.

https://rocicorp.slack.com/archives/C013XFG80JC/p1757003614580649

https://github.com/user-attachments/assets/2462cb48-1224-4570-938e-ca402d9fdfbb

### DIFF
--- a/packages/zero-react/src/use-query.test.tsx
+++ b/packages/zero-react/src/use-query.test.tsx
@@ -1,4 +1,12 @@
-import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest';
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  test,
+  vi,
+  type Mock,
+} from 'vitest';
 import {Suspense} from 'react';
 import {createRoot, type Root} from 'react-dom/client';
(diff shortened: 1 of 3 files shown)


packages/zero-react/src/use-query.test.tsx
```diff
@@ -875,5 +884,236 @@ describe('useSuspenseQuery', () => {
...
+
+      root.render(
+        <ZeroProvider zero={zero}>
+          <Suspense fallback={<>loading</>}>
+            <Comp />
+          </Suspense>
+        </ZeroProvider>,
+      );
+
+      await expect.poll(() => element.textContent).toBe('loading');
```
arv, line 920: TIL

I've been using waitFor and waitUntil

PR #4866, opened 2025-09-05
chore(zero): do not apply `one` for singular relationships

tantaman:
This is a ~15% drag on perf (when joining 4 extra tables) to track take data for relationships that should never be more than 1.

If the user put in the wrong cardinality in their schema that is on them.


aboodman:
I agree w/ this from a product sense, but I think we did this because something bad happens when a `one()` relationship turns out to be wrong. I think `ArrayView` freaks out when it receives a second value for a relationship it expected to be `one()`.

Can you add a test for this case and confirm that a reasonable error messages is emitted that gives the user a clue of what went wrong?

tantaman:
ArrayView gives detailed errors about this these days:

https://github.com/rocicorp/mono/blob/76a32e48fe0828b335920436c2571336d87d5064/packages/zql/src/ivm/view-apply-change.ts#L134

existing test is here:

https://github.com/rocicorp/mono/blob/76a32e48fe0828b335920436c2571336d87d5064/packages/zql/src/ivm/array-view.test.ts#L157

aboodman:
Great

a (phone)

PR #4867, opened 2025-09-05
chore(zero-cache): OTEL logging config option

cesara:


packages/zero-cache/src/server/otel-start.ts
```diff
@@ -28,21 +28,15 @@ class OtelManager {
...
-        warn: (msg: string, ...args: unknown[]) => log.warn?.(msg, ...args),
-        error: (msg: string, ...args: unknown[]) => log.error?.(msg, ...args),
-      });
-    }
     if (this.#started || !otelEnabled()) {
       return;
     }
     this.#started = true;
 
+    // Store and temporarily remove OTEL_LOG_LEVEL to prevent NodeSDK from setting its own logger
```
darkgnotic, line 36: Nit: Maybe put this right around the `sdk.start()` call to be clearer about what it's working around, i.e.

```ts
    // Store and temporarily remove OTEL_LOG_LEVEL to prevent NodeSDK from setting its own logger
    const otelLogLevel = process.env.OTEL_LOG_LEVEL;
    delete process.env.OTEL_LOG_LEVEL;
    try {
      sdk.start();
    } finally {
      if (otelLogLevel) {
        process.env.OTEL_LOG_LEVEL = otelLogLevel;
      }
    }
    setupOtelDiagnosticLogger(lc, true);
```

PR #4870, opened 2025-09-06
feat(zero): Add support for StandardSchema to synced queries.

aboodman:
This was surprisingly difficult because StandardSchema supports async validation. This meant that SyncedQuery needed to get forked into ValidatedQuery, but fortunately that didn't spiral too far.

The other crazy unfortunate collision was that the async validation means that `withValidation` becomes async. But the return value `withValidation` is a query. And queries themselves are promise-like. This means that when code that calls `withValidation` uses await, it will unwrap recursively because that's what await does.

The result is that awaiting the result of `withValidation` in order to return the query over the http response actually tries to run the query.

The one lucky break was that `handleGetQueriesRequest` actually wants {query: Q}, not Q as an input. So I was able to leverage that and have `withValidation` return that shape not a raw query. This stops the recursive unwrapping of promises from happening. I think the previous API was actually better, but it's the best I could think of.

We probably need to remove the ability to run queries directly, as we've now added zero.run and friends. But we cannot do that until we add tx.run. That will likely have usability impacts on reading data inside transactions.

### DIFF
--- a/apps/zbugs/api/index.ts
+++ b/apps/zbugs/api/index.ts
@@ -213,7 +213,7 @@ async function getQueriesHandler(
   await withAuth(request, reply, async authData => {
     reply.send(
       await handleGetQueriesRequest(
-        (name, args) => ({query: getQuery(authData, name, args)}),
+        (name, args) => getQuery(authData, name, args),
         schema,
         request.body,
       ),
--- a/package-lock.json
+++ b/package-lock.json
@@ -78,6 +78,7 @@
         "remark-gfm": "^4.0.0",
         "use-debounce": "^10.0.4",
         "usehooks-ts": "^3.1.0",
+        "valibot": "^1.1.0",
         "wouter": "^3.7.1",
         "zod": "^3.21.4"
       },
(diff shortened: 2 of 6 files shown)


tantaman:
nit: this is a breaking change. so `feat(zero)!:`

Maybe complete the refactor so we do not need to do`{query:` first?

Related:
> Will tx in custom mutators return runnable queries or will it require the user to tx.run(query)? - https://github.com/rocicorp/mono/pull/4871#issuecomment-3266389599

aboodman:
Yeah my plan is to complete the refactor first.

arv:
https://standardschema.dev/#:~:text=How%20to%20only%20allow%20synchronous%20validation%3F

tantaman:
turns out we must allow `withValidation` to be async given the user will want to run DB queries on the server side impl of their synced query.

I forgot about this when adding all the new APIs. Before there wasn't any wrapping of the synced query so the user could create any function they wanted (sync or asnyc) on the backend to implement the synced query.

aboodman:
> https://standardschema.dev/#:~:text=How%20to%20only%20allow%20synchronous%20validation%3F

I actually tried this (hopefully - i hadn't seen this link) but it didn't seem to work. The schema validators I tried always seemed to return a promise.

PR #4871, opened 2025-09-06
refactor(zql): extract query execution into separate run.ts module

aboodman:
For discussion. At first I felt like this refactor was necessary but as I got to understand the code more, it seemed less-so. Basically what it does is move the 'execution methods' (run, preload, materialize) into separate bare functions and make the Query objects dumber.

There is no actual problem w/ current code however. @grgbkr do you see any value to this refactor?

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -82,13 +82,8 @@ import {
   isClientMetric,
 } from '../../../zql/src/query/metrics-delegate.ts';
 import type {QueryDelegate} from '../../../zql/src/query/query-delegate.ts';
+import {newQuery, type AnyQuery} from '../../../zql/src/query/query-impl.ts';
 import {
-  materialize,
-  newQuery,
-  type AnyQuery,
-} from '../../../zql/src/query/query-impl.ts';
-import {
-  delegateSymbol,
   type HumanReadable,
   type MaterializeOptions,
   type PreloadOptions,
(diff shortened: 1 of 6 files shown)


aboodman:
Update: I'm going to go ahead and propose we land this. I think it makes some other refactoring I want to do easier.

I want to pull out a separate runnable query interface and implementation that the base query interface does not have the run method, materialize, or preload methods on it. It's already kind of ugly having to have these methods through exceptions and various implementations. But it has become an actual problem on the server-side that `Query` has `run()` and implicitly calls it when you `await`.

In the current code, the core implementation of run, materialize, preload, etc. are in the base query implementation. But then if we want a version of the query interface that does not have these features, we have to hide them using the private symbol approach. And then if we want a higher-level interface that does have these features, we have to re-expose them with a public interface. The whole thing just seems kind of indirect, hiding these implementations and re-exposing them, and it feels a lot simpler to put them in a separate function that you can compose.

aboodman:
See https://github.com/rocicorp/mono/pull/4870 for the context.

I'm trying to get the `run` method and the associated `then()` implementation out of the query interface. And moreover, I don't want just the implementation of these methods out of the interface. I need there to be no `then` method on `QueryImpl` at all, because if there is a `then` method, `await` will call it.

tantaman:
>  I need there to be no then method on QueryImpl at all, because if there is a then method, await will call it.

We can clean up `getQueries` after that. `getQueries` requires the user to return `{query: getQuery(...)}` because otherwise `then` is called.

https://github.com/rocicorp/mono/blob/76a32e48fe0828b335920436c2571336d87d5064/apps/zbugs/api/index.ts#L216

tantaman:
Will `tx` in custom mutators return `runnable queries` or will it require the user to `tx.run(query)`?

aboodman:
i was thinking the latter. wdyt?

aboodman:
I think that eventually `RunnableQuery` will go away. But the reason to introduce it here is to limit the amount of bustage in one release 😅. So I was thinking that for a little while - maybe until beta, `tx.query.foo.run()` will continue to work. As will `zero.query.foo.run()`.

PR #4872, opened 2025-09-08
feat(zbugs): cursor based pagination rather than unbounded limit growth

grgbkr:


apps/zbugs/shared/queries.ts
```diff
@@ -210,25 +215,28 @@ function buildListQuery(
...
-            exists('comments', q =>
-              q.where('body', 'ILIKE', `%${escapeLike(textFilter)}%`),
-            ),
-          )
-        : undefined,
-      ...(labels ?? []).map(label =>
-        exists('labels', q => q.where('name', label)),
+  return q
+    .orderBy(sortField, orderByDir)
+    .orderBy('id', orderByDir)
```
grgbkr, line 243: Bad diff from github here.  The only difference is the addition of  `.orderBy('id', orderByDir)`.

Embarrassingly this is what took the longest with this change.  Without this, we were almost but not exactly reversing the list with `orderByDir`, leading to lots of subtle bugs in the scrolling code when changing direction of paging, which for a long time I assumed was due to a bug in the scrolling code.

apps/zbugs/shared/queries.ts
```diff
@@ -149,22 +151,47 @@ export const queries = {
...
           .limit(limit)
           .related('viewState', q => q.where('userID', userID).one())
           .related('labels'),
         auth?.role,
-      ),
+      );
+    },
+  ),
+
+  issueListV2: syncedQueryWithContext(
```
grgbkr, line 177: The need to version queries like this is an ugly wrinkle of custom queries.

apps/zbugs/shared/queries.ts
tantaman: ah, that is subtle since `id` would not be placed in the right direction without it explicitly being set.

apps/zbugs/src/pages/list/list-page.tsx
```diff
@@ -28,9 +28,38 @@ import {mark} from '../../perf-log.ts';
...
+});
+
+const toIssueArrayIndex = (index: number, anchor: Anchor) =>
+  anchor.direction === 'forward' ? index - anchor.index : anchor.index - index;
+
+const toBoundIssueArrayIndex = (
+  index: number,
+  anchor: Anchor,
+  length: number,
+) => Math.min(length - 1, Math.max(0, toIssueArrayIndex(index, anchor)));
```
tantaman, line 58: ` Math.max(0, toIssueArrayIndex(index, anchor))`?

`toIssueArrayIndex` can return a negative number?

apps/zbugs/src/pages/list/list-page.tsx
```diff
@@ -72,29 +101,71 @@ export function ListPage({onReady}: {onReady: () => void}) {
...
+      ? {
+          id: anchor.startRow.id,
+          modified: anchor.startRow.modified,
+          created: anchor.startRow.created,
+        }
+      : null,
+    anchor.direction,
   );
 
+  const baseQ = queries.issueListV2(
```
tantaman, line 136: makes sense. Maybe share the `{
      sortDirection,
      sortField,
      assignee,
      creator,
      labels,
      open,
      textFilter,
    },` args with `q` so it is slightly dryer?

apps/zbugs/src/pages/list/list-page.tsx
grgbkr: Yes, it may be logically before the 0th index of the current issueArray, or past its length.

grgbkr:
> complicated but it looks like we can factor this pattern out to a set of hooks for people to use.
> 
> What does the interaction look like when you:
> 
> 1. Scroll several pages down
> 2. Grab the scroll handle and try to move all the way to the beginning
> 
> unbounded limit growth probably handles that case well.

It works quite well, you can try it out here:
https://zbugs-6yzchhpfl.preview.rocicorp.dev/?status=all

PR #4873, opened 2025-09-08
feat(zero-solid): return errors to `useQuery` users

tantaman:


packages/zero-solid/package.json
```diff
@@ -20,8 +20,9 @@
     "@rocicorp/eslint-config": "^0.7.0",
     "@rocicorp/prettier-config": "^0.3.0",
     "@solidjs/testing-library": "^0.8.10",
-    "vite-plugin-solid": "^2.11.7",
-    "solid-js": "^1.9.4"
+    "jsdom": "^26.1.0",
```
arv, line 23: Can this be removed? We are using real browsers so we should never have a dependency on jsdom.

packages/zero-solid/src/solid-view.ts
```diff
@@ -8,17 +8,41 @@ import {
...
+  | {
+      type: 'complete';
+    }
+  | {
+      type: 'unknown';
+    }
+  | QueryErrorDetails
+>;
+
+type QueryErrorDetails = {
```
arv, line 31: I wonder if we should put this in some shared place since we have this in a few places.

arv (commented):
LGTM

Please fix the package.json

PR #4874, opened 2025-09-08
fix(zero-cache): start with smaller CVR purge sizes, make configurable

darkgnotic:
Default to a smaller CVR purge batch size (now 25 instead of 1000), and make CVR purging more generally configurable with two new options:
* `ZERO_CVR_GARBAGE_COLLECTION_INITIAL_BATCH_SIZE`
* `ZERO_CVR_GARBAGE_COLLECTION_INITIAL_INTERVAL_SECONDS`

Users should generally not need to tune these as the logic is adaptive, but they are made available in case tuning is desired. Notably, garbage collection can be disabled entirely by setting the batch size to 0.

The large initial batch size of 1000 can take a long time for apps with large CVRs, causing "returning" CVRs that are in the process of being purged in being locked out. (Detecting and responding to loading being-purged CVRs is forthcoming in a separate PR).

User report:

* https://discord.com/channels/830183651022471199/1288232858795769917/1414648438393536773

### DIFF
--- a/packages/zero-cache/src/config/zero-config.test.ts
+++ b/packages/zero-cache/src/config/zero-config.test.ts
@@ -122,6 +122,20 @@ test('zero-cache --help', () => {
                                                                    Note that garbage collection is an incremental, periodic process which does not                   
                                                                    necessarily purge all eligible CVRs immediately.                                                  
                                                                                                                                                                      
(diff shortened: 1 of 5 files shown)


arv:
LGTM

PR #4875, opened 2025-09-09
chore(zbugs): set mutateURL and getQueriesURL for previews

grgbkr:


grgbkr:
This is wrong ill fix in a follow up.

PR #4877, opened 2025-09-09
fix(zero-cache): deprecate deletes for other client groups

darkgnotic:
Deprecate deletes of clients or client groups other than the connected one (making them no-ops).

(These requests will also be removed from the client side by @arv).

The operation on rows of other client groups was incompatible with the row locking invariants assumed by the CVR garbage collection (and other CVR) logic. The client/client-group logic cleanup is now encompassed by CVR garbage collection anyway, so the former is no longer necessary. 

User report:
* https://discord.com/channels/830183651022471199/1414648438393536773/1414753413178327181


arv:
LGTM

PR #4879, opened 2025-09-09
feat(zbugs): Discord aside message in issue composer

alexhking:
Adding aside to new issue modal with Discord link, messaging.


aboodman:
CleanShot 2025-09-09 at 19 49 59@2x

Should we also have something like this in the comment box?

alexhking:
> CleanShot 2025-09-09 at 19 49 59@2x

1. The capitalization makes sense. Sometimes we style zero-related things as lowercase so I had it this way, but I think it makes more sense to be capitalized.
2. I think "issue" is the right word to use.
3. I can update and show you what it looks like for CTA to be a link with no button

> Should we also have something like this in the comment box?

Yeah, we can add this. I just don't want things to be too noisy. In the comment area I think we should just be showing the second sentence: "Want a faster response? Join us on Discord."

alexhking:
https://github.com/user-attachments/assets/1e40a6e4-9390-4a83-84e2-0276d8714124

alexhking:
@aboodman Circled back on this from last week... made some edits to the issue composer and then added a slim line to the comment composer that only shows up on form field focus. Video above. Let me know what you think when you get a chance.

aboodman:
A little bummed that this has sat unfinished for a few weeks. One more comment then let's get 'er in.

In this case, even just landing and tweaking async would have been better than letting it sit for so long.


https://github.com/user-attachments/assets/e65e0c48-8c77-437f-b57b-c302688565d3

PR #4881, opened 2025-09-10
refactor(zero): move typedoc to dev deps

tmm:
Moves `typedoc` and `typedoc-plugin-markdown` to `"devDependencies"` (both seem unused) to trim [dependency tree](https://npmgraph.js.org/?q=%40rocicorp%2Fzero#select=exact%3Atypedoc%400.28.12) a bit.


arv:
Thanks

PR #4882, opened 2025-09-10
chore(zero): Don't send deleted clientGroupIDs to server

arv:
This PR refactors the handling of deleted clients on the client side. Instead of tracking deleted clients by separate IDs, we now use tuples of `{clientGroupID, clientID}`. This change allows for more precise filtering: only deleted clients from the current client group are sent to the server.

The previous callback for garbage collecting client groups has been removed, as it was only triggered when deleting a group with no clients—something we no longer need to handle.

Additionally, the `clientGroupIDs` field in the protocol is now always an empty array. The server already tolerates its absence, and this makes it clearer what the message can contain.

### DIFF
--- a/packages/zero-client/src/client/active-clients-manager.test.ts
+++ b/packages/zero-client/src/client/active-clients-manager.test.ts
@@ -10,6 +10,7 @@ import {
 import {Queue} from '../../../shared/src/queue.ts';
 import {nanoid} from '../util/nanoid.ts';
 import {ActiveClientsManager} from './active-clients-manager.ts';
+import {waitForPostMessage} from './test-utils.ts';
 
 afterEach(() => {
   vi.restoreAllMocks();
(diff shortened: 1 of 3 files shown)


packages/replicache/src/deleted-clients.ts
```diff
@@ -97,3 +98,34 @@ export async function removeDeletedClients(
...
+    for (const [clientID, client] of clients) {
+      if (client.clientGroupID === clientGroupID) {
+        clientIDs.add(clientID);
+      }
+    }
+  }
+
+  return {
+    clientIDs: Array.from(clientIDs).sort(),
+    clientGroupIDs: [],
```
darkgnotic: Should we consider removing this field from the protocol altogether? I think the server already handles the absence of the field, and I think that would make it clearer about what this message can contain.

Just a thought ... up to you!

darkgnotic (approved):
Thank you @arv !

arv:
This turned out to be a lot bigger than I thought. I'm still working through the failure and it requires some more changes

packages/replicache/src/deleted-clients.ts
arv: Let's keep it for now... and remove it in the future

arv:
PTAL

darkgnotic:
> This turned out to be a lot bigger than I thought. I'm still working through the failure and it requires some more changes

Wow, indeed! TBH I don't have a lot of familiarity with this part of the client so I can't be a good verifier of correctness here. Thank you for doing this!

aboodman:
Maybe Greg should review instead or Matt? If there’s a high chance of
destabilizing.

a (phone)

arv:
@grgbkr Please Take A Look

PR #4883, opened 2025-09-10
perf(zbugs): optimize assignee and creator filters

grgbkr:
This does manual join flipping for assignee and creator filters in the get-queries api-server endpoint.

### DIFF
--- a/apps/zbugs/server/get-query.ts
+++ b/apps/zbugs/server/get-query.ts
@@ -49,11 +49,11 @@ const queries = {
    * If the assignee filter is highly selective, many non-matching issues will
    * be checked.
    *
-   * The original form is used on the client because it returns the data shape
-   * useful for rendering in the UI.  It will do the less efficient iteration
-   * of issues on the client, but this is ok because the number of rows synced
-   * to the client is small enough that the less efficient approach is still
-   * very fast.
+   * The form used on the client is structured as it is because it returns the
+   * data shape useful for rendering in the UI.  It will do the less efficient
+   * iteration of issues on the client, but this is ok because the number of
+   * rows synced to the client is small enough that the less efficient approach
+   * is still very fast.
    *
    * This code assumes a selectivity order, from most to least selective, of:
    * - assignee,


apps/zbugs/server/get-query.ts
```diff
@@ -18,6 +80,43 @@ export function getQuery(
   if (name in validated) {
     return validated[name](context, ...args);
   }
-
   throw new Error(`Unknown query: ${name}`);
 }
+
+function serverOptimizedListQuery(
```
grgbkr: // TODO add explanatory comment

apps/zbugs/server/get-query.ts
```diff
@@ -1,6 +1,64 @@
...
+  type ReadonlyJSONValue,
+} from '@rocicorp/zero';
+import {buildListQuery, queries as sharedQueries} from '../shared/queries.ts';
 import type {AuthData} from '../shared/auth.ts';
+import {builder} from '../shared/schema.ts';
+
+const queries = {
+  ...sharedQueries,
+  /**
+   * Replace issueListV2 with a server optimized version.
```
grgbkr, line 13: TODO document what the optimization does

apps/zbugs/server/get-query.ts
```diff
@@ -1,6 +1,110 @@
...
+   * This code assumes a selectivity order, from most to least selective, of:
+   * - assignee,
+   * - creator
+   */
+  issueListV2: syncedQueryWithContext(
+    sharedQueries.issueListV2.queryName,
+    sharedQueries.issueListV2.parse,
+    (auth: AuthData | undefined, listContext, userID, limit, start, dir) => {
+      if (!listContext) {
+        return builder.issue.where(({or}) => or());
```
tantaman, line 67: we should really add an `ALWAYS_FALSE` utility.

`builder.issue.where(alwaysFalse)`

apps/zbugs/shared/queries.ts
```diff
@@ -177,90 +170,116 @@ export const queries = {
...
+  }
 
+  const {sortField, sortDirection} = listContext;
   const orderByDir =
     dir === 'forward'
       ? sortDirection
       : sortDirection === 'asc'
         ? 'desc'
         : 'asc';
+  q.orderBy(sortField, orderByDir).orderBy('id', orderByDir);
```
tantaman, line 252: don't you need to assign this to `q`? Looks like this `orderBy` is being dropped.

grgbkr: good spot, fixed here https://github.com/rocicorp/mono/pull/4885

apps/zbugs/server/get-query.ts
grgbkr: Also having to still set up the relationships so that the result shape matches in this case is also annoying.  Its not necessary server side, but is necessary on the client when you actually use the result type.

tantaman: yeah, I ended up doing this:

CleanShot 2025-09-11 at 14 58 44

PR #4884, opened 2025-09-10
refactor(zbugs): deprecate queries.prevNext, instead use queries.issueListV2

grgbkr:
The only differences are:
1. `prevNext` uses `.one()` 
2. and `prevNext` doesn't include related `labels` and `viewState`

#1 is trivial and #2 does not have a meaningful performance difference.

This simplification simplifies the code structure for future query optimization for gigabugs.


tantaman:
lgtm

PR #4885, opened 2025-09-11
fix(zbugs): fix order by for issueListV2

grgbkr:
Bug introduced here https://github.com/rocicorp/mono/commit/d1bbfd6bd2e7a8c4fbe112022d615206fa4bcebe

### DIFF
--- a/apps/zbugs/shared/queries.ts
+++ b/apps/zbugs/shared/queries.ts
@@ -249,7 +249,7 @@ export function buildListQuery(args: ListQueryArgs) {
       : sortDirection === 'asc'
         ? 'desc'
         : 'asc';
-  q.orderBy(sortField, orderByDir).orderBy('id', orderByDir);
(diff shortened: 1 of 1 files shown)


apps/zbugs/shared/queries.ts
```diff
@@ -249,7 +249,7 @@ export function buildListQuery(args: ListQueryArgs) {
       : sortDirection === 'asc'
         ? 'desc'
         : 'asc';
-  q.orderBy(sortField, orderByDir).orderBy('id', orderByDir);
+  q = q.orderBy(sortField, orderByDir).orderBy('id', orderByDir);
```
aboodman, line 252: :sad trombone:

a little surprised our eslint config doesn't complain about this. but  i guess there are too many ignored returned values in idiomatic js.

tantaman: made a rule for this today

PR #4887, opened 2025-09-11
feat(replicache): Expose replicache/expo-sqlite and replicache/op-sqlite

arv:
This PR reintroduces the `SQLiteStore` abstraction, now using a single connection per store instead of multiple. The previous approach with multiple connections was incompatible with async `op-sqlite`, so this update resolves that issue.

The PR adds and exports two new modules: `replicache/expo-sqlite` and `replicache/op-sqlite`. Each provides an implementation of the `StorageProvider` interface, which can be supplied via the `kvStore` option. This should make it easier to use SQLite-based storage across different environments.

Additionally, there is a third implementation using `@rocicorp/zero-sqlite3`, intended only for testing purposes. It is not exported for production use.

### DIFF
--- a/packages/zero-react-native/src/mod.ts
+++ b/packages/zero-react-native/src/mod.ts
@@ -2,3 +2,7 @@ export {
   expoSQLiteStoreProvider,
   type ExpoSQLiteStoreOptions,
 } from '../../replicache/src/kv/expo-sqlite/store.ts';
+export {
+  opSQLiteStoreProvider,
+  type OpSQLiteStoreOptions,
+} from '../../replicache/src/kv/op-sqlite/store.ts';


goatrenterguy:
@arv Just saw this was merged. How long before its available on npm? I had written my own expo sqlite adapter but am running into some crashes.

arv:
You can check out the repo and do npm pack from packages/replicache

arv:
Or if you want the zero version then it will be soon. We do pretty frequent releases.

goatrenterguy:
@arv thanks! Could I convince you to publish it for the ease of CI/CD on my end? Otherwise I would have to fork and publish it on github packages

PR #4888, opened 2025-09-11
fix(react): view caching race condition

tantaman:
For some reason we were replacing the view in the view store when `onMaterialized` was called.

There's no need for this. All we need is a valid view in the view store which we have as soon as we create the view. Now we only remove from the map when a view is destroyed.

user report: https://discord.com/channels/830183651022471199/1415395645203812424/1415395649205305434

Trying on having Claude review changes as well.

# Claude Review

  Key Changes Made:

  1. Removed the onMaterialized callback entirely (lines 375-382 in the new version)
    - The ViewWrapper constructor no longer takes an onMaterialized parameter
    - The problematic error throw on line 385 is completely removed
    - The onDematerialized callback now directly deletes from the map
  2. Removed the #materializeIfNeeded call from onDematerialized (line 521)
    - The callback no longer tries to re-add the view to the store
  3. Added assertion in destroy() method (lines 540-543)
    - Ensures views are not destroyed while they still have listeners
  4. Refactored cleanup logic (lines 529-530)
    - Calls this.destroy() directly instead of inline cleanup

  Analysis:

  ✅ This SHOULD resolve the "View already exists" error because:

  1. Eliminates the race condition entirely - The error was thrown in the onMaterialized callback when checking if a
  different view already existed. By removing this callback and the check, the error can no longer occur.
  2. Simplifies the lifecycle - Views are now added to the store immediately when created (line 383) and removed when
  destroyed (line 381), with no intermediate re-registration step.
  3. Maintains single ownership - Each ViewWrapper is added to the store exactly once when created and removed exactly
  once when destroyed.

  ✅ Correctness appears maintained because:

  1. Reference counting still works - The #reactInternals Set still tracks all React listeners
  2. Cleanup timing preserved - The 10ms setTimeout delay for cleanup is maintained
  3. View reuse still works - Existing views are still returned and their TTL updated (lines 384-386)
  4. Assertion adds safety - The new assertion ensures views aren't destroyed while in use

  Potential Concerns:

  1. Strict Mode behavior - The original design with onMaterialized seemed intended to handle React Strict Mode's
  mount/unmount/remount cycle. The new approach should still work since:
    - The view stays in the store between rapid unmount/remount cycles
    - The 10ms delay gives time for remounting to occur
    - If a component remounts, it will find and reuse the existing ViewWrapper
  2. Memory management - Views might stay in the store slightly longer, but the cleanup mechanism via onDematerialized
  still ensures they're removed when truly unused.

  Conclusion: Yes, these changes should resolve the "View already exists" error while maintaining correctness. The
  simplified lifecycle eliminates the race condition that caused the error, and the core functionality of view
  management, reference counting, and cleanup remains intact.

### DIFF
--- a/packages/zero-react/src/use-query.test.tsx
+++ b/packages/zero-react/src/use-query.test.tsx
@@ -1116,4 +1116,120 @@ describe('useSuspenseQuery', () => {
       expect((capturedDetails as any).refetch).toBeUndefined();
     });
   });
+
+  describe('view management after fix', () => {
+    beforeEach(() => {
+      vi.useFakeTimers();
+    });
+
+    afterEach(() => {
+      vi.useRealTimers();
+    });
+
+    test('concurrent getView calls ideally share the same view', async () => {
+      const viewStore = new ViewStore();
+      const zero = newMockZero('client1');
+      const query = newMockQuery('query1');
+
+      // Simulate concurrent calls
+      const promises = Array.from({length: 10}, () =>
+        Promise.resolve().then(() =>
+          viewStore.getView(zero, query, true, 'forever'),
+        ),
+      );
+
+      const views = await Promise.all(promises);
+
+      // Check if views are shared (ideal case)
+      const uniqueViews = new Set(views);
+      expect(uniqueViews.size).toBe(1);
+
+      // Subscribe to all views
+      const cleanups = views.map(v => v.subscribeReactInternals(() => {}));
+
+      // Clean up all
+      cleanups.forEach(cleanup => cleanup());
+      vi.advanceTimersByTime(100);
+
+      // Verify all views are eventually cleaned up
+      expect(getAllViewsSizeForTesting(viewStore)).toBe(0);
+    });
+
+    test('rapid mount/unmount/remount reuses view when possible', () => {
+      const viewStore = new ViewStore();
+      const zero = newMockZero('client1');
+      const query = newMockQuery('query1');
+
+      const views = [];
+
+      // Simulate React strict mode double-mounting
+      for (let i = 0; i < 5; i++) {
+        const view = viewStore.getView(zero, query, true, 'forever');
+        views.push(view);
+        const cleanup = view.subscribeReactInternals(() => {});
+
+        // Immediate cleanup (unmount)
+        cleanup();
+
+        // Immediate remount before timeout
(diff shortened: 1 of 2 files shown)


packages/zero-react/src/use-query.test.tsx
```diff
@@ -1116,4 +1116,120 @@ describe('useSuspenseQuery', () => {
       expect((capturedDetails as any).refetch).toBeUndefined();
     });
   });
+
+  describe('view management after fix', () => {
```
arv, line 1120: this description is silly

packages/zero-react/src/use-query.test.tsx
```diff
@@ -1116,4 +1116,120 @@ describe('useSuspenseQuery', () => {
...
+      // Simulate concurrent calls
+      const promises = Array.from({length: 10}, () =>
+        Promise.resolve().then(() =>
+          viewStore.getView(zero, query, true, 'forever'),
+        ),
+      );
+
+      const views = await Promise.all(promises);
+
+      // Check if views are shared (ideal case)
```
arv, line 1143: what does it mean ideal case?

packages/zero-react/src/use-query.test.tsx
```diff
@@ -1116,4 +1116,120 @@ describe('useSuspenseQuery', () => {
...
+        cleanup();
+
+        // Immediate remount before timeout
+        const view2 = viewStore.getView(zero, query, true, 'forever');
+        views.push(view2);
+        const cleanup2 = view2.subscribeReactInternals(() => {});
+
+        // In ideal case, should reuse the same view
+        // There can be an edge case where we do not share the view.
+        // If this test is able to trigger that we should change expectation
```
arv, line 1181: This is really odd... Are these tests flaky or why does it say ideal case?

packages/zero-react/src/use-query.test.tsx
```diff
@@ -1116,4 +1116,120 @@ describe('useSuspenseQuery', () => {
...
+        // that ~99% of the time we share the view.
```
arv, line 1182: Tests should handle 100%

tantaman: I was thinking there could be a case where a view is removed while a registration is coming in. There was a variant of the implementation where that was possible but it is not now. Will update.

PR #4889, opened 2025-09-12
feat(zero-cache)!: require --admin-password in production mode

arv:
BREAKING CHANGE: In production mode (when NODE_ENV is not set to "development"), the `--admin-password` configuration option is now required. If this option is not set, the application will refuse to start.


arv:
@darkgnotic I had to set `process.env.NODE_ENV` to `'development'` in zero-cache-dev.ts. Please verify that this is the right place to set this?

darkgnotic (commented):
LGTM!

PR #4893, opened 2025-09-12
chore(ci): fix zql benchmarks

tantaman:


tantaman:
We need docker available in the runner so we can install postgres.

tantaman:
running now, but now we get: `Failed to validate the model for the latency Measure Threshold: Invalid threshold model: Invalid statistical boundary: 1.1`

may need to swap output from latency to throughput

.github/actions/run-zql-benchmarks/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    description: 'The BMF formatted benchmark output'
+    value: ${{ steps.run-benchmarks.outputs.result }}
+runs:
+  using: 'composite'
+  steps:
+    - name: Run ZQL benchmarks
+      id: run-benchmarks
+      working-directory: packages/zql-benchmarks
+      shell: bash
+      run: |
```
tantaman, line 14: keeping this since it is useful if something ever goes wrong in the conversion.

.github/workflows/bencher-benchmarks-pr.yml
```diff
@@ -90,3 +90,48 @@ jobs:
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22.x
+          cache: 'npm'
+      - uses: bencherdev/bencher@main
+
+      - run: npm ci
+
+      - name: Run ZQL benchmarks
+        uses: ./.github/actions/run-zql-benchmarks
```
arv, line 115: TIL

arv (commented):
LGTM

PR #4894, opened 2025-09-12
chore(zero): Add analyze to inspector query

arv:
Add Query Analyse to Inspector
---

This PR adds a new `analyze()` method to Query objects in the Zero inspector, enabling detailed performance analysis of queries.

This uses the server side AST to analyze the query.

What's New
---

- **Query.analyze() method**: New async method that returns query analysis including timing, row counts, execution plans, and warnings
- **Server-side analysis**: New `analyzeQuery()` service function that leverages the existing `analyze-query` package
- **Protocol support**: Extended inspect protocol with `analyze-query` RPC (protocol version 32)
- **Development mode**: Inspector authentication bypassed in dev mode for easier debugging

Usage
---

```typescript
const inspector = await zero.inspect();
const query = await inspector.client.queries()[0];

const result = await query.analyze({
  syncedRows: true,   // Include synced row data
  vendedRows: true    // Include database access data
});

console.log({
  timing: `${result.end - result.start}ms`,
  rowCount: result.syncedRowCount,
  plans: result.plans,
  warnings: result.warnings
});
```

### DIFF
--- a/packages/zero-cache/src/server/inspector-delegate.ts
+++ b/packages/zero-cache/src/server/inspector-delegate.ts
@@ -98,10 +98,7 @@ export class InspectorDelegate implements MetricsDelegate {
    * per "worker".
    */
   isAuthenticated(clientGroupID: ClientGroupID): boolean {
-    if (isDevelopmentMode()) {
-      return true;
-    }
-    return authenticatedClientIDs.has(clientGroupID);
+    return isDevelopmentMode() || authenticatedClientIDs.has(clientGroupID);
   }
 
   setAuthenticated(clientGroupID: ClientGroupID): void {
--- a/packages/zero-cache/src/services/analyze.ts
+++ b/packages/zero-cache/src/services/analyze.ts
@@ -3,7 +3,10 @@ import auth from 'basic-auth';
 import type {FastifyReply, FastifyRequest} from 'fastify';
 import {explainQueries} from '../../../analyze-query/src/explain-queries.ts';
 import {runAst} from '../../../analyze-query/src/run-ast.ts';
+import * as valita from '../../../shared/src/valita.ts';
+import type {AnalyzeQueryResult} from '../../../zero-protocol/src/analyze-query-result.ts';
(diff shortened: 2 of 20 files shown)


arv:
I didn't add `inspector.analyze(query, options)` yet. That can be done when the API is redesigned.

arv:
@copilot can you remove the unused type in types.ts?

packages/zero-cache/src/server/inspector-delegate.ts
```diff
@@ -98,10 +98,7 @@ export class InspectorDelegate implements MetricsDelegate {
    * per "worker".
    */
   isAuthenticated(clientGroupID: ClientGroupID): boolean {
-    if (isDevelopmentMode()) {
-      return true;
-    }
-    return authenticatedClientIDs.has(clientGroupID);
+    return isDevelopmentMode() || authenticatedClientIDs.has(clientGroupID);
```
tantaman, line 101: it is a little odd that we call this `authenticatedClientID` but pass in a `clientGroupID`

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -1908,6 +1905,17 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
         break;
       }
 
+      case 'analyze-query': {
+        const ast = body.value;
+        const result = await analyzeQuery(lc, this.#config, ast, body.options);
```
tantaman, line 1910: We should probably thread this out to a worker since `analyzeQuery` will block while running the hydration.

packages/zero-cache/src/services/analyze.ts
```diff
@@ -86,6 +104,5 @@ export async function handleAnalyzeQueryRequest(
   });
 
   result.plans = explainQueries(result.vendedRowCounts ?? {}, db);
-
-  await res.send(result);
+  return result;
```
tantaman, line 107: this was a problem in the prior impl too. Lets add:

```
db.close();
```

PR #4895, opened 2025-09-13
chore(zero-cache): add active user gauge for telemetry

cesara:


packages/zero-cache/src/server/anonymous-otel-start.ts
```diff
@@ -203,6 +204,13 @@ class AnonymousTelemetryManager {
       },
     );
 
+    const activeUsersGauge = this.#meter.createObservableGauge(
+      'zero.gauge_active_users',
```
darkgnotic: Since the value of the gauge is implicitly tied to the duration that it measures, I think we should put it in the name, like `zero.active_users_last_day`.

Then we could theoretically add a `last_hour` equivalent (or whatever interval we feel is useful).

Also, no need to put the metric type (i.e. "gauge") in the name.

packages/zero-cache/src/server/anonymous-otel-start.ts
```diff
@@ -271,6 +279,12 @@ class AnonymousTelemetryManager {
         `telemetry: gauge_active_client_groups=${activeClientGroups}`,
       );
     });
+
+    activeUsersGauge.addCallback((result: ObservableResult) => {
+      const activeUsers = this.#activeUsersGetter?.() ?? 0;
```
darkgnotic: Is it possible to just not call observe if there's no `#activeUsersGetter`? Or do we have to call `observe()` in the callback. I'd prefer not to put in a dummy value if possible (as it would look like a weird drop).

packages/zero-cache/src/server/reaper.ts
```diff
@@ -44,6 +47,11 @@ export default async function runWorker(
       initialBatchSize: cvr.garbageCollectionInitialBatchSize,
       initialIntervalMs: cvr.garbageCollectionInitialIntervalSeconds * 1000,
     }),
+    // Periodically computes and exports active users gauge to anonymous telemetry
+    new ActiveUsersGauge(lc, cvrDB, shard, {
+      // Default 60s refresh; can be made configurable later if needed
+      updateIntervalMs: 60_000,
```
darkgnotic: Since this is a value over the last day, updating every minute might be overkill. (While `lastActive` is indexed, I think COUNT(*) still has to scan that number of rows)

Maybe every 5 or 10 minutes? wdyt?

packages/zero-cache/src/services/view-syncer/active-users-gauge.ts
```diff
@@ -0,0 +1,67 @@
...
+    shard: ShardID,
+    opts: Options = {},
+  ) {
+    this.#lc = lc;
+    this.#db = db;
+    this.#schema = cvrSchema(shard);
+    this.#updateIntervalMs = opts.updateIntervalMs ?? 60 * 1000; // default 1 minute
+
+    // Expose value to anonymous telemetry exporter
+    setActiveUsersGetter(() => this.#value);
```
darkgnotic: Similarly, can we delay setting this getter until we've set `this.#value` at least once?

packages/zero-cache/src/services/view-syncer/active-users-gauge.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  async run(): Promise<void> {
+    while (this.#state.shouldRun()) {
+      try {
+        const since = Date.now() - DAY;
+        await this.#db.begin(async sql => {
```
darkgnotic: I don't think you need a `begin()` transaction here. It's just a single read so you can do the bare query (that you have inside of the transaction).

darkgnotic (approved):
Beautiful!   ✨ 

Nice work @cesara!

PR #4896, opened 2025-09-13
refactor(zbugs): update zbugs to use drizzle

grgbkr:
The drizzle schema was mainly created by using
`npx drizzle-kit pull`.

Confirmed a match between the current prod and sandbox dbs and this schema, by also running npx drizzle-kit pull against prod and sandbox and making sure it matched.

The results of 
`npx drizzle-kit pull`
had some small bugs (incorrect ops for boolean index, an unterminated string for a default value of '', and double quoting "issue_shortID_seq"). I fixed these manually.

I also created drizzle._drizzle_migrations tables on prod and sandbox dbs and manually recorded the 0000_init.sql migration as applied.  

image

image

I've killed the 1gb setup with this change.  I'll try to revive it next.


cesara (commented):
lgtm

PR #4897, opened 2025-09-15
feat(zero)!: remove `then` from `Query`

tantaman:
Remove `then`.

Having a `runnableQuery` will be a bit problematic as it will require duplicating the `Query` interface as each `limit, where, related, etc.` call must return `RunnableQuery` rather than `Query`.


packages/zql-integration-tests/src/chinook/chinook.pg-test.ts
```diff
@@ -32,6 +32,7 @@ describe(
         {
           suiteName: 'compiler_chinook',
           pgContent,
+          only: 'compare primary key',
```
arv, line 35: why did this change?

PR #4899, opened 2025-09-16
feat: surface zero update reasons in client callbacks

Karavil:
### Why
When zero-cache forced a reload we only logged the enum. Our Sentry breadcrumb looked like:

```ts
capture(new Error("zero-update-needed"), {
  reason,              // VersionNotSupported
  user_id,
  zero_server_url: ZERO_SERVER_URL,
});
```

No clue whether the cache rejected us for a schema mismatch, a missing table, or something else—so we always had to inspect websocket frames manually.

### What
- Thread the server-provided `message` through `UpdateNeededReason`, so both the default reload path and custom `onUpdateNeeded` handlers now receive:

  ```json
  {
    "type": "VersionNotSupported",
    "message": "table assignments missing on remote"
  }
  ```

- Reuse that same text when the client fires `onError`, keeping console logs and telemetry in sync.

### Result
Incidents now tell us *why* the client reloaded. The next time Zero reports "VersionNotSupported", the breadcrumb already includes the server’s explanation—no more spelunking in DevTools.

## Testing

- `npm install`
- `npx playwright install --with-deps`
- `npx vitest run packages/zero-client/src/client/zero.test.ts`


aboodman:
Thank you for this contribution. The only issue with this approach is that actually `onError` was purposely undocumented because we had some questions about its API design. I think this change is okay, but I think we're also going to change the way error handling happens in Zero in the future. On error may go away in favor of something else that is more unified with the overall state of the connection. 

We can land this for now, but I'm not sure if the API changes that I'm imagining will happen before the next release or not. If they do, this change may never be released. But at least checking it in will help to ensure that it makes it to the new API design. 

Do you need a canary with this change in it?

packages/zero-client/src/client/zero.test.ts
```diff
@@ -3903,6 +3911,23 @@ test('onError is called on error', async () => {
...
 });
 
+test('onError includes server error reason', async () => {
+  const onErrorSpy = vi.fn();
+  const z = zeroForTest({onError: onErrorSpy});
+
+  await z.triggerConnected();
+  expect(z.connectionState).toBe(ConnectionState.Connected);
+
+  const server_message = 'table missing on remote';
```
arv:
```suggestion
  const serverMessage = 'table missing on remote';
```

arv (changes requested):
LGTM

Just a naming nit. I'll deal with that

PR #4910, opened 2025-09-18
feat(zero-client)!: Make inspector less async

arv:
BREAKING CHANGE:

zero.inspector is now a synchronous getter.

The API is otherwise mostly unchanged, except that ClientGroup id is now a Promise<string>.


arv:
Maybe I should add back `inspect(): Promise<Inspector>` but mark it as deprecated?

PR #4914, opened 2025-09-19
chore(zero-client): Basic styling of the password prompt

arv:
When using `await zero.inspect.client.queries()` we prompt for a password.

This styles that dialog a little bit.


alexhking (commented):
Looks good!

PR #4917, opened 2025-09-19
chore(zql): factor out `push-accumulated`

tantaman:
`fan-in` and `union-fan-in` will do nearly identical things in push accumulated changes. Factor the code out.

The one big change is that `union-fan-in` will merge relationships. Added two arguments to handle relationship merging

### DIFF
--- a/packages/zql/src/ivm/push-accumulated.ts
+++ b/packages/zql/src/ivm/push-accumulated.ts
@@ -1,0 +1,395 @@
+import {assert, unreachable} from '../../../shared/src/asserts.ts';
+import {must} from '../../../shared/src/must.ts';
+import {emptyArray} from '../../../shared/src/sentinels.ts';
+import type {Change} from './change.ts';
+import type {Node} from './data.ts';
+import type {Output} from './operator.ts';
+import type {SourceSchema} from './schema.ts';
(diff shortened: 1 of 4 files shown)


packages/zql/src/ivm/push-accumulated.ts
```diff
@@ -0,0 +1,395 @@
...
+ *
+ * Pushes the changes that were accumulated by
+ * [fan-out, fan-in] or [ufo, ufi] sub-graphs.
+ *
+ * This function is called at the end of the sub-graph.
+ *
+ * The sub-graphs represents `OR`s.
+ *
+ * Changes that can enter the subgraphs:
+ * 1. child (due to exist joins being above the sub-graph)
```
grgbkr, line 21: this is all possible changes right?

packages/zql/src/ivm/push-accumulated.ts
```diff
@@ -0,0 +1,395 @@
...
+ * 2. add
+ * 3. remove
+ * 4. edit
+ *
+ * # Changes that can exit into `pushAccumulatedChanges`:
```
grgbkr, line 26: exit into?

PR #4919, opened 2025-09-19
chore(zql): union fan in

tantaman:


packages/zql/src/ivm/union-fan-in.ts
```diff
@@ -0,0 +1,153 @@
...
+  push(change: Change): void {
+    if (!this.#fanOutPushStarted) {
+      // if fan out has not started pushing but we receive a push, then it
+      // must be from an input to flip join. This input is the child of the output
+      // of flip join.
+      assert(
+        change.type === 'child',
+        'Only child changes allowed to come from internal nodes in union structures',
+      );
+      this.#output.push(change);
```
grgbkr, line 117: Do we need to fetch the node for the change row from all other branches and merge in their relationships?  

Perhaps we need some way of knowing which branches have flipped joins in them so we only fetch from those for efficency?

PR #4923, opened 2025-09-20
feat(replicache): bun sqlite store

Dizzzmas:
## What this does

Enables `bun:sqlite` as the `kvStore` for zero
e.g.
```typescript
import { Zero } from "@rocicorp/zero";
import { bunSQLiteStoreProvider } from "@rocicorp/zero/bun";

const zero = new Zero({
  kvStore: bunSQLiteStoreProvider(),
  ...
});
```

Looking forward to some cool use-cases in non-browser environments!
Really wanna use this for my projects

## How was this tested?

Mock tests similar to op/expo-sqlite were added

But also tests running against real `bun:sqlite` were implemented in https://github.com/Dizzzmas/zero-bun-sqlite-store - a separate package i published to use in the meantime


packages/replicache/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "lib": ["dom", "es2022"]
+    "lib": ["dom", "es2022"],
+    "types": ["bun"]
```
arv: I wonder if we could do another tsconfig.bun.ts file? I'm a little bit concerned about adding Bun globals to all of Replicache. It could cause us to not catch things at compile time.

packages/zero-bun/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "type": "module",
+  "module": "true",
+  "scripts": {
+    "check-types": "tsc",
+    "check-types:watch": "tsc --watch",
+    "test": "vitest run",
+    "test:watch": "vitest watch",
+    "format": "prettier --write .",
+    "check-format": "prettier --check .",
+    "lint": "eslint --ext .ts,.tsx,.js,.jsx src/"
```
arv: please update this.

See the recent oxlint change

packages/zero-bun/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "@rocicorp/eslint-config": "^0.7.0",
+    "@rocicorp/prettier-config": "^0.3.0",
+    "@types/bun": "latest"
+  },
+  "peerDependencies": {
+    "bun": ">=1.0.0"
+  },
+  "eslintConfig": {
+    "extends": "../../eslint-config.json"
+  },
```
arv: ditto

This can be removed now

packages/zero-bun/vitest.config.ts
```diff
@@ -0,0 +1,3 @@
+import sharedConfig from '../shared/src/tool/vitest-config.ts';
+
+export default sharedConfig;
```
arv: export {default} from '...'

packages/replicache/src/kv/bun-sqlite/store.ts
```diff
@@ -0,0 +1,108 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+// @ts-ignore - Bun modules are available at runtime in Bun environments
+import {Database, type Statement} from 'bun:sqlite';
+// @ts-ignore - fs module is available at runtime in Bun environments
+import {unlinkSync, existsSync} from 'fs';
```
arv: use `node:fs`

packages/replicache/src/kv/bun-sqlite/store.ts
```diff
@@ -0,0 +1,108 @@
...
+    return Promise.resolve().then(() => {
+      const row = this.#statement.get(params);
+
+      if (row === null || row === undefined) {
+        return undefined;
+      }
+
+      // Handle different row structures based on the query
+      // For has() queries: SELECT 1 FROM entry WHERE key = ? LIMIT 1
+      // Returns: {1: 1} when key exists
```
arv: I think you are better off using `values()`

https://bun.com/docs/api/sqlite#values

we do not need extra objects

packages/replicache/src/kv/bun-sqlite/store.ts
```diff
@@ -0,0 +1,108 @@
...
+class BunSQLitePreparedStatement implements PreparedStatement {
+  readonly #statement: Statement;
+
+  constructor(statement: Statement) {
+    this.#statement = statement;
+  }
+
+  // eslint-disable-next-line require-await -- Required by PreparedStatement interface
+  async firstValue(params: string[]): Promise<string | undefined> {
+    return Promise.resolve().then(() => {
```
arv: Just use `async` and no promise wrapping.

packages/replicache/src/kv/bun-sqlite/store.ts
```diff
@@ -0,0 +1,108 @@
...
+      return undefined;
+    });
+  }
+
+  async exec(params: string[]): Promise<void> {
+    await Promise.resolve().then(() => {
+      this.#statement.run(params);
+    });
+  }
+}
```
arv, line 39:
```suggestion
  // eslint-disable-next-line require-await
  async exec(params: string[]): Promise<void> {
    this.#statement.run(params);
  }
}
```

packages/replicache/src/kv/bun-sqlite/store.ts
```diff
@@ -0,0 +1,108 @@
...
+      stmt.finalize();
+    }
+    this.#db.close();
+  }
+
+  destroy(): void {
+    if (
+      (this.#db as {destroy?: () => void})?.destroy &&
+      process.env.NODE_ENV === 'test'
+    ) {
```
arv: Can you explain this?

arv (changes requested):
Very cool.

I wonder if we should add tests that run in bun... Not in this PR but it would be amazing to support that.

packages/replicache/src/kv/bun-sqlite/store.ts
Dizzzmas: all the other sqlite stores expose `destroy` method to delete the sqlite db, but bun doesn't

tests that use mocks seem to manipulate the db in a way where to properly delete it - we rely on calling the mock `destroy` - so this was just a workaround to have the mock usable

maybe there are better ways to handle this like extending `Database` from bun - and using that or directly patching our `BunSqliteStore.destroy` in tests

arv: I prefer changing the mock in the tests.

Dizzzmas:
thanks for the review @arv!

yeah, ideally there'd be tests running against real bun
probably indeed best as a separate PR, to properly integrate with the existing tests setup

it's of little help, but can confirm all tests passing against real bun in this repo https://github.com/Dizzzmas/zero-bun-sqlite-store

packages/replicache/src/kv/bun-sqlite/store.ts
Dizzzmas: had a couple attempts at solving this mocking situation elegantly, but couldn't get it right

would appreciate some help here

arv:
@Dizzzmas Changes are here https://github.com/rocicorp/mono/pull/4941

aboodman:
Sorry to reply after this is landed. I replied earlier,but github blocked my email message and i didn't noice.

I am wondering if we should really check this into the core? It should be possible to implement this outside of core shouldn't it? What is the value of having in the zero package?

I get that it means we will maintain it but that is also a downside for us. And I think the relevant interfaces will change so infrequently that it should be easy to maintain outside.

Dizzzmas:
> Sorry to reply after this is landed. I replied earlier,but github blocked my email message and i didn't noice.
> 
> 
> 
> I am wondering if we should really check this into the core? It should be possible to implement this outside of core shouldn't it? What is the value of having in the zero package?
> 
> 
> 
> I get that it means we will maintain it but that is also a downside for us. And I think the relevant interfaces will change so infrequently that it should be easy to maintain outside.

@aboodman I get the maintenance concern. Maybe then a solution would be to expose the underlying SQLite primitives, that these store providers can be built on top of?

I've published https://github.com/Dizzzmas/zero-bun-sqlite-store but I basically had to copy-paste a bunch of code from this repo(cannot import directly from what i've tried), which felt a bit whack.

On direct bun SQLite store support in zero - guess a reason to have it could be adoption and hype(for lack of a better word) related? I know Dax and the sst team are working on https://github.com/sst/opentui, built on top of bun. bun:sqlite could be a great way to bring zero into the tui world, and having it as a part of the main package could make things smoother.

arv:
No matter what we do, we should expose the primitives to make this work as a stand alone package.

I was also excited about the "hype".

... another thing that argues against bundling it in replicache/zero is that bun:sqlite only supports sync access so the extra async is pure overhead.

arv:
@Dizzzmas I'm going to revert this one and create a new PR that exposes the needed interfaces instead.

PR #4926, opened 2025-09-23
fix(zql): use correct comparator for generateWithStart in MemorySource

grgbkr:
Pass the connectionComparator rather than the indexComparator to genereateWithStart.

To understand the bug consider this test case

```
  function t(c: {
    columns?: Record<string, SchemaValue> | undefined;
    startData: Row[];
    start: Start;
    constraint: Constraint;
    reverse?: boolean | undefined;
  }) {
    const sort = [['a', 'asc']] as const;
    const s = createSource(
      lc,
      testLogConfig,
      'table',
      c.columns ?? {
        a: {type: 'number'},
        b: {type: 'boolean'},
      },
      ['a'],
    );
    for (const row of c.startData) {
      s.push({type: 'add', row});
    }
    const out = new Catch(s.connect(sort));
    return out.fetch({
      constraint: c.constraint,
      start: c.start,
      reverse: c.reverse,
    });
  }

  test('c1', () => {
    expect(
      t({
        startData: [
          {a: 1, b: '1000'},
          {a: 2, b: '3000'},
          {a: 3, b: '2000'},
          {a: 5, b: '1000'},
          {a: 6, b: '4000'},
          {a: 7, b: '0000'},
        ],
        start: {
          row: {a: 3, b: '2000'},
          basis: 'at',
        },
        constraint: {b: '1000'},
      }),
    ).toMatchInlineSnapshot(`
      [
        {
          "relationships": {},
          "row": {
            "a": 5,
            "b": "1000",
          },
        },
      ]
    `);
  });
```

connectionComparator is [['a', 'asc']], 
while indexComparator is [['b', 'asc'], ['a', 'asc']]

So the index orders the rows like
          {a: 7, b: '0000'},       
          {a: 1, b: '1000'},
          {a: 5, b: '1000'},
          {a: 3, b: '2000'},
          {a: 2, b: '3000'},
          {a: 6, b: '4000'},

We start scanning the index from (this is the startAt passed to generateWithOverlay):
{ b: '1000', a: Symbol(min-value) }
this is because constraint: {b: '1000'}, and so we start from the cluster matching the constraint.

generateWithStart skips nodes until its comparator tells it the node.row is >= start (or > then if basis: 'after').

If we use indexComparator ([['b', 'asc'], ['a', 'asc']])
 {a: 1, b: '1000'} < {a: 3, b: '2000'} so generateWithStart skips it
 {a: 5, b: '1000'} < {a: 3, b: '2000'} so generateWithStart skips it 
 
no other rows match the constraint of 1000, so no rows are yielded, which is a mistake  {a: 5, b: '1000'}  should have been yielded.

If we use connectionComparator ([['a', 'asc']])
 {a: 1, b: '1000'} < {a: 3, b: '2000'} so generateWithStart skips it
 {a: 5, b: '1000'} > {a: 3, b: '2000'} so generateWithStart yields it

### DIFF
--- a/packages/zql/src/ivm/memory-source.ts
+++ b/packages/zql/src/ivm/memory-source.ts
@@ -329,6 +329,15 @@ export class MemorySource implements Source {
       this.#overlay,
       this.#splitEditOverlay,
       callingConnectionIndex,
+      // Use indexComparator, generateWithOverlayInner has a subtle dependency
+      // on this.  Since generateWithConstraint is done after
+      // generateWithOverlay, the generator consumed by generateWithOverlayInner
+      // does not end when the constraint stops matching and so the final
+      // check to yield an add overlay if not yet yielded is not reached.
+      // However, using the indexComparator the add overlay will be less than
+      // the first row that does not match the constraint, and so any
+      // not yet yielded add overlay will be yielded when the first row
+      // not matching the constraint is reached.
       indexComparator,
       conn.filters?.predicate,
     );
--- a/packages/zql/src/ivm/memory-source.ts
+++ b/packages/zql/src/ivm/memory-source.ts
@@ -284,7 +284,7 @@ export class MemorySource implements Source {
 
     const index = this.#getOrCreateIndex(indexSort, from);
     const {data, comparator: compare} = index;
-    const comparator = (r1: Row, r2: Row) =>
+    const indexComparator = (r1: Row, r2: Row) =>
       compare(r1, r2) * (req.reverse ? -1 : 1);
 
     const startAt = req.start?.row;
@@ -329,7 +329,7 @@ export class MemorySource implements Source {
       this.#overlay,
       this.#splitEditOverlay,
       callingConnectionIndex,
-      comparator,
+      indexComparator,
       conn.filters?.predicate,
     );
 
--- a/packages/zql/src/ivm/memory-source.ts
+++ b/packages/zql/src/ivm/memory-source.ts
@@ -251,7 +251,9 @@ export class MemorySource implements Source {
     const callingConnectionIndex = this.#connections.indexOf(from);
     assert(callingConnectionIndex !== -1, 'Output not found');
     const conn = this.#connections[callingConnectionIndex];
-    const {sort: requestedSort} = conn;
+    const {sort: requestedSort, compareRows} = conn;
+    const connectionComparator = (r1: Row, r2: Row) =>
+      compareRows(r1, r2) * (req.reverse ? -1 : 1);
 
     const pkConstraint = primaryKeyConstraintFromFilters(
       conn.filters?.condition,
@@ -332,7 +334,7 @@ export class MemorySource implements Source {
     );
 
     const withConstraint = generateWithConstraint(
-      generateWithStart(withOverlay, req.start, comparator),
+      generateWithStart(withOverlay, req.start, connectionComparator),
       // we use `req.constraint` and not `fetchOrPkConstraint` here because we need to
       // AND the constraint with what could have been the primary key constraint
       req.constraint,
--- a/packages/zql/src/ivm/source.test.ts
+++ b/packages/zql/src/ivm/source.test.ts
@@ -362,7 +362,7 @@ suite('fetch-with-constraint-and-start', () => {
           "relationships": {},
           "row": {
             "a": 5,
-            "b": "1000",
+            "b": "3000",
           },
         },
       ]
--- a/packages/zql/src/ivm/source.test.ts
+++ b/packages/zql/src/ivm/source.test.ts
@@ -339,6 +339,67 @@ suite('fetch-with-constraint-and-start', () => {
     });
   }
 
+  test('c1', () => {
+    expect(
+      t({
+        startData: [
+          {a: 1, b: '1000'},
+          {a: 2, b: '3000'},
+          {a: 3, b: '2000'},
+          {a: 5, b: '1000'},
+          {a: 6, b: '4000'},
+          {a: 7, b: '0000'},
+        ],
+        start: {
+          row: {a: 3, b: '2000'},
+          basis: 'at',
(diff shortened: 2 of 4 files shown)


packages/zql/src/ivm/memory-source.ts
```diff
@@ -327,12 +329,12 @@ export class MemorySource implements Source {
       this.#overlay,
       this.#splitEditOverlay,
       callingConnectionIndex,
-      comparator,
+      indexComparator,
```
grgbkr, line 341: Still trying to reason through if this should be indexComparator or connectionComparator.  

TableSource uses connectionComparator.

packages/zql/src/ivm/source.test.ts
```diff
@@ -339,6 +339,67 @@ suite('fetch-with-constraint-and-start', () => {
     });
   }
 
+  test('c1', () => {
```
grgbkr, line 342: These new tests failed before this change.

packages/zql/src/ivm/memory-source.ts
tantaman: It is fishy to me that `withOverlay` is created via `indexComparator` but passed to `generateWithStart` which uses `connectionComparator`.

tantaman: Can you describe the bug again?

And why is having the index in index order:

```ts
const index = this.#getOrCreateIndex(indexSort, from);
```

but then doing `generateWithStart` over it (with a different comparator, the connection comparator) is ok?

grgbkr: Consider this test case

```
  function t(c: {
    columns?: Record<string, SchemaValue> | undefined;
    startData: Row[];
    start: Start;
    constraint: Constraint;
    reverse?: boolean | undefined;
  }) {
    const sort = [['a', 'asc']] as const;
    const s = createSource(
      lc,
      testLogConfig,
      'table',
      c.columns ?? {
        a: {type: 'number'},
        b: {type: 'boolean'},
      },
      ['a'],
    );
    for (const row of c.startData) {
      s.push({type: 'add', row});
    }
    const out = new Catch(s.connect(sort));
    return out.fetch({
      constraint: c.constraint,
      start: c.start,
      reverse: c.reverse,
    });
  }

  test('c1', () => {
    expect(
      t({
        startData: [
          {a: 1, b: '1000'},
          {a: 2, b: '3000'},
          {a: 3, b: '2000'},
          {a: 5, b: '1000'},
          {a: 6, b: '4000'},
          {a: 7, b: '0000'},
        ],
        start: {
          row: {a: 3, b: '2000'},
          basis: 'at',
        },
        constraint: {b: '1000'},
      }),
    ).toMatchInlineSnapshot(`
      [
        {
          "relationships": {},
          "row": {
            "a": 5,
            "b": "1000",
          },
        },
      ]
    `);
  });
```

connectionComparator is [['a', 'asc']], 
while indexComparator is [['b', 'asc'], ['a', 'asc']]

So the index orders the rows like
          {a: 7, b: '0000'},       
          {a: 1, b: '1000'},
          {a: 5, b: '1000'},
          {a: 3, b: '2000'},
          {a: 2, b: '3000'},
          {a: 6, b: '4000'},

We start scanning the index from (this is the startAt passed to generateWithOverlay):
{ b: '1000', a: Symbol(min-value) }
this is because constraint: {b: '1000'}, and so we start from the cluster matching the constraint.

generateWithStart skips nodes until its comparator tells it the node.row is >= start (or > then if basis: 'after').

If we use indexComparator ([['b', 'asc'], ['a', 'asc']])
 {a: 1, b: '1000'} < {a: 3, b: '2000'} so generateWithStart skips it
 {a: 5, b: '1000'} < {a: 3, b: '2000'} so generateWithStart skips it 
 
no other rows match the constraint of 1000, so no rows are yielded, which is a mistake  {a: 5, b: '1000'}  should have been yielded.

If we use connectionComparator ([['a', 'asc']])
 {a: 1, b: '1000'} < {a: 3, b: '2000'} so generateWithStart skips it
 {a: 5, b: '1000'} > {a: 3, b: '2000'} so generateWithStart yields it

grgbkr: > We start scanning the index from (this is the startAt passed to generateWithOverlay):
> { b: '1000', a: Symbol(min-value) }
> this is because constraint: {b: '1000'}, and so we start from the cluster matching the constraint.

BTW I think we can improve our selection of startAt so that it takes into account both fetch constraint and start.  For the above test case I think the best startAt is { b: '1000', a: 3 }

I'm not sure yet if that means we can get rid of generateWithStart.

grgbkr: I think it doesn't matter which we pass to `generateWithOverlay` because we drop the overlay if it doesn't match the constraint (see https://github.com/rocicorp/mono/blob/d8a6f4d59ef5cfee7f654f041ee181e7cda62c76/packages/zql/src/ivm/memory-source.ts#L667).  So overlay will only exist if it matches the constraint, so the first part of the indexComparator will always match so we are effectively ordering it by the rest which is the same as the connectionComparator.  Switch to use the connectionComparator since it is consistent with TableSource and is logically equivalent.

tantaman: Ok, this makes sense.

If a constraint is set, we must return things that match the constraint. This swapping of comparators is ok, conceptually, as the connection comparator is finding a row _within_ the group returned by the constraint.

Do we return early if we leave the group created by the constraint? We should not return a row that matches `start` (via connection comparator) but does not match the constraint (via index comparator).

grgbkr: generateWithConstraint stops generating when it receives a row that doesn't match the constraint.  

indexComparator is needed here... for very subtle reasons.  I added a comment.

tantaman:
lgtm

PR #4929, opened 2025-09-24
test,fix(zqlite-zql-test): reintroduce zqlite-sql-test

grgbkr:
These were removed in https://github.com/rocicorp/mono/pull/4186/files#diff-a6505b28eeee3c0fae0650416c0adc0cb2073c80b0d5929c0ab83000478f0f78L6 by mistake.


packages/zqlite/src/table-source.ts
```diff
@@ -484,7 +484,7 @@ export class TableSource implements Source {
     if (constraint) {
       for (const [key, value] of Object.entries(constraint)) {
         constraints.push(
-          sql`${sql.ident(key)} IS ${toSQLiteType(
+          sql`${sql.ident(key)} = ${toSQLiteType(
```
grgbkr, line 487: @tantaman this was changed in https://github.com/rocicorp/mono/commit/c72d614afeb5cd24f18b82e082183752c63caa73

without changing it back this test fails

image

https://github.com/rocicorp/mono/blob/d8a6f4d59ef5cfee7f654f041ee181e7cda62c76/packages/zql/src/ivm/source.test.ts#L202

tantaman: hmm, there was a bug related to `null` where we swapped this to `IS` but `=` does still seem more correct to me. Will search slack.

PR #4930, opened 2025-09-24
fix(zero-cache): use explicit deletes (instead of fk references) in cvr gc

darkgnotic:
When purging rows from CVR tables, explicitly delete the rows from child tables up to parent tables, rather than only deleting rows in parent tables and relying on foreign key cascades.

For some tables (e.g. the "desires" table), the foreign key reference is not a prefix of the primary key, so a cascading delete from the parent table can result in a quadratic scan of a clientGroup's rows. 

Instead, explicitly deleting by clientGroupID ensures that it is linear.

User report:
* https://discord.com/channels/830183651022471199/1420477677227348105/1420478261875577036


arv:
LGTM

PR #4931, opened 2025-09-24
 chore(zql): rework source split edit to be more consistent

grgbkr:
Currently when source splits an edit, it does it just for the connection where it would impact a join.  

This means some connections see the change as an edit and some see it as a remove followed by an add.  This creates challenges for UnionFanIn..  

This change makes it so that if any connection has an impacted join, we split the edit for all connections.  First a remove is sent to all connections, followed by an add to all connections.


packages/zql/src/builder/builder.test.ts
```diff
@@ -1938,6 +1938,136 @@ test('not exists self join', () => {
         },
         "type": "remove",
       },
+      {
```
grgbkr, line 1941: The big increase in pushes here is from us splitting edit into remove and add for all connections.  The end result is the same (a bunch of rows are added, and then removed).

tantaman:
This is the specific challenge for union-fan-in fixed by this pr:

image

After this PR the flow of events is:
- green remove
- blue remove
- green add
- blue add

which stops us from re-ordering add and remove events in `union-fan-in`

PR #4934, opened 2025-09-25
chore(zero-cache): yield in analyzeQuery to avoid blocking

arv:
Follow up to #4894

https://github.com/rocicorp/mono/pull/4894#discussion_r2350268492

### DIFF
--- a/packages/analyze-query/src/run-ast.ts
+++ b/packages/analyze-query/src/run-ast.ts
@@ -3,6 +3,7 @@ import {astToZQL} from '../../ast-to-zql/src/ast-to-zql.ts';
 import {formatOutput} from '../../ast-to-zql/src/format.ts';
 import {assert} from '../../shared/src/asserts.ts';
 import {must} from '../../shared/src/must.ts';
+import {sleep} from '../../shared/src/sleep.ts';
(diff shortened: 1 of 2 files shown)


packages/analyze-query/src/run-ast.ts
```diff
@@ -82,6 +83,15 @@ export async function runAst(
   const seenByTable: Set<string> = new Set();
   for (const rowChange of hydrate(pipeline, hashOfAST(ast), tableSpecs)) {
     assert(rowChange.type === 'add');
+
+    // yield to other tasks to avoid blocking for too long
+    if (syncedRowCount % 10 === 0) {
+      await Promise.resolve();
+    }
+    if (syncedRowCount % 100 === 0) {
+      await sleep(1);
```
tantaman, line 92: `Promise.resolve()` to yield the micro task loop and `sleep(1)` to yield the event loop? Why 1 instead of 0?

arv: 0 is special cased in sleep.

```ts
export function sleep(ms: number, signal?: AbortSignal): Promise<void> {
  const newAbortError = () => new AbortError('Aborted');

  if (signal?.aborted) {
    return Promise.reject(newAbortError());
  }

  if (ms === 0) {
    return promiseVoid;
  }
```

PR #4937, opened 2025-09-25
chore(zbugs): upgrade zod to 4.x

arv:
And use zod/mini for smaller bundle size.


arv:
Main motivation was to get "standard schema" support.

PR #4940, opened 2025-09-26
chore(bencher-viz): see all benchmark statuses at a glance

tantaman:
See all the things at a glance and if there are problems.

CleanShot 2025-09-25 at 21 21 52@2x


tantaman:
put this in its own repo

PR #4941, opened 2025-09-26
Dizzzmas/main

arv:
@dizzzmas I was not able to push to your repo.

Here is the work to fix the PR.

- The tests should not have renamed the files.

- I also made sure that we can type check everything

- And also export replicache/bun-sqlite


Dizzzmas:
amazing, thanks for all the help to push this through @arv !

i should've had more courage to move away from file renaming pattern in sqlite tests 😅

arv:
Do you want to apply these changes to your PR?

I can also take over if you prefer?

Dizzzmas:
> Do you want to apply these changes to your PR?
> 
> I can also take over if you prefer?

applied changes to the original PR

PR #4942, opened 2025-09-26
feat(zql): added query return type helper

0xcadams:
A few users have asked for a type helper since 0.23 release. This uses `HumanReadableRecursive` to flatten the types.

```tsx
import {
  type Query,
  type QueryFnReturn,
  syncedQuery,
} from '@rocicorp/zero';
import * as z from 'zod/mini';

export const queries = {
  issueList: syncedQuery(
    'issueList',
    z.tuple([listContextParams, z.string(), z.number()]),
    (auth: AuthData | undefined, listContext, userID, limit) =>
      issueListV2(listContext, limit, userID, auth, null, 'forward'),
  ),
} as const;

type IssueList = QueryFnReturn<typeof queries.issueList>;
// {
//   readonly id: string;
//   readonly shortID: number | null;
//   readonly title: string;
//   readonly open: boolean;
//   readonly modified: number;
//   readonly created: number;
//   readonly creatorID: string;
//   readonly assigneeID: string | null;
//   readonly description: string;
//   readonly visibility: "internal" | "public";
//   readonly viewState: {
//       readonly issueID: string;
//       readonly userID: string;
//       readonly viewed: number;
//   } | undefined;
//   readonly labels: readonly {
//       readonly id: string;
//       readonly name: string;
//   }[];
// }[]
```


0xcadams:
> lgtm. How were we able to get away with `HumanReadable` on `run` vs `HumanReadableRecursive` though?
> 
> Or should we be using `Recursive` on run and friends?

I think we should use Recursive on run - the completion types in intellisense don't collapse well with relationships:

```tsx
const t = await z.run(commentQuery(z, displayed));

{
    readonly id: string;
    readonly issueID: string;
    readonly created: number;
    readonly body: string;
    readonly creatorID: string;
    readonly creator: PullRow<"user", {
        tables: {
            readonly emoji: {
                name: "emoji";
                columns: {
                    readonly id: {
                        type: "string";
                        optional: false;
                        customType: string;
                    };
                    readonly value: {
                        type: "string";
                        optional: false;
                        customType: string;
                    };
                      ...;
                };
                primaryKey: readonly [...];
            } & {
                ...;
            };
            ... 7 more ...;
            readonly issueNotifications: {
                ...;
            } & {
                ...;
            };
        };
        relationships: {
            ...;
        };
        enableLegacyQueries: false;
        enableLegacyMutators: false;
    }> | undefined;
    readonly emoji: readonly (PullRow<...> & {
        ...;
    })[];
}[]
```

With Recursive:

```
{
    readonly id: string;
    readonly issueID: string;
    readonly created: number;
    readonly body: string;
    readonly creatorID: string;
    readonly creator: {
        readonly id: string;
        readonly login: string;
        readonly name: string | null;
        readonly avatar: string;
        readonly role: "user" | "crew";
    } | undefined;
    readonly emoji: readonly {
        readonly id: string;
        readonly value: string;
        readonly annotation: string;
        readonly subjectID: string;
        readonly creatorID: string;
        readonly created: number;
        readonly creator: {
            readonly id: string;
            readonly login: string;
            readonly name: string | null;
            readonly avatar: string;
            readonly role: "user" | "crew";
        } | undefined;
    }[];
}[]
```

I see a note on `HumanReadableRecursive` about opaque types expanding incorrectly, but I don't think this is an issue for query types?

packages/zql/src/query/query.ts
```diff
@@ -97,7 +97,7 @@ export type PullRow<TTable extends string, TSchema extends ZeroSchema> = {
   >['columns']]: SchemaValueToTSType<
     PullTableSchema<TTable, TSchema>['columns'][K]
   >;
-};
+} & {};
```
0xcadams, line 100: Actually, @tantaman on second thought, this should be a less invasive type change and accomplishes the same thing.

Image

PR #4943, opened 2025-09-26
chore(zql): flipped-join review feedback https://github.com/rocicorp/mono/pull/4855

grgbkr:


packages/zql/src/ivm/flipped-join.ts
```diff
@@ -127,7 +126,7 @@ export class FlippedJoin implements Input {
       const removedNode = this.#inprogressChildChange.change.node;
       const compare = this.#child.getSchema().compareRows;
       const insertPos = binarySearch(childNodes.length, i =>
-        compare(childNodes[i].row, removedNode.row),
+        compare(removedNode.row, childNodes[i].row),
```
grgbkr, line 129: Really glad you caught this one.  I'm sure however this would have manifested in a customers app it would have been hard as hell to debug.

PR #4949, opened 2025-09-30
feat: add offline-pending state and reject offline writes

0xcadams:
Adds `online`, `offline`, and `offline-pending` states to OnlineManager and rejects all writes with OfflineError after `offlineDelayMs`.

Starts out in `offline-pending` to allow for any on-mount mutators to run successfully, then if there are no successful connections on timeout moves to `offline`.

Breaking changes:
1. Removed the previously deprecated `onOnlineChange`.
2. Updated `online` to be an enum instead of boolean.
3. Writes in CRUD and CM reject when offline for `offlineDelayMs = 300_000` (default 5 minutes).

Work to update zbugs will be done in a subsequent PR.

### DIFF
--- a/packages/zero-client/src/client/client-error.ts
+++ b/packages/zero-client/src/client/client-error.ts
@@ -4,17 +4,24 @@ import type {OnlineManager} from './online-manager.ts';
  * Client-side error thrown when the Zero client is in offline mode.
  * Offline mode is a local state that disables mutators until connectivity
  * is restored (or the mode is cancelled).
+ *
+ * @see {@link https://zero.rocicorp.dev/docs/offline Offline mode docs}
  */
 export class OfflineError extends Error {
   readonly name = 'OfflineError';
   readonly code = 'OFFLINE';
-  constructor(message = 'Offline mode: Mutations are disabled while offline') {
-    super(message);
+  constructor() {
+    super(
+      `Offline mode: mutations are disabled while offline. See https://zero.rocicorp.dev/docs/offline for more information.`,
+    );
   }
 }
 
-export function assertNotOffline(onlineManager: OnlineManager): void {
-  if (onlineManager.status === 'offline') {
-    throw new OfflineError();
+export const offlinePromiseRejection = (
+  onlineManager: OnlineManager | undefined,
+): Promise<never> | undefined => {
+  if (onlineManager?.status === 'offline') {
+    return Promise.reject(new OfflineError());
   }
-}
+  return undefined;
+};
--- a/packages/zero-client/src/client/client-error.ts
+++ b/packages/zero-client/src/client/client-error.ts
@@ -1,0 +1,20 @@
+import type {OnlineManager} from './online-manager.ts';
+
+/**
+ * Client-side error thrown when the Zero client is in offline mode.
+ * Offline mode is a local state that disables mutators until connectivity
+ * is restored (or the mode is cancelled).
+ */
+export class OfflineError extends Error {
+  readonly name = 'OfflineError';
+  readonly code = 'OFFLINE';
+  constructor(message = 'Offline mode: Mutations are disabled while offline') {
+    super(message);
+  }
+}
+
+export function assertNotOffline(onlineManager: OnlineManager): void {
+  if (onlineManager.status === 'offline') {
+    throw new OfflineError();
+  }
+}
--- a/packages/zero-client/src/client/crud.ts
+++ b/packages/zero-client/src/client/crud.ts
@@ -23,6 +23,8 @@ import type {
   UpsertValue,
   DeleteID,
 } from '../../../zql/src/mutate/custom.ts';
+import type {OnlineManager} from './online-manager.ts';
+import {assertNotOffline} from './client-error.ts';
 
 /**
  * This is the type of the generated mutate.<name>.<verb> function.
@@ -81,6 +83,7 @@ type ZeroCRUDMutate = {
 export function makeCRUDMutate<const S extends Schema>(
   schema: S,
   repMutate: ZeroCRUDMutate,
+  onlineManager: OnlineManager,
 ): {mutate: DBMutator<S>; mutateBatch: BatchMutator<S>} {
   const {[CRUD_MUTATION_NAME]: zeroCRUD} = repMutate;
 
@@ -88,7 +91,7 @@ export function makeCRUDMutate<const S extends Schema>(
     const ops: CRUDOp[] = [];
     const m = {} as Record<string, unknown>;
     for (const name of Object.keys(schema.tables)) {
-      m[name] = makeBatchCRUDMutate(name, schema, ops);
+      m[name] = makeBatchCRUDMutate(name, schema, ops, onlineManager);
     }
 
     const rv = await body(m as DBMutator<S>);
@@ -98,7 +101,12 @@ export function makeCRUDMutate<const S extends Schema>(
 
   const mutate: Record<string, TableMutator<TableSchema>> = {};
   for (const [name, tableSchema] of Object.entries(schema.tables)) {
-    mutate[name] = makeEntityCRUDMutate(name, tableSchema.primaryKey, zeroCRUD);
+    mutate[name] = makeEntityCRUDMutate(
+      name,
+      tableSchema.primaryKey,
+      zeroCRUD,
+      onlineManager,
+    );
   }
   return {
     mutate: mutate as DBMutator<S>,
@@ -114,9 +122,11 @@ function makeEntityCRUDMutate<S extends TableSchema>(
   tableName: string,
   primaryKey: S['primaryKey'],
   zeroCRUD: CRUDMutate,
+  onlineManager: OnlineManager,
 ): TableMutator<S> {
   return {
     insert: (value: InsertValue<S>) => {
+      assertNotOffline(onlineManager);
       const op: InsertOp = {
         op: 'insert',
         tableName,
@@ -126,6 +136,7 @@ function makeEntityCRUDMutate<S extends TableSchema>(
       return zeroCRUD({ops: [op]});
     },
     upsert: (value: UpsertValue<S>) => {
+      assertNotOffline(onlineManager);
       const op: UpsertOp = {
         op: 'upsert',
         tableName,
@@ -135,6 +146,7 @@ function makeEntityCRUDMutate<S extends TableSchema>(
       return zeroCRUD({ops: [op]});
     },
     update: (value: UpdateValue<S>) => {
+      assertNotOffline(onlineManager);
       const op: UpdateOp = {
         op: 'update',
         tableName,
@@ -144,6 +156,7 @@ function makeEntityCRUDMutate<S extends TableSchema>(
       return zeroCRUD({ops: [op]});
     },
     delete: (id: DeleteID<S>) => {
+      assertNotOffline(onlineManager);
       const op: DeleteOp = {
         op: 'delete',
         tableName,
@@ -156,17 +169,19 @@ function makeEntityCRUDMutate<S extends TableSchema>(
 }
 
 /**
- * Creates the `{inesrt, upsert, update, delete}` object for use inside a
+ * Creates the `{insert, upsert, update, delete}` object for use inside a
  * batch.
  */
 export function makeBatchCRUDMutate<S extends TableSchema>(
   tableName: string,
   schema: Schema,
   ops: CRUDOp[],
+  onlineManager: OnlineManager,
 ): TableMutator<S> {
   const {primaryKey} = schema.tables[tableName];
   return {
     insert: (value: InsertValue<S>) => {
+      assertNotOffline(onlineManager);
       const op: InsertOp = {
         op: 'insert',
         tableName,
@@ -177,6 +192,7 @@ export function makeBatchCRUDMutate<S extends TableSchema>(
       return promiseVoid;
     },
     upsert: (value: UpsertValue<S>) => {
+      assertNotOffline(onlineManager);
       const op: UpsertOp = {
         op: 'upsert',
         tableName,
@@ -187,6 +203,7 @@ export function makeBatchCRUDMutate<S extends TableSchema>(
       return promiseVoid;
     },
     update: (value: UpdateValue<S>) => {
+      assertNotOffline(onlineManager);
       const op: UpdateOp = {
         op: 'update',
         tableName,
@@ -197,6 +214,7 @@ export function makeBatchCRUDMutate<S extends TableSchema>(
       return promiseVoid;
     },
     delete: (id: DeleteID<S>) => {
+      assertNotOffline(onlineManager);
       const op: DeleteOp = {
         op: 'delete',
         tableName,
--- a/packages/zero-client/src/client/crud.ts
+++ b/packages/zero-client/src/client/crud.ts
@@ -23,8 +23,8 @@ import type {
   UpsertValue,
   DeleteID,
 } from '../../../zql/src/mutate/custom.ts';
+import {offlinePromiseRejection} from './client-error.ts';
 import type {OnlineManager} from './online-manager.ts';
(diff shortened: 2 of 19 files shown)


aboodman:
@arv should review but i will add some product/api feedback.

packages/zero-client/src/client/client-error.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+export class OfflineError extends Error {
+  readonly name = 'OfflineError';
+  readonly code = 'OFFLINE';
+  constructor(message = 'Offline mode: Mutations are disabled while offline') {
+    super(message);
+  }
+}
+
+export function assertNotOffline(onlineManager: OnlineManager): void {
```
aboodman: The word `assert` in our codebase mean an invariant was violated - basically a bug in our code. Use a word like checkOffline or mustBeOnline or something.

packages/zero-client/src/client/client-error.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Client-side error thrown when the Zero client is in offline mode.
+ * Offline mode is a local state that disables mutators until connectivity
+ * is restored (or the mode is cancelled).
+ */
+export class OfflineError extends Error {
+  readonly name = 'OfflineError';
+  readonly code = 'OFFLINE';
+  constructor(message = 'Offline mode: Mutations are disabled while offline') {
```
aboodman: The message should probably include a link to the relevant doc page.

packages/zero-client/src/client/online-manager.ts
```diff
@@ -0,0 +1,65 @@
...
+  #offlineDelayMs: number;
+  #pendingOfflineTimer: ReturnType<typeof setTimeout> | undefined;
+  #lc: ZeroLogContext;
+
+  constructor(offlineDelayMs: number, lc: ZeroLogContext) {
+    super();
+    this.#offlineDelayMs = offlineDelayMs;
+    this.#lc = lc;
+
+    // we start in offline-pending state so that writes aren't rejected immediately
```
aboodman: I prefer to start in `online` mode. People will hook up status UI to this and it looks funny to start in offline mode. Better to assume happy state. That solves your problem with allowing initial mutations too, so I think we do not need the third state. Right?

packages/zero-client/src/client/client-error.ts
arv: This can be called throwIfOffline

packages/zero-client/src/client/crud.ts
```diff
@@ -114,9 +122,11 @@ function makeEntityCRUDMutate<S extends TableSchema>(
   tableName: string,
   primaryKey: S['primaryKey'],
   zeroCRUD: CRUDMutate,
+  onlineManager: OnlineManager,
 ): TableMutator<S> {
   return {
     insert: (value: InsertValue<S>) => {
+      assertNotOffline(onlineManager);
```
arv: Careful. This should be reject if offline since it returns a Promise.

We should not throw sync errors in functions that return promises.

packages/zero-client/src/client/custom.test.ts
```diff
@@ -773,6 +781,35 @@ test('run waiting for complete results throws in custom mutations', async () =>
...
+    } as const,
+  });
+
+  const resultOnline = z.mutate.issue.create();
+  await resultOnline.client;
+
+  await vi.advanceTimersByTimeAsync(offlineDelayMs);
+  expect(z.online).equal('offline');
+  const resultOffline = z.mutate.issue.create();
+  await expect(resultOffline.client).rejects.toBeInstanceOf(OfflineError);
```
arv, line 805: Maybe test the message too?

packages/zero-client/src/client/client-error.ts
arv: We never call `new OfflineError` with an argument. You can remove the parameter.

packages/zero-client/src/client/custom.test.ts
```diff
@@ -773,6 +781,35 @@ test('run waiting for complete results throws in custom mutations', async () =>
...
+test('offline rejects writes', async () => {
+  const offlineDelayMs = 1_000;
+  const z = zeroForTest({
+    schema,
+    offlineDelayMs,
+    mutators: {
+      issue: {
+        create: async (tx: MutatorTx) => {
+          await tx.query.issue;
+        },
```
arv, line 793: also try this:

```ts
create: (tx: MutatorTx) => tx.query.issue,
```

packages/zero-client/src/client/online-manager.test.ts
```diff
@@ -0,0 +1,133 @@
+import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
+
+import {OnlineManager} from './online-manager.ts';
+import type {OnlineStatus} from './online-manager.ts';
+import type {ZeroLogContext} from './zero-log-context.ts';
+
+type MockLogger = {
```
arv: We have testing classes for the loggers

packages/zero-client/src/client/online-manager.test.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  return {manager, lc, events, unsubscribe, offlineDelayMs};
+}
+
+describe('OnlineManager', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
```
arv, line 24: FYI, you can return a cleanup function in beforeEach.

packages/zero-client/src/client/online-manager.test.ts
```diff
@@ -0,0 +1,133 @@
...
+    vi.useRealTimers();
+  });
+
+  it('starts offline', () => {
```
arv: we use `test` in all other files.

packages/zero-client/src/client/options.ts
```diff
@@ -158,6 +147,14 @@ export interface ZeroOptions<
    */
   hiddenTabDisconnectDelay?: number | undefined;
 
+  /**
+   * The number of milliseconds to wait after a disconnection/error before
+   * entering offline mode. While in offline mode, mutators throw immediately.
+   *
+   * Default is 300_000 (5 minutes).
+   */
+  offlineDelayMs?: number | undefined;
```
arv: hiddenTabDisconnectDelay has no ms prefix. For consistency remove it here as well.

packages/zero-client/src/client/zero.ts
```diff
@@ -680,15 +684,43 @@ export class Zero<
...
+              then: (onFulfilled: unknown, onRejected: unknown) =>
+                (rejected as Promise<unknown>).then(
+                  onFulfilled as (value: unknown) => unknown,
+                  onRejected as (reason: unknown) => unknown,
+                ),
+            } as unknown as ReturnType<F>;
+          }
+          return (f as (...a: Parameters<F>) => ReturnType<F>)(...args);
+        }) as (...args: Parameters<F>) => ReturnType<F>;
+      };
```
arv, line 714: Maybe move this code to online-manager.ts?

arv: Do we need all these casts? Seems a it excessive.

arv: Also, you are wrapping custom mutators. Maybe wrap the cruds too?

packages/zero-react/src/use-zero-online.tsx
```diff
@@ -1,14 +1,28 @@
...
+ * The status can be one of:
+ * - `online` - the client is online and accepting writes.
+ * - `offline` - the client is offline and will reject writes until the client becomes online again.
+ * - `offline-pending` - the client cannot reach the server and will enter offline mode after `offlineDelayMs` milliseconds,
+ *   but will accept writes until then.
+ *
+ * @example
+ * const online = useZeroOnline();
+ *
+ * <span>
```
arv: not sure what the span is for  :-)

I think you should use markdown here instead.

arv: Actually, I don't think this example provides much.

packages/zero-solid/src/use-zero-online.ts
arv, line 1: A crazy (?) idea would be to make `zero.online` a signal itself. Just throwing it out there.

packages/zero-client/src/client/online-manager.ts
0xcadams: That's not quite what I was thinking - as a dev I would want to know when Zero is about to stop allowing mutations. If this state isn't exposed to developers, they will have to show an abrupt "You are now offline" as opposed to "You are about to be offline - please reconnect before you lose work". The `offline-pending` state is intended to communicate that - Zero failed to connect to a remote server, you may have shoddy connection, reconnect within the next `x` minutes or else.

Aside from that though, we can start in `online` mode - that makes sense to me.

aboodman:
I see. I get what you're trying to do, and it's a good impulse, but I don't
think that it's necessary.

0 won't lose work when you go offline. The existing mutations remain queued
and will be sent when 0 reconnects.

The idea is that 0 allows mutations to be accumulated in the normal way and
sent asynchronously, and when it notices that it's been offline for 5
minutes, it stops allowing new mutations.

But the mutations accumulated in the 5 mins are still queued and they will
still be sent on reconnect. Just as before. What we're trying to do is
prevent a lot of work from being accumulated while offline. Not prevent any
work from happening while offline.

Does this make sense?

a (phone)

0xcadams:
> I see. I get what you're trying to do, and it's a good impulse, but I don't think that it's necessary. 0 won't lose work when you go offline. The existing mutations remain queued and will be sent when 0 reconnects. The idea is that 0 allows mutations to be accumulated in the normal way and sent asynchronously, and when it notices that it's been offline for 5 minutes, it stops allowing new mutations. But the mutations accumulated in the 5 mins are still queued and they will still be sent on reconnect. Just as before. What we're trying to do is prevent a lot of work from being accumulated while offline. Not prevent any work from happening while offline. Does this make sense? a (phone)

Yeah, that makes sense from a technical perspective. But let's say I'm building an email client with zero. Now that offline writes throw, I would want to show a warning to users to reconnect before they are not able to make any more changes. It would be a bad experience for my users to randomly not be able to make any more changes, and I would get complaints saying "please warn me that I can't save draft emails before I can't save draft emails anymore". They would lose work and complain. The only way I could show this warning would be manually tracking connectivity state.

packages/zero-client/src/client/crud.ts
0xcadams: Good call - will update to return a rejected promise.

aboodman:
On Tue, Sep 30, 2025 at 8:43 AM Chase Adams ***@***.***>
wrote:

> *0xcadams* left a comment (rocicorp/mono#4949)
> <https://github.com/rocicorp/mono/pull/4949#issuecomment-3353374225>
>
> I see. I get what you're trying to do, and it's a good impulse, but I
> don't think that it's necessary. 0 won't lose work when you go offline. The
> existing mutations remain queued and will be sent when 0 reconnects. The
> idea is that 0 allows mutations to be accumulated in the normal way and
> sent asynchronously, and when it notices that it's been offline for 5
> minutes, it stops allowing new mutations. But the mutations accumulated in
> the 5 mins are still queued and they will still be sent on reconnect. Just
> as before. What we're trying to do is prevent a lot of work from being
> accumulated while offline. Not prevent any work from happening while
> offline. Does this make sense? a (phone)
>
> Yeah, that makes sense from a technical perspective. But let's say I'm
> building an email client with zero. Now that offline writes throw, I would
> want to show a warning to users to reconnect before they are not able to
> make any more changes. It would be a bad experience for my users to
> randomly not be able to make any more changes, and I would get complaints
> saying "please warn me that I can't save draft emails before I can't save
> draft emails anymore". They would lose work and complain. The only way I
> could show this warning would be manually tracking connectivity state.
>

Aha. This is what the `connecting` state is meant to do in the new error
handling doc:

https://www.notion.so/replicache/Polishing-Error-Handling-Auth-Revalidation-2683bed8954580769bf2f942ac13fb26?source=copy_link#26f3bed8954580eba6b3fcdfeca7fec0

The idea is that in UI terms:

- "connecting": sync spinner
- "connected": green check
- "disconnected": grey disconnected icon - mutations throw
- "needs-auth": some sort of auth icon - mutations throw
- "error": sad red icon with error message - mutations throw

It could make sense to simplify this change to only be able throwing errors
while offline and not about changing connection states at all. Then leave
that to a separate change that implements above design.

Or alternately, we could implement above design first then circle back to
implement offline behavior once the new connection states are in place.

aboodman:
On Tue, Sep 30, 2025 at 11:56 AM Aaron Boodman ***@***.***>
wrote:

> Aha. This is what the `connecting` state is meant to do in the new error
> handling doc:
>
>
> https://www.notion.so/replicache/Polishing-Error-Handling-Auth-Revalidation-2683bed8954580769bf2f942ac13fb26?source=copy_link#26f3bed8954580eba6b3fcdfeca7fec0
>
> The idea is that in UI terms:
>
> - "connecting": sync spinner
> - "connected": green check
>

mutations don't throw for either of these states.

and for the 5 minutes zero is trying to reconnect, it is in this connecting
state.

0xcadams:
> Aha. This is what the `connecting` state is meant to do in the new error handling doc: https://www.notion.so/replicache/Polishing-Error-Handling-Auth-Revalidation-2683bed8954580769bf2f942ac13fb26?source=copy_link#26f3bed8954580eba6b3fcdfeca7fec0 The idea is that in UI terms: - "connecting": sync spinner - "connected": green check - "disconnected": grey disconnected icon - mutations throw - "needs-auth": some sort of auth icon - mutations throw - "error": sad red icon with error message - mutations throw It could make sense to simplify this change to only be able throwing errors while offline and not about changing connection states at all. Then leave that to a separate change that implements above design. Or alternately, we could implement above design first then circle back to implement offline behavior once the new connection states are in place.

Right - okay, so in the end this offline change feels like a half-baked version of the new error handling.

Is it worthwhile to make this offline change ASAP? I would prefer to work with Erik to get that nailed down, since this will probably get reworked anyways with the error handling changes. Adding rejected writes on offline will probably cause headache without better error handling and docs on how errors bubble up.

aboodman:
On Tue, Sep 30, 2025 at 12:12 PM Chase Adams ***@***.***>
wrote:

> *0xcadams* left a comment (rocicorp/mono#4949)
> <https://github.com/rocicorp/mono/pull/4949#issuecomment-3353957969>
>
> Aha. This is what the connecting state is meant to do in the new error
> handling doc:
> https://www.notion.so/replicache/Polishing-Error-Handling-Auth-Revalidation-2683bed8954580769bf2f942ac13fb26?source=copy_link#26f3bed8954580eba6b3fcdfeca7fec0
> The idea is that in UI terms: - "connecting": sync spinner - "connected":
> green check - "disconnected": grey disconnected icon - mutations throw -
> "needs-auth": some sort of auth icon - mutations throw - "error": sad red
> icon with error message - mutations throw It could make sense to simplify
> this change to only be able throwing errors while offline and not about
> changing connection states at all. Then leave that to a separate change
> that implements above design. Or alternately, we could implement above
> design first then circle back to implement offline behavior once the new
> connection states are in place.
>
> Right - okay, so in the end this offline change feels like a half-baked
> version of the new error handling.
>
> Is it worthwhile to make this offline change ASAP? I would prefer to work
> with Erik to get that nailed down, since this will probably get reworked
> anyways with the error handling changes. Adding rejected writes on offline
> will probably cause headache without better error handling and docs on how
> errors bubble up.
>

Not quite following your wording here, but I think it at least makes sense
for you and I to meet and go through the new error handling. Then you can
decide what order to do these two in.

0xcadams:
Closing in favor of https://www.notion.so/replicache/Polishing-Error-Handling-Auth-Revalidation-2683bed8954580769bf2f942ac13fb26#26f3bed8954580eba6b3fcdfeca7fec0

PR #4951, opened 2025-09-30
chore(replicache/zero): Export SQLite Store factory

arv:
This exports the types and the factory class for SQLite based KV stores for Replicache and Zero client.

The exports are:

```ts
import {SQLiteStore} from 'replicache/sqlite';
```

and

```ts
import {SQLiteStore} from '@rocicorp/zero/sqlite';
```


arv:
@Dizzzmas I created a sample repo here: https://github.com/arv/replicache-bun-sqlite. I was hoping you could take this and apply it to yours.

If this works I will push new versions of Replicache and Zero.

packages/zero/src/sqlite.ts
```diff
@@ -0,0 +1 @@
+export * from '../../replicache/src/sqlite.ts';
```
0xcadams, line 1: Does this need a new package.json export to expose?

0xcadams (commented):
LGTM just one comment

packages/zero/src/sqlite.ts
arv: Yes it does. Good catch

Dizzzmas:
> @Dizzzmas I created a sample repo here: https://github.com/arv/replicache-bun-sqlite. I was hoping you could take this and apply it to yours.
> 
> If this works I will push new versions of Replicache and Zero.

@arv 
tried it  - works well. thank you!
sorry for the late reply

looking forward to new replicache/zero versions being published
will then release a new version of my package

PR #4952, opened 2025-09-30
given plan, apply flips

tantaman:
This is the "flipper" that flips join nodes after we've arrived at a plan.

The planner is still just a doc: https://www.notion.so/replicache/Simple-Flip-2713bed8954580dba401f6bb362252cf

The planner-via-flip only has access to 2^(n-1) possible plans vs the complete n! search space.


tantaman:
We can flip at the same time we create the query plan by adding a `pin & flip` step (step 8). Closing this.

PR #4955, opened 2025-10-01
chore: Add copilot instructions

arv:
Not only ClAUDE.md.


0xcadams:
@arv what do you think about using https://agents.md/ instead?

arv:
> @arv what do you think about using https://agents.md/ instead?

soundd good to me

aboodman:
Is there a way to make this a symlink or something? Are we going to have to copy the instructions for every single one of these AIs?

PR #4956, opened 2025-10-02
fix(zero-cache): traverse change log entries chronologically

darkgnotic:


packages/zero-cache/src/services/view-syncer/snapshotter.ts
```diff
@@ -288,7 +288,7 @@ class Snapshot {
 
   changesSince(prevVersion: string) {
     const cached = this.db.statementCache.get(
-      'SELECT * FROM "_zero.changeLog" WHERE stateVersion > ?',
+      'SELECT * FROM "_zero.changeLog" WHERE stateVersion > ? ORDER BY stateVersion ASC, pos ASC',
```
darkgnotic, line 291: Note: I believe this ordering was already the case, but I feel better about specifying it explicitly.

erikmunson:
any chance we can get this over the finish line soon? it's not fully breaking my app but it would be nice to not have customers re-hydrating expensive queries super frequently and the error looks scary in the console

darkgnotic:
> any chance we can get this over the finish line soon? it's not fully breaking my app but it would be nice to not have customers re-hydrating expensive queries super frequently and the error looks scary in the console

sorry for the delay. schedules were a bit fragmented last week. @grgbkr and I will figure out the way forward soon.  🙏

darkgnotic:
> > any chance we can get this over the finish line soon? it's not fully breaking my app but it would be nice to not have customers re-hydrating expensive queries super frequently and the error looks scary in the console
> 
> sorry for the delay. schedules were a bit fragmented last week. @grgbkr and I will figure out the way forward soon. 🙏

okay @erikmunson, we've brainstormed and come up with a better solution. @grgbkr will be implementing it. thank you for your patience!

erikmunson:
yay thanks guys!

PR #4958, opened 2025-10-02
feat(zero): defineQuery

arv:
This introduces defineQuery. It is the replacement for syncedQuery and syncedQueryWithArgs.

At this point only look at the changes to apps/zbugs. The internals of the implementation is about to change.


arv:
@aboodman Please look at the apps/zbugs changes. The rest will change behind the scenes

packages/zero-client/src/client/zero.ts
```diff
@@ -830,42 +829,43 @@ export class Zero<
...
+  run<
+    TTable extends keyof S['tables'] & string,
+    TReturn extends PullRow<TTable, S>,
+    Q extends WithContext<S, TTable, TReturn, TContext>,
+  >(newQuery: Q, runOptions?: RunOptions): Promise<HumanReadable<TReturn>> {
+    const q = newQuery.withContext(this.context);
+    return q[delegateSymbol](this.#zeroContext).run(runOptions);
+  }
+
+  get context(): TContext {
```
arv, line 904: `context` or `ctx`

It felt too strange to call this ctx here... which kind of implies we should call it context everywhere

apps/zbugs/api/index.ts
```diff
@@ -213,7 +212,7 @@ async function getQueriesHandler(
   await withAuth(request, reply, async authData => {
     reply.send(
       await handleGetQueriesRequest(
-        (name, args) => ({query: getQuery(authData, name, args)}),
+        (name, args) => getQuery(authData, name, args),
```
arv: This is a breaking change... Maybe I can undo this.

This was needed when Queries were `then`able.

apps/zbugs/shared/queries.ts
```diff
@@ -41,19 +42,51 @@ const issueRowSort = z.object({
 
 type IssueRowSort = z.infer<typeof issueRowSort>;
 
+const defineAuthQuery = <
```
arv: Turned out that adding this local helper made things much cleaner.

Without it one has to type the context and the args parameters.

One thing I was hoping was that one would be able to write:

```ts
defineQuery<AuthData>(...)
```

...but that does not work. If you provide a type parameter you have to provide all of them. You can make the rest of them optional but then we do not get type inference.

https://github.com/microsoft/TypeScript/issues/26242

apps/zbugs/shared/queries.ts
```diff
@@ -41,19 +42,51 @@ const issueRowSort = z.object({
...
+    },
+    ({args: userID}) => builder.user.where('id', userID).one(),
   ),
 
-  issuePreload: syncedQueryWithContext(
+  issuePreload: defineAuthQuery(
     'issuePreload',
     idValidator,
-    (auth: AuthData | undefined, userID) =>
+    ({args: userID, ctx: auth}) =>
```
arv: here args get inferred from the validator and ctx is hard coded in defineAuthQuery.

PR #4962, opened 2025-10-03
feat(zbugs): introduce project into the zbugs urls and queries

grgbkr:


grgbkr:
Seeing some bug on preview when navigating from issue page to list page.  Investigating.

grgbkr:
> Seeing some bug on preview when navigating from issue page to list page. Investigating.

done

PR #4965, opened 2025-10-03
chore(zero-events): package zero-events publicly as @rocicorp/zero-events

darkgnotic:


darkgnotic:
@arv I tested this locally by doing an `npm pack` and then installing that as a `file:...tgz` another repo; the types and imports work as expected.

But as a sanity check, can you confirm that I'm doing this correctly?

packages/zero-events/src/status.ts
```diff
@@ -41,12 +40,14 @@ export const statusEventSchema = zeroEventSchema.extend({
   description: v.string().optional(),
 
   /** Structured data describing the state of the component. */
-  state: jsonObjectSchema.optional(),
+  state: v.object({}).optional(),
```
arv: These are not compatible.

packages/zero-events/package.json
```diff
@@ -1,21 +1,45 @@
...
   "dependencies": {
-    "shared": "0.0.0"
+    "@badrap/valita": "0.3.11"
   },
   "devDependencies": {
+    "@rocicorp/eslint-config": "^0.7.0",
     "@rocicorp/prettier-config": "^0.3.0",
     "typescript": "~5.8.2"
   },
+  "eslintConfig": {
```
arv: no more eslint config.

packages/zero-events/src/status.ts
```diff
@@ -1,8 +1,7 @@
-import {jsonObjectSchema} from '../../shared/src/json-schema.ts';
-import * as v from '../../shared/src/valita.ts';
+import * as v from '@badrap/valita';
```
arv: This is also not entirely the same.

I think we need to figure out how we want to do this.

arv (changes requested):
This is not going to work. The json schema is not compatible.

packages/zero-events/src/status.ts
darkgnotic: At a high level, the idea here is that the consuming logic will parse known types with structured state, like the ReplicationStatusEvent and the ReplicationState object, but can also handle unknown types in a generic manner, i.e. by storing whatever is passed into `state` as is (e.g. parsing the object with "passthrough" mode and storing it in a json column).

I agree that jsonObjectSchema is better but I was hoping to avoid pulling in a lot of extra stuff just for that.  🤔

packages/zero-events/src/status.ts
darkgnotic: Yeah, same motivation here ... I'm trying to make the package more self-contained to facilitate exporting/publishing. Your thoughts / ideas are appreciated!

(Also, I maybe we can update to a newer valita version in mono soon. I like some of the stuff in 0.4.6 ...)

darkgnotic:
New plan ... I realize that the only thing that needs to be "published" are the types. In fact, that's all that zero-cache uses anyway.

We can leave the parsing implementation (e.g. valita, zod, or whatever) to the consumers of the library, which can then use those types to validate that their parsing is correct.

(It didn't make sense for the type specification to assume the parsing implementation anyway)

darkgnotic:
Okay, please have another look. Here is the output of `npm pack`:

```
npm notice
npm notice 📦  @rocicorp/zero-events@0.0.1
npm notice Tarball Contents
npm notice 914B out/index.d.ts
npm notice 235B out/index.d.ts.map
npm notice 90B out/mod.d.ts
npm notice 140B out/mod.d.ts.map
npm notice 222B out/mod.js
npm notice 2.6kB out/status.d.ts
npm notice 1.5kB out/status.d.ts.map
npm notice 908B out/util.d.ts
npm notice 504B out/util.d.ts.map
npm notice 936B package.json
npm notice Tarball Details
npm notice name: @rocicorp/zero-events
npm notice version: 0.0.1
npm notice filename: rocicorp-zero-events-0.0.1.tgz
npm notice package size: 3.2 kB
npm notice unpacked size: 8.0 kB
npm notice shasum: 69eb6621f879b10152482785a6c21e7c21078b4a
npm notice integrity: sha512-ZWFTVLj1kE7sp[...]CbSZSMpKTVWjw==
npm notice total files: 10
npm notice
rocicorp-zero-events-0.0.1.tgz
```

packages/zero-events/src/mod.ts
```diff
@@ -0,0 +1,2 @@
+export * from './index.ts';
```
arv, line 1: if this are only types then 

```ts
export type * from './index.ts';
```

packages/zero-events/tool/build.ts
```diff
@@ -0,0 +1,19 @@
...
+  const external = Object.keys(pkg.dependencies);
+
+  await esbuild.build({
+    bundle: true,
+    target: 'es2022',
+    format: 'esm',
+    platform: 'neutral',
+    external,
+    outdir: 'out',
+    entryPoints: ['src/mod.ts'],
```
arv, line 15: what is this building now? An empty file?

arv (changes requested):
I think the problem is that I do not understand what problem we are trying to solve?

packages/zero-events/tool/build.ts
darkgnotic: Parsing code will use both the types and the string constants defined in status.ts.

packages/zero-events/src/status.test.ts
arv, line 1: What does this test :-)

Does vitest not complain when there are no tests?

arv: vitest has type tests. 

https://github.com/rocicorp/mono/blob/b481eea051be2f46deb8924fb8da57f4ba38cf77/packages/shared/src/immutable.test.ts#L17-L22

packages/zero-events/src/status.ts
```diff
@@ -18,95 +17,79 @@ export type Status = v.Infer<typeof statusSchema>;
...
+
+export type ReplicationStatusEvent = Extend<
+  StatusEvent,
+  {
+    type: 'zero/events/status/replication/v1';
+    component: 'replication';
+    stage: ReplicationStage;
+    state: ReplicationState;
+  }
 >;
```
arv, line 95: Can't you do?

```ts
export type ReplicationStatusEvent = StatusEvent & {
  type: 'zero/events/status/replication/v1';
  component: 'replication';
  stage: ReplicationStage;
  state: ReplicationState;
}
```

or use interface extends?

I don't think we need this Extends type helper.

packages/zero-events/package.json
```diff
@@ -1,18 +1,36 @@
...
+    ".": {
+      "types": "./out/mod.d.ts",
+      "default": "./out/mod.js"
+    }
+  },
+  "files": [
+    "out",
+    "!*.tsbuildinfo"
+  ],
   "scripts": {
```
arv, line 24: missing test script

packages/zero-events/src/status.test.ts
darkgnotic: The file was a static check that the type in question did not evaluate to `never`. It was not actually a runtime test.

But ... thank you for the pointer to the vitest type tests! This is exactly what I was looking for. I will use this instead.  🙏

packages/zero-events/src/status.ts
darkgnotic: There's a slight difference and I'd be interested in your opinion. Originally the code was as you suggested:

```ts
export type ReplicationStatusEvent = StatusEvent & {
  type: 'zero/events/status/replication/v1';
  component: 'replication';
  stage: ReplicationStage;
  state: ReplicationState;
}
```

But I realized that this declaration does not actually guarantee that the `ReplicationStatusEvent` is assignable to (i.e. is a subtype of) `StatusEvent`. For instance, if I mistakenly add a leading slash to the `type` field:

```ts
export type ReplicationStatusEvent = StatusEvent & {
  type: '/zero/events/status/replication/v1';
  component: 'replication';
  stage: ReplicationStage;
  state: ReplicationState;
}
```

it wouldn't catch the fact that the `ReplicationStatusEvent` is not actually a subtype of `StatusEvent`.

The `Extend` helper, on the other hand, has supertype / subtype semantics and checks that the result is assignable to the supertype, and otherwise resolves to `never`.

I guess in either case, you would still need test it though (e.g. check against `never`), but the `never` behavior is an easier error to notice ...

wdyt? Is `Extend` too obtuse?

arv: Interesting.

What about interface? Does it have the same problem?

darkgnotic: Interface works! I'll change it to use this. Thank you.

PR #4966, opened 2025-10-04
fix(zero-cache-dev): pass environment variables and CLI args to subprocess

alizain:
**Problem:**

When running `npx zero-cache-dev --upstream-db=<connection-string>`, the main process starts successfully, but the `zero-cache` subprocess fails with:

```
TypeError: Missing required option ZERO_UPSTREAM_DB at upstream.db. Got undefined
```

**Root Cause:**

The `zero-cache-dev` script has two issues:

1. **Missing environment variables in subprocess**: The `deployPermissionsScript` subprocess is spawned without passing `env: process.env`, so it doesn't inherit environment variables from the parent process.

2. **Incorrect argument filtering**: Lines 39-43 attempt to filter CLI arguments between the two subprocesses using `parseOptionsAdvanced`, but this logic is flawed:
   - `zeroCacheArgs` is computed by parsing `deployPermissionsOptions` (backwards)
   - `deployPermissionsArgs` is computed by parsing `zeroOptions` (backwards)
   - This filtering removes shared options like `--upstream-db` that both subprocesses need

**Solution:**

Two changes in `zero-cache-dev.js`:

1. Add `env: process.env` to the `permissionsProcess` spawn call (line 61-65)
2. Replace the argument filtering logic (lines 39-43) to pass all CLI args to both subprocesses:

```javascript
// Pass all CLI args to both subprocesses - they'll each parse what they need
const cliArgs = process.argv.slice(2);
const zeroCacheArgs = cliArgs;
const deployPermissionsArgs = cliArgs;
```

Both subprocesses can parse the args they need and ignore what they don't recognize.

---


aboodman:
I have noticed this bug too. Thank you for the fix.  @darkgnotic I do not know this code very well, can you take a look please? Landing optimistically.

alizain:
important note: this code was generated by claude

aboodman:
Did you run it?

a (phone)

alizain:
Yup

darkgnotic:
Thanks @alizain.

As a sanity check, can you run this with a flag that only one of the scripts accepts, like `npx zero-cache-dev --output-format=sql`?

(I do understand the original issue but I'm unclear if the current solution is correct ... will think about this more).

darkgnotic:
The original flag parsing logic relied on the fact that the first and second processes had disjoint args (it used to be `gen-schema` rather than `deploy-permissions`).

Now that the first script is deploy-permissions, the problem arises because the two processes share some but not all args.

I think the proper solution is to do something like:
* use parseOptionsAdvance() for each set of options to determine which flags were `unknown` to that set of options
* run each process with their respective `unknown` flags excluded from the original cliArgs

@grgbkr any other ideas?

alizain:
Forgive my ignorance, but why does it matter if some unapplicable flags are passed to each underlying process?On Oct 4, 2025, at 1:29 PM, Darick Tong ***@***.***> wrote:﻿darkgnotic left a comment (rocicorp/mono#4966)
The original flag parsing logic relied on the fact that the first and second processes had disjoint args (it used to be gen-schema rather than deploy-permissions).
Now that the first script is deploy-permissions, the problem arises because the two processes share some but not all args.
I think the proper solution is to do something like:

use parseOptionsAdvance() for each set of options to determine which flags were unknown to that set of options
run each process with their respective unknown flags excluded from the original cliArgs

@grgbkr any other ideas?

darkgnotic:
I believe the default arg parsing logic of each process rejects unknown flags (a common defensive policy to catch typos, etc.).

alizain:
Gotcha, I've updated the solution based on your feedback. I'm relying on the fact that `parseOptionsAdvanced` will filter and format options correctly. From there, I wrote a serializer that converts this nested object back into command line args to pass down to each individual process.

Also, not sure how to do this in development...
> As a sanity check, can you run this with a flag that only one of the scripts accepts, like npx zero-cache-dev --output-format=sql?

packages/zero/src/zero-cache-dev.ts
```diff
@@ -49,16 +52,19 @@ async function main() {
     process.exit(-1);
   });
 
-  const {unknown: zeroCacheArgs} = parseOptionsAdvanced(
```
darkgnotic, line 52: Your approach jogged my memory on what `parseOptionsAdvanced()` returns, and I believe we can achieve the desired behavior without modifying the `options.ts` library.

What we want to use is the `env` object returned from `parseOptionsAdvanced()`. That is an object containing all parsed options (via argument or env), mapped from its corresponding env variable name.

So I think what you can do is:

```ts
const {env: deployPermissionsEnv} = parseOptionsAdvanced(deployPermissionsOptions, {... allowUnknown: true});
const {env: zeroCacheEnv} = parseOptionsAdvanced(zeroOptions, {... allowUnknown: true});
```

And then when spawning each process, pass both the original and parsed env objects to the child process. 

(subsequent comments to clarify ...)

packages/zero/src/zero-cache-dev.ts
```diff
@@ -86,6 +92,7 @@ async function main() {
       deployPermissionsScript,
       deployPermissionsArgs ?? [],
       {
+        env: process.env,
```
darkgnotic: Then this would be:

```ts
   env: {...process.env, ...deployPermissionsEnv}
```


And for `zero-cache`, line 133 would become:

```ts
  ...process.env,
  ...zeroCacheEnv,
```

darkgnotic:
> 
> Also, not sure how to do this in development...
> 
> > As a sanity check, can you run this with a flag that only one of the scripts accepts, like npx zero-cache-dev --output-format=sql?

What I want to verify is that with the fix, we are able to specify a flag that only one of the processes accepts. Another example would be the `--replica-file=...` flag, which ordinarily `deploy-permissions` would reject.

So something like:

```sh
$ npx zero-cache-dev --replica-file=/tmp/foobar.db
```

should work with this fix.

Does that make sense?

packages/zero/src/zero-cache-dev.ts
```diff
@@ -49,14 +49,19 @@ async function main() {
...
 
-  const {unknown: zeroCacheArgs} = parseOptionsAdvanced(
+  // Parse options for each subprocess to get environment variables
+  const {env: deployPermissionsEnv} = parseOptionsAdvanced(
     deployPermissionsOptions,
-    {envNamePrefix: ZERO_ENV_VAR_PREFIX, allowUnknown: true},
+    {
+      envNamePrefix: ZERO_ENV_VAR_PREFIX,
+      allowUnknown: true,
+      allowPartial: true,
```
darkgnotic: Is `allowPartial` necessary? I don't believe that's what we actually want.  🤔

packages/zero/src/zero-cache-dev.ts
```diff
@@ -49,14 +49,19 @@ async function main() {
...
+    },
   );
-
-  const {unknown: deployPermissionsArgs} = parseOptionsAdvanced(zeroOptions, {
+  const {env: zeroCacheEnv} = parseOptionsAdvanced(zeroOptions, {
     envNamePrefix: ZERO_ENV_VAR_PREFIX,
     allowUnknown: true,
+    allowPartial: true,
```
darkgnotic: Same question here ...

darkgnotic (commented):
Awesome, this looks great!

To clarify, perhaps you can test by adding the connection string for upstream as you normally do, with the process-specific flags in addition, e.g.

```sh
$ npx zero-cache-dev --upstream-db=<connection-string> --replica-file=/tmp/foobar.db
```

and

```sh
$ npx zero-cache-dev --upstream-db=<connection-string> --output-format=sql
```

If you can confirm both of those work, we'll have confirmation that we've found the right solution.  🙏

alizain:
updated, and this command works correctly

`npx -w packages/zero zero-cache-dev --output-format=json`

```
❯ npx -w packages/zero zero-cache-dev --output-format=json
[ 'pid=50727' ] 
[ 'pid=50727' ] Running zero-deploy-permissions.
[ 'pid=50738' ] 
Loading permissions from schema.ts
No schema found at schema.ts, so could not deploy permissions. Replicating data, but no tables will be syncable. Create a schema file with permissions to be able to sync data.
[ 'pid=50727' ] zero-deploy-permissions completed successfully.
[ 'pid=50727' ] Running zero-cache at

	http://localhost:4848


ZERO_CHANGE_STREAMER_PROTOCOL is deprecated:
Set the ZERO_CHANGE_STREAMER_URI on view-syncers instead.


ZERO_TARGET_CLIENT_ROW_COUNT is deprecated:
This option is no longer used and will be removed in a future version.
The client-side cache no longer enforces a row limit. Instead, TTL-based expiration
automatically manages cache size to prevent unbounded growth.

TypeError: Missing required option ZERO_UPSTREAM_DB at upstream.db. Got undefined

pid 50739 exiting with error TypeError: Missing required option ZERO_UPSTREAM_DB at upstream.db. Got undefined
    at Module.parse (file:///Users/alizain/Code/mono/packages/zero/out/shared/src/valita.js:132:15)
    at parseOptionsAdvanced (file:///Users/alizain/Code/mono/packages/zero/out/shared/src/options.js:228:23)
    at parseOptions (file:///Users/alizain/Code/mono/packages/zero/out/shared/src/options.js:96:12)
    at getZeroConfig (file:///Users/alizain/Code/mono/packages/zero/out/zero-cache/src/config/zero-config.js:580:24)
    at runWorker (file:///Users/alizain/Code/mono/packages/zero/out/zero-cache/src/server/runner/run-worker.js:21:17)
    at file:///Users/alizain/Code/mono/packages/zero/out/zero-cache/src/server/runner/main.js:6:26
    at exitAfter (file:///Users/alizain/Code/mono/packages/zero/out/zero-cache/src/services/life-cycle.js:180:15)
    at file:///Users/alizain/Code/mono/packages/zero/out/zero-cache/src/server/runner/main.js:6:10
    at ModuleJob.run (node:internal/modules/esm/module_job:329:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:644:26)
pid=50727 zero-cache exited. Exiting.
```

alizain:
`npx -w packages/zero --upstream-db=<connection-string> --output-format=sql` doesn't work, and technically shouldn't work. `replica-file` is required for one of the processes and we haven't passed it

packages/zero/src/zero-cache-dev.ts
alizain: yup good catch

alizain:
Thank you guys!

PR #4972, opened 2025-10-06
Surface JSON parse context in table-source logs

Karavil:
## Summary\n- wrap table-source row hydration in guarded conversions so JSON parsing errors log table and column metadata\n- raise JSONColumnParseError with raw value snippet to make SyntaxError root causes debuggable\n- log unsupported value conversions against the current table for parity\n\n## Testing\n- goblins lint --files=".external/zero-mono/packages/zqlite/src/table-source.ts" *(fails: existing lint warnings about assignment expressions)*\n

### DIFF
--- a/packages/zqlite/src/table-source.ts
+++ b/packages/zqlite/src/table-source.ts
@@ -56,6 +56,8 @@ type Statements = {
 
 let eventCount = 0;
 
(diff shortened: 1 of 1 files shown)


Karavil:
superseded by #4973 with isolated JSON logging changes

PR #4975, opened 2025-10-07
chore(zero): Rename analyze-query result fields for clarity

arv:
Renames fields in `AnalyzeQueryResult` to be more descriptive:

`vendedRowCounts` -> `readRowCountsByQuery`
`vendedRows` -> `readRows`
Added `readRowCount` for total rows read

Old fields are deprecated but still work for backward compatibility. Protocol version bumped to 35.

"Vended" was unclear - "read" better describes what these fields represent (rows read from the database during query execution).

Closes https://bugs.rocicorp.dev/p/zero/issue/4108

### DIFF
--- a/packages/analyze-query/src/bin-analyze.ts
+++ b/packages/analyze-query/src/bin-analyze.ts
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import chalk from 'chalk';
 import fs from 'node:fs';
 import {astToZQL} from '../../ast-to-zql/src/ast-to-zql.ts';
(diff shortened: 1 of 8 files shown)


arv:
> Did you ever implement support for analyzing arbitrary queries? That would be really useful and would allow us to get rid of the analyze query cli tool.

No. It should be straight forward to add. The RPC endpoint already supports it. It is just a matter of extracting the AST.

https://github.com/rocicorp/mono/blob/b481eea051be2f46deb8924fb8da57f4ba38cf77/packages/zero-protocol/src/inspect-up.ts#L43-L47

arv:
> No. It should be straight forward to add. The RPC endpoint already supports it. It is just a matter of extracting the AST.

I looked briefly at this. The backend needs more work.

When you get a `Query` (as in `zero.query.issue`) on the client we have the client side AST. For analyzeQuery I think what we want is the server side AST. For legacy ad-ho queries those are the same but for named/synced queries we want to analyze the server side AST. I will need to pass the name/args to the RPC end point and have the RPC endpoint call the API server to get the AST.

PR #4976, opened 2025-10-08
fix(zero): do not load a module that loads zero-sqlite3

tantaman:
Bun users would get errors when including `@rocicorp/zero/server` because it would import an error from mutagen which imported zero-sqlite3.

My 2 cents: we should create and enforce explicit package boundaries (making it impossible to zero/server to import from zero-cache) but its a bit late to claw that one back.

https://discord.com/channels/830183651022471199/1425144484337025104/1425144488287801556

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,3 +92,4 @@ WARP.md
 *.db-wal
 
 1gb
+bun.lock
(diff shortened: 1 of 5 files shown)


arv:
LGTM

We do have ways to restrict import across boundaries using our lint rules.

In this case, would the rule be that packages/zero-server/ cannot import from packages/zero-cache?

Should the error file be moved to a shared location so that it can be used from zero-cache?

PR #4978, opened 2025-10-08
fix(zero-client): refetch() api doesn't need to be |undefined.

aboodman:


tantaman:
lgtm.

tantaman:
initial thought was to only allow refetch on error since why else would you ever refetch?

PR #4979, opened 2025-10-08
Revert "chore(zero): Rename analyze-query result fields for clarity"

arv:
Reverts rocicorp/mono#4975


arv:
I had to revert this because we cannot have protocol version changes until https://github.com/rocicorp/mono/actions/runs/18063743288/job/51403670511#step:12:148 is fixed.

PR #5094, opened 2025-10-28
feat(zero-cache): adopt ProtocolError and propagate push/transform errors

0xcadams:
* Changed the `fetchFromAPIServer` helper to emit `ProtocolError` w/ the different failure paths (HTTP error, parsing, unknown failures)
* Updated transform/push paths to emit `ProtocolError`
* Refactor `ErrorForClient` to use `ProtocolError`


packages/zero-cache/src/workers/connection.test.ts
```diff
@@ -1,30 +1,59 @@
...
+import {ErrorKind} from '../../../zero-protocol/src/error-kind.ts';
+import {ErrorOrigin} from '../../../zero-protocol/src/error-origin.ts';
+import {ProtocolErrorWithLevel} from '../types/error-with-level.ts';
 import {send} from './connection.ts';
 
 class MockSocket implements Pick<WebSocket, 'readyState' | 'send'> {
   readyState: WebSocket['readyState'] = WebSocket.OPEN;
-  send(_message: string) {}
+
+  send(data: RawData, cb?: (err?: Error) => void): void;
```
arv: Maybe restrict this to the standard WebSocket interface instead of the extended `npm:ws` interface.

arv (approved):
LGTM

packages/zero-cache/src/custom-queries/transform-query.ts
```diff
@@ -160,7 +160,7 @@ export class CustomQueryTransformer {
         message:
           e instanceof Error
             ? e.message
-            : 'An unknown error occurred while transforming queries.',
+            : `An unknown error occurred while transforming queries: ${String(e)}`,
```
arv, line 163: No need to wrap in `String(e)` here ...

We should fix the linter.

arv (commented):
LGTM

packages/zero-cache/src/types/error-for-client.ts
```diff
@@ -1,41 +0,0 @@
...
-
-export function getLogLevel(error: unknown): LogLevel {
-  return error instanceof ErrorWithLevel ? error.logLevel : 'error';
-}
-
-export class ErrorForClient extends ErrorWithLevel {
-  readonly errorBody;
-  constructor(
-    errorBody: ErrorBody,
-    logLevel: LogLevel = 'warn', // 'warn' by default since these are generally not server issues
```
darkgnotic, line 25: Hmmm ... IIUC we lost this default in this refactoring. I think this means that the zero-cache logs will have a lot more "error" level messages for situations that are not really zero-cache issues.

@0xcadams do you have a recommendation for adding back an API to capture this intent? (i.e. a protocol error to convey to the client that's not really a problem on the server, but something we should still be alerted about, hence the default to 'warn').

I guess one option is explicitly set these to 'warn' ...  😬 

Let me know your thoughts!   🙏

packages/zero-cache/src/services/view-syncer/cvr-store.ts
```diff
@@ -925,18 +928,22 @@ export async function checkVersion(
...
 
-export class ConcurrentModificationException extends ErrorWithLevel {
+export class ConcurrentModificationException extends ProtocolErrorWithLevel {
   readonly name = 'ConcurrentModificationException';
 
   constructor(expectedVersion: string, actualVersion: string) {
     super(
-      `CVR has been concurrently modified. Expected ${expectedVersion}, got ${actualVersion}`,
+      {
+        kind: ErrorKind.Internal,
```
tantaman, line 937: why did we change this to `Internal`?

tantaman: I'm asking because of a customer getting "CVR Concurrent Modification Exceptions"

https://rocicorp.slack.com/archives/C013XFG80JC/p1778185039370299?thread_ts=1778182430.166939&cid=C013XFG80JC

Their client gets stuck and doesn't un-stick till a refresh. IIUC, prior to this change the websocket would close and the client would auto-reconnect.

0xcadams: Oh yeah that's kinda the spec we went with. We have some exceptions to this but generally any zero-cache error results in error state on the client and the app has to reconnect. We only have a few exceptions like Rehome errors which result in a retry on the client

0xcadams: It is odd though for errors like this. It's a zero internal error that's not expected, but we still treat it as if the app needs to handle it.

darkgnotic: Ah, I guess there was a miscommunication there. This particular error not an internal error. When two view-syncers operate on the same client group ID, thisis the "control flow" error by which we resolve that and redirect one of the browsers to the other view-syncer.

PR #5150, opened 2025-11-07
refactor(zql): Update Join and FlippedJoin to always fetch parent with partition constraint.

grgbkr:


packages/zql/src/ivm/join-utils.ts
```diff
@@ -126,3 +127,112 @@ export function isJoinMatch(
...
+   * @param storage The underlying key-value storage implementation.
+   * @param name A unique name for this set (used as the first part of the key).
+   * @param setKey The row properties that define the "partition" or "set".
+   * @param primaryKey The row properties that uniquely identify an entry *within* the set.
+   * @param valueKey (Optional) The row properties to be stored *in* the key,
+   * which can be retrieved with `getValues`.
+   */
+  constructor(
+    storage: Storage,
+    name: string,
```
tantaman, line 177: why do we need a unique name? I thought all operator storage was already isolated?

packages/zql/src/ivm/join-utils.ts
```diff
@@ -126,3 +127,112 @@ export function isJoinMatch(
...
+        primaryKeyValues,
+      ]);
+    }
+    const valueKeyValues: Value[] = [];
+    for (const key of this.#valueKey) {
+      valueKeyValues.push(row[key]);
+    }
+    return KeySet.#makeKeySetStorageKeyForValues(this.#name, [
+      setKeyValues,
+      valueKeyValues,
```
tantaman, line 246: why do valueKeyValues come before primaryKeyValues?

packages/zql/src/ivm/join-utils.ts
```diff
@@ -126,3 +127,112 @@ export function isJoinMatch(
...
+    name: string,
+    setKey: CompoundKey,
+    primaryKey: CompoundKey,
+    valueKey: V,
+  ) {
+    this.#storage = storage;
+    this.#name = name;
+    this.#setKey = setKey;
+    this.#primaryKey = primaryKey;
+    this.#valueKey = valueKey;
```
tantaman, line 186: is it expected that these columns will never overlap? E.g., `id` will not be present twice or three times in `setKey, primaryKey, valueKey`?

packages/zql/src/ivm/join-utils.ts
```diff
@@ -126,3 +127,112 @@ export function isJoinMatch(
...
+  add(row: Row): void {
+    this.#storage.set(this.#makeKeySetStorageKey(row), true);
+  }
+
+  delete(row: Row): void {
+    this.#storage.del(this.#makeKeySetStorageKey(row));
+  }
+
+  *getValues(row: Row): Iterable<{
+    readonly [key: string]: Value;
```
tantaman, line 198: what is this `key` here? It doesn't look like it would be a column named based on types (`V extends CompoundKey`)

packages/zql/src/ivm/join-utils.ts
```diff
@@ -126,3 +127,112 @@ export function isJoinMatch(
   }
   return true;
 }
+
+export class KeySet<V extends CompoundKey | undefined> {
```
tantaman, line 160: a unit test for this class would help to understand how it is intended to be used.

packages/zql/src/ivm/join.ts
```diff
@@ -276,37 +314,16 @@ export class Join implements Input {
...
-          const empty =
-            [
-              ...take(
-                this.#storage.scan({
-                  prefix: makeStorageKeyPrefix(parentNodeRow, this.#parentKey),
-                }),
-                1,
-              ),
-            ].length === 0;
+          this.#primaryKeySet.delete(parentNodeRow);
```
tantaman, line 300: should we go ahead and remove `cleanup` while we're here? If it isn't a simple matter of deleting the cleanup section then in a follow up pr?

PR #5297, opened 2025-12-08
chore(zero-cache): add error states for got queries

grgbkr:


packages/zero-client/src/client/zero-poke-handler.test.ts
```diff
@@ -1350,12 +1350,18 @@ describe('poke handler', () => {
           {
             op: 'put',
             key: 'd/c1/h1',
-            value: null,
+            value: {
```
grgbkr, line 1333: test with an errorVersion

PR #5380, opened 2025-12-28
feat(replicache): Add bulk insertion optimization with putMany

arv:
# feat(replicache): Add bulk insertion optimization with putMany

## Overview

This PR adds bulk insertion optimization to Replicache's BTree and database layer, significantly improving performance for large batch operations like sync patches.

### Core BTree Changes

**`packages/replicache/src/btree/node.ts`**

- Added `putMany()` method to `DataNodeImpl` for efficient merging of sorted entries
- Added `putMany()` method to `InternalNodeImpl` with child grouping and rebalancing
- Added `putManyMergeAndPartition()` helper for node rebalancing during bulk operations
- Extracted `binarySearchFrom()` to enable optimized searching from a start index
- Refactored `readTreeData()` to accept `getEntrySize` parameter and moved it outside the `describe` block (test helper improvement)

**`packages/replicache/src/btree/write.ts`**

- Added `BTreeWrite.putMany()` method with two optimized paths:
  - **Fast path**: Bulk loads empty trees bottom-up using optimal partitioning
  - **Slow path**: Merges entries into existing trees with efficient rebalancing
- Validates entries are sorted and converts to sized entries in a single pass
- Reuses arrays to minimize allocations during tree construction

### Database Layer

**`packages/replicache/src/db/write.ts`**

- Added `Write.putMany()` method that delegates to `BTreeWrite.putMany()`
- Handles index updates for all entries before bulk insertion
- Maintains compatibility with existing `put()` semantics

### Sync Layer Optimization

**`packages/replicache/src/sync/patch.ts`**

- Added `optimizePatch()` function to eliminate redundant operations:
  - Drops all operations before the last `clear`
  - For each key, a later `put` or `del` replaces all earlier operations on that key
  - Removes standalone `del` operations after a `clear` (deleting from an empty tree is a no-op)
  - If any `update` operation is present, returns the original patch unmodified (from the last `clear` onward) — `update` is not used in Zero/Replicache, so optimization is skipped to stay safe
  - Sorts the resulting operations by key for optimal bulk loading
- Modified `apply()` to use optimized patches with bulk loading
- Extracted `mergeUpdate()` helper for update operation handling
- Added `bulkLoadPuts()` to handle the leading run of consecutive `put` operations efficiently

### Other

**`packages/zero/package.json`**

- Removed unused `zero-protocol` devDependency

## Performance Impact

The optimization targets common sync patterns:

1. **Initial sync**: Loading thousands of entries into an empty tree
2. **Large patches**: Applying batches of updates from the server
3. **Snapshot application**: Replacing entire datasets

### Benchmark Results

Comparison of `putMany()` vs sequential `put()` operations:

| Scenario          | Entries | Value Size | Speedup    |
| ----------------- | ------- | ---------- | ---------- |
| Empty tree        | 100     | small      | **3.36x**  |
| Empty tree        | 100     | large      | **1.49x**  |
| Empty tree        | 1,000   | small      | **5.30x**  |
| Empty tree        | 1,000   | large      | **1.58x**  |
| Empty tree        | 10,000  | small      | **4.15x**  |
| Empty tree        | 10,000  | large      | **1.13x**  |
| Construction only | 10,000  | small      | **53.73x** |
| Update existing   | 1,000   | mixed      | **4.47x**  |

**Key findings:**

- Small values show 3-5x speedup consistently
- Construction-only (no flush) shows dramatic 53x improvement
- Large values benefit less due to serialization overhead
- Updating existing entries shows 4.5x improvement

Additional benefits:

- Reduces chunk writes through optimal tree construction
- Minimizes redundant operations through patch optimization

## Testing

**New test files:**

- `packages/replicache/src/btree/write.bench.ts` - Performance benchmarks comparing sequential put() vs putMany()
- Extensive test coverage in:
  - `packages/replicache/src/btree/node.test.ts` - 17 new tests for putMany() behavior
  - `packages/replicache/src/db/write.test.ts` - 3 new tests for database-level putMany()
  - `packages/replicache/src/sync/patch.test.ts` - 24 new tests for patch optimization

**Test scenarios covered:**

- Empty tree bulk loading
- Merging with existing entries
- Tree rebalancing and partitioning
- Index updates
- Update operation merging
- Patch optimization edge cases

## Compatibility

- No breaking changes to public APIs
- Existing `put()` and `del()` methods remain unchanged
- `putMany()` is an additive optimization that can be adopted incrementally
- Works with both FormatVersion.V6 and FormatVersion.V7

### Key Algorithm Improvements

1. **Bottom-up tree construction**: When building from scratch, constructs the optimal tree structure in a single pass
2. **Batch rebalancing**: Groups entries by affected child node and rebalances once per group
3. **Restricted binary search**: Uses previous search results to narrow search ranges for sorted input
4. **Patch deduplication**: Eliminates redundant operations before applying to the tree

### Memory Efficiency

- Reuses arrays during tree construction to minimize allocations
- Mutates entries in-place during tree node creation (for immutable node pattern)
- Batch processes operations to reduce intermediate tree states

## Future Work

- Consider adding `delMany()` for bulk deletions
- Explore parallel index updates for large batches

### DIFF
--- a/packages/replicache/src/btree/node.test.ts
+++ b/packages/replicache/src/btree/node.test.ts
@@ -37,6 +37,45 @@ import {BTreeWrite} from './write.ts';
 
 type FormatVersion = Enum<typeof FormatVersion>;
 
+type TreeData = {
+  $level: number;
+  [key: string]: TreeData | ReadonlyJSONValue;
+};
+
+async function readTreeData(
+  rootHash: Hash,
+  dagRead: Read,
+  formatVersion: FormatVersion,
+  getEntrySize: <K, V>(k: K, v: V) => number,
+): Promise<Record<string, unknown>> {
+  const chunk = await dagRead.getChunk(rootHash);
+  const node = parseBTreeNode(chunk?.data, formatVersion, getEntrySize);
+  let lastKey: string | undefined;
+  const rv: Record<string, unknown> = {
+    $level: node[NODE_LEVEL],
+  };
+
+  if (node[NODE_LEVEL] === 0) {
+    for (const [k, v] of node[NODE_ENTRIES]) {
+      if (lastKey !== undefined) {
+        assert(lastKey < k, () => `Expected lastKey ${lastKey} < key ${k}`);
+        lastKey = k;
+      }
+      rv[k] = v;
+    }
+    return rv;
+  }
+
+  for (const [k, hash] of (node as InternalNode)[NODE_ENTRIES]) {
+    if (lastKey !== undefined) {
+      assert(lastKey < k, () => `Expected lastKey ${lastKey} < key ${k}`);
+      lastKey = k;
+    }
+    rv[k] = await readTreeData(hash, dagRead, formatVersion, getEntrySize);
+  }
+  return rv;
+}
+
 describe('btree node', () => {
   function createSizedEntry<K, V>(
     key: K,
@@ -44,11 +83,6 @@ describe('btree node', () => {
   ): [key: K, value: V, sizeOfEntry: number] {
     return [key, value, getEntrySize(key, value)];
   }
-
-  type TreeData = {
-    $level: number;
-    [key: string]: TreeData | ReadonlyJSONValue;
-  };
 
   function makeTree(
     node: TreeData,
@@ -99,39 +133,6 @@ describe('btree node', () => {
     }
   }
 
-  async function readTreeData(
-    rootHash: Hash,
-    dagRead: Read,
-    formatVersion: FormatVersion,
-  ): Promise<Record<string, unknown>> {
-    const chunk = await dagRead.getChunk(rootHash);
-    const node = parseBTreeNode(chunk?.data, formatVersion, getEntrySize);
-    let lastKey: string | undefined;
-    const rv: Record<string, unknown> = {
-      $level: node[NODE_LEVEL],
-    };
-
-    if (node[NODE_LEVEL] === 0) {
-      for (const [k, v] of node[NODE_ENTRIES]) {
-        if (lastKey !== undefined) {
-          assert(lastKey < k, () => `Expected lastKey ${lastKey} < key ${k}`);
-          lastKey = k;
-        }
-        rv[k] = v;
-      }
-      return rv;
-    }
-
-    for (const [k, hash] of (node as InternalNode)[NODE_ENTRIES]) {
-      if (lastKey !== undefined) {
-        expect(lastKey < k);
-        lastKey = k;
-      }
-      rv[k] = await readTreeData(hash, dagRead, formatVersion);
-    }
-    return rv;
-  }
-
   async function expectTree(
     rootHash: Hash,
     dagStore: Store,
@@ -139,9 +140,9 @@ describe('btree node', () => {
     expected: TreeData,
   ) {
     await withRead(dagStore, async dagRead => {
-      expect(await readTreeData(rootHash, dagRead, formatVersion)).toEqual(
-        expected,
-      );
+      expect(
+        await readTreeData(rootHash, dagRead, formatVersion, getEntrySize),
+      ).toEqual(expected);
     });
   }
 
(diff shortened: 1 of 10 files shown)


rajczi:
We've now been using this build of Replicache in our expo react native mobile app using the sqlite kvStore against op-sqlite@15.2 for quite a while and are very happy with it.  It is saving us over half of our initial sync snapshot time (download complete to ready).

rajczi:
We've discovered an issue with this branch where the output of watchers doesn't match the patch from the server.  I've prepared a unit test to demonstrate the issue which passes on main and fails here.  I will send it to @arv.

packages/replicache/src/btree/node.test.ts
```diff
@@ -1704,3 +1705,849 @@ describe('Write nodes using ChainBuilder', () => {
...
+      }
+
+      const structure1 = await getTreeStructure(hash1, dagStore1);
+      const structure2 = await getTreeStructure(hash2, dagStore2);
+
+      // The structures should be identical
+      expect(structure2).toEqual(structure1);
+
+      // Also verify both trees have the same data
+      await withRead(dagStore1, async dagRead1 => {
```
grgbkr: instead of nesting the reads like this maybe a helper, and do them in sequence
```

function checkContents(dagStore, hash) {
    return withRead(dagStore, async dagRead => {
          const tree = new BTreeRead(
            dagRead,
            formatVersion,
            hash,
            getEntrySize,
            chunkHeaderSize,
          );
          for (let i = 0; i < 500; i++) {
            const key = `key${i.toString().padStart(4, '0')}`;
            expect(await tree.get(key)).toBe(i);
          }
    });
}
await checkContents(dagStore1, hash1);
await checkContents(dagStore2, hash2);

````

File: packages/replicache/src/btree/node.test.ts [hunk 2]
```diff
@@ -1704,3 +1705,849 @@ describe('Write nodes using ChainBuilder', () => {
...
+  const minSize = 8 * 1024;
+  const maxSize = 16 * 1024;
+  const chunkHeaderSize = NODE_HEADER_SIZE;
+
+  function getEntrySize<K, V>(key: K, value: V): number {
+    return getSizeOfEntry(key, value);
+  }
+
+  for (const formatVersion of [FormatVersion.V6, FormatVersion.V7] as const) {
+    test(`putMany empty entries > v${formatVersion}`, async () => {
```
grgbkr, line 1719: instead of having all the `> v${formatVersion}`, you could do
```
for (const formatVersion of [FormatVersion.V6, FormatVersion.V7] as const) {
  describe(`v${formatVersion}`, () => {
    test(`putMany empty entries`, async () => {
    });
    // etc
  });
}
```

packages/replicache/src/btree/node.test.ts
```diff
@@ -1704,3 +1705,849 @@ describe('Write nodes using ChainBuilder', () => {
...
+          for (let i = 0; i < 500; i++) {
+            const key = `key${i.toString().padStart(4, '0')}`;
+            expect(await tree1.get(key)).toBe(i);
+            expect(await tree2.get(key)).toBe(i);
+          }
+        });
+      });
+    });
+
+    test(`putMany triggers merge and partition > v${formatVersion}`, async () => {
```
grgbkr, line 2094: How do we know in the tests that a merge or partition are being triggered?

Comments describing how these are triggering the various merges/partitions could help to clarify.

packages/replicache/src/btree/write.ts
```diff
@@ -135,6 +135,121 @@ export class BTreeWrite extends BTreeRead {
...
+          // Partition parent entries
+          const parentPartitions = partition(
+            currentLevel as Entry<Hash>[],
+            e => e[2],
+            contentMin,
+            contentMax,
+          );
+
+          // Create internal nodes
+          currentLevel = parentPartitions.map(entries =>
```
grgbkr, line 218: Above you comment `reuse array to avoid allocations`, but doesn't this map create a new array?

Should this be

```
for (let i = 0; i< entries.length; i++ ){
  currentLevel[i] = this.newInternalNodeImpl(entries, level);
}
currentLevel.length = entries.length;
```

Or maybe track the logical length of currentLevel in its own variable instead of resizing the array.

```
for (let i = 0; i< entries.length; i++ ){
  currentLevel[i] = this.newInternalNodeImpl(entries, level);
}
currentLevelLength= entries.length;
```

packages/replicache/src/btree/write.ts
```diff
@@ -135,6 +135,121 @@ export class BTreeWrite extends BTreeRead {
...
+          );
+        }
+
+        this.rootHash = (
+          currentLevel[0] as DataNodeImpl | InternalNodeImpl
+        ).hash;
+        return;
+      }
+
+      // Slow path: merge with existing tree
```
grgbkr, line 229: Is this slow path still faster than putting each entry with a separate put call?

packages/replicache/src/db/write.test.ts
```diff
@@ -327,3 +329,219 @@ test('mutationID on newWriteLocal', async () => {
...
+        deepFreeze({id: i}),
+      ]);
+    }
+
+    await w.putMany(lc, entries);
+    await w.commit(DEFAULT_HEAD_NAME);
+  });
+
+  // Verify entries are accessible
+  await withWriteNoImplicitCommit(ds, async dagWrite => {
```
grgbkr, line 525: maybe just use a dagRead to read contents?

packages/replicache/src/db/write.ts
```diff
@@ -84,6 +84,32 @@ export class Write extends Read {
     await this.map.put(key, value);
   }
 
+  /**
+   * Efficiently insert multiple entries. Uses BTreeWrite.putMany which
+   * automatically uses fromEntries for empty trees or multiPut for existing
```
grgbkr: What do fromEntries and multiPut reference?  I can find these functions.

packages/replicache/src/db/write.ts
```diff
@@ -84,6 +84,32 @@ export class Write extends Read {
...
+    lc: LogContext,
+    entries: ReadonlyArray<readonly [string, FrozenJSONValue]>,
+  ): Promise<void> {
+    if (entries.length === 0) {
+      return;
+    }
+
+    // Update indexes for all entries
+    if (this.indexes.size > 0) {
+      // TODO(arv): Indexes can use the optimizePatch pattern too.
```
grgbkr, line 110: Is this worth doing?

packages/replicache/src/btree/node.test.ts
```diff
@@ -1704,3 +1705,849 @@ describe('Write nodes using ChainBuilder', () => {
...
+describe('BTreeWrite.putMany', () => {
+  const minSize = 8 * 1024;
+  const maxSize = 16 * 1024;
+  const chunkHeaderSize = NODE_HEADER_SIZE;
+
+  function getEntrySize<K, V>(key: K, value: V): number {
+    return getSizeOfEntry(key, value);
+  }
+
+  for (const formatVersion of [FormatVersion.V6, FormatVersion.V7] as const) {
```
grgbkr, line 1718: Can you remind me why we need to test both format versions, and not just the latest?

packages/replicache/src/sync/patch.test.ts
```diff
@@ -352,3 +352,386 @@ describe('patch', () => {
...
+      {op: 'put', key: 'a', value: 1},
+      {op: 'put', key: 'b', value: 2},
+      {op: 'del', key: 'a'}, // del replaces put
+      {op: 'clear'}, // Everything above gets dropped
+      {op: 'put', key: 'c', value: 3},
+      {op: 'put', key: 'd', value: 4},
+      {op: 'put', key: 'c', value: 5}, // Overwrites c=3
+      {op: 'del', key: 'e'}, // Del after clear without prior put, removed
+      {op: 'put', key: 'f', value: 6},
+      {op: 'del', key: 'f'}, // del replaces put
```
grgbkr, line 460: del replace put, and since the del is after a clear it is removed?

packages/replicache/src/sync/patch.test.ts
```diff
@@ -352,3 +352,386 @@ describe('patch', () => {
...
+
+  test('update operations orders position', () => {
+    const patch: PatchOperationInternal[] = [
+      {op: 'put', key: 'b', value: {b: 1}},
+      {op: 'put', key: 'a', value: {a: 2}},
+      {op: 'update', key: 'a', merge: {a2: 3}},
+      {op: 'update', key: 'b', merge: {b2: 4}},
+    ];
+    // put + update should be merged for 'a' (object value)
+    // but 'b' keeps separate update (non-object value)
```
grgbkr: This comment doesn't match the test behavior.  I'm unsure if the test name matches the test.

packages/replicache/src/sync/patch.test.ts
```diff
@@ -352,3 +352,386 @@ describe('patch', () => {
...
+  test('complex optimization', () => {
+    const patch: PatchOperationInternal[] = [
+      {op: 'put', key: 'a', value: 1},
+      {op: 'put', key: 'b', value: 2},
+      {op: 'del', key: 'a'}, // del replaces put
+      {op: 'clear'}, // Everything above gets dropped
+      {op: 'put', key: 'c', value: 3},
+      {op: 'put', key: 'd', value: 4},
+      {op: 'put', key: 'c', value: 5}, // Overwrites c=3
+      {op: 'del', key: 'e'}, // Del after clear without prior put, removed
```
grgbkr: Why is `without prior put` important?, it seems the del of f below id also removed despite their being a prior put.

packages/replicache/src/sync/patch.test.ts
```diff
@@ -352,3 +352,386 @@ describe('patch', () => {
...
+      {op: 'put', key: 'a', value: {b: 1}},
+      {op: 'update', key: 'a', merge: {c: 2}},
+      {op: 'update', key: 'a', merge: {d: 3}},
+      {op: 'del', key: 'a'},
+    ];
+    expect(optimizePatch(patch)).toEqual([{op: 'del', key: 'a'}]);
+  });
+
+  test('updates without initial put', () => {
+    const patch: PatchOperationInternal[] = [
```
grgbkr: could this be merged to
`{op: 'update', key: 'a', merge: {b: 1, c:2}}`

packages/replicache/src/sync/patch.test.ts
```diff
@@ -352,3 +352,386 @@ describe('patch', () => {
...
+      {op: 'update', key: 'a', merge: {b: 1}},
+      {op: 'update', key: 'a', merge: {c: 2}},
+    ];
+    expect(optimizePatch(patch)).toEqual(patch);
+  });
+
+  test('removes del after clear without put', () => {
```
grgbkr: what is meant by `without put`?

the del b is removed even though there is a put b.

i think it is safe to remove it.

packages/replicache/src/sync/patch.test.ts
```diff
@@ -352,3 +352,386 @@ describe('patch', () => {
...
+    const patch: PatchOperationInternal[] = [
+      {op: 'clear'},
```
grgbkr, line 512: Suggest testing the sequence:

{op: 'clear'},
{op: 'del', key: 'b'},
{op: 'put', key: 'b', value: 1},

packages/replicache/src/sync/patch.test.ts
```diff
@@ -352,3 +352,386 @@ describe('patch', () => {
...
+  });
+
+  test('maintains relative order of operations', () => {
+    const patch: PatchOperationInternal[] = [
+      {op: 'put', key: 'a', value: 1},
+      {op: 'put', key: 'b', value: 2},
+      {op: 'put', key: 'c', value: 3},
+    ];
+    const optimized = optimizePatch(patch);
+    // Order should be preserved
```
grgbkr: Is this true, doesn't it sort by key?  That happens to preserve order here since they were already sorted by key.

packages/replicache/src/sync/patch.test.ts
```diff
@@ -352,3 +352,386 @@ describe('patch', () => {
...
+    expect(optimizePatch(patch)).toEqual([{op: 'clear'}]);
+  });
+});
+
+describe('patch with optimization', () => {
+  const formatVersion = FormatVersion.Latest;
+  const clientID = 'client-id';
+  const lc = new LogContext();
+
+  test('optimized patch produces same result as non-optimized', async () => {
```
grgbkr: Maybe just
'applying optimized patch produced correct results'

its not really comparing to applying non-optimized (there doesn't seem to be a way to apply non-optimized anymore since apply does it by default).

packages/replicache/src/sync/patch.test.ts
```diff
@@ -352,3 +352,386 @@ describe('patch', () => {
...
+        b.chain[0].chunk.hash,
+        {[clientID]: 1},
+        'cookie',
+        dagWrite,
+        clientID,
+        formatVersion,
+      );
+      await apply(lc, dbWrite, patch);
+
+      expect(await dbWrite.get('a')).toBe(2);
```
grgbkr, line 592: Maybe these test should use scan to read and assert on entire content of db?

packages/replicache/src/sync/patch.ts
```diff
@@ -1,76 +1,216 @@
...
   | {
       op: 'clear';
     };
 
+/**
+ * Optimizes a patch array by:
+ * 1. Dropping all operations before the last 'clear'
+ * 2. For each key: put/del replace all previous operations; updates accumulate
+ * 3. Removing standalone 'del' operations after a clear (deleting from empty tree)
+ * 4. Merging updates after puts into a single put operation
```
grgbkr: why can't updates that don't have a put before them be merged?

packages/replicache/src/sync/patch.ts
```diff
@@ -1,76 +1,216 @@
...
+ * Note: Order is preserved for operations on the same key, but operations
```
grgbkr, line 29: If we also merged updates, would there always be just one operation per key?

packages/replicache/src/sync/patch.ts
```diff
@@ -1,76 +1,216 @@
...
+  const entries: [string, FrozenJSONValue][] = puts.map(p => {
+    if (p.op !== 'put') {
+      throw new Error('Expected put operation');
+    }
+    return [p.key, deepFreeze(p.value)];
+  });
+
+  // already sorted
+
+  // Use putMany which will use BTreeWrite.fromEntries if the map is empty
```
grgbkr: What is `BTreeWrite.fromEntries`?

packages/replicache/src/sync/patch.ts
```diff
@@ -1,76 +1,216 @@
...
+      }
+      case 'update': {
+        // update accumulates with previous operations
+        const existing = keyOps.get(p.key);
+        if (existing) {
+          // Merge with existing put if possible
+          if (existing.length === 1 && existing[0].op === 'put') {
+            const {value} = existing[0];
+            assertObject(value);
+            const merged = mergeUpdate(p, value);
```
grgbkr: It would be good to add some tests around merging updates that have a constrain property.  I feel uncertain if those are being merged correctly.

grgbkr (commented):
Very nice optimization work!

packages/replicache/src/btree/node.test.ts
arv: I'm also thinking it is time to remove the old format.

arv: I'm also thinking it is time to remove the old format.

packages/replicache/src/db/write.ts
arv: Out of date comment. 

Thanks

packages/replicache/src/db/write.ts
arv: I looked at this and the win would be huge. But we are going away from indexes so that is why it is a todo. If people want to continue to use replicache indexes we should do this.

packages/replicache/src/sync/patch.test.ts
arv: If there is a del after the put the put can be ignored.

If there is a del after clear the del can be removed.

If there is a put after a put, the first put can be replaced. 

If there is a del after a del it can be ignored/replaced.

And a few more cases for update/merge.

arv: The main reason to do this optimization is to be able to use more multiPut.

packages/replicache/src/sync/patch.test.ts
arv: Not sure. I need to go through these comments. Thanks.

packages/replicache/src/sync/patch.ts
arv: A put + merge -> put

I don't recall if I did

merge + merge -> merge

But it should be correct to do.

packages/replicache/src/sync/patch.ts
arv: I think that is correct. 

In theory we could get a del followed by a merge but I have to check again if that is an error or ignored.

packages/replicache/src/sync/patch.ts
arv: Old reference. I removed it and dealt with it without a new API.

packages/replicache/src/sync/patch.ts
arv: Good call.

packages/replicache/src/btree/write.ts
arv: I added a todo, partition should create the correct objects directly

packages/replicache/src/btree/write.ts
arv: Yes It allows a single pass through the tree instead of n passes (where n is the number of items being inserted)

packages/replicache/src/db/write.test.ts
arv: The test is ensuring that we can do gets in the write transaction.

packages/replicache/src/btree/node.test.ts
arv: cargo culting?

packages/replicache/src/sync/patch.test.ts
arv: Bad comment. After a clear a free del is dropped

packages/replicache/src/sync/patch.test.ts
arv: This tests that case. I think at some point I incorrectly didn't remove the del. Improved the test name.

packages/replicache/src/sync/patch.test.ts
arv: sorts by key.

packages/replicache/src/sync/patch.ts
arv: We can merge multiple updates in almost all cases. The only case is when you get 

```
{op: update, key, merge, constrain: ['a']}
{op: update, key, merge, constrain: ['b']}
```

which should lead to an empty object but an empty array `[]` is the same as no constraint.

arv:
@grgbkr I removed support for optimizing "update" ops. If there is one then we do not optimize.

We should probably remove all support for "update" ops in zero and replicache since we no longer use them anywhere

PR #5392, opened 2026-01-08
publish `zeroForTest` as potential alternative to #5225

tantaman:
Trying to accomplish the same goal as https://github.com/rocicorp/mono/pull/5225 but with our existing test harnesses.


apps/zbugs/src/local-mutate.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        name: args.name,
+        role: args.role,
+        avatar: '',
+      });
+    }),
+  },
+});
+
+test('local mutate', async () => {
+  const zero = zeroForTest({
```
tantaman, line 40: example of how you could use `zeroForTest`

apps/zbugs/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "jsx": "react-jsx",
     "paths": {
       "@rocicorp/zero": ["../../packages/zero/src/zero.ts"],
-      "@rocicorp/zero/react": ["../../packages/zero/src/react.ts"]
+      "@rocicorp/zero/react": ["../../packages/zero/src/react.ts"],
+      "@rocicorp/zero/test-helpers": ["../../packages/zero/src/test-helpers.ts"]
```
tantaman: idk why we need these in zbugs. presumably you won't need this in an external project.

apps/zbugs/vitest.config.ts
```diff
@@ -19,6 +19,10 @@ export function configForVersion(version: number, url: string) {
     plugins: [tsconfigPaths()],
     resolve: {
       alias: [
+        {
+          find: '@rocicorp/zero/test-helpers',
+          replacement: resolve(packagesDir, 'zero/src/test-helpers.ts'),
```
tantaman: 🫠

presumably this is only needed in our mono-repo and not by actual users

packages/zero/src/testing.ts
```diff
@@ -0,0 +1 @@
+export {zeroForTest} from '../../zero-client/src/client/test-utils.ts';
```
tantaman: @kvnkusch -  did you try `zeroForTest`? / does it do what you need?

apps/zbugs/src/local-mutate.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    mutators: {
+      ...mutators,
+      ...mutatorsForTest,
+    },
+    userID: 'user-1',
+    context: {sub: 'user-1', role: 'user'} as AuthData,
+  });
+
+  await zero.mutate(
+    // @ts-ignore - wtf?
```
tantaman: I have no idea what is going on here...

tantaman:
@arv / @0xcadams  - @kvnkusch is trying to use this but hitting errors with properties referenced by the testing symbol being undefined.

```
test/zeroForTest-repro.test.ts:
23 |   get connectingStart() {
24 |     return this[exposedToTestingSymbol].connectStart;
25 |   }
26 |   constructor(options) {
27 |     super(options);
28 |     this[exposedToTestingSymbol].connectionManager().subscribe((state) => {
              ^
TypeError: undefined is not an object (evaluating 'this[exposedToTestingSymbol].connectionManager')
      at new TestZero (/Users/kkusch/Code/ccs/cc1/node_modules/@rocicorp/zero/out/zero-client/src/client/test-utils.js:28:10)
      at zeroForTest (/Users/kkusch/Code/ccs/cc1/node_modules/@rocicorp/zero/out/zero-client/src/client/test-utils.js:174:10)
      at <anonymous> (/Users/kkusch/Code/ccs/cc1/packages/core/test/zeroForTest-repro.test.ts:23:15)
```

iirc, you guys fixed something like this recently? and that it has to do with our build?

apps/zbugs/tsconfig.json
arv: Without the paths the lookup would be through node_modules which would  require rebuilding packages/zero every time something changes.

packages/zero/tsconfig.server.json
```diff
@@ -14,6 +14,7 @@
     "src/zqlite.ts",
     "src/analyze-query.ts",
     "src/transform-query.ts",
-    "src/zero-out.ts"
+    "src/zero-out.ts",
+    "src/test-helpers.ts"
```
arv: Should this be in tsconfig.server.json?

Seems like it should be in tsconfig.client.json instead.

apps/zbugs/vitest.config.ts
```diff
@@ -47,7 +51,23 @@ export function configForVersion(version: number, url: string) {
...
+        },
+        {
+          find: '@rocicorp/zero',
+          replacement: resolve(packagesDir, 'zero/src/zero.ts'),
+        },
+      ],
+    },
```
arv: I don't think this is needed. `tsconfigPaths()` should deal with this?

apps/zbugs/src/local-mutate.test.ts
arv: Hard to tell. What is the type error?

Without pulling this PR to my local env I would guess it is an issue with the Context?

arv (commented):
I like where this is going.

How about `@rocicorp/zero/test` or `@rocicorp/zero/testing`?

apps/zbugs/src/local-mutate.test.ts
tantaman: It is a context type error:

```
Types of property '$context' are incompatible.
                Type '{ readonly sub: string; readonly role: "user" | "crew"; } | undefined' is not assignable to type 'AuthData'.
                  Type 'undefined' is not assignable to type 'AuthData'.ts(2345)
```

tantaman:
@kvnkusch - does this work for you now?

apps/zbugs/src/local-mutate.test.ts
```diff
@@ -0,0 +1,67 @@
...
+test('local mutate', async () => {
+  const zero = zeroForTest({
+    cacheURL: null,
+    kvStore: 'mem',
+    schema,
+    mutators: mutatorsForTest,
+    userID: 'user-1',
+    // can't make this work using the proper type. Context always throws an error
+    // no matter what we do.
+    context: {sub: 'user-1', role: 'user'} as any,
```
tantaman: @arv - ready for re-review. I cannot, for the life of me, figure out how to make this work however.

Changing 

```ts
context: {sub: 'user-1', role: 'user'} as any,
```

to

```ts
context: {sub: 'user-1', role: 'user'} as AuthData | undefined
```

causes a type error at lines 43 and 52:

```
Types of property '$context' are incompatible.
                Type '{ readonly sub: string; readonly role: "user" | "crew"; } | undefined' is not assignable to type 'AuthData'.
                  Type 'undefined' is not assignable to type 'AuthData'.ts(2345)
```

Claude is also stumped fwiw.

kvnkusch:
> @kvnkusch - does this work for you now?

Getting the same error. For reference, here's a minimal repro:
```
import { createBuilder, defineMutator, defineMutators } from '@rocicorp/zero';
import { zeroForTest } from '@rocicorp/zero/testing';
import { expect, test } from 'vitest';
import { z } from 'zod';

const schema = {
    tables: {
        item: {
            name: 'item',
            columns: {
                id: { type: 'string' as const, optional: false },
                name: { type: 'string' as const, optional: false },
            },
            primaryKey: ['id'] as const,
        },
    },
    relationships: {},
};

declare module "@rocicorp/zero" {
    interface DefaultTypes {
        schema: typeof schema;
    }
}

const zql = createBuilder(schema);

const zero = zeroForTest({
    schema,
    mutators: defineMutators({
        item: {
            insert: defineMutator(z.object({ id: z.string(), name: z.string() }), async ({ args: { id, name }, tx }) => {
                await tx.mutate.item.insert({ id, name });
            }),
        },
    }),
});

test("basic insert and query", async () => {
    await zero.mutate.item.insert({
        id: "1",
        name: "test",
    });

    const result = await zero.run(zql.item);
    expect(result).toHaveLength(1);
});
```

Types on `zero.mutate` also don't seem to be working.

tantaman:
this is almost done... one more symbol problem.

tantaman:
ok @kvnkusch, I was able to get this to work in `hello-zero` so it should be good to go now...

kvnkusch:
The symbol error is gone, and queries are working via `zero.run(zql.item)`! But looks like `zero.mutate` is not an object:
```
45 | test("basic insert", async () => {
46 |     await zero.mutate.item.insert({
                           ^
TypeError: undefined is not an object (evaluating 'zero.mutate.item.insert')
      at <anonymous> (/Users/kkusch/Code/ccs/cc2/packages/test-tmp/test/repro.test.ts:46:23)
      at <anonymous> (/Users/kkusch/Code/ccs/cc2/packages/test-tmp/test/repro.test.ts:45:22)
✗ basic insert [0.21ms]
```
There's also a Typescript error on this line that matches this, so might be something I'm setting up the schema/mutators wrong? (using the same example as above)
```
test/repro.test.ts:46:23 - error TS2339: Property 'item' does not exist on type 'ZeroMutate<{ tables: { item: { name: string; columns: { id: { type: "string"; optional: boolean; }; name: { type: "string"; optional: boolean; }; }; primaryKey: readonly ["id"]; }; }; relationships: {}; }, CustomMutatorDefs | undefined, unknown>'.
  Property 'item' does not exist on type '(mr: MutateRequest<any, { tables: { item: { name: string; columns: { id: { type: "string"; optional: boolean; }; name: { type: "string"; optional: boolean; }; }; primaryKey: readonly ["id"]; }; }; relationships: {}; }, unknown, any>) => { ...; }'.

46     await zero.mutate.item.insert({
                         ~~~~


Found 1 error in test/repro.test.ts:46
```
Is the `hello-zero` branch up?

arv:
zero.mutate is only available if enableLegacyMutators is true.

kvnkusch:
Oh duh my bad. All working on my end.

packages/zero/tool/build.ts
```diff
@@ -143,6 +171,58 @@ async function getViteConfig(): Promise<InlineConfig> {
   };
 }
 
+// Modules that should be kept external in the testing build to share
+// symbol instances with the main bundle
+const testingExternalModules = ['query-internals'];
```
arv, line 176: I'm not sure this is correct? Don't we want to share everything?

It is a bit strange because we do not want to get copies of the code but the code needs have have TESTING true.

I'm really not sure how to square this.

arv: Maybe we should get rid of the TESTING flag and use import.meta.env or process.env?

arv: Maybe we should get rid of the TESTING flag and use import.meta.env or process.env?

PR #5404, opened 2026-01-09
feat(zero): added initZero to client/react/solidjs

0xcadams:


apps/zbugs/shared/zero-hooks.ts
```diff
@@ -0,0 +1,15 @@
+import type {Row, Zero} from '@rocicorp/zero';
+import {initZero} from '@rocicorp/zero/react';
+import type {AuthData} from './auth.ts';
+import type {schema} from './schema.ts';
+
+export type ZbugsRow = Row<typeof schema>;
+export type ZbugsZero = Zero<typeof schema, undefined, AuthData | undefined>;
```
aboodman, line 7: This one can just be returned by `initZero` as it is a class.

apps/zbugs/shared/zero-hooks.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const {
+  ZeroProvider,
+  useZero,
+  useQuery,
+  useSuspenseQuery,
+  useConnectionState,
+} = initZero<typeof schema, AuthData | undefined>();
```
aboodman, line 15: I think for clarity maybe we'd call this `initZeroReact()`.

apps/zbugs/shared/zero-hooks.ts
```diff
@@ -0,0 +1,15 @@
+import type {Row, Zero} from '@rocicorp/zero';
+import {initZero} from '@rocicorp/zero/react';
+import type {AuthData} from './auth.ts';
+import type {schema} from './schema.ts';
+
+export type ZbugsRow = Row<typeof schema>;
```
aboodman, line 6: I think the pattern would be to do it the other way:

```ts
import {Row as ZeroRow} ...;

export Row = ZeroRow<typeof schema>;
```

aboodman (commented):
some quick thoughts, still reviewing/thinking

apps/zbugs/shared/zero-hooks.ts
0xcadams: The class is created by the ZeroProvider though, not directly in `initZero`, so those type-only returns aren't really handled in `initZero` very easily

PR #5518, opened 2026-02-04
chore(zero-cache): define and handle `backfill` change messages

darkgnotic:
Defines a new `backfill` message in the change stream protocol, and implements the handling of them at the replication level, as specified in the [backfill algorithm](https://www.notion.so/replicache/Auto-backfill-Schema-Change-Replication-2cb3bed8954580ee8c92e3a730c38772?source=copy_link#2cc3bed89545801c9f74dd255c6ffdef).

Key points:
* The changeLog metadata tracked in https://github.com/rocicorp/mono/pull/5505 is used to determine which rows / column values to backfill, avoiding re-introducing deleted rows or overwriting columns from newer replicated updates.
* Care is taken not to change row versions during the backfill. In other words, no changes to any syncable columns should occur while the backfill is in progress.
* To avoid unnecessary IVM processing, no replica-changed notifications are sent if nothing has been added to the changeLog (e.g. transactions consisting only of backfill messages).
* A completed backfill is handled similarly to a schema change: all row versions in the table are bumped, and a reset signal is inserted into the changeLog so that pipelines are rehydrated.

Upcoming:
* change-source logic for producing backfill messages

### DIFF
--- a/packages/zero-cache/src/services/replicator/change-processor.ts
+++ b/packages/zero-cache/src/services/replicator/change-processor.ts
@@ -821,7 +821,6 @@ class TransactionProcessor {
         skipped++;
         continue; // the row was deleted after the backfill snapshot
       }
-      const fullRow = {...row.row, [ZERO_VERSION_COLUMN_NAME]: watermark};
       const updates =
         rowOp?.op === SET_OP
           ? columns.filter(
@@ -840,7 +839,8 @@ class TransactionProcessor {
           ON CONFLICT (${rowKeyColsStr})
           DO UPDATE SET ${updateStmts.join(',')};
       `,
-        Object.values(fullRow),
+        ...Object.values(row.row),
+        watermark, // the _0_version for new rows (i.e. table backfill)
       );
       backfilled++;
     }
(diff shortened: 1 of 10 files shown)


packages/zero-cache/src/services/replicator/change-processor.ts
```diff
@@ -554,7 +561,17 @@ class TransactionProcessor {
       );
     }
 
-    this.#logResetOp(table.name);
+    if (
+      Object.keys(create.backfill ?? {}).length ===
+      Object.keys(create.spec.columns).length
+    ) {
```
arv, line 567: What does this test?

arv (commented):
LGTM

packages/zero-cache/src/services/replicator/change-processor.ts
darkgnotic: * If all of the columns of the table are marked for backfill, we reload table specs for internal book-keeping, but do not log a reset op in the change-log because there is nothing material for IVM to process (i.e. downstream client won't see any changes).
* Otherwise (common case), we log the reset-op (which also loads internal table specs) so that all pipelines reset to "see" the new table.

The test condition is probably more complicated than it needs to be because it checks if the number of backfill columns equals the number of table columns. In practice, a `create-table` will either backfill all of the columns or none of them; there isn't really a situation in which a new table would only require a subset of the columns to be backfilled.

darkgnotic: FYI, you were right to question this @arv.  👍 

https://github.com/rocicorp/mono/pull/6454

PR #5557, opened 2026-02-13
zero-cache: respect ZERO_CHANGE_STREAMER_QUEUE_SIZE_BACK_PRESSURE_THRESHOLD

Karavil:
Adds change streamer config option queueSizeBackPressureThreshold.

Env var: ZERO_CHANGE_STREAMER_QUEUE_SIZE_BACK_PRESSURE_THRESHOLD

Behavior: if set, the change streamer applies back pressure when either the existing heap based queued bytes threshold is exceeded or the queued change count exceeds this threshold. Release requires falling below both thresholds.

### DIFF
--- a/packages/zero-cache/src/config/zero-config.ts
+++ b/packages/zero-cache/src/config/zero-config.ts
@@ -578,6 +578,17 @@ export const zeroOptions = {
         `throughput.`,
       ],
     },
+
+    queueSizeBackPressureThreshold: {
+      type: v.number().optional(),
+      desc: [
+        `Optional queued change count threshold for applying back pressure.`,
+        ``,
+        `If set, back pressure is applied when either the estimated queued bytes exceeds`,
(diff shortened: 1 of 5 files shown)


Karavil:
Cleanup: removed duplicated 0.8 by introducing BACK_PRESSURE_RELEASE_RATIO constant in storer, and restored backward compatible initializeStreamer signature so tests and existing callsites do not need to pass undefined.

Karavil:
Tweaked #maybeReleaseBackPressure: added isBelowThreshold helper to remove duplication, and moved the release comment to describe the combined condition.

Karavil:
Moved the release comment above BACK_PRESSURE_RELEASE_RATIO and removed the in-condition comment that claimed a specific percent.

Karavil:
Tweaked #maybeReleaseBackPressure to precompute release thresholds once (releaseBytesThreshold, releaseQueueSizeThreshold) instead of multiplying by BACK_PRESSURE_RELEASE_RATIO multiple times.

PR #5608, opened 2026-02-25
perf(zql): optimize Join processParentNode closure and spread

Karavil:
> **Note**: This PR was generated by Claude (Anthropic) and has not been
> reviewed by @Karavil. It is part of an upstream contribution effort from
> the Goblins team.

## Summary
Optimize `Join#processParentNode` by extracting the child stream closure to a method and skipping the object spread when `parentNodeRelations` is empty.

## Motivation
`#processParentNode` is called for every parent row during join fetch and push operations. In a workload with 135 IVM pipelines, many of which involve joins with ~200 child rows per parent, this method is called tens of thousands of times per page render.

Two specific costs:
1. **Closure allocation**: The current code creates a new closure inside `#processParentNode` on every call. Extracting the logic to a class method (`#fetchChildStream`) means we create a lightweight arrow function that delegates to the method, rather than a closure capturing local constraint-building logic.
2. **Object spread on empty**: For leaf joins (the most common case in deep join pipelines), `parentNodeRelations` is an empty `{}`. The `{...parentNodeRelations, [name]: stream}` spread still allocates a new object and copies all (zero) properties. Checking emptiness with `for...in` and skipping the spread avoids this overhead.

## Changes
* Extract inline closure to `#fetchChildStream(parentNodeRow)` private method
* Add fast path in `#processParentNode`: check if `parentNodeRelations` has any keys via `for...in`, skip the spread if empty

For leaf joins (bottom of a join chain), the spread skip avoids unnecessary object creation on every parent row. The method extraction reduces closure allocation overhead. Together, these reduce per-call overhead in the most frequently called join method.

## Testing
* All 336 existing join tests pass (fetch/push/sibling + flipped variants)

### DIFF
--- a/packages/zql/src/ivm/join.ts
+++ b/packages/zql/src/ivm/join.ts
@@ -250,47 +250,54 @@ export class Join implements Input {
     }
   }
 
+  #fetchChildStream(parentNodeRow: Row): Stream<Node | 'yield'> {
+    const constraint = buildJoinConstraint(
+      parentNodeRow,
+      this.#parentKey,
+      this.#childKey,
+    );
+    const stream = constraint ? this.#child.fetch({constraint}) : [];
+
+    if (
+      this.#inprogressChildChange &&
+      isJoinMatch(
+        parentNodeRow,
+        this.#parentKey,
+        this.#inprogressChildChange.change.node.row,
+        this.#childKey,
+      ) &&
+      this.#inprogressChildChange.position &&
+      this.#schema.compareRows(
+        parentNodeRow,
+        this.#inprogressChildChange.position,
+      ) > 0
+    ) {
+      return generateWithOverlay(
+        stream,
+        this.#inprogressChildChange.change,
+        this.#child.getSchema(),
+      );
+    }
+    return stream;
+  }
+
   #processParentNode(
     parentNodeRow: Row,
     parentNodeRelations: Record<string, () => Stream<Node | 'yield'>>,
   ): Node {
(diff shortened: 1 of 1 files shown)


arv:
Very nice.

I found the same one :-)

I'm adding a few more performance tests and then we can start to apply these.

packages/zql/src/ivm/join.ts
```diff
@@ -250,47 +250,54 @@ export class Join implements Input {
...
+      break;
+    }
+    const relationships = hasExisting
+      ? {
+          ...parentNodeRelations,
+          [this.#relationshipName]: () => this.#fetchChildStream(parentNodeRow),
+        }
+      : {
+          [this.#relationshipName]: () => this.#fetchChildStream(parentNodeRow),
+        };
```
arv, line 300: When I microbenchmark this the extra check makes things slower than always spreading

packages/zql/src/ivm/join.ts
```diff
@@ -250,47 +250,54 @@ export class Join implements Input {
...
+  ): Node {
+    let hasExisting = false;
+    for (const _ in parentNodeRelations) {
+      hasExisting = true;
+      break;
+    }
+    const relationships = hasExisting
+      ? {
+          ...parentNodeRelations,
+          [this.#relationshipName]: () => this.#fetchChildStream(parentNodeRow),
```
arv, line 296: It still creates a closure here.

@Karavil Are you sure this has any real effect?

PR #5611, opened 2026-02-25
perf(zql): comparator fast paths + compareBounds null fix

Karavil:
> **Note**: This PR is part of an upstream contribution effort from the Goblins team (@goblinshq). Co-authored with Claude by Anthropic.

## Summary
Multiple comparator optimizations targeting the hottest paths in the IVM pipeline, plus a bug fix for nullable column comparisons.

## Motivation
In a production app with 45 parent rows x ~200 related rows across 135 IVM pipelines, comparator functions are invoked millions of times per page render (every BTree node comparison during index scans calls `makeBoundComparator`, and every row sort calls `makeComparator`). These optimizations target the innermost loops:

- **String comparison**: 99%+ of real-world values are ASCII strings. The default `compareUTF8` processes full UTF-8 encoding even for simple ASCII, adding unnecessary overhead per comparison.
- **Single-key sorts**: Most IVM pipeline sorts use a single column. The generic multi-key loop has per-iteration overhead (destructuring, array access) that compounds across millions of calls.
- **BTree comparator call depth**: `makeBoundComparator` -> `compareBounds` -> `compareValues` -> `compareUTF8` is a 4-deep call chain on every BTree node visit. Flattening this into a single `compareBoundValue` helper eliminates 3 function call frames.

### data.ts
* Add `compareStringUTF8Fast()`: ASCII fast-path that compares char codes directly, falling back to `compareUTF8` only when non-ASCII characters are encountered
* Reorder `compareValues()` to check strings before nulls (strings are the most common type in practice)
* Add single-key fast path in `makeComparator()`: for single-column orderings, return a direct comparator avoiding the loop

### memory-source.ts
* **Bug fix**: Add null/undefined guards before delegating to `compareValues()`. `compareValues` asserts type homogeneity (e.g., calls `assertString(b)` when `a` is a string), but nullable database columns can produce `null` vs `string` comparisons that trigger assertion failures.
* Add `compareBoundValue()` helper: merges sentinel handling (minValue/maxValue), null guards, and type-specific comparison (string via `compareStringUTF8Fast`, number via subtraction) into a single function. Well within V8 TurboFan's ~460-bytecode inlining threshold, so it gets inlined at call sites.
* Simplify `makeBoundComparator()` to delegate to `compareBoundValue`: single-key path is ~2 lines instead of 35, multi-key path also uses `compareBoundValue` directly (replacing the old `compareBounds` wrapper).

These changes primarily reduce CPU time in BTree index scans. In profiling:
- `compareUTF8` calls dropped significantly due to ASCII fast-path (most string comparisons resolve without the UTF-8 library call)
- Single-key comparator eliminates loop overhead for the common single-column sort case
- `compareBoundValue` flattens the 4-deep call chain into 1 function that V8 inlines, eliminating frame overhead across thousands of BTree node visits per fetch

Combined with other IVM optimizations in this series, contributed to reducing page freeze from ~7.7s to <1s across the full optimization series.

## Testing
* All existing data and memory-source tests pass
* New tests for `compareStringUTF8Fast` (ASCII, Unicode, empty, prefix, property-based)
* New tests for `makeComparator` single-key path (asc, desc, reverse, equality)
* Existing `compareValues` property test updated to use `Math.sign` (the ASCII fast-path returns char code differences rather than -1/0/1, but the sign is always correct)
---
## Stack Order
This PR is part of a stacked series of IVM performance optimizations. Merge in order:
1. #5609 - Allocation reduction
2. #5610 - Index caching
3. **#5611** (this PR) - Comparator fast paths
4. #5612 - Fetch pipeline fusion

Independent PRs (no conflicts): #5607 (BTree iterators), #5608 (Join optimizations)

### DIFF
--- a/packages/zql/src/ivm/data.ts
+++ b/packages/zql/src/ivm/data.ts
@@ -23,6 +23,28 @@ export type Node = {
 };
 
 /**
+ * Fast-path string comparison that handles the common ASCII case
+ * without calling into compareUTF8. Falls back to compareUTF8 for
+ * non-ASCII characters.
+ *
+ * Returns a sign-only contract: negative if a < b, 0 if equal, positive
+ * if a > b. Callers must NOT rely on the magnitude of the return value.
+ */
+export function compareStringUTF8Fast(a: string, b: string): number {
+  if (a === b) return 0;
+  const len = a.length < b.length ? a.length : b.length;
+  for (let i = 0; i < len; i++) {
+    const ac = a.charCodeAt(i);
+    const bc = b.charCodeAt(i);
+    if (ac !== bc) {
+      if (ac < 128 && bc < 128) return ac - bc;
+      return compareUTF8(a, b);
+    }
+  }
+  return a.length - b.length;
+}
+
+/**
  * Compare two values. The values must be of the same type. This function
  * throws at runtime if the types differ.
  *
@@ -41,6 +63,15 @@ export function compareValues(a: Value, b: Value): number {
   if (a === b) {
     return 0;
   }
+  // String check before null: strings are the most common value type in
+  // practice, so testing them first reduces branch mispredictions. The
+  // null sub-check inside handles the string-vs-null comparison without
+  // falling through to the generic null checks below.
+  if (typeof a === 'string') {
+    if (b === null) return 1;
+    assertString(b);
+    return compareStringUTF8Fast(a, b);
+  }
   if (a === null) {
     return -1;
   }
@@ -54,18 +85,6 @@ export function compareValues(a: Value, b: Value): number {
   if (typeof a === 'number') {
     assertNumber(b);
     return a - b;
-  }
-  if (typeof a === 'string') {
-    assertString(b);
-    // We compare all strings in Zero as UTF-8. This is the default on SQLite
-    // and we need to match it. See:
-    // https://blog.replicache.dev/blog/replicache-11-adventures-in-text-encoding.
-    //
-    // TODO: We could change this since SQLite supports UTF-16. Microbenchmark
-    // to see if there's a big win.
-    //
-    // https://www.sqlite.org/c3ref/create_collation.html
-    return compareUTF8(a, b);
   }
   throw new Error(`Unsupported type: ${a}`);
 }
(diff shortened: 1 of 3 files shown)


packages/zql/src/ivm/data.ts
```diff
@@ -84,6 +103,18 @@ export function normalizeUndefined(v: Value): NormalizedValue {
 export type Comparator = (r1: Row, r2: Row) => number;
 
 export function makeComparator(order: Ordering, reverse?: boolean): Comparator {
+  if (order.length === 1) {
```
arv, line 106: Note to self.

This needs to be done

packages/zql/src/ivm/memory-source.ts
```diff
@@ -801,12 +807,51 @@ type MinValue = typeof minValue;
...
 type MaxValue = typeof maxValue;
 
+/**
+ * Compares two Bound values, handling minValue/maxValue sentinels,
+ * null, and delegating to type-specific comparison. This merges the
+ * logic of compareBounds + compareValues into a single function that
+ * V8 can inline at the call site (well within TurboFan's 460-bytecode
+ * inlining threshold).
+ */
+function compareBoundValue(a: Bound, b: Bound): number {
```
arv, line 817: I solved this differently. 

Note to self. Compare approches.

arv (commented):
My goal is to merge these improvements.

Some already landed as part of my investigations.

packages/zql/src/ivm/memory-source.ts
```diff
@@ -277,15 +288,27 @@ export class MemorySource implements Source {
...
-    const index = this.#getOrCreateIndex(indexSort, conn);
+    // Patch 4: Cache index lookup per connection
+    let constraintShapeKey = '';
+    if (fetchOrPkConstraint) {
+      for (const key of Object.keys(fetchOrPkConstraint)) {
+        constraintShapeKey += key + '|';
+      }
+    }
+    const indexCacheKey = `${constraintShapeKey}::${includeRequestedSort ? conn.requestedSortKey : 'pk'}`;
+    let index = conn.indexCache.get(indexCacheKey);
```
arv, line 314: This means that the Connection owns the index, I need to verify that it is ok to do that .

packages/zql/src/ivm/memory-source.ts
```diff
@@ -535,31 +539,34 @@ function* genPush(
       unreachable(change);
   }
 
+  // Reuse a small set of objects across the connection loop below to avoid
```
arv, line 572: This is good.

arv: The only question is if we should have them be module level. Then they never gets GCed

arv: @grgbkr / @tantaman To double check that it is safe to reuse these.

packages/zql/src/ivm/memory-source.ts
arv: https://github.com/rocicorp/mono/pull/5791

PR #5673, opened 2026-03-20
fix(zqlite): skip IS NULL in cursor queries for non-nullable columns

utkarsh-611:
Cursor pagination queries generate IS NULL checks for non-nullable columns, causing SQLite to abandon its MULTI-INDEX OR optimization and fall back to full table scans.

This is the cursor-pagination counterpart to the getRows fix in [#5542](https://github.com/rocicorp/mono/pull/5542) — same SQLite NULL+OR issue, different code path (gatherStartConstraints in query-builder.ts).

Fix: check columnTypes[field].optional and skip IS NULL branches for non-nullable columns.


utkarsh-611:
@aboodman any comments on this PR, saw your similar PR https://github.com/rocicorp/mono/pull/5542 , Is this PR valid and required any comments ?

PR #5754, opened 2026-04-03
perf(zql): fix V8 deopt in compareValues and compareBounds 

arv:
## Problem

Running the IVM hot path under Node's `--trace-deopt` flag revealed two
deopt sites in the comparison code causing repeated bailouts from
Maglev/Turbofan back to the interpreter:

```sh
NODE_OPTIONS=--trace-deopt npm --workspace=zql-benchmarks run bench -- memory-ivm-deopt
```

**`compareValues`** — The original implementation narrowed the type of
`a` with `typeof` but then called assertion helpers (`assertBoolean`,
`assertNumber`, `assertString`) on `b` separately. These helpers
introduce additional call sites with mixed type feedback, preventing V8
from compiling a specialised fast path for the common string/number
cases.

**`compareBounds`** — The sentinel values `minValue`/`maxValue` are
Symbols. Checking `a === minValue` before knowing the type of `a` meant
that when a Symbol appeared, V8 deopted the integer/string fast path it
had compiled for the common case.

## Fix

**`compareValues`**: Check `typeof a === 'string' && typeof b ===
'string'` (and similarly for number/boolean) so each branch gives V8 a
clear monomorphic profile. The null sentinel check is moved after the
type checks since it is the less common case.

**`compareBounds`**: Guard with `typeof a === 'symbol'` first, giving
V8 a clear type discriminant so the non-symbol path stays compiled as a
numeric/string specialisation.

## Benchmarks

Measured with the new memory-ivm-deopt.bench.ts benchmark (2 runs
each, WIP vs. main):

| benchmark | main | WIP | speedup |
|---|---|---|---|
| `compareValues(string, string)` | 305 ns | 190 ns | **1.6×** |
| `compareValues(number, number)` | 286 ns | 125 ns | **2.3×** |
| `compareValues(mixed)` | 35 ns | 21 ns | **1.7×** |
| MemorySource push add/remove | 2.66 µs | 2.39 µs | **1.11×** |
| MemorySource push edit | 3.65 µs | 3.18 µs | **1.15×** |
| Join push add/remove | 4.22 µs | 3.83 µs | **1.10×** |
| Join push edit | 3.69 µs | 3.28 µs | **1.13×** |

The IVM operator improvements are a downstream effect of the faster
comparator. Fetch/scan paths are unchanged as expected since they do not
exercise the comparison hot path heavily.

### DIFF
--- a/packages/shared/src/btree-set.bench.ts
+++ b/packages/shared/src/btree-set.bench.ts
@@ -105,3 +105,57 @@ describe('BTreeSet iterator next() in isolation', () => {
     use(sum);
   });
 });
+
+// Lookup and mutation benchmarks exercise internal node traversal,
+// triggering the BNode vs BNodeInternal shape polymorphism.
+describe('BTreeSet lookups', () => {
+  bench('has() hit', () => {
+    // Spread across the tree to exercise multiple internal nodes
+    use(tree.has(100));
+    use(tree.has(500));
+    use(tree.has(900));
+  });
+
+  bench('has() miss', () => {
+    use(tree.has(NUM_ENTRIES + 1));
+    use(tree.has(NUM_ENTRIES + 2));
+    use(tree.has(NUM_ENTRIES + 3));
+  });
+
+  bench('get() hit', () => {
+    use(tree.get(100));
+    use(tree.get(500));
+    use(tree.get(900));
+  });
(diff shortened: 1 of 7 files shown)


aboodman:
Nice

PR #5766, opened 2026-04-07
feat(zero-cache): add auth maintenance intervals

0xcadams:
## Add auth maintenance config knobs

Adds `auth.revalidateIntervalSeconds` and `auth.retransformIntervalSeconds` to `zero-cache`.


grgbkr:
I think these should default to on, so things are more secure by default.

0xcadams:
> I think these should default to on, so things are more secure by default.

Understood - they're a bit app-specific but perhaps 5 minutes for each is a reasonable default? I was worried about breaking how users expect auth to behave.

grgbkr:
> Understood - they're a bit app-specific but perhaps 5 minutes for each is a reasonable default? I was worried about breaking how users expect auth to behave.

Agree on 5 mins.  Can you say more about how this might break how users expect auth to behave?  Is this just because it is a change in behavior?

0xcadams:
I changed my mind. If things were working before, and the behavior is changed, they were probably wrong and should fix it. I'll add the default to five minutes.

PR #5767, opened 2026-04-07
feat(zero-cache): scope auth to websocket connections

0xcadams:
## Move `zero-cache` auth to per-connection runtime state

This PR introduces `ConnectionContextManager`, which is core to the refactor: it's a state machine that manages the state of the connection, and provides helpers to share auth context across push and query.

- Replaces shared client-group auth session state with per-connection auth, query, and push context managed by `ConnectionContextManager`.
- Threads the connection auth snapshot through syncer, view-syncer, pusher, custom query transform, and inspect flows.
- Uses the originating connection's auth for connection-triggered work, and one validated background connection for shared query work.
- Adds runtime handling for provisional vs validated connections, auth revision checks, targeted auth failures, and scheduled revalidation/retransform work.

### DIFF
--- a/packages/zero-cache/src/services/mutagen/pusher.ts
+++ b/packages/zero-cache/src/services/mutagen/pusher.ts
@@ -20,8 +20,9 @@ import {
   type PushBody,
   type PushResponse,
 } from '../../../../zero-protocol/src/push.ts';
+import {isAuthErrorBody} from '../../auth/auth.ts';
 import {type ZeroConfig} from '../../config/zero-config.ts';
-import {compileUrlPattern, fetchFromAPIServer} from '../../custom/fetch.ts';
+import {fetchFromAPIServer} from '../../custom/fetch.ts';
 import {getOrCreateCounter} from '../../observability/metrics.ts';
 import {recordMutation} from '../../server/anonymous-otel-start.ts';
 import {ProtocolErrorWithLevel} from '../../types/error-with-level.ts';
@@ -29,26 +30,23 @@ import type {Source} from '../../types/streams.ts';
 import {Subscription} from '../../types/subscription.ts';
 import type {HandlerResult, StreamResult} from '../../workers/connection.ts';
 import type {RefCountedService, Service} from '../service.ts';
+import type {
+  ConnectionContext,
+  ConnectionContextManager,
+  ConnectionSelector,
+} from '../view-syncer/connection-context-manager.ts';
 
 export interface Pusher extends RefCountedService {
-  readonly pushURL: string | undefined;
-
-  initConnection(
-    clientID: string,
-    wsID: string,
-    userPushURL: string | undefined,
-    userPushHeaders: Record<string, string> | undefined,
-    onAuthFailure?: () => void,
-  ): Source<Downstream>;
-  enqueuePush(
-    clientID: string,
-    push: PushBody,
-    auth: string | undefined,
-    httpCookie: string | undefined,
-    origin: string | undefined,
-  ): HandlerResult;
-  ackMutationResponses(upToID: MutationID): Promise<void>;
-  deleteClientMutations(clientIDs: string[]): Promise<void>;
+  initConnection(selector: ConnectionSelector): Source<Downstream>;
+  enqueuePush(selector: ConnectionSelector, push: PushBody): HandlerResult;
+  ackMutationResponses(
+    requester: ConnectionSelector,
+    upToID: MutationID,
+  ): Promise<void>;
+  deleteClientMutations(
+    requester: ConnectionSelector,
+    clientIDs: string[],
+  ): Promise<void>;
 }
 
 type Config = Pick<ZeroConfig, 'app' | 'shard'>;
@@ -68,78 +66,58 @@ type Config = Pick<ZeroConfig, 'app' | 'shard'>;
  */
 export class PusherService implements Service, Pusher {
   readonly id: string;
+  readonly #contextManager: ConnectionContextManager;
   readonly #pusher: PushWorker;
   readonly #queue: Queue<PusherEntryOrStop>;
-  readonly #pushConfig: ZeroConfig['push'] & {url: string[]};
   readonly #config: Config;
   readonly #lc: LogContext;
-  readonly #pushURLPatterns: URLPattern[];
   #stopped: Promise<void> | undefined;
   #refCount = 0;
   #isStopped = false;
 
   constructor(
     appConfig: Config,
-    pushConfig: ZeroConfig['push'] & {url: string[]},
     lc: LogContext,
     clientGroupID: string,
+    contextManager: ConnectionContextManager,
   ) {
+    this.#contextManager = contextManager;
     this.#config = appConfig;
     this.#lc = lc.withContext('component', 'pusherService');
-    this.#pushURLPatterns = pushConfig.url.map(compileUrlPattern);
     this.#queue = new Queue();
     this.#pusher = new PushWorker(
       appConfig,
       lc,
-      pushConfig.url,
-      pushConfig.apiKey,
-      pushConfig.allowedClientHeaders,
+      this.#contextManager,
       this.#queue,
     );
     this.id = clientGroupID;
-    this.#pushConfig = pushConfig;
   }
 
-  get pushURL(): string | undefined {
-    return this.#pusher.pushURLs[0];
-  }
-
-  initConnection(
-    clientID: string,
-    wsID: string,
-    userPushURL: string | undefined,
-    userPushHeaders: Record<string, string> | undefined,
-    onAuthFailure?: () => void,
-  ) {
-    return this.#pusher.initConnection(
-      clientID,
-      wsID,
-      userPushURL,
-      userPushHeaders,
-      onAuthFailure,
-    );
+  initConnection(selector: ConnectionSelector) {
+    return this.#pusher.initConnection(selector);
   }
 
   enqueuePush(
-    clientID: string,
+    selector: ConnectionSelector,
     push: PushBody,
-    auth: string | undefined,
-    httpCookie: string | undefined,
-    origin: string | undefined,
   ): Exclude<HandlerResult, StreamResult> {
-    if (!this.#pushConfig.forwardCookies) {
-      httpCookie = undefined; // remove cookies if not forwarded
-    }
-    this.#queue.enqueue({push, auth, clientID, httpCookie, origin});
+    this.#pusher.enqueuePush(
+      this.#contextManager.mustGetConnectionContext(selector),
+      push,
+    );
 
     return {
       type: 'ok',
     };
   }
 
-  async ackMutationResponses(upToID: MutationID) {
-    const url = this.#pusher.effectivePushURL;
-    if (!url) {
+  async ackMutationResponses(
+    requester: ConnectionSelector,
+    upToID: MutationID,
+  ): Promise<void> {
+    const ctx = this.#contextManager.getConnectionContext(requester);
+    if (!ctx?.pushContext?.url) {
       // No push URL configured, skip cleanup
       return;
     }
@@ -173,10 +151,8 @@ export class PusherService implements Service, Pusher {
         pushResponseSchema,
         'push',
         this.#lc,
-        url,
-        this.#pushURLPatterns,
+        ctx,
         {appID: this.#config.app.id, shardNum: this.#config.shard.num},
-        {apiKey: this.#pushConfig.apiKey},
         cleanupBody,
       );
     } catch (e) {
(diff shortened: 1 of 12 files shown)


packages/zero-cache/src/services/mutagen/pusher.ts
```diff
@@ -566,32 +497,54 @@ class PushWorker {
...
+          this.#contextManager.failConnection(
+            entry.context,
+            entry.context.revision,
+          );
+        }
+        return response;
+      }
+      // A successful push also validates this connection's current auth snapshot.
+      // That lets later shared work reuse it without trusting stale credentials.
+      this.#contextManager.validateConnection(
```
0xcadams, line 526: We validate the connection's auth on successful push too, since this is a signal from the API that auth is OK.

packages/zero-cache/src/services/mutagen/pusher.ts
```diff
@@ -68,78 +66,58 @@ type Config = Pick<ZeroConfig, 'app' | 'shard'>;
...
   #isStopped = false;
 
   constructor(
     appConfig: Config,
-    pushConfig: ZeroConfig['push'] & {url: string[]},
     lc: LogContext,
     clientGroupID: string,
+    contextManager: ConnectionContextManager,
   ) {
+    this.#contextManager = contextManager;
```
grgbkr, line 84: There are a lot of places were ConnectionContextManager and ConnectionContext are name contextManager and context respectively, I think connContextManager and connContext would help readability a bit.

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -0,0 +1,742 @@
...
+import {compileUrlPattern} from '../../custom/fetch.ts';
+import {ProtocolErrorWithLevel} from '../../types/error-with-level.ts';
+import type {ConnectParams} from '../../workers/connect-params.ts';
+import type {UpdateAuthBody} from '../../../../zero-protocol/src/update-auth.ts';
+import type {ZeroConfig} from '../../config/zero-config.ts';
+
+export type ConnectionState = 'provisional' | 'validated';
+
+/**
+ * Identifies one live websocket for a client slot.
```
grgbkr, line 19: what is a client slot?

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -0,0 +1,742 @@
...
+
+/**
+ * State machine for the auth state of a single `ViewSyncerService`.
+ *
+ * Connections are registered as `provisional`, optionally backfilled with
+ * `initConnection` metadata, and then promoted to `validated` once their
+ * stored `userID` is confirmed as valid. The manager also tracks which
+ * validated connection currently serves as the group's background connection.
+ *
+ * This is intentionally side-effect free.
```
grgbkr, line 155: can you say more about "This is intentionally side-effect free."  Is your point that it doesnt make the calls to the api server or interact with the client websockets?

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -0,0 +1,742 @@
...
+      userID: connectParams.userID,
+      auth,
+
+      profileID: connectParams.profileID,
+      baseCookie: connectParams.baseCookie,
+      protocolVersion: connectParams.protocolVersion,
+
+      revalidateAt: undefined,
+
+      queryContext: {
```
grgbkr, line 237: looks like this code for queryContext and pushContext could be dry'd up a bit.

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -0,0 +1,742 @@
...
+    }
+    if (body.userPushURL) {
+      connection.pushContext.url = body.userPushURL;
+    }
+    if (body.userPushHeaders) {
+      connection.pushContext.headerOptions.customHeaders = body.userPushHeaders;
+    }
+
+    connection.revision++;
+    this.#demoteConnection(connection);
```
grgbkr, line 297: maybe this should only be allowed when in provisional state

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -0,0 +1,742 @@
...
+  markBackgroundRetransformSuccess(
+    selector: ConnectionSelector,
+    revision: number,
+  ): void {
+    const backgroundConnection = this.#getBackgroundConnectionContext();
+    if (!backgroundConnection) {
+      return;
+    }
+    if (
+      selector !== undefined &&
```
grgbkr, line 417: by the typescript types selector cannot be undefined here

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -0,0 +1,742 @@
...
+   * A material auth change demotes the connection back to provisional until it
+   * is validated again.
+   */
+  async updateAuth(
+    selector: ConnectionSelector,
+    body: UpdateAuthBody,
+  ): Promise<Readonly<ConnectionContext>> {
+    const connection = this.#mustGetConnectionContext(selector);
+
+    const nextAuth = await resolveAuth(
```
grgbkr, line 312: What happens if another updateAuth is received on this connection while resolveAuth is still pending?

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -0,0 +1,742 @@
...
+};
+
+export type ConnectionFetchContext = {
+  url: string | undefined;
+  allowedUrlPatterns: URLPattern[] | undefined;
+  headerOptions: HeaderOptions;
+};
+
+/**
+ * A snapshot of one live connection tracked by the manager.
```
grgbkr, line 43: I think thing could be simplified a bit by making ConnectionFetchContext and ConnectionContext fully readonly, and doing copy-on-write anytime they are modified by the manager.  Than we don't need to snapshot them when returning them.

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -644,62 +629,102 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
+  }
+
+  /**
+   * Schedules the auth maintenance wakeup from coordinator-derived
+   * deadlines. The timer plumbing is intentionally separate from the actual
+   * revalidation/retransform work so future policy changes only need to update
+   * the maintenance workers, not the wakeup logic.
+   *
+   * This is intentionally cheap & idempotent, so it can be called frequently
+   * when upstream state might have changed.
```
grgbkr, line 647: What is the relevance of "upstream state might have changed." here?

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -644,62 +629,102 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
+
+    const delay = Math.max(0, plan.earliestDeadlineAt - Date.now());
+    lc.debug?.(
+      `Scheduling auth maintenance timer at ${new Date(plan.earliestDeadlineAt).toISOString()}`,
+      {
+        delay,
+        earliestDeadlineAt: plan.earliestDeadlineAt,
+      },
+    );
+    this.#authMaintenanceTimer = this.#setTimeout(() => {
```
grgbkr, line 666: we can pass an async function here and use await / try / catch

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -0,0 +1,742 @@
...
+        {
+          kind: ErrorKind.InvalidConnectionRequest,
+          message:
+            'No validated connection is available for shared query work.',
+          origin: ErrorOrigin.ZeroCache,
+        },
+        'warn',
+      );
+    }
+    return backgroundConnection;
```
grgbkr, line 459: I believe this is returning the mutable context rather than a snapshot as getBackgroundConnectionContext does.

Making ConnectionFetchContext and ConnectionContext fully readonly, and doing copy-on-write anytime they are modified by the manager would structurally avoid this type of bug.

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -714,14 +739,14 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
+        ctx.wsID,
         this.#shard,
-        baseCookie,
+        ctx.baseCookie,
         downstream,
       );
-      this.#clients.get(clientID)?.close(`replaced by wsID: ${wsID}`);
-      this.#clients.set(clientID, newClient);
+      this.#clients.get(ctx.clientID)?.close(`replaced by wsID: ${ctx.wsID}`);
+      this.#clients.set(ctx.clientID, newClient);
```
grgbkr, line 749: I think there is a race here.  We are adding a client handler for this socket before validating the connection in runInLockForClient below.  If an advance, or changeDesiredQueries->hydrate gets the lock before we validate, data may be poked to this client before it is validated.

I think the safest thing is to delay adding the client handler to clients until after the #validateConnection, but we need to make the downstream cleanup still works, see the comment below

// Note: initConnection() must be synchronous so that `downstream` is
      // immediately returned to the caller (connection.ts). This ensures
      // that if the connection is subsequently closed, the `downstream`
      // subscription can be properly canceled even if #runInLockForClient()
      // has not had a chance to run.

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -743,6 +768,11 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
               'warn',
             );
           }
+          // Every new websocket must revalidate so shared maintenance always
```
grgbkr, line 771: // Validate auth before any data is sent to this connection.
            // the#handleConfigUpdate call below will also transform
            // queries, but that may hit the transform cache so do not rely on
            // it for validation. This also ensures shared maintenance always has 
            // a validated connection to fall back to.

grgbkr (commented):
Nice work!

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
0xcadams: This is called when auth changed using updateAuth on an existing connection and so we now have a new token that hasn't hit an API server yet so it can be called from validated

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
0xcadams: I think this is processed serially because of the websocket queueing in Connection

packages/zero-cache/src/services/view-syncer/view-syncer.ts
0xcadams: Yeah we do this in https://github.com/rocicorp/mono/blob/069c6450b44d0a04ac5632c8cf71b2208c628a47/packages/zero-cache/src/workers/syncer.ts#L249-L254 too - there's a note there but it's a bigger lift to have websockets start out as provisional and only once they are validated do they replace the current client ID and join the client group.

0xcadams: I agree this is a problem but I think there should be a separate design doc for it since it's a bigger rework.

0xcadams: (please let me know if you disagree or if I'm conflating these two problems)

0xcadams:
I addressed most of this feedback in #5767. There's still a few outstanding points that I left comments on.

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
grgbkr: I think this is worth fixing, and the best way to do it is to make ConnectionFetchContext and ConnectionContext really immutable so that we can't forget to snapshot.

0xcadams: I fixed this in https://github.com/rocicorp/mono/pull/5837

0xcadams:
Other fixes are also included in https://github.com/rocicorp/mono/pull/5837

packages/zero-cache/src/services/view-syncer/view-syncer.ts
grgbkr: Was not aware of that other issue.  It is related, but different, and I think less dangerous.  If it is exploited a user's healthy connection is killed when it shouldn't be.  If this one is exploited user data can be leaked.

I think this one is easier to fix as well.  I think the fix doesn't require the Syncer-level "pending connection" concept.

I think this fix should work:
https://github.com/rocicorp/mono/pull/5893

PR #5769, opened 2026-04-08
feat(zero-client): omit userID for logged-out clients

0xcadams:
## Let logged-out `Zero` clients omit `userID`

Makes `userID` optional for logged-out clients, while still requiring it whenever `auth` is set. This is because we now send an `X-User-ID` header to the API, and the API should have a better representation of logged in/logged out than 'anon' or an empty string.

The semantics are now: on the API, if the header exists, the user is logged in, and this is a user ID. If the header doesn't exist, the user is not logged in.

- Stops sending the legacy `'anon'` / empty sentinel on connect URLs; logged-out connections now omit `userID` entirely, matching the new zero-cache auth contract.
- Keeps logged-out local storage stable by using a private storage-only sentinel for IDB naming.

### DIFF
--- a/packages/zero-client/src/client/options.ts
+++ b/packages/zero-client/src/client/options.ts
@@ -55,15 +55,17 @@ export type ZeroOptions<
   auth?: string | null | undefined;
 
   /**
-   * A unique identifier for the user. Must be non-empty.
+   * A unique identifier for the user.
    *
    * Each userID gets its own client-side storage so that the app can switch
    * between users without losing state.
    *
+   * Omit this only for logged-out clients.
+   *
    * This must match the `sub` claim of the `auth` token if
    * `auth` is provided.
    */
-  userID: string;
+  userID?: string | undefined;
 
   /**
    * Distinguishes the storage used by this Zero instance from that of other
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -256,6 +256,8 @@ const NULL_LAST_MUTATION_ID_SENT = {clientID: '', id: -1} as const;
 
 const DEFAULT_QUERY_CHANGE_THROTTLE_MS = 10;
 
+export const LOGGED_OUT_STORAGE_USER_ID = '__anonymous__';
+
 function convertOnUpdateNeededReason(
   reason: ReplicacheUpdateNeededReason,
 ): UpdateNeededReason {
@@ -330,7 +332,7 @@ export class Zero<
 
   readonly #rep: ReplicacheImpl<WithCRUD<MutatorDefs>>;
   readonly #server: HTTPString | null;
-  readonly userID: string;
+  readonly userID: string | undefined;
   readonly storageKey: string;
 
   readonly #lc: LogContext;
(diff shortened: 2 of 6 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -470,12 +472,16 @@ export class Zero<
       maxRecentQueries = 0,
       slowMaterializeThreshold = 5_000,
     } = options;
-    if (!userID) {
+
+    if (userID === '' || userID === 'anon') {
       throw new ClientError({
         kind: ClientErrorKind.Internal,
-        message: 'ZeroOptions.userID must not be empty.',
+        message: `ZeroOptions.userID should not be ${userID === '' ? 'empty' : userID}. Omit it entirely for logged-out clients.`,
```
0xcadams, line 479: We throw for clients passing this, to help migration to omit instead of pass a sentinel.

arv: this is a breaking change.

It is not worth it.

Just log a warning instead.

packages/zero-client/src/client/zero.ts
```diff
@@ -2451,6 +2462,18 @@ export class Zero<
       ...(args as ClientMetricMap[keyof ClientMetricMap]),
     );
   };
+
+  #checkAuthValid(
```
arv, line 2466: Nit: Use local module level functions when possible.

arv (commented):
I do not think this is worth breaking the API. Let's just log a warning.

0xcadams:
Agreed. I actually changed this in the downstream branch to log a warning.

packages/zero-client/src/client/options.ts
```diff
@@ -55,15 +55,17 @@ export type ZeroOptions<
...
    * Each userID gets its own client-side storage so that the app can switch
    * between users without losing state.
    *
+   * Omit this only for logged-out clients.
+   *
    * This must match the `sub` claim of the `auth` token if
    * `auth` is provided.
    */
-  userID: string;
+  userID?: string | undefined;
```
grgbkr, line 68: The  
```
 * This must match the `sub` claim of the `auth` token if `auth` is provided.
```
is really only true for deprecated zero-cache verifies jwt mode.

We should update this comment to reflect the new requirement re customer api returning verified user id for tokens.

packages/zero-client/src/client/zero.ts
```diff
@@ -256,6 +256,8 @@ const NULL_LAST_MUTATION_ID_SENT = {clientID: '', id: -1} as const;
 
 const DEFAULT_QUERY_CHANGE_THROTTLE_MS = 10;
 
+export const LOGGED_OUT_STORAGE_USER_ID = '__anonymous__';
```
grgbkr, line 259: Some small chance of collision, maybe we use '', since that is a value we have always disallowed.

grgbkr:
LGTM

packages/zero-client/src/client/options.ts
0xcadams: Screenshot 2026-04-24 at 12 11 13 PM
This was fixed in a later PR

PR #5770, opened 2026-04-08
feat(zbugs): stop using anon user IDs

0xcadams:
## Adopt the new auth model in zbugs

- Stops threading `userID` through preload, issue detail, and list queries; user-scoped relations now derive from authenticated query context (`auth.sub`) instead.
- Treats logged-out state as `undefined` instead of `'anon'`, updating app init, issue UI, list navigation, and emoji helpers to stop pretending there is an anonymous user.
- Keeps logged-out queries valid by returning no `viewState` or `notificationState` rows until a real authenticated user exists.
- Tightens server-side auth checks by requiring `X-User-ID` to match the JWT on server-driven requests, while leaving the direct client upload endpoint free to skip that header check.
- Aligns zbugs with the new zero-client and zero-cache contract so authenticated requests are keyed to real identity and logged-out requests no longer invent one.

### DIFF
--- a/apps/zbugs/api/index.ts
+++ b/apps/zbugs/api/index.ts
@@ -132,10 +132,11 @@ async function withAuth<T extends {headers: IncomingHttpHeaders}>(
   request: T,
   reply: FastifyReply,
   handler: (authData: JWTData | undefined) => Promise<void>,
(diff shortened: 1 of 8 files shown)


grgbkr:
LGTM

PR #5772, opened 2026-04-08
fix(zero-client): fixed types for connection API to not allow null

0xcadams:
We previously allowed null/undefined for the Connection API, but this was not semantically correct, since a new Zero instance must be created under this condition. So this is removing the types which may confuse users.

### DIFF
--- a/packages/zero-client/src/client/connection.ts
+++ b/packages/zero-client/src/client/connection.ts
@@ -90,26 +90,26 @@ export interface Connection {
    * Resumes the connection from a terminal state.
    *
    * If called when not in a terminal state, this method does nothing.
+   * To clear existing auth credentials, create a new Zero instance with `auth` omitted.
    *
    * @param opts - Optional connection options
    * @param opts.auth - Token to use for authentication. If provided, this overrides
    *                    the stored auth credential for this connection attempt.
-   *                    If `null` or `undefined`, the stored auth credential is cleared.
    * @returns A promise that resolves once the connection state has transitioned to connecting.
    */
-  connect(opts?: {auth: string | null | undefined}): Promise<void>;
+  connect(opts?: {auth: string}): Promise<void>;
 }
 
 export class ConnectionImpl implements Connection {
   readonly #connectionManager: ConnectionManager;
   readonly #lc: LogContext;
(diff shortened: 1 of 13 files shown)


packages/zero-client/src/client/connection.ts
```diff
@@ -90,26 +90,26 @@ export interface Connection {
    * Resumes the connection from a terminal state.
    *
    * If called when not in a terminal state, this method does nothing.
+   * To clear existing auth credentials, create a new Zero instance with `auth` omitted.
```
grgbkr, line 93: I think the documentation comment for this api is stale.  I believe this method is also how you update the auth token for the connection, even when not in a terminal state.  

In this case it doesn't really reconnect, it updates the auth token via a message on the existing connection.

grgbkr, line 93: Possibly a bit clearer:

Note: A Zero client's `userID` cannot be changed. To log out or authenticate as a different user, close the current Zero client and create a new instance.

grgbkr (commented):
LGTM

packages/zero-client/src/client/connection.ts
0xcadams: No they’ll need to create a new Zero instance. That’s what the ZeroProvider does under the hood.

This is only for updating a logged in user w/ new auth on the wire. And there must be a value that you’re passing - you can’t be logged out.

0xcadams: So this should be accurate since they can’t go from logged in -> logged out without creating a new Zero instance and reconnecting.

0xcadams: Yeah that’s definitely clearer.

grgbkr: I was referring to the part of the comment that didn't change in this pr:
```
   * Resumes the connection from a terminal state.
   *
   * If called when not in a terminal state, this method does nothing.
```

Pretty sure that is no longer accurate.

0xcadams: Oh I see - yeah that's true it _can_ be called from terminal state but doesn't _need_ to be. I will update that copy.

PR #5795, opened 2026-04-10
fix(zero-cache): fix backfill completed state version

darkgnotic:
Ensure that the db's `stateVersion` is at least up to the backfill LSN, which may constitute some of the versions of rows that it backfilled. The backfill-manager already ensures that the replication stream has reached or exceeded the backfill LSN, but the actual transaction version (which is based on the latest commit) may be before then. So the final version of the committed transaction is the greater of the original commit version and the watermark of the completed backfill.

Context:
* https://rocicorp.slack.com/archives/C0ABMK92H47/p1775589780500379


darkgnotic:
I'm not super happy with this. I'm gonna try a different approach.

darkgnotic:
Abandoned in favor of https://github.com/rocicorp/mono/pull/5796

PR #5799, opened 2026-04-10
chore(zql): Optimize join

arv:
Get rid of one allocation in every join push


arv:
also adds summary to bench

PR #5802, opened 2026-04-11
Nullable Array column "null" value handling

kvnkusch:
Includes two fixes for nullable array columns :

- Preserve `NULL` for nullable array column writes instead of converting to `[]`
- Temporal nullable array columns no longer convert `NULL` to `[]` on reads


tantaman:
Nice. This is only a problem in custom mutators, correct?

tantaman:
lgtm. Pulling down to test one more thing.

tantaman (approved):
lgtm

PR #5804, opened 2026-04-13
chore: Let fmt sort imports

arv:


.oxfmtrc.json
```diff
@@ -5,7 +5,11 @@
   "bracketSpacing": false,
   "quoteProps": "consistent",
   "printWidth": 80,
-  "sortPackageJson": false,
+  "sortImports": {
+    "newlinesBetween": false,
+    "partitionByNewline": true
+  },
+  "sortPackageJson": true,
```
arv, line 12: These are the only real changes. Everything else is a result of reformattin.

arv (commented):
Now we can disable organize imports on save and rely on format on save instead.

PR #5807, opened 2026-04-14
fix(zero-cache): preserve compound primary key column order

arv:
Stop alphabetically sorting primaryKey, uniqueKeys, and allPotentialPrimaryKeys in computeZqlSpecs. The sorting caused compound primary key columns to be reordered alphabetically (e.g. [callId, userId, connectionId] became [callId, connectionId, userId]), which resulted in incorrect ORDER BY clauses and index column ordering in SQLite.

This mismatch between the ORDER BY column order and the actual index column order forces SQLite to fall back to full table scans instead of using the index, causing severe performance degradation.

The sorting was originally added as an optimization so that normalizedKeyOrder() in row-key.ts could take its fast path (already-sorted check) and avoid allocating new objects. However, these two concerns are independent:

- Row identity (normalizedKeyOrder) sorts RowKey object properties alphabetically at each point of use (change-log, snapshotter, CVR) for deterministic stringification. It handles unsorted input fine.
- Query ordering (completeOrdering -> addPrimaryKeys) appends PK columns to ORDER BY in the PK array order from the tableSpec. This is where the bug manifested.

Removing the pre-sort is safe: row identity still gets normalized at point of use, and query ordering now preserves the index-defined column order.

Fixes: https://bugs.rocicorp.dev/p/zero/issue/246641

### DIFF
--- a/packages/z2s/src/compiler.output.test.ts
+++ b/packages/z2s/src/compiler.output.test.ts
@@ -117,6 +117,17 @@ const alternateUser = table('alternate_user')
   })
   .primaryKey('id');
 
+// Table with compound PK where user-defined order differs from alphabetical.
+// PK order: callId, userId, connectionId
+// Alphabetical: callId, connectionId, userId
+const connectedCalls = table('connected_calls')
+  .columns({
+    callId: string(),
+    userId: string(),
+    connectionId: string(),
+  })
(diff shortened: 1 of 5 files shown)


darkgnotic (commented):
LGTM. Nice investigation and fix!

PR #5810, opened 2026-04-14
chore(zero-cache): sampled sync for use in test reset

tantaman:
`initialSync` runs only when a customer first onboards. If a customer ever
needs a full reset (schema drift, corruption, disaster recovery, support
request), we rely on that code path working — but it hasn't been exercised
on their data in months, and bugs can sit dormant until the moment they
matter most.

Shadow initial sync is a periodic canary. It exercises the real
`initialSync` code path against a small sample of every published table,
writes to a throwaway SQLite database, and verifies the resulting replica
is structurally sound. If the production code path is broken, shadow fails
before a real reset ever needs to run.

## For a follow up PR

- The scheduler that fires `shadowInitialSync` on an interval.

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/initial-sync.ts
+++ b/packages/zero-cache/src/services/change-source/pg/initial-sync.ts
@@ -152,7 +152,10 @@ export async function initialSync(
     let lsn: string;
 
     if (shadow) {
-      const acquired = await acquireExportedSnapshot(lc, upstreamURI);
+      const acquired = await acquireExportedSnapshotForShadowSync(
+        lc,
+        upstreamURI,
+      );
       snapshot = acquired.snapshot;
       lsn = acquired.lsn;
       releaseShadowSnapshot = acquired.release;
@@ -517,7 +520,7 @@ type ReplicationSlot = {
  * Idle-in-transaction timeout is disabled locally so the exporter doesn't
  * get killed while workers are still importing.
  */
-async function acquireExportedSnapshot(
+async function acquireExportedSnapshotForShadowSync(
   lc: LogContext,
   upstreamURI: string,
 ): Promise<{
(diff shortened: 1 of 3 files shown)


packages/zero-cache/src/services/change-source/pg/initial-sync.ts
```diff
@@ -384,6 +524,60 @@ type ReplicationSlot = {
...
+    connection: {['application_name']: 'shadow-initial-sync-snapshot'},
+  });
+  const ready = resolver<{snapshot: string; lsn: string}>();
+  const release = resolver<void>();
+  const held = holder
+    .begin(Mode.READONLY, async tx => {
+      await tx`SET LOCAL idle_in_transaction_session_timeout = 0`.execute();
+      const [row] = await tx<{snapshot: string; lsn: string}[]>`
+        SELECT pg_export_snapshot() AS snapshot,
+               pg_current_wal_lsn()::text AS lsn`;
```
darkgnotic, line 563: Note that this `pg_current_wal_lsn()` not pinned to the snapshot. If you call it twice it can give you a different value. Is that okay?

The only way to correctly get the LSN for a snapshot is by creating a replication slot.

In backfill, for example, we create a temporary replication slot (and then throw it away) specifically for this purpose:

https://github.com/rocicorp/mono/blob/c7e3188abe685544e265a9e17c9b0ce017b73a88/packages/zero-cache/src/services/change-source/pg/backfill-stream.ts#L220

darkgnotic: That said, if you're not actually using the lsn for anything semantic, then this is of course fine.

tantaman: I think that is fine. I call this once for shadow sync and use it to:

1. `SET TRANSACTION SNAPSHOT` in the copy readers.
2. set `initialVersion = toStateVersionString(lsn)` for the throwaway shadow replica

I was thinking we wouldn't want to create replication slots for the shadow sync process. Unless you think otherwise.

PR #5814, opened 2026-04-15
fix(zero-cache): distinguish metrics across processes

grgbkr:
## Fix metric clobbering across syncer worker processes

Each syncer worker process independently starts the OTel SDK and exports metrics to the collector. Without a distinguishing identifier, all N workers on a pod emit metrics with identical resource labels, causing the collector to arbitrarily overwrite values on each scrape — resulting in oscillating metrics.

image

### Fix

Add `process.worker` and `process.worker_index` as OTel Resource attributes in `startOtelAuto`. These are automatically attached to every metric, trace, and log from that SDK instance — no changes needed to individual metric call sites.

- `process.worker` mirrors the `worker` key in log contexts for correlation
- `process.worker_index` is a stable 0-based index (not PID) to avoid label churn in Prometheus
-  Added `workerIndex` to log context for being able to correlate metrics and logs.

### Cleanup

- Made `workerName` and `workerIndex` required params on `startOtelAuto` and `createLogContext` — every worker must declare its identity
- Added missing `startOtelAuto` calls in `replicator.ts` and `mutator.ts` (they were relying on `OtelLogSink`'s lazy fallback)
- Removed `OtelLogSink`'s lazy `startOtelAuto()` call — SDK initialization is the responsibility of each worker entry point, not the log sink

### DIFF
--- a/packages/zero-cache/src/server/otel-start.ts
+++ b/packages/zero-cache/src/server/otel-start.ts
@@ -25,7 +25,11 @@ class OtelManager {
     return OtelManager.#instance;
   }
 
-  startOtelAuto(lc?: LogContext, workerName?: string | undefined) {
+  startOtelAuto(
+    lc: LogContext | undefined,
+    workerName: string,
+    workerIndex: number,
+  ) {
     if (this.#started || !otelEnabled()) {
       return;
     }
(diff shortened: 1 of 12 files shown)


packages/zero-cache/src/server/otel-start.ts
```diff
@@ -43,6 +47,15 @@ class OtelManager {
...
       [ATTR_SERVICE_VERSION]: process.env.ZERO_SERVER_VERSION ?? 'unknown',
+      // Tag every metric/trace/log with the worker name and index so each
+      // worker process in a multi-worker pod is distinguishable. Without
+      // this, N syncer workers sharing the same pod labels clobber each
+      // other in the OTel collector on every scrape interval.
+      // These mirror the 'worker' and 'workerIndex' keys in every log
+      // context so logs and metrics can be correlated on the same fields.
+      // Using a stable index instead of PID avoids label churn in Prometheus.
+      'process.worker': workerName,
+      'process.worker_index': workerIndex,
```
darkgnotic, line 58: This is a nice workaround to the label explosion issue.

packages/zero-cache/src/server/runner/run-worker.ts
```diff
@@ -25,7 +25,7 @@ export async function runWorker(
   // Note: Deprecation warnings are only emitted at this top-level parse;
   //       they are suppressed when parsed in subprocesses.
   const cfg = getZeroConfig({env, emitDeprecationWarnings: true});
-  const lc = createLogContext(cfg, {worker: 'runner'});
+  const lc = createLogContext(cfg, {worker: 'runner', workerIndex: 0});
```
darkgnotic: Would it be possible to default workerIndex to 0 so that most places don't need to pass it in? AFAICT there's only one worker that runs in multiples.

darkgnotic (approved):

packages/zero-cache/src/server/runner/run-worker.ts
grgbkr: done

PR #5816, opened 2026-04-15
feat(zero): analyze query library

tantaman:
https://rocicorp.slack.com/archives/C0A8NUDMDM1/p1776274371975709?thread_ts=1776272794.639749&cid=C0A8NUDMDM1

This will replace the old analyze-query CLI. This also leverages the existing inspector protocol rather than creating a new path.

Getting this to Goblins so they can test out their re-normalization without a re-deploy.

---

## Usage

Consumers wire up the CLI with a three-line wrapper that hands it their schema:

```ts
// apps/zbugs/scripts/analyze.ts
import '../../../packages/shared/src/dotenv.ts';
import {runAnalyzeCli} from '@rocicorp/zero/analyze';
import {schema} from '../shared/schema.ts';

await runAnalyzeCli({schema});
```

Register it as an npm script:

```json
"analyze-query": "tsx scripts/analyze.ts"
```

All other options (server URL, admin password, query, etc.) are read from argv (or `ZERO_*` env vars). `--help` prints the full flag list.

### Three input modes

**1. ZQL chain form** — the ergonomic everyday case:

```bash
npm run analyze-query -- \
  --zero-cache-url=http://localhost:4848 \
  --auth-token="$TOKEN" \
  --query='issue.related("comments", c => c.limit(5)).related("creator").limit(3)'
```

alternative using cookies

```bash
npm run analyze-query -- \
  --zero-cache-url=http://localhost:4848 \
  --cookie="$COOKIE" \
  --query='issue.related("comments", c => c.limit(5)).related("creator").limit(3)'
```

**2. Named (server-registered) query** — analyze what your app actually sends:

```bash
npm run analyze-query -- \
  --zero-cache-url=http://localhost:4848 \
  --query-name=allProjects --query-args='[]'

npm run analyze-query -- \
  --zero-cache-url=http://localhost:4848 \
  --query-name=issueDetail \
  --query-args='[{"idField":"shortID","id":1}]' \
  --output-vended-rows
```

**3. Raw AST** — paste a transformed AST from a log line straight in:

```bash
npm run analyze-query -- \
  --zero-cache-url=http://localhost:4848 \
  --ast='{"table":"issue","limit":5}'
```

### Env-var shorthand

Since options fall back to `ZERO_*` env vars (the wrapper's `dotenv` import picks up `.env`), most flags can be dropped:

```bash
ZERO_ZERO_CACHE_URL=http://localhost:4848 npm run analyze-query -- --query='issue.limit(10)'
```

---


tantaman:
this should use a zero-client instance imo. would cut the pr in half

packages/zero-client/src/client/inspector/lazy-inspector.ts
```diff
@@ -356,6 +356,48 @@ export async function analyzeQuery(
...
+    inspectAnalyzeQueryDownSchema,
+  );
+}
+
+/**
+ * Sends an `authenticate` op to the server. Returns `true` if the password
+ * is accepted (or the server is in a development mode that bypasses the
+ * check). Use this from Node contexts where the default interactive HTML
+ * prompt used by other inspector RPCs is not available.
+ */
```
arv, line 389: duplicated comment

packages/analyze-query/src/analyze-cli.ts
```diff
@@ -0,0 +1,423 @@
...
+import type {AST} from '../../zero-protocol/src/ast.ts';
+import type {Schema} from '../../zero-types/src/schema.ts';
+import {createBuilder} from '../../zql/src/query/create-builder.ts';
+import type {AnyQuery} from '../../zql/src/query/query.ts';
+import type {SchemaQuery} from '../../zql/src/query/schema-query.ts';
+
+export type AnalyzeCliOptions = {
+  schema: Schema;
+  /** Defaults to `process.argv.slice(2)`. */
+  argv?: readonly string[];
```
arv, line 21: `| undefined`

packages/analyze-query/src/analyze-cli.ts
```diff
@@ -0,0 +1,423 @@
...
+ * Entry point for a user's `cli.ts`. Parses argv, connects to a remote
+ * zero-cache by standing up an in-process Zero client (in-memory storage,
+ * no subscriptions), calls the inspector's `analyze-query` RPC, and
+ * renders the result. Intended to be called as:
+ *
+ * ```ts
+ * import {schema} from './schema.ts';
+ * import {runAnalyzeCli} from '@rocicorp/zero/analyze';
+ * await runAnalyzeCli({schema});
+ * ```
```
arv, line 129: This makes me think we should provide a `.bin`. The only thing this does is provide the schema and a CLI can take the path of that as an argument.

arv (commented):
LGTM

Very cool

packages/analyze-query/src/analyze-cli.ts
tantaman: I was worried about import problems if we try to take the schema as an argument.

PR #5818, opened 2026-04-15
chore(zero-cache): move TransactionPool args into an Options object

darkgnotic:
No function change; just preparing to add a new Option.


darkgnotic:
Looks like I already accidentally checked this into main.  😝

arv:
> Looks like I already accidentally checked this into main. 😝

I was trying to disable (non --force) pushes to main. I thought I got that working but it is possible someone enabled it again?

PR #5819, opened 2026-04-16
chore(zero-cache): add statement response timeout to change-log transaction statements

darkgnotic:
Adds logic in the TransactionPool to check for orphaned statements for which responses do not arrive. This is to investigate/work-around a pathological scenario in which change-streamer progress halts on a Promise (either being awaited in the storer loop or in the postgres.js begin() code) even though Postgres itself believes the transaction to be idle:

* https://rocicorp.slack.com/archives/C0AQDEA9N10/p1775232642218719
* https://rocicorp.slack.com/archives/C0AQDEA9N10/p1776270576471669

### DIFF
--- a/packages/zero-cache/src/db/transaction-pool.ts
+++ b/packages/zero-cache/src/db/transaction-pool.ts
@@ -70,6 +70,16 @@ export type Options = {
    * workers will be shut down after an idle timeout of 5 seconds.
    */
   maxWorkers?: number;
+
+  /**
+   * Aborts the transaction if the response for a statement is not received
+   * within the specified timeout. Note that this is different from a Postgres
+   * `statement_timeout` or `lock_timeout` in that it is specifically intended
+   * to detect situations in which Postgres either never received the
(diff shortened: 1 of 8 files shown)


packages/zero-cache/src/db/transaction-pool.ts
```diff
@@ -760,3 +853,11 @@ export const TIMEOUT_TASKS: TimeoutTasks = {
 // The slice of information from the Query object in Postgres.js that gets logged for debugging.
 // https://github.com/porsager/postgres/blob/f58cd4f3affd3e8ce8f53e42799672d86cd2c70b/src/connection.js#L219
 type Query = {string: string; parameters: object[]};
+
+export class ResponseTimeoutError extends Error {
+  static readonly name = 'ResponseTimeoutError';
+
+  constructor(msg: string) {
+    super(msg);
+  }
```
arv, line 862: useless code

arv (commented):
LGTM

Too bad we had to do this advanced dance just to detect this.

PR #5822, opened 2026-04-16
Add input validation schema error to error `details` so that frontend can read it

tjenkinson:
It would be useful to be able to get access the raw `issues` from the validation error in order to build a more custom error message, so this adds a `InputValidationError` type that contains a `result` property, but also puts the error on a `details` property so that `getErrorDetails` in error.ts finds it when creating an `ApplicationError`

Let me know if there's a better way to do this

Tested locally and in the frontend now get an error like

```json
{
  "type": "error",
  "error": {
    "type": "app",
    "message": "Validation failed for mutator integrations.insert: custom error",
    "details": {
      "type": "InputValidationError",
      "result": {
        "issues": [
          {
            "code": "custom",
            "message": "custom error",
            "path": [
              "data",
              "host"
            ]
          }
        ]
      }
    }
  }
}
```


packages/zql/src/query/query-registry.test.ts
```diff
@@ -85,7 +85,7 @@ describe('createQuery', () => {
     expect(() =>
       addContextToQuery(query('boom'), {}),
     ).toThrowErrorMatchingInlineSnapshot(
-      `[Error: Validation failed for query test: bad]`,
+      `[InputValidationError: Validation failed for query test: bad]`,
```
tjenkinson, line 88: hoping the name changing isn't a breaking change 🙏

aboodman:
Thank you! I really like this feature.

PR #5823, opened 2026-04-16
feat(zero-cache): shadow initial sync

tantaman:
## Summary
- Adds a `shadow-syncer` worker that periodically re-runs `initialSync` against
  a throwaway SQLite file using a sampled subset of every published table. Acts
  as a canary so we catch schema-drift, PG version quirks, or encoding
  regressions before a real customer needs a full reset.
- Runs only on the replication-manager (`runChangeStreamer`) so view-syncers
  don't stampede the upstream with redundant canaries.
- Disabled by default (`ZERO_SHADOW_SYNC_ENABLED=true` to opt in).

## Behavior
- **Interval**: `shadowSync.intervalHours` (default 24h).
- **First run**: sleeps `[intervalMs, 2·intervalMs)` — waits at least one full
  interval on task startup *plus* jitter, so a fleet restart doesn't stampede
  and a fresh task never shadow-syncs immediately.
- **Sampling**: `TABLESAMPLE BERNOULLI(sampleRate) LIMIT maxRowsPerTable`.
  Defaults: `sampleRate=0.1`, `maxRowsPerTable=10000` (~100k rows scanned per
  big table, same upstream cost as earlier 1%/1k defaults but 10× more rows
  exercised downstream through COPY → SQLite).
- **Failure handling**: logs and keeps running; one bad run never disables the
  canary.

---

Note on alternate approach:

## Future work: CTID-window sampling

Commit `92562f4ad` ("ctid windodws?") prototypes an alternative to
`TABLESAMPLE BERNOULLI`: a deterministic sliding window over the main-heap
block range, parameterized by a `runNumber` (e.g.
`floor(Date.now() / cron_period_ms)`).

```
start = (runNumber * windowBlocks) % relpages
end   = min(start + windowBlocks, relpages)
WHERE ctid >= '(start,0)'::tid AND ctid < '(end,0)'::tid
```

Tables with `relpages <= windowBlocks` are scanned in full; larger tables
slide forward each run until the whole heap has been covered, then wrap.

**Why it's compelling**:
- **Predictable scan size** — each run reads a fixed number of heap blocks
  regardless of row density, versus BERNOULLI where scan cost grows with
  `maxRowsPerTable / sampleRate`.
- **Eventual full coverage** — every heap block gets visited over
  `ceil(relpages / windowBlocks)` runs, unlike random sampling which never
  guarantees coverage.
- **Cheaper probe** — replaces per-run `COUNT(*)` / `pg_column_size()` sums
  (full table scans) with a single `pg_relation_size(regclass)` lookup.
- **Consistent data loads** — byte volume is dominated by block count, not
  sampling luck; run-to-run variance mostly reflects TOAST fetches.

No state to persist: `runNumber` is derived from the wall clock as
`floor(Date.now() / intervalMs)`, so a restarted task picks up the same
window its predecessor would have used without any extra bookkeeping.

Not included in this PR — landing the canary on BERNOULLI first, since the
windowing scheme still needs more soak time.

### DIFF
--- a/packages/zero-cache/src/config/zero-config.ts
+++ b/packages/zero-cache/src/config/zero-config.ts
@@ -956,7 +956,7 @@ export const zeroOptions = {
     },
 
     sampleRate: {
-      type: v.number().default(0.01),
+      type: v.number().default(0.1),
       desc: [
         `The BERNOULLI sampling rate for each table (0 < rate <= 1). A value of`,
         `1 disables sampling and copies all rows (still subject to`,
@@ -965,7 +965,7 @@ export const zeroOptions = {
     },
 
     maxRowsPerTable: {
-      type: v.number().default(1000),
+      type: v.number().default(10000),
       desc: [
         `The hard upper bound on rows copied per table per shadow run. Guards`,
         `against unexpectedly large tables consuming disk / upstream bandwidth.`,
--- a/packages/zero-cache/src/config/zero-config.ts
+++ b/packages/zero-cache/src/config/zero-config.ts
@@ -934,6 +934,45 @@ export const zeroOptions = {
     },
   },
 
+  shadowSync: {
+    enabled: {
+      type: v.boolean().default(false),
+      desc: [
+        `Periodically exercises the initial-sync code path against a sample of`,
+        `rows from every published table, writing to a throwaway SQLite database.`,
+        `This acts as a canary: if the real initial-sync path ever breaks (schema`,
+        `drift, PG version quirks, etc.), the shadow run fails before a customer`,
+        `actually needs a full reset.`,
+      ],
+    },
+
+    intervalHours: {
+      type: v.number().default(24),
+      desc: [
+        `The interval between shadow initial-sync runs, in hours. The first run`,
+        `is additionally staggered by a random fraction of this interval so that`,
+        `a fleet restart does not cause all tasks to canary simultaneously.`,
+      ],
+    },
+
+    sampleRate: {
+      type: v.number().default(0.01),
+      desc: [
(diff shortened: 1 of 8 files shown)


packages/zero-cache/src/config/zero-config.ts
```diff
@@ -934,6 +934,45 @@ export const zeroOptions = {
...
+    intervalHours: {
+      type: v.number().default(24),
+      desc: [
+        `The interval between shadow initial-sync runs, in hours. The first run`,
+        `is additionally staggered by a random fraction of this interval so that`,
+        `a fleet restart does not cause all tasks to canary simultaneously.`,
+      ],
+    },
+
+    sampleRate: {
```
tantaman, line 958: another option I've been exploring in the background is to query by CTID range.

The advantage of this is:
1. we can slide the CTID window each run to eventually get full coverage
2. CTID window guarantees a fixed number of pages scanned per table so wide tables don't blow us up and narrow tables are covered quicker.

packages/zero-cache/src/server/shadow-syncer.ts
```diff
@@ -0,0 +1,56 @@
...
+  parent: Worker,
+  env: NodeJS.ProcessEnv,
+  ...argv: string[]
+): Promise<void> {
+  const config = getNormalizedZeroConfig({env, argv});
+
+  startOtelAuto(createLogContext(config, {worker: 'shadow-syncer'}, false));
+  const lc = createLogContext(config, {worker: 'shadow-syncer'}, true);
+  initEventSink(lc, config);
+  startAnonymousTelemetry(lc, config);
```
tantaman: should we start anon telemetry here? we do it in reaper and syncer but I would have thought we only want one per process?

PR #5825, opened 2026-04-16
feat!: rework API helpers/response protocol schemas to add `userID`

0xcadams:
**BREAKING: this PR adds breaking changes to `handleQueryRequest` and `handleMutateRequest` to add a required new parameter, `userID`, as well as adds `queryParams` to the `handleQueryRequest`.**

This PR adds a new backwards compatible query/mutate response schema (to add `userID`), as well as refactors the old schemas and zero-client/zero-cache WS schemas for clarity.

The `fetch.ts` will continue to accept old tuple-type responses and will handle gracefully.

Also adds tests for zero-cache/API response protocol backwards compatibility in both `zero-protocol` and `fetch.ts`.

This *does not* change any of the zero-cache/API request protocol.


packages/zero-cache/src/custom/fetch.ts
```diff
@@ -236,7 +236,9 @@ export async function fetchFromAPIServer<TValidator extends Type>(
 
       try {
         const json = await response.json();
-        const result = validator.parse(json);
+        const result = validator.parse(json, {
+          mode: 'strip',
```
0xcadams, line 240: We add `strip` here 😅

packages/zero-protocol/src/custom-queries.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        queries: [
+          {
+            id: 'q1',
+            name: 'issues',
+            ast: {
+              table: 'issue',
+            },
+          },
+        ],
+        futureQueryPriority: 'future-field',
```
0xcadams, line 23: We also test the underlying schemas for `strip` backwards compatibility.

PR #5828, opened 2026-04-17
fix(zero-cache): handle concurrently issued DDL statements

darkgnotic:
Concurrently issued ddl statement can wedge the replica by introducing incorrect (e.g. redundant) schema changes due to inconsistent before/after schema snapshots.

Fix this by:
* looking up schema (table and index) information in a single statement
* serializing ddl statements using a transaction-scoped advisory lock.

Context:
* https://rocicorp.slack.com/archives/C0ARBDKG3D4/p1776291713832139


packages/zero-cache/src/services/change-source/pg/schema/published.ts
```diff
@@ -170,8 +169,12 @@ export function indexDefinitionsQuery(publications: readonly string[]) {
       GROUP BY "schema", "tableName", "name", "unique", 
          "isPrimaryKey", "isReplicaIdentity", "isImmediate")
 
-    SELECT COALESCE(json_agg("index"), '[]'::json) as "indexes" FROM indexes
-  `;
+    SELECT json_build_object(
```
grgbkr, line 172: why is the lock and reading table+index in a single select both needed?

grgbkr (commented):
LGTM

packages/zero-cache/src/services/change-source/pg/schema/published.ts
darkgnotic: Good question. I think the lock is generally enough for our practical purposes.

Reading from a single select is more like good hygiene, recommended by Claude, I suppose against changes to the catalog tables that might bypass DDL triggers? I don't know if that can actually happen in practice though.

PR #5829, opened 2026-04-17
fix(zero-cache): handle nested ddl events

darkgnotic:
Correctly handle nested DDL events in schema change detection.

External event triggers that execute their own DDL statements (e.g. an event trigger that enables RLS on `CREATE TABLE` statements) can cause a nesting of zero events, e.g.

* ddlStart
  * ddlStart
  * ddlUpdate
  * ddlStart
  * ddlUpdate
* ddlUpdate

The previous logic of always comparing a ddlUpdate with the preceding ddlStart would result in missing schema changes when nested in this fashion.

The updated logic treats all events uniformly, each producing a schema snapshot in a linear timeline. Any two snapshots within a transaction are diffed with the immediately preceding snapshot. This ensures that the diff logic correctly matches the actual sequence of potential ddl changes.

Context:
* https://rocicorp.slack.com/archives/C0ARBDKG3D4/p1776376302026549

### Miscellaneous

Coalesce the previous per-command functions / triggers into a single `ddl_end` function / trigger pair, using the built in `TG_TAG` variable to determine the tag instead.

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/change-source.ts
+++ b/packages/zero-cache/src/services/change-source/pg/change-source.ts
@@ -88,11 +88,7 @@ import type {
 } from './logical-replication/pgoutput.types.ts';
 import {subscribe, type StreamMessage} from './logical-replication/stream.ts';
 import {fromBigInt, toBigInt, toStateVersionString, type LSN} from './lsn.ts';
-import {
-  replicationEventSchema,
-  type DdlUpdateEvent,
-  type SchemaSnapshotEvent,
-} from './schema/ddl.ts';
+import {replicationEventSchema, type ReplicationEvent} from './schema/ddl.ts';
 import {updateShardSchema} from './schema/init.ts';
 import {
   getPublicationInfo,
@@ -1041,7 +1037,7 @@ class ChangeMaker {
         }
 
       case 'commit':
-        this.#lastSnapshotInTx = undefined;
+        this.#lastReplicationEventInTx = undefined;
         return [
           [
             'commit',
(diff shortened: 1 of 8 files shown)


grgbkr:
Very nice fix!

packages/zero-cache/src/services/change-source/pg/change-source.ts
```diff
@@ -1064,49 +1060,56 @@ class ChangeMaker {
...
     }
 
-    // Store the schema (from either a ddlUpdate or schemaSnapshot) to
-    // diff against the next schemaSnapshot.
-    this.#lastSnapshotInTx = event.schema;
-    if (!previousSchema) {
-      this.#lc.info?.(`received ${msg.prefix}/${type} event`);
-      return []; // First schemaSnapshot in the tx.
+    // Store the new event to diff against the next event.
+    this.#lastReplicationEventInTx = event;
```
arv, line 1087: I was expecting some kind of stack here?

Maybe this works?

arv (commented):
LGTM

packages/zero-cache/src/services/change-source/pg/change-source.ts
darkgnotic: Yeah, it's not a stack, instead will always perform a diff chronologically, and never "skip ober" a snapshot in a diff. So in the case of:

```
start_a -> start_b -> end_b ->  end_a
```

diffs will be computed between `[start_a, start_b]`, then `[start_b, end_b]`, then `[end_b, end_a]`, i.e. always diffing consecutive snapshots.

PR #5835, opened 2026-04-17
feat(zero-client): log sink option to redirect client output

tantaman:
If a user runs zero-client inside a node process this would be a generally desired thing to do. Even in the browser if the user wants to collect zero-client logs in a special way.

### DIFF
--- a/packages/analyze-query/src/analyze-cli.ts
+++ b/packages/analyze-query/src/analyze-cli.ts
@@ -1,6 +1,8 @@
 import '../../shared/src/dotenv.ts';
 
+import {Console} from 'node:console';
 import {styleText} from 'node:util';
+import type {LogSink} from '@rocicorp/logger';
 import {WebSocket as NodeWebSocket} from 'ws';
 import {logLevel, logOptions} from '../../otel/src/log-options.ts';
 import {colorConsole} from '../../shared/src/logging.ts';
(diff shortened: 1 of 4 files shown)


packages/analyze-query/src/analyze-cli.ts
```diff
@@ -116,6 +118,24 @@ type QueryPlan =
   | {kind: 'zql'; text: string}
   | {kind: 'named'; name: string; args: ReadonlyArray<unknown>};
 
+// Route all Zero client log output to stderr so stdout contains only the
```
arv, line 121: It is kind of weird to route info and debug to stderr but 🤷🏼 

Another option might be to just silence all the non errors?

packages/zero-client/src/client/options.ts
```diff
@@ -88,6 +88,19 @@ export type ZeroOptions<
...
+   * `console.log`/`info`/`warn`/`error` etc.
+   *
+   * Provide a custom {@linkcode LogSink} to redirect logs — for example, to
+   * `process.stderr` in a CLI so that the tool's structured stdout output
+   * isn't polluted by log lines.
+   *
+   * Note: this overrides only the console sink. When analytics logging is
+   * enabled, the Datadog sink is still installed alongside this one.
+   */
+  logSink?: LogSink | undefined;
```
arv, line 102: This is an addition to the public API.

This brings back memories.

Also, are we still using the Datadog logsink for anything?

@aboodman

arv (commented):
LGTM

PR #5836, opened 2026-04-17
feat: add `userID` to server helpers & API protocol

0xcadams:
This PR adds backwards compatible changes to `handleQueryRequest` and `handleMutateRequest` to add a new parameter, `userID`.

This also refactors the old schemas and zero-client/zero-cache WS schemas for clarity.

`fetch.ts` will continue to accept old tuple-type responses and will handle gracefully.

Also adds tests for zero-cache/API response protocol backwards compatibility in both `zero-protocol` and `fetch.ts`.

This *does not* change any of the zero-cache/API request protocol, or any of the auth semantics. Auth semantics are in https://github.com/rocicorp/mono/pull/5837 - this is protocol ONLY. 

**THIS WILL BREAK PARSING in older zero-cache talking to a newer API server and zero-cache MUST be rolled out before server.** (we can fix this with a query param? zbugs will be down for a few minutes during rollout because of this.)


packages/zero-cache/src/custom/fetch.ts
```diff
@@ -236,7 +236,9 @@ export async function fetchFromAPIServer<TValidator extends Type>(
 
       try {
         const json = await response.json();
-        const result = validator.parse(json);
+        const result = validator.parse(json, {
+          mode: 'strip',
```
grgbkr: We've preferred passthrough over strip so far.   It's slightly more performant since valita doensnt have to do structural cloning to remove unknown fields.

packages/zero-protocol/src/push.ts
```diff
@@ -217,36 +205,117 @@ const zeroPusherErrorSchema = v.object({
...
+  id: mutationIDSchema,
+  result: apiMutationResultSchema,
+});
+export type APIMutationResponse = v.Infer<typeof apiMutationResponseSchema>;
+
+const apiPushSuccessSchema = v.object({
+  mutations: v.array(apiMutationResponseSchema),
+});
+
+const apiPushResponseSchema = v.union(
```
grgbkr: Isn't this part of the legacy API?

packages/zero-protocol/src/custom-queries.ts
```diff
@@ -51,25 +59,56 @@ export const transformRequestMessageSchema = v.tuple([
...
   transformFailedBodySchema,
 ]);
-const transformOkMessageSchema = v.tuple([
+const legacyTransformSuccessMessageSchema = v.tuple([
   v.literal('transformed'),
-  transformResponseBodySchema,
+  queryResponseBodySchema,
 ]);
 
 export const transformResponseMessageSchema = v.union(
```
grgbkr, line 69: might be clearer to just call this legacyTransformResponseMessageSchema

packages/zero-protocol/src/custom-queries.ts
```diff
@@ -51,25 +59,56 @@ export const transformRequestMessageSchema = v.tuple([
...
+
+export const queryResponseSchema = v.union(
+  querySuccessSchema,
+  transformFailedBodySchema,
+);
+export type QueryResponse = v.Infer<typeof queryResponseSchema>;
+
+export const apiQueryResponseSchema = v.union(
+  queryResponseSchema,
+  transformResponseMessageSchema,
```
grgbkr: might be clearer to refer to this by legacyTransformResponseMessageSchema here.

packages/zero-protocol/src/custom-queries.ts
```diff
@@ -51,25 +59,56 @@ export const transformRequestMessageSchema = v.tuple([
...
 );
 export type TransformResponseMessage = v.Infer<
   typeof transformResponseMessageSchema
 >;
+
+export const legacyTransformResponseMessageSchema =
+  transformResponseMessageSchema;
+export type LegacyTransformResponseMessage = TransformResponseMessage;
+
+export const transformResponseBodySchema = queryResponseBodySchema;
```
grgbkr: Suggests getting rid of these aliases transformResponseBodySchema, TransformResponseBody, and just consistently use queryResponseBodySchema and QueryResponseBody.

packages/zero-protocol/src/push.ts
```diff
@@ -217,36 +205,117 @@ const zeroPusherErrorSchema = v.object({
...
+
+export const mutateResponseSchema = v.union(
+  mutateSuccessSchema,
+  pushFailedBodySchema,
+);
+export type MutateResponse = v.Infer<typeof mutateResponseSchema>;
+
+export const apiMutateResponseSchema = v.union(
+  mutateResponseSchema,
+  apiPushResponseSchema,
```
grgbkr: I'm struggling to follow the changes to this api. 

The new response API is mutateResponseSchema.

It looks like pushResponseSchema is the old response API.  

What is the apiPushResponseSchema?  

Why is this a union of mutateResponseSchema and apiPushResponseSchema, and not a union of mutateResponseSchema and pushResponseSchema?


I think it could also be useful to prefix the legacy stuff with legacy or deprecated as you started to do in custom-queries.ts.

packages/zero-server/src/queries/process-queries.ts
```diff
@@ -58,47 +65,257 @@ export function handleTransformRequest<S extends Schema>(
...
+    requestOrJsonBody,
+    'transform',
+    logLevel,
+  );
 }
 
+type UserID = string | null | undefined;
+type QueryString = URLSearchParams | Record<string, string>;
+
+type HandleQueryRequestArgs = [
```
grgbkr: The overloaded signatures here are getting pretty hard to follow, and the normalizeQueryRequestArgs / isUserID / isQueryString runtime sniffing is fragile — e.g. isUserID matches string | null | undefined, which overlaps with ReadonlyJSONValue, so correctness depends on check ordering. Same risk with isQueryString misidentifying a Record<string, string> JSON body as query params.

How about an options bag instead? The pre-existing signature has logLevel as the 4th positional arg, so we can extend that slot to accept LogLevel | Options:

```
type QueryRequestOptions = {
  userID?: string | null;
  queryParams?: URLSearchParams | Record<string, string>;
  logLevel?: LogLevel;
};
export function handleQueryRequest<S extends Schema>(
  transformQuery: TransformQueryFunction,
  schema: S,
  requestOrJsonBody: Request | ReadonlyJSONValue,
  logLevelOrOptions?: LogLevel | QueryRequestOptions,
): Promise<QueryResponse> {
  const options = typeof logLevelOrOptions === 'string'
    ? { logLevel: logLevelOrOptions }
    : logLevelOrOptions ?? {};
  // ...
}
```

This is fully backwards compatible (existing callers pass nothing or a LogLevel string), eliminates the ambiguous type sniffing, and is easy to extend later if we need to thread more context through without another round of overload surgery.

packages/zero-server/src/process-mutations.ts
```diff
@@ -104,6 +104,36 @@ const applicationErrorWrapper = async <T>(fn: () => Promise<T>): Promise<T> => {
  */
 export const handleMutationRequest = handleMutateRequest;
 
+type UserID = string | null | undefined;
+type QueryString = URLSearchParams | Record<string, string>;
+
+type HandleMutateRequestArgs = [
```
grgbkr: Same feedback as on process-queries.ts — the options bag approach works here too. The existing two overloads already have logLevel as the last arg, so we can widen that slot:
```
typescript
type MutateRequestOptions = {
  userID?: string | null;
  logLevel?: LogLevel;
};
// Overload 1: pre-parsed (shape unchanged)
export function handleMutateRequest(
  dbProvider, cb, queryString, body, logLevelOrOptions?: LogLevel | MutateRequestOptions,
): Promise<MutateResponse>;
// Overload 2: Request (shape unchanged)
export function handleMutateRequest(
  dbProvider, cb, request, logLevelOrOptions?: LogLevel | MutateRequestOptions,
): Promise<MutateResponse>;
```
This keeps the existing two overloads (which are clean — instanceof Request is an unambiguous guard)

PR #5837, opened 2026-04-17
feat(zero-cache): added server-authoritative `userID` check in ctx manager

0xcadams:
This changes validated auth state reflect server-confirmed identity instead of trusting the websocket client's claimed `userID`.

- Bind validated connection and client-group auth state in `zero-cache` to the `userID` confirmed by the API server.
- New semantics:
  - `userID: "<id>"` means the server confirmed that exact user.
  - `userID: null` means the server confirmed the connection is logged out.
  - No returned `userID` (or `userID: undefined`) means fall back to the connection's stored `userID` for legacy compatibility.
- Apply this binding across connect, transform, revalidation, and push flows.
- Reject and invalidate connections when the client-claimed `userID` does not match the server-confirmed identity.

Builds on https://github.com/rocicorp/mono/pull/5836 where protocol was updated.


packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -365,7 +367,30 @@ export class ConnectionContextManagerImpl implements ConnectionContextManager {
       return undefined;
     }
 
-    if (this.#group.validated && this.#group.userID !== connection.userID) {
+    const effectiveValidatedUserID = normalizeValidatedUserID(
```
aboodman: I had a hard time following this logic. I think it is for the following reasons:

* the meaning of `null` and `undefined` is not consistent. The userID coming in from server can be string|null|undefined and `null` means the server has asserted the user is logged out. But in the implementation here `undefined` means that.
* The `effectiveValidatedUserID` can have the client-claimed userID mixed in in some cases.
* The naming is a little unclear to me in cases. In this kind of auth code I like to be really clear about which userID is which.

I suggest the following changes:

* Call the userID that comes from the server the `authenticatedUserID`, because that's what it is. It's what the server has authenticatd the user ID to be, from the auth token.
* Normalize `authenticatedUserID` to `string|null|undefined` early on. `null` meaning server is asserting user is logged out, `undefined` meaning server is not asserting anything.
* Call the userID on the group `requiredUserID` because that is what it is, it's the `userID` that the user must be in order to access the group.
* Call the userID on the connection `claimedUserID` because that is what it is, it's just what the client claims, it's not necessarily true.
* Change the type of group.requiredUserID to `string|null|undefined` with same semantics as `authenticatedUserID`: string means it must be this userID. null means it must be null userID (logged out). undefined means there is no required userID (the client group is new and hasn't been locked yet).
* Change the logic to flow in three linear steps:
  1. if authenticatedUserID is not undefined (server has auth'd the user), validate that claimedUserID is the same.
  2. if the group's requiredUserID is not undefined, validate that the claimedUserID (now authenticated to the extent possible) is the same.
  3. otherwise, if the group's requiredUserID is undefined, set the group's requiredUserID to the claimedUserID.

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -16,6 +16,21 @@ import type {ConnectParams} from '../../workers/connect-params.ts';
...
+ * `id: null` means logged out.
+ */
+export type UserState = {id: string | null};
+
+/**
+ * Delineates the two paths for validating a connection: either server can validate
+ * the user's identity and return a definitive userID to trust, or we fall back to
+ * trusting the one provided by the client in the incoming query params.
+ */
+export type ConnectionValidation =
```
0xcadams, line 30: Added these types for clarity.

packages/zero-cache/src/custom-queries/transform-query.ts
```diff
@@ -38,11 +39,23 @@ type TransformResponse =
...
+  | ({
+      kind: 'success';
+      result: (TransformedAndHashed | ErroredQuery)[];
+    } & (
+      | {
+          cached: true;
+        }
+      | {
+          cached: false;
+          validation: ConnectionValidation;
```
0xcadams, line 58: These types ensure that if a transform is cached, it will *not* validate the connection.

packages/zero-cache/src/services/view-syncer/connection-context-manager.ts
```diff
@@ -74,13 +89,9 @@ export type ConnectionContext = {
  * The background connection is the validated connection currently used for
  * shared background work. Retransform happens on a group level, and uses
  * the background connection's credential to refetch the latest queries.
- *
- * Since auth can be pinned to logged-out users, `userID` can be undefined
- * even after validation.
  */
 export type GroupAuthState = {
-  userID: string | undefined;
-  validated: boolean;
```
grgbkr, line 83: did you get rid of `validated` because its redundant with `pinnedUser !== undefined`

0xcadams: It was less clear. Before:
- pinned to logged out: validated: true & userID: undefined
- no pin yet: validated: false & userID: undefined

Now:
- pinned to logged out: pinnedUser: {id: null}
- no pin yet: pinnedUser: undefined

PR #5838, opened 2026-04-17
feat(zero-cache): added five min default for revalidate/retransform

0xcadams:
This PR makes periodic auth revalidation and shared query retransform run every 5 minutes by default in `zero-cache` instead of requiring explicit env configuration.

The intervals remain configurable via `ZERO_AUTH_REVALIDATE_INTERVAL_SECONDS` and `ZERO_AUTH_RETRANSFORM_INTERVAL_SECONDS`.


grgbkr:
LGTM

PR #5839, opened 2026-04-17
feat(zero-server): add kysely db adapter

0xcadams:
- add a `zeroKysely` adapter alongside the existing Prisma and Drizzle adapters so Zero server mutators can run against a Kysely transaction
- lean on Kysely's own transaction and raw-query helpers by using `transaction().execute(...)` and `CompiledQuery.raw(...)`
- export the adapter from `@rocicorp/zero/server/adapters/kysely` and add `kysely` as an optional peer dependency on `@rocicorp/zero`
- add Kysely adapter coverage plus rollback coverage across all server adapters


arv:
This seems fine conceptually.

At some point we might want to move this to a different repo.

0xcadams:
> At some point we might want to move this to a different repo.

Agreed

arv (approved):
LGTM

PR #5840, opened 2026-04-17
chore(zero-cache): catch query drift by hashing query result

tantaman:
required to re-introduce CAP which can be non-deterministic on re-hydration


tantaman:
submitting as split up prs

PR #5842, opened 2026-04-18
fix(zero-cache): potentially controversial: log all ws errors as warning

aboodman:
I got to this by handling an alert for a ws message that was too large:

https://app.incident.io/rocicorp/on-call/alerts/01KPF3T0C4A30GDCV91J3MK388/details

But this lead to the question: why do we ever want to alert for errors from web sockets? They are always some kind of problem with either a client or upstream server.


darkgnotic:
Good question ... what was the conclusion for the `RangeError: Max payload size exceeded` alert? Was it something we were doing wrong in the server code, or was it an external cause, outside of our control?

The intention is that the error-level logs alert us of problems in the code that we should fix. If they are instead producing noise from issues outside of our control, then downgrading them to warn makes sense to me.

darkgnotic (approved):

PR #5844, opened 2026-04-18
feat(zql): add BTreeSet.fromSorted and use it in #getOrCreateIndex

arv:
## What

Adds `BTreeSet.fromSorted(comparator, sortedEntries)` — a static factory that builds the tree bottom-up in O(N) from a pre-sorted iterable, without materializing a flat intermediate array. Leaf nodes are flushed as they fill to `MAX_NODE_SIZE`; internal nodes are assembled level-by-level from the leaves.

Uses it in `MemorySource.#getOrCreateIndex` to replace the prior one-at-a-time `add()` loop (which even had a comment acknowledging the missing fast path).

## Why

`#getOrCreateIndex` fires on the client side (via `IVMSourceBranch`) the first time a query requests a sort order not yet indexed on a `MemorySource` that already has data — i.e. the cold-start / first-render scenario. The prior approach did N individual `add()` calls, each O(log N), for O(N log N) total with large constant factors (binary search + potential node splits per insert).

The new path: collect primary index rows → `Array.sort(comparator)` → `fromSorted`. The sort step is unavoidable (the primary index uses a different comparator; if they matched, the early-return in `#getOrCreateIndex` would have already returned the existing index). But the tree construction itself drops from O(N log N) to O(N).

## Perf (1 000 rows, Node.js)

| Benchmark | Before | After | Speedup |
|---|---|---|---|
| Build tree from 100 pre-sorted keys | 8.68 µs | 476 ns | **18×** |
| Build tree from 1 000 pre-sorted keys | 90.22 µs | 2.56 µs | **35×** |
| `#getOrCreateIndex` pattern (sort + build, 1 000 rows) | 143.11 µs | 18.86 µs | **7.6×** |

The pure `fromSorted` numbers show the ceiling when data arrives in the right order. The `#getOrCreateIndex` number is the real-world case — the sort now dominates, and construction is essentially free.

### DIFF
--- a/packages/shared/src/btree-set.bench.ts
+++ b/packages/shared/src/btree-set.bench.ts
@@ -149,12 +149,54 @@ describe('BTreeSet mutations', () => {
     };
   });
 
+  bench('fromSorted() 100 sequential keys', function* () {
+    const keys = Array.from({length: 100}, (_, i) => i);
+    yield () => {
+      const t = BTreeSet.fromSorted(numericComparator, keys);
+      use(t.size);
+    };
+  });
+
   bench('add() 1000 sequential keys', function* () {
     yield () => {
       const t = new BTreeSet<number>(numericComparator);
       for (let i = 0; i < NUM_ENTRIES; i++) {
         t.add(i);
       }
+      use(t.size);
+    };
+  });
+
+  bench('fromSorted() 1000 sequential keys', function* () {
+    const keys = Array.from({length: NUM_ENTRIES}, (_, i) => i);
+    yield () => {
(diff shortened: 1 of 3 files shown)


packages/shared/src/btree-set.bench.ts
```diff
@@ -149,6 +149,14 @@ describe('BTreeSet mutations', () => {
     };
   });
 
+  bench('fromSorted() 100 sequential keys', function* () {
```
arv: Remove this test.

tantaman (commented):
lgtm

PR #5846, opened 2026-04-19
chore(zql): improve asQueryInternals assertion message

arv:
Improve asQueryInternals assertion message to hint at duplicate Zero copies.


0xcadams (commented):
Great!

PR #5852, opened 2026-04-21
Fix hang creating replication slot

aboodman:
We saw two instances of zero-cache refusing to startup due to hang creating a replication slot. This was caused by an idle transaction hanging open.

"Fixed" immediate issue by adding a timeout to this statement. This will cause zero-cache to restart, but if the offending process is not zero-cache this won't help.

Also required all pg connections to be named so that if the offending transaction is part of zero cache we'll be able to identify it next time.

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/initial-sync.ts
+++ b/packages/zero-cache/src/services/change-source/pg/initial-sync.ts
@@ -122,13 +122,13 @@ export async function initialSync(
     shadow,
   } = syncOptions;
   const copyProfiler = profileCopy ? await CpuProfiler.connect() : null;
-  const sql = pgClient(lc, upstreamURI);
+  const sql = pgClient(lc, upstreamURI, 'initial-sync');
   // Replication session is only needed to create a replication slot in the
   // real path. In shadow mode we export a snapshot on a normal connection
   // instead, so no replication session is opened.
   const replicationSession = shadow
     ? undefined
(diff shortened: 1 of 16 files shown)


packages/zero-cache/src/services/change-source/pg/initial-sync.ts
```diff
@@ -529,6 +529,10 @@ type ReplicationSlot = {
   output_plugin: string;
 };
 
+// Successful CREATE_REPLICATION_SLOT calls have always completed within 1s in
+// observed production runs; use 5s as a hard stop for pathological hangs.
+const CREATE_REPLICATION_SLOT_TIMEOUT_MS = 5_000;
```
darkgnotic, line 534: FWIW I have seen legitimate delays in replication slot creation, particularly when I was testing backfill by running schema changes that touched rows in all tables (e.g. adding a column with a random value).

The explanation from Claude was that postgres has to wait for a large operation to finish writing to the wal before it can create a new replication slot, which requires creating a transaction snapshot consistent with a wal position.

I feel like 5 seconds might be an aggressive timeout, but I guess if the server is stuck waiting anyway it doesn't hurt to crash and retry.  🤷

darkgnotic: (It has also be failing tests, which is how I noticed this  😉 )

PR #5856, opened 2026-04-21
chore(ci) try oxfmt pinning

tantaman:
claude seems to think the hang is from using a different oxfmt version 🤷‍♂️


tantaman:
looks like it worked for https://github.com/rocicorp/mono/pull/5849

PR #5857, opened 2026-04-21
 chore(zero-cache): detect row-set signature drift and re-execute drifted queries

tantaman:
The change is pretty tiny (all in view-syncer.ts + 1 method in cvr.ts). The bulk of this is tests.

- here -> https://github.com/rocicorp/mono/pull/5857
- https://github.com/rocicorp/mono/pull/5849
- https://github.com/rocicorp/mono/pull/5848

## Summary

Builds on the row-set signature XOR introduced in #5849. When the view-syncer re-hydrates queries at the same CVR version (e.g. on startup), it now compares each query's freshly-computed signature against the signature persisted in the CVR. A mismatch means deterministic re-execution produced a different row set — for example, a `Cap` operator picking a different N-row subset — and the query is re-executed through the normal diff-emitting path so clients see the corrected rows.

## What changes

- **Drift detection in `#hydrateUnchangedQueries`** (`view-syncer.ts`): after hydrating each query, compare `#pipelines.rowSetSignature(queryID)` against `cvr.queries[queryID].rowSetSignature`. On mismatch, log a warning, bump the `query.row-set-signature-drifts` OTEL counter, remove the query from the pipelines, and collect its ID in a `driftedQueryIDs` set.
- **Re-execution via the CVR-driven path**: `driftedQueryIDs` is threaded into `#syncQueryPipelineSet('missing', …)` and on to `#addAndRemoveQueries`. Because drifted queries appear as "missing" from the pipelines, the existing add path re-hydrates them through `CVRQueryDrivenUpdater`, which emits the row diff to clients.
- **Forced `configVersion` bump** (`cvr.ts`): a new public `CVRQueryDrivenUpdater.ensureNewVersion()` alias for `_ensureNewVersion` lets the view-syncer force a version bump when the only reason for re-execution is drift (no transformation-hash change, no add/remove). Pending `queryPatches` are re-stamped to the post-bump cookie so pokers don't treat them as stale.
- **Legacy-signature handling**: queries whose stored signature is `null` (from before #5849 deployed) skip the comparison. Their signature gets initialized on the next normal re-execution, at which point drift detection becomes effective for subsequent cycles.

## Observability

- `query.row-set-signature-drifts` counter — expected to be near-zero in steady state. Persistent non-zero values indicate non-deterministic query execution (e.g. `Cap` without a stable tiebreaker).

## Test plan

- [x] `npm --workspace=zero-cache run test -- --project=zero-cache/pg-18 view-syncer`
- [x] New `view-syncer.pg.test.ts` cases cover:
  - signature drift at startup triggers re-execution and client poke
  - legacy (null) signature skips drift check on first cycle
  - `configVersion` bumps when drift is the sole cause of re-execution
  - pending `queryPatches` reach clients at the post-bump cookie
- [x] `npm run lint && npm run format && npm run check-types`

## Follow-ups

- Consider adding a histogram of drifted-queries-per-hydration-cycle so we can tell "one chronic offender" apart from "broad drift."
- If drift counter ever goes non-zero in production, the root cause is almost certainly a `Cap` operator lacking a deterministic tiebreaker.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/view-syncer.pg.test.ts
+++ b/packages/zero-cache/src/services/view-syncer/view-syncer.pg.test.ts
@@ -4955,6 +4955,153 @@ describe('view-syncer/service', () => {
     }
   });
 
+  test('rowSetSignature: emptied-by-advance query rehydrates to "0" without drift', async () => {
+    // Phase 1: hydrate non-empty. Signature is sig(1) ^ sig(2).
+    pruneIssues('3', '4', '5');
+    const client1 = connect(SYNC_CONTEXT, [
+      {op: 'put', hash: 'query-hash1', ast: ISSUES_QUERY2},
+    ]);
+    await nextPoke(client1);
+    stateChanges.push({state: 'version-ready'});
+    await nextPoke(client1);
+    expect(await loadStoredSig('query-hash1')).toEqual(
+      expectedIssuesSig(issueRowID('1'), issueRowID('2')),
+    );
+
+    // Advance: delete both rows via the replicator. #trackRowSetSignatures
+    // XORs each REMOVE with the same unit the ADD contributed, so the live
+    // signature returns to 0n and the CVR persists hex "0".
+    replicator.processTransaction(
+      '123',
+      messages.delete('issues', {id: '1'}),
+      messages.delete('issues', {id: '2'}),
+    );
+    stateChanges.push({state: 'version-ready'});
+    const phase1Cookie = cookieOf(await nextPoke(client1));
+    expect(await loadStoredSig('query-hash1')).toEqual('0');
+
+    // Phase 2: restart. Rehydration yields 0 rows → the map entry is never
+    // created → provider returns undefined → candidate = undefined ?? 0n = 0n.
+    // Stored "0" parses to 0n. Match: no drift, no re-execution.
+    const {
+      queue,
+      stateChanges: rsc,
+      cleanup,
+    } = await restartAfter({
+      baseCookie: phase1Cookie,
+      mutate: () => {},
+      queriesPatch: [{op: 'put', hash: 'query-hash1', ast: ISSUES_QUERY2}],
+    });
+    const removeSpy = vi.spyOn(PipelineDriver.prototype, 'removeQuery');
+    try {
+      rsc.push({state: 'version-ready'});
+
+      expect(await drainUntilRowsPatchOrQuiet(queue)).toBeUndefined();
+      expect(await loadStoredSig('query-hash1')).toEqual('0');
+
+      // Single call = internal removeQuery from the one addQuery in
+      // hydrateUnchangedQueries. Drift would be 3.
+      const calls = removeSpy.mock.calls.filter(
+        ([id]) => id === 'query-hash1',
+      ).length;
+      expect(calls).toBe(1);
+    } finally {
+      removeSpy.mockRestore();
+      await cleanup();
+    }
+  });
+
+  test('rowSetSignature: transformationHash change bypasses the drift check', async () => {
+    // Direct-mock the transformer so we can drive the transformationHash
+    // explicitly (the real transformer caches by query id for 5s, which would
+    // re-serve phase 1's hash in phase 2).
+    pruneIssues('3', '4', '5');
+    const transformSpy = vi
+      .spyOn(customQueryTransformer!, 'transform')
+      .mockResolvedValue(
+        transformAttempt([
+          {
+            id: 'custom-1',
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/services/view-syncer/view-syncer.pg.test.ts
```diff
@@ -4480,9 +4620,321 @@ describe('view-syncer/service', () => {
...
+      queriesPatch: [{op: 'put', hash: 'query-hash1', ast: ISSUES_QUERY}],
+    });
+    try {
+      rsc.push({state: 'version-ready'});
+
+      // CVRQueryDrivenUpdater flush writes the corrected signature
+      // asynchronously. Poll until it matches the driver-side signature.
+      const deadline = Date.now() + 10_000;
+      let finalSig = await loadStoredSig('query-hash1');
+      while (finalSig !== correctSig && Date.now() < deadline) {
```
tantaman: we can use vitest waitFor, no?

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -1468,7 +1491,39 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
       this.#hydrationTime.recordMs(elapsed);
       this.#addQueryMaterializationServerMetric(transformationHash, elapsed);
       lc.debug?.(`hydrated ${count} rows for ${queryID} (${elapsed} ms)`);
+
+      // Drift detection: compare the just-computed candidate signature against
+      // the signature stored in the CVR. They should match for a deterministic
+      // query at the same db state. A mismatch indicates a query containing
+      // the Cap operator picked a different N-row subset on re-execution.
```
grgbkr, line 1498: or a bug

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -1468,7 +1491,39 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
+      // Skip when the stored signature is absent — legacy queries from before
+      // this feature was deployed have no signature to compare against, and a
+      // forced re-execution would needlessly resend rows to the client. Such
+      // queries get their signature initialized whenever they next re-execute
+      // via the normal path (transformation hash change, etc.), at which
+      // point drift detection becomes effective for subsequent cycles.
+      const storedSigHex = cvr.queries[queryID]?.rowSetSignature;
+      if (storedSigHex !== undefined && storedSigHex !== null) {
+        const priorSig = parseSignature(storedSigHex);
+        const candidateSig = this.#pipelines.rowSetSignature(queryID) ?? 0n;
```
grgbkr, line 1512: should this be an assert, the rowSetSignature should alway exist for queryID here, correct?

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -1468,7 +1491,39 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
+        if (priorSig !== candidateSig) {
+          lc.warn?.(
```
grgbkr, line 1514: maybe we should check the ast for cap.  if cap is not in the ast, this indicates a bug that needs investigation

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -1868,15 +1926,27 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
+      // startPoke so the pokers see the final cookie version.
+      if (addQueries.some(q => driftedQueryIDs.has(q.id))) {
+        updater.ensureNewVersion();
+      }
+      const newVersion = updater.updatedVersion();
       const clients = this.#getClients();
       const pokers = startPoke(clients, newVersion);
       for (const patch of queryPatches) {
-        await pokers.addPatch(patch);
+        // Bump patches' toVersion to the post-drift-bump version so that
```
grgbkr, line 1947: This seems fishy.

Should we be doing `updater.ensureNewVersion();` before calling 
```
 const {queryPatches} = updater.trackQueries(
        lc,
        addQueries,
        removeQueries,
      );
```
so that the query patches have the right version and we don't have to override them here.

tantaman: yeah, that would be cleaner. good catch.

packages/zero-cache/src/services/view-syncer/view-syncer.ts
tantaman: if the query returns no row this is undefined which should map to a hash of 0

PR #5862, opened 2026-04-22
fix(zero-cache,replicache): preserve sqlite tx failures via Error.cause

aboodman:
Some SQLite errors auto-abort the transaction. In this case, calling ROLLBACK throws a new error, obscuring the original one.

Fix this pattern throughout the codebase.

Original report: https://discord.com/channels/830183651022471199/1495148037935206620/1496448522860630017

### DIFF
--- a/packages/replicache/src/with-transactions.ts
+++ b/packages/replicache/src/with-transactions.ts
@@ -49,9 +49,31 @@ export async function using<TX extends Release, Return>(
   fn: (tx: TX) => Return | Promise<Return>,
 ): Promise<Return> {
   const write = await x;
+  let result: Return | undefined;
+  let operationError: unknown;
   try {
-    return await fn(write);
-  } finally {
-    write.release();
+    result = await fn(write);
+  } catch (e) {
+    operationError = e;
   }
+
+  try {
+    write.release();
+  } catch (releaseError) {
+    if (operationError !== undefined) {
+      const combinedError = new Error(
+        `Transaction operation failed and release also failed: operation error = ${String(
+          operationError,
+        )}; release error = ${String(releaseError)}`,
+      );
+      combinedError.cause = operationError;
+      throw combinedError;
+    }
+    throw releaseError;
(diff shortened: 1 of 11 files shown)


packages/replicache/src/with-transactions.ts
```diff
@@ -49,9 +49,31 @@ export async function using<TX extends Release, Return>(
...
-  } finally {
     write.release();
+  } catch (releaseError) {
+    if (operationError !== undefined) {
+      const combinedError = new Error(
+        `Transaction operation failed and release also failed: operation error = ${String(
+          operationError,
+        )}; release error = ${String(releaseError)}`,
+      );
+      combinedError.cause = operationError;
```
arv, line 69: FYI,

```js
new Error(`...`, {cause: operationError})
```

packages/replicache/src/with-transactions.ts
arv, line 47: Seems subtle enough to warrant a test at this layer.

packages/replicache/src/with-transactions.ts
arv: but...

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SuppressedError

would be a lot better if present.

My goal was always to make this use JS resource management but it is not supported in Safari yet.

packages/zero-cache/src/db/migration-lite.ts
```diff
@@ -336,7 +336,18 @@ async function runTransaction<T>(
     return result;
   } catch (e) {
     log.error?.('Aborted transaction due to error', e);
-    db.prepare('ROLLBACK').run();
```
arv, line 339: At this point I would have added a helper.

```js
rollbackTransaction(() => db.prepare('ROLLBACK').run());
```

you can put this in `with-transactions.ts`

arv (commented):
LGTM with nits

This is great.

PR #5865, opened 2026-04-22
feat(zql): boolean absorption, duplication elimination, or -> in

tantaman:


tantaman:
https://github.com/rocicorp/mono/pull/5894 will fix most of our problems. Boolean absorption still a thing to do though.

PR #5866, opened 2026-04-22
fix(zero-cache): improve resolution of latency histograms

grgbkr:
## Fix histogram bucket boundaries for latency metrics

All latency histograms (`hydration-time`, `advance-time`, `ivm.advance-time`, `lock-wait-time`, `query.transformation-time`, `poke.time`, `cvr.flush-time`) were using OTel's default bucket boundaries (`[0, 5, 10, 25, …, 10000]` seconds). Since our operations typically complete in 1–5000ms, 100% of observations landed in the first bucket, producing flat useless quantile lines.

**Fix:**
- Introduce `getOrCreateLatencyHistogram()` → `LatencyHistogram` with a `recordMs()` method that bakes in `unit: 's'`, appropriate bucket boundaries, and the ms→s conversion — making it impossible for future metrics to repeat this mistake.
- Delete the now-unused `getOrCreateHistogram()` so there's no footgun to reach for.
- Migrate all 7 latency histograms to the new API.

Bucket set: `1ms, 2ms, 5ms, 10ms, 20ms, 50ms, 100ms, 200ms, 500ms, 1s, 2s, 5s, 10s, 30s`

### DIFF
--- a/packages/zero-cache/src/observability/metrics.ts
+++ b/packages/zero-cache/src/observability/metrics.ts
@@ -75,8 +75,7 @@ export function getOrCreateUpDownCounter(
  * durations and converts them to seconds internally.
  *
  * Use {@link getOrCreateLatencyHistogram} to create one — the unit (`'s'`),
- * bucket boundaries, and ms→s conversion are all baked in and cannot be
- * forgotten.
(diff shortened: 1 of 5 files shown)


tantaman:
thanks

PR #5867, opened 2026-04-22
fix(zero-cache): handle PG array defaults in ALTER TABLE ADD COLUMN

arv:
## Problem

When a PostgreSQL column with an array default (e.g. `ARRAY[]::text[]` or `ARRAY['a','b']::text[]`) is added via `ALTER TABLE ADD COLUMN`, the default expression was not handled by `mapPostgresToLiteDefault`, causing it to fall through to SQLite which doesn't understand PG array constructor syntax and would reject the statement.

## Fix

**`pg-to-lite.ts`**: Add explicit handling for PG array constructor expressions (`ARRAY[...]::type[]`):
- Empty arrays (`ARRAY[]::text[]`) are mapped to the JSON string literal `'[]'` in SQLite.
- Non-empty arrays (`ARRAY['a','b']::text[]`) throw `UnsupportedColumnDefaultError`, triggering the existing backfill path in `change-source`.

**`change-processor.ts`**: If SQLite rejects an `ADD COLUMN` with a `DEFAULT` (i.e. a PG-specific expression slipped through), throw `UnsupportedColumnDefaultError` instead of failing hard. This causes `change-source` to backfill the column from upstream rather than crashing.

### DIFF
--- a/packages/zero-cache/src/db/pg-to-lite.test.ts
+++ b/packages/zero-cache/src/db/pg-to-lite.test.ts
@@ -414,6 +414,10 @@ test.each([
   ['CURRENT_TIMESTAMP'],
   ['Current_Time'],
   ['current_DATE'],
+  // Non-empty array constructors are unsupported (trigger backfill path)
+  ["ARRAY['foo']::text[]"],
+  ["ARRAY['a','b']::text[]"],
+  ['ARRAY[1,2]::integer[]'],
 ])('unsupported column default %s', value => {
   expect(() =>
     mapPostgresToLiteDefault('foo', 'bar', 'boolean', value),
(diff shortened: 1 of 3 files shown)


aboodman:
How have we not hit this before?

arv:
> How have we not hit this before?

Nope. First time.

aboodman:
@copilot how have we not seen this before?

aboodman:
Thank you for this but on investigating I think the function `mapPostgresToLiteDefault` needs to be rewritten. It is operating on a denylist approach rather than allowlist and we're going to keep having these kind of problems. Postgrss default value declarations can be arbitrarily complex, for example all of these are valid default values that the current function would barf on:

```ts
# special pg goop
LOCALTIME, LOCALTIMESTAMP
CURRENT_USER, CURRENT_ROLE, CURRENT_SCHEMA, CURRENT_CATALOG
SESSION_USER

# boolean logic
true AND false
NOT true

...etc...
```

We should allowlist for just the specific values we can definitely propagate to SQLite and fallback to backfill for everything else. Backfill is super powerful! the only downside is that it's slow. The mapping of values is an optimization.

aboodman:
I was talking this through w/ claude I'll have him just update the function w/ that context it will be fast.

aboodman:
superseded by https://github.com/rocicorp/mono/pull/5868

PR #5868, opened 2026-04-22
fix(zero-cache): Rewrite `mapPostgresToLiteDefault` to use allowlist.

aboodman:
The previous implementation used a denylist and had many cases of valid PG syntax that would not be mapped correctly to SQLite values:

- LOCALTIME, LOCALTIMESTAMP
- CURRENT_USER, CURRENT_ROLE, CURRENT_SCHEMA, CURRENT_CATALOG
- SESSION_USER
- true AND false
- NOT true
- etc...

Now we only map a small set of common known-supported values and everything else falls back to backfill:

- Numeric literals: 123, -45.67
- Boolean literals: true/false → 1/0
- Quoted strings with simple type casts: 'foo'::text
- Empty arrays: ARRAY[]::type[] and '{}'::type[] → '[]'

Removed the dataType parameter since we trust PG's type validation - PG already enforces that defaults match column types at schema definition time.

This supersedes https://github.com/rocicorp/mono/pull/5867

### DIFF
--- a/packages/zero-cache/src/db/pg-to-lite.ts
+++ b/packages/zero-cache/src/db/pg-to-lite.ts
@@ -63,49 +63,83 @@ export function warnIfDataTypeSupported(
   }
 }
 
-// As per https://www.sqlite.org/lang_altertable.html#altertabaddcol,
-// expressions with parentheses are disallowed ...
-const SIMPLE_TOKEN_EXPRESSION_REGEX = /^[^'()]+$/; // e.g. true, false, 1234, 1234.5678
+// Numeric literals: integers and decimals, optionally negative
+const NUMERIC_LITERAL_REGEX = /^-?\d+(\.\d+)?$/;
 
-// as well as current_time, current_date, and current_timestamp ...
-const UNSUPPORTED_TOKENS = /\b(current_time|current_date|current_timestamp)\b/i;
+// Boolean literals (PG emits lowercase)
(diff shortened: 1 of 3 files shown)


darkgnotic (approved):
I did not explicitly research and confirm the structure of the allow list, but I support the general approach of converting to a restrictive allow-list and letting backfill cover the rest.  👍

packages/zero-cache/src/db/pg-to-lite.ts
```diff
@@ -63,49 +63,83 @@ export function warnIfDataTypeSupported(
   }
 }
 
-// As per https://www.sqlite.org/lang_altertable.html#altertabaddcol,
-// expressions with parentheses are disallowed ...
-const SIMPLE_TOKEN_EXPRESSION_REGEX = /^[^'()]+$/; // e.g. true, false, 1234, 1234.5678
+// Numeric literals: integers and decimals, optionally negative
```
arv, line 66: This is only a subset of supported numeric literals, which I think is fine. We will just backfill if not matching the regexp.

OOC.

Here is the PG BNF:

```
digits
digits.[digits][e[+-]digits]
[digits].digits[e[+-]digits]
digitse[+-]digits
```

as well as

```
0xhexdigits
0ooctdigits
0bbindigits
```

and the SQLite BNF:

 
[Image](https://www.sqlite.org/syntax/numeric-literal.html)

arv (approved):
LGTM

PR #5869, opened 2026-04-22
chore(zero-cachee): log error (with context) before asserting on missing ddl_start

darkgnotic:
Log an error-level with message with full context before asserting on a missing ddlStart message to collect/confirm the situations in which this happens.

This is a short-term data gathering step before the fix is in place.


tantaman (commented):
lgtm

PR #5874, opened 2026-04-23
fix(zero-cache): replication slot timeout improvement

grgbkr:
## Add server-side `lock_timeout` to `CREATE_REPLICATION_SLOT`

`CREATE_REPLICATION_SLOT` can hang indefinitely when long-running transactions block snapshot creation. We already have a client-side `orTimeout` that closes the connection after 5s, but this leaves an orphaned walsender on the Postgres server — the backend is asleep in `LockAcquire` and won't notice the closed socket until TCP keepalive fires (~2h default) or the blocking transaction finishes.

The orphaned walsender is actively harmful because the replication slot has already been created by this point. It pins WAL retention and `catalog_xmin`, and because it's marked `active`, the existing inactive-slot cleanup logic can't drop it.

This adds `SET lock_timeout` before the slot creation. `statement_timeout` does not apply to replication commands, but `lock_timeout` does — it governs the heavyweight lock wait inside `LockAcquire` (the call chain is `SnapBuildWaitSnapshot` → `XactLockTableWait` → `LockAcquire`). When it fires, Postgres raises `ERRCODE_LOCK_NOT_AVAILABLE` and cleanly tears down the walsender and slot.

Note that `lock_timeout` applies per `LockAcquire` call, not across the entire operation — `SnapBuildWaitSnapshot` loops over all running XIDs and waits on each one individually. In theory, many short-but-not-quite-expired waits could add up. In practice the pathological case is a single stuck transaction, and the client-side `orTimeout` (retained as defense-in-depth) provides an absolute wall-clock cap regardless.


aboodman:
OK this seems better but do we still want/need the client-side timeout?

aboodman:
I would rather remove the client-side timeout if it's not doign anything to reduce complexity.

PR #5875, opened 2026-04-23
feat(inspector): remove old analyze-query cli, always enable the query planner in the inspector

tantaman:
The query planner defaulted to off in the inspector which is a bit problematic. Default to on.


packages/zero-cache/src/services/analyze.ts
```diff
@@ -46,10 +46,13 @@ export async function analyzeQuery(
     fullTables,
   );
 
-  const planDebugger = joinPlans ? new AccumulatorDebugger() : undefined;
-  const costModel = joinPlans
+  // Mirror production: the planner runs iff ZERO_ENABLE_QUERY_PLANNER is set
+  // on the server, so the analysis reflects what actually executes. Diagnostic
+  // event collection is orthogonal and opt-in via `joinPlans`.
+  const costModel = config.enableQueryPlanner
```
tantaman, line 52: always create the cost model unless the user disabled the query planner in their zero cache config. This matches the `view-syncer`.

packages/zero-cache/src/services/analyze.ts
```diff
@@ -46,10 +46,13 @@ export async function analyzeQuery(
...
     ? createSQLiteCostModel(db, tableSpecs)
     : undefined;
+  const planDebugger = joinPlans ? new AccumulatorDebugger() : undefined;
```
tantaman, line 55: the `joinPlans` arg no only controls whether or not we accumulate debug information about the considered plans.

aboodman:
CleanShot 2026-04-23 at 16 05 42@2x

@copilot can you fix the breaking package dependency check

arv (commented):
LGTM

PR #5876, opened 2026-04-23
fix(zero-cache): prevent autoreset startup slot deadlock

aboodman:
When startup acquires PurgeLocker and then hits AutoResetSignal, the retry path was deleting the replica but keeping the purge-lock transaction open.

On deployments where CHANGE_DB == UPSTREAM_DB, the retry's initial sync creates a new logical replication slot on the same Postgres. Postgres waits for older transactions before slot creation, so startup could block behind its own still-held purge-lock transaction.

Release and clear purgeLock before retrying initial sync on AutoResetSignal.

Add an integration test that drives real change-streamer startup, forces the missing-slot autoreset path, and verifies startup succeeds by creating a new slot. With the lock release removed, this test reproduces the lock-timeout failure.

### DIFF
--- a/packages/zero-cache/src/server/change-streamer.ts
+++ b/packages/zero-cache/src/server/change-streamer.ts
@@ -168,6 +168,19 @@ export default async function runWorker(
         // TODO: Make deleteLiteDB work with litestream. It will probably have to be
         //       a semantic wipe instead of a file delete.
         deleteLiteDB(replica.file);
+        // Release the purge lock before retrying. This is safe because the
+        // purge lock exists to preserve change-log entries so the new
+        // change-streamer can resume from the backup replica's watermark.
+        // An AutoResetSignal means we cant resume from the backup replica
+        // (e.g. its replication slot is gone), so the change-log entries the lock
+        // was protecting are no longer needed. The retry performs a fresh
+        // initial sync with a new replication slot, independent of the old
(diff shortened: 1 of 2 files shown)


grgbkr (commented):
LGTM

packages/zero-cache/src/server/change-streamer.ts
```diff
@@ -168,6 +168,10 @@ export default async function runWorker(
         // TODO: Make deleteLiteDB work with litestream. It will probably have to be
         //       a semantic wipe instead of a file delete.
         deleteLiteDB(replica.file);
+        // We need to release the purge lock in order to allow the
```
grgbkr:
```suggestion
      // Release the purge lock before retrying. This is safe because the
      // purge lock exists to preserve change-log entries so the new
      // change-streamer can resume from the backup replica's watermark.
      // An AutoResetSignal means we cant resume from the backup replica
      // (e.g. its replication slot is gone), so the change-log entries the lock
      // was protecting are no longer needed. The retry performs a fresh
      // initial sync with a new replication slot, independent of the old 
      // change-log. Releasing is also necessary to avoid a
      // self-deadlock when CHANGE_DB == UPSTREAM_DB:
      // CREATE_REPLICATION_SLOT waits for all older transactions to
      // finish, including this lock's open transaction.
```

PR #5881, opened 2026-04-24
[wip] chore(zql): quicksort of flipped-join fetches on unique indices

tantaman:
### Summary

When `FlippedJoin.fetch()` runs and the `parentKey` matches a unique
index on the parent (primary key or any `UNIQUE` index), each child →
parent lookup returns at most one row. The existing N-way merge-sort
path opens N simultaneous prepared-statement iterators, each holding a
single-row cursor — wasted work that also defeats the statement-cache
reuse a sequential loop would get.

This PR adds a `#fetchQuicksort` path for that case:

1. Iterate child nodes sequentially, fetching the (≤1) parent for each
   so the same prepared statement is reused.
2. Sort the resulting `(child, parent)` pairs by `parent.row` using the
   schema's `compareRows` (respecting `req.reverse`).
3. Group consecutive equal parents and yield each parent once with its
   matched children.

The existing N-way merge-sort path (`#fetchMergeSort`) is preserved
unchanged for the non-unique case. Overlay handling for in-progress
child changes is extracted into `#yieldParentWithOverlay` so both paths
share it.

---


tantaman:
I'm replacing quicksort.

Turns out merge-sort with `IN (...)` batching and a min-heap outperforms quicksort.

PR #5883, opened 2026-04-24
chore(rm): first shadow sync at ~12 hrs

tantaman:
We restart the RM every 24 hrs. So shadow-sync needs to run before then.

https://rocicorp.slack.com/archives/C013XFG80JC/p1777048675512559

### DIFF
--- a/packages/zero-cache/src/services/shadow-sync/shadow-sync-service.ts
+++ b/packages/zero-cache/src/services/shadow-sync/shadow-sync-service.ts
@@ -40,11 +40,16 @@ export class ShadowSyncService implements Service {
   async run() {
     const {intervalMs, sampleRate, maxRowsPerTable, textCopy} = this.#options;
 
-    // Why: wait at least one full interval before the first run so shadow
-    // sync never fires immediately on task startup, and add a random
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/services/shadow-sync/shadow-sync-service.ts
```diff
@@ -40,11 +40,16 @@ export class ShadowSyncService implements Service {
...
-    // sync never fires immediately on task startup, and add a random
-    // fraction of the interval on top so a fleet-wide restart does not
-    // cause every task to canary simultaneously.
-    const firstRunDelay = intervalMs + Math.floor(Math.random() * intervalMs);
+    // Why: first run fires in [intervalMs * 2/3, intervalMs) — late enough
+    // that shadow sync never fires immediately on task startup, but always
+    // before one full interval elapses so the canary completes at least once
+    // per task lifetime (the replication manager is restarted every ~24h).
+    // The last third is randomized so a fleet-wide restart does not cause
+    // every task to canary simultaneously.
```
tantaman, line 48: iiuc, this is possible if someone has many apps/shards against the same upstream db

0xcadams (commented):
LGTM

PR #5885, opened 2026-04-24
fix(zql): added optional ctx when context is unknown

0xcadams:
Currently when you don't have a context type defined it will default to `unknown` and that means that you need to do this weird workaround:

```tsx
const result = await handleMutateRequest(
  dbProvider,
  transact =>
    transact((tx, name, args) => {
      const mutator = mustGetMutator(mutators, name)
      return mutator.fn({
        args,
        tx,
        // workaround
        ctx: undefined,
      })
    }),
  request
)
```

This allows you to omit `ctx` when `unknown`:

```tsx
const result = await handleMutateRequest(
  dbProvider,
  transact =>
    transact((tx, name, args) => {
      const mutator = mustGetMutator(mutators, name)
      return mutator.fn({
        args,
        tx,
      })
    }),
  request
)
```

But require when you define a context type (with either the global types or `defineMutatorWithType`).

### DIFF
--- a/packages/zql/src/mutate/mutator.ts
+++ b/packages/zql/src/mutate/mutator.ts
@@ -5,6 +5,7 @@ import type {
   DefaultContext,
   DefaultSchema,
   DefaultWrappedTransaction,
+  IsUnknown,
 } from '../../../zero-types/src/default-types.ts';
 import type {Schema} from '../../../zero-types/src/schema.ts';
 import type {AnyTransaction, Transaction} from './custom.ts';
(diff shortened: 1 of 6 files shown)


packages/zql/src/mutate/mutator.ts
```diff
@@ -234,7 +245,7 @@ export type Mutator<
    * during rebase (from stored JSON) and on the server (from wire format).
    * Validation happens internally before the recipe function runs.
    */
-  readonly 'fn': MutatorDefinitionFunction<
+  readonly 'fn': MutatorExecutionFunction<
```
arv, line 248: Is this a breaking change?

0xcadams: These types weren't a public API surface so it shouldn't be. And it's an expansion not contraction

arv: Cool.

I added the API snapshot test so we will know if something changed in the public API.

arv (commented):
LGTM

PR #5886, opened 2026-04-28
feat(zero): advertise Cloud Zero in zero-cache startup

aboodman:
Is this the right place to put this code @darkgnotic ?

image


darkgnotic:
I think it might have a better home in: https://github.com/rocicorp/mono/blob/main/packages/zero/src/zero-cache-dev.ts

so that it doesn't show up in server logs, etc. wdyt?

packages/zero/src/zero-cache-dev.ts
```diff
@@ -137,6 +138,10 @@ async function main() {
       // But let the developer override any of these dev defaults.
       ...process.env,
       ...zeroCacheEnv,
+
+      // Used by zero-cache to print the dev-only startup message after the
+      // server is listening.
+      [startupMessageEnv]: '1',
```
darkgnotic: Should this be set before the developer overrides so that the developer can set it to 0?

darkgnotic (approved):

packages/zero/src/zero-cache-dev.ts
aboodman: good call fixed

aboodman:
That's how it was originally. I moved it to this location because I wanted to print after all the startup goo has happened, to increase prob of being seen.

The implementation that I've uploaded there does ensure that it's only shown in development, by using an env var that is set by -dev.

PR #5887, opened 2026-04-28
Remove `select *` against Postgres where there can be pooled connections

aboodman:
`select *` + pgbouncer + transaction mode can cause the error "cached plan must not change result type". See: https://www.pgbouncer.org/config#max_prepared_statements.

This PR removes all such queries from the codebase (there are not that many) and replaces them w/ explicit column lists.


tools/oxlint-plugin-zero/index.js
```diff
@@ -0,0 +1,88 @@
+const selectStar = /\bselect\s+\*\s+from\b/i;
```
arv: I would think `select *, ... ` would also be problem.

tools/oxlint-plugin-zero/index.js
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        return {
+          Literal(node) {
+            if (typeof node.value === 'string') {
+              check(node, node.value);
+            }
+          },
+          TemplateLiteral(node) {
+            check(node, node.quasis.map(element => element.value.raw).join(''));
```
arv, line 54: or check for each

tools/oxlint-plugin-zero/index.js
```diff
@@ -0,0 +1,88 @@
...
+    },
+  },
+};
+
+function stripCommentOnlyLines(source) {
+  return source
+    .split('\n')
+    .map(line => {
+      const trimmed = line.trimStart();
+      return trimmed.startsWith('//') || trimmed.startsWith('--') ? '' : line;
```
arv, line 67: `//` is not a valid comment in SQL

tools/oxlint-plugin-zero/index.js
```diff
@@ -0,0 +1,88 @@
...
+    })
+    .join('\n');
+}
+
+function normalizePath(source) {
+  return source.replaceAll('\\', '/');
```
arv: I don't think you will ever get `\\` here.

arv (approved):
LGTM

packages/zero-cache/src/services/change-source/pg/schema/shard.ts
```diff
@@ -273,7 +273,16 @@ export async function getReplicaAtVersion(
...
-    SELECT * FROM ${schema}.replicas JOIN ${schema}."shardConfig" ON true
+    SELECT
+      replicas."slot",
+      replicas."version",
+      replicas."initialSchema",
+      replicas."initialSyncContext",
+      replicas."subscriberContext",
+      "shardConfig"."publications",
+      "shardConfig"."ddlDetection",
+      "shardConfig"."lock"
```
darkgnotic: While we're making things more optimal, can you omit the "lock"? It's a singleton-table implementation detail and not actually used in the app logic.

packages/zero-cache/src/services/change-streamer/storer.ts
```diff
@@ -258,7 +258,12 @@ export class Storer implements Service {
         ) SELECT COUNT(*) as deleted FROM purged;`;
 
       const [{owner}] = await sql<ReplicationState[]>`
-        SELECT * FROM ${this.#cdc('replicationState')} FOR SHARE`;
+        SELECT
+          "lastWatermark",
+          "owner",
+          "ownerAddress",
+          "lock"
```
darkgnotic: omit "lock"

packages/zero-cache/src/services/change-streamer/storer.ts
```diff
@@ -471,7 +476,12 @@ export class Storer implements Service {
           // a concurrent ownership change.
           void tx.pool.process(tx => {
             tx<ReplicationState[]> /*sql*/ `
-          SELECT * FROM ${this.#cdc('replicationState')} FOR UPDATE`.then(
+          SELECT
+            "lastWatermark",
+            "owner",
+            "ownerAddress",
+            "lock"
```
darkgnotic: omit "lock"

packages/zero-cache/src/services/change-streamer/storer.ts
```diff
@@ -569,7 +579,12 @@ export class Storer implements Service {
       // snapshot for the REPEATABLE_READ transaction.
       [{lastWatermark}] = await reader.processReadTask(
         sql => sql<ReplicationState[]>`
-        SELECT * FROM ${this.#cdc('replicationState')}
+        SELECT
+          "lastWatermark",
+          "owner",
+          "ownerAddress",
+          "lock"
```
darkgnotic: omit "lock"

darkgnotic (approved):
Thank you!

So it is just a coincidence that we started seeing these now?

aboodman:
> So it is just a coincidence that we started seeing these now?

It happens because we added a column to the cvr table.

tools/oxlint-plugin-zero/index.js
aboodman: It's catching cases where a SQL statement is in a JS comment.

tools/oxlint-plugin-zero/index.js
aboodman: Fixed in `31a84c8f2`: the rule now inspects the SELECT list instead of matching only `SELECT * FROM`, so it catches `SELECT *, ...`, `SELECT foo, * FROM`, and qualified stars like `SELECT foo.* FROM`. I also checked it does not flag `count(*)`.

tools/oxlint-plugin-zero/index.js
aboodman: Updated in `31a84c8f2`: the rule now walks each `SELECT` it finds in the literal/template text and inspects that select list for star tokens.

tools/oxlint-plugin-zero/index.js
aboodman: Fixed in `31a84c8f2`: removed the `//` handling. The rule is AST-based so it does not scan JS comments; it only strips SQL `--` comment-only lines inside SQL strings/templates.

tools/oxlint-plugin-zero/index.js
aboodman: Fixed in `31a84c8f2`: removed the path backslash normalization.

tools/oxlint-plugin-zero/index.js
aboodman: Updated in `242d19e24`: after follow-up, I kept this intentionally simple as a regex checking for a projection star between `SELECT` and `FROM`, with normal `oxlint-disable-next-line zero/no-select-star` for false positives.

tools/oxlint-plugin-zero/index.js
aboodman: Updated in `242d19e24`: simplified the rule back to a regex, but it still catches `SELECT *, ...` and `SELECT foo, * FROM ...`. I verified those cases with a temporary fixture before pushing.

tools/oxlint-plugin-zero/index.js
aboodman: Updated in `242d19e24`: kept `//` stripping intentionally because this scans string/template text and we also want to ignore SQL snippets inside JS comment-looking lines. It also still strips SQL `--` comment-only lines.

tools/oxlint-plugin-zero/index.js
aboodman: Fixed in `242d19e24`: removed the backslash normalization.

packages/zero-cache/src/services/change-source/pg/schema/shard.ts
aboodman: Fixed in `242d19e24`: removed `"lock"` from the metadata join selection.

packages/zero-cache/src/services/change-streamer/storer.ts
aboodman: Fixed in `242d19e24`: this now selects only `"owner"`, which is the field used by this check.

packages/zero-cache/src/services/change-streamer/storer.ts
aboodman: Fixed in `242d19e24`: this now selects only `"owner"` and uses a narrower row type.

packages/zero-cache/src/services/change-streamer/storer.ts
aboodman: Fixed in `242d19e24`: this catchup snapshot read now selects only `"lastWatermark"`.

aboodman:
@darkgnotic ptal carefully

PR #5891, opened 2026-04-28
chore: avoid swallowing trigger update errors if ddlDetection is on

darkgnotic:
The initial trigger event setup logic handles the Postgres "insufficient privilege" error code by proceeding without ddl detection, for Postgres providers that do not support event triggers.

This same error can happen when attempting to upgrade existing event triggers (on a Postgres provider that _does_ support event triggers) if, for example, the zero-cache is running as a different user.

In such cases, the error should be propagated instead of swallowed, as the application is essentially unable to perform the schema upgrade that it needs to correctly proceed.

Context:
* https://rocicorp.slack.com/archives/C0AK5KX8HGE/p1777052557306029


0xcadams (commented):
Great!!

PR #5892, opened 2026-04-28
chore: disable /keepalive timeouts by default when not in ECS

darkgnotic:
Makes the `/keepalive` timeout + graceful-shutdown logic an explicit option, defaulting to 20 seconds when run in ECS and disabled otherwise. (For non ECS environments, SIGINT/SIGTERM is already an appropriate drain signal).

### DIFF
--- a/packages/zero-cache/src/config/normalize.ts
+++ b/packages/zero-cache/src/config/normalize.ts
@@ -28,6 +28,13 @@ export function isDevelopmentMode(): boolean {
   return process.env.NODE_ENV === 'development';
 }
 
+function isRunningInECS(): boolean {
+  // https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-environment-variables.html
+  return process.env.ECS_CONTAINER_METADATA_URI_V4 !== undefined;
+}
+
+const DEFAULT_ECS_KEEPALIVE_TIMEOUT_MS = 20_000;
+
 export function assertNormalized(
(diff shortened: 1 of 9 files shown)


aboodman:
Are you concerned that other zero servers could have same problem we saw where they shut down too eagerly? I guess this makes sense.

aboodman:
Hm, I'm not so sure about this. At least on Fly.io this looks like a breaking change:

https://fly.io/docs/reference/configuration/#kill_signal-option

Fly.io does support the clean shutdown sequence but it looks like it must be explicitly configured. It seems like this change would cause zero-caches hosted on Fly to stop working (they would stop draining).

I think the current `/keepalive` behavior is a reasonable feature, it just needs to be documented.

aboodman:
If you agree I will update the docs.

darkgnotic:
> Hm, I'm not so sure about this. At least on Fly.io this looks like a breaking change:
> 
> https://fly.io/docs/reference/configuration/#kill_signal-option
> 
> Fly.io does support the clean shutdown sequence but it looks like it must be explicitly configured. It seems like this change would cause zero-caches hosted on Fly to stop working (they would stop draining).

The documentation states that the default behavior is to send a `SIGINT`:

> When shutting down a Fly Machine, Fly.io sends a SIGINT signal to the running process by default.

The configuration is to change the signal to something else, but we support both SIGINT and SIGTERM as graceful drain signals.

PR #5893, opened 2026-04-28
fix(zero-cache): fix init connection race

grgbkr:
Problem
======
In `ViewSyncerService.initConnection()`, the `ClientHandler` is added to `#clients` synchronously (line 802), before the async `#runInLockForClient` validates the connection's auth. If another lock-held operation (e.g., `#advancePipelines`, `#addAndRemoveQueries`) runs between the synchronous add and the async validation, it will see the unvalidated client in #clients and send pokes containing real user data before the connection is authorized.

Solution
=======
Move `#clients.set` from the synchronous part into the locked callback, after `#validateConnection` succeeds.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/view-syncer.ts
+++ b/packages/zero-cache/src/services/view-syncer/view-syncer.ts
@@ -799,16 +799,21 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
         connCtx.baseCookie,
         downstream,
       );
-      this.#clients
-        .get(connCtx.clientID)
-        ?.close(`replaced by wsID: ${connCtx.wsID}`);
-      this.#clients.set(connCtx.clientID, newClient);
+
 
       // Note: initConnection() must be synchronous so that `downstream` is
       // immediately returned to the caller (connection.ts). This ensures
       // that if the connection is subsequently closed, the `downstream`
       // subscription can be properly canceled even if #runInLockForClient()
       // has not had a chance to run.
+      //
+      // The client is NOT added to #clients here — it is deferred until
+      // after auth validation inside the lock below. This prevents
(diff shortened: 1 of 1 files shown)


0xcadams:
LGTM but it will still be a problem that someone can DoS another client by sending their client ID and force disconnecting them so I was thinking we could solve this at a higher level and start the connection out as `provisional` and then transition to `validated` and only then have them join the client group/send pokes to them/replace the existing client. I trust your judgement here, you would know more than me.

grgbkr:
> LGTM but it will still be a problem that someone can DoS another client by sending their client ID and force disconnecting them so I was thinking we could solve this at a higher level and start the connection out as `provisional` and then transition to `validated` and only then have them join the client group/send pokes to them/replace the existing client. I trust your judgement here, you would know more than me.

It would be great to fix the DoS vulnerability, and I can see how there might be fix that covers both that and this vulnerability. 

That fix just seems bigger, and I think this issue is high enough priority that its worth plugging with this small fix, even if this small fix is made unnecessary by a future fix to the DoS issue.

grgbkr:
> > LGTM but it will still be a problem that someone can DoS another client by sending their client ID and force disconnecting them so I was thinking we could solve this at a higher level and start the connection out as `provisional` and then transition to `validated` and only then have them join the client group/send pokes to them/replace the existing client. I trust your judgement here, you would know more than me.
> 
> It would be great to fix the DoS vulnerability, and I can see how there might be fix that covers both that and this vulnerability.
> 
> That fix just seems bigger, and I think this issue is high enough priority that its worth plugging with this small fix, even if this small fix is made unnecessary by a future fix to the DoS issue.

This didn't work in the end.  Too many assumptions in view-syncer about the client being added synchronously.  It is going to require a bigger refactor to fix.

PR #5894, opened 2026-04-28
perf(zql): fix table scan on `or(cmp, exists)` query patterns

tantaman:
`or(cmp(pk), exists)` would do a table scan rather than a primary key lookup for the first branch :( 

https://rocicorp.slack.com/archives/C013XFG80JC/p1777385619898799

CleanShot 2026-07-14 at 20 03 09@2x


test-gaps.md
arv, line 1: remove file

packages/zql/src/planner/planner-connection.ts
```diff
@@ -91,6 +92,17 @@ export class PlannerConnection {
...
+  /**
+   * Per-branch filters registered by `PlannerFilter` nodes (simple OR
+   * branches) when the enclosing FanIn is in UFI mode. Key is the same
+   * `branchPattern.join(',')` used by `#constraints`.
+   *
+   * These are AND-merged with `#filters` when calling the cost model so
+   * the model sees the true filter that the runtime would push to the
+   * source for this specific branch.
+   */
+  readonly #perBranchFilters: Map<string, NoSubqueryCondition> = new Map();
```
arv, line 104: Using `CustomKeyMap` might be slightly cleaner.

packages/zql/src/planner/planner-filter.ts
```diff
@@ -0,0 +1,192 @@
...
+      // simple branch's filter is the one feeding the outer FanOut, but
+      // from a nested FanIn we don't have a single input to walk through.
+      // Be conservative: don't register, accept that deeply nested OR
+      // shapes lose this optimization.
+      return undefined;
+    case 'filter':
+      // Shouldn't occur in graphs built by `processOr` — a PlannerFilter's
+      // input is always the FanOut it's a branch of, never another filter.
+      return undefined;
+  }
```
arv, line 191:
```
default:
  unreachable(node)
```

so we do not miss this when the kind changes.

arv (dismissed):
LGMT

Copilots comments makes sense as well.

tantaman:
resurrecting this. I didn't land before since it felt high risk and I wanted to add more tests. Margins hit this and we've improved the fuzzer drastically since this was originally written so I'm now confident enough to land it.

arv (commented):
LGTM with comments resolved

PR #5896, opened 2026-04-29
Fix: Exclude null clientAST values from statz query

KidlingYT:
## Summary

- `biggestAstHash` in `statz.ts` returns a result with `ast_length: null` instead of the largest AST whenever any `queries` row has a `NULL` `clientAST`. The cause is PostgreSQL's default NULL ordering: with `ORDER BY ... DESC`,
  NULLs sort before non-null values, so a single NULL returns from `LIMIT 1`.
- [Custom queries intentionally store `NULL` in `clientAST`](https://github.com/rocicorp/mono/blob/f52c6d2ef340bf9c9937523e2a77d2eab66217b0/packages/zero-cache/src/services/view-syncer/schema/cvr.ts#L124-L129) because custom queries don't persist an AST.
- **Fix**: Added WHERE "clientAST" IS NOT NULL to the biggestAstHash query.

## Why

The current largest AST statz endpoint is unhelpful.
image

## Validation 

- `npm test`

### DIFF
--- a/packages/zero-cache/src/services/statz.ts
+++ b/packages/zero-cache/src/services/statz.ts
@@ -166,6 +166,7 @@ async function cvrStats(lc: LogContext, config: ZeroConfig) {
         'biggestAstHash',
         sql`SELECT "queryHash", length("clientAST"::text) AS "ast_length"
       FROM ${sql(schema)}."queries"
+      WHERE "clientAST" IS NOT NULL
       ORDER BY length("clientAST"::text) DESC
       LIMIT 1;`,
       ],


aboodman:
Thank you for the contribution. Unfortunately I dont think this will do what you want - the purpose of largest AST output here is for debugging large queries. But when we moved to custom queries this got broken. We no longer store the query AST in the database at all. So if you filter out NULLs you'll not get useful data either.

aboodman:
If we still want this feature it needs to be implemented differently.

KidlingYT:
> Thank you for the contribution. Unfortunately I dont think this will do what you want - the purpose of largest AST output here is for debugging large queries. But when we moved to custom queries this got broken. We no longer store the query AST in the database at all. So if you filter out NULLs you'll not get useful data either.

Ah, okay. Thanks for the feedback! I hadn't realized it wasn't stored anywhere anymore, we just moved off of the legacy system.

PR #5900, opened 2026-04-30
chore(zero-cache): stringify change streamer payloads before subscriber fanout

darkgnotic:
Reduce the amount of JSON stringification performed in the replication-manager by:
* stringifying each change payload once before fanning out to subscribers (i.e. view-syncers and backup replicator)
* avoiding parsing and re-stringifying JSON payloads when catching up subcribers from the change DB

A local (MacBook) load test was run with replication-managers connected to 20 view-syncers. With the caveat that the MacBook has to time share its 16 cores across 42+ busy processes, the data shows a meaningful improvement in performance.

replication-manager CPU (with 20 connected view-syncers)

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/storer.ts
+++ b/packages/zero-cache/src/services/change-streamer/storer.ts
@@ -25,15 +25,17 @@ import {
   type SchemaChange,
   type TableMetadata,
 } from '../change-source/protocol/current.ts';
-import {type Commit} from '../change-source/protocol/current/downstream.ts';
+import {
+  type ChangeStreamData,
+  type Commit,
+} from '../change-source/protocol/current/downstream.ts';
 import type {
   DownstreamStatusMessage,
   UpstreamStatusMessage,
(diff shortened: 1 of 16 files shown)


packages/zero-cache/src/services/change-streamer/storer.ts
```diff
@@ -489,7 +487,9 @@ export class Storer implements Service {
           watermark: tag === 'commit' ? watermark : tx.preCommitWatermark,
           precommit: tag === 'commit' ? tx.preCommitWatermark : null,
           pos: tx.pos,
-          change: json,
+          // For backwards compatibility, only the change message is stored
+          // in the cdc changeLog.
+          change: extractChangeSubstring(json, tag),
```
darkgnotic, line 492: FWIW, I know this is not great, but the longer term plan (RM v2) is to ditch PG and store the change-log in a local SQLite replica.

At that point, we can store the full change JSON as is and not just a part of it.

arv (commented):
LGTM

PR #5901, opened 2026-04-30
chore(zero-cache): improve error handling for rep slot creation timeout

grgbkr:
Increase timeout to 30 seconds, slot creation does often take sometime as it waits for all open transactions to complete.

Also only log errors for timeouts during slot creation for backfill once the maximum retry backoff has been reached: https://github.com/rocicorp/mono/blob/604f104b41dad4b91533d0c1af30e82822f6e23e/packages/zero-cache/src/services/change-source/common/backfill-manager.ts#L191

Timeout errors during initial sync immediately log error and crash the server.

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/initial-sync.ts
+++ b/packages/zero-cache/src/services/change-source/pg/initial-sync.ts
@@ -529,8 +529,14 @@ type ReplicationSlot = {
   output_plugin: string;
 };
 
-// Successful CREATE_REPLICATION_SLOT calls have always completed within 1s in
-// observed production runs; use 30s as a hard stop for pathological hangs.
+// When creating a replication slot, Postgres waits for open transactions
+// to complete before reserving a consistent_point (LSN) in the WAL and creating
+// a matching transaction snapshot. As such, it can technically take an arbitrary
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/services/change-source/pg/initial-sync.ts
```diff
@@ -530,8 +530,8 @@ type ReplicationSlot = {
 };
 
 // Successful CREATE_REPLICATION_SLOT calls have always completed within 1s in
```
darkgnotic: Suggested update to this comment:

```ts
// When creating a replication slot, Postgres waits for open transactions
// to complete before reserving a consistent_point (LSN) in the WAL and creating
// a matching transaction snapshot. As such, it can technically take an arbitrary
// amount of time (e.g. DDL operations, table-wide operations, etc.).
//
// However, to detect pathological situations, bound the amount of time that
// the server waits for replication slot creation, so that a continual failure to
// create a replication slot is surfaced by errors / alerts.
```

darkgnotic (approved):
Thanks @grgbkr !

darkgnotic:
@grgbkr I'm going to merge this because I'm working in this area of the code (for the slot management stuff).

PR #5903, opened 2026-05-01
fix(zero-client): rotate zero instance w/o page reload

0xcadams:
We currently do a full page refresh when a client comes back to an app after:

- 24h local client GC: stale tab cleaned up by another active tab
- 48h server purge: inactive CVR state cleanup

It causes flashing and odd issues that are unexpected for developers. It comes up with both zbugs and cloudzero - the full page refresh ends up causing more churn than is necessary. It's why coming back to CZ after a couple days is super slow and flashes.

**Description**
- teach the React and Solid `ZeroProvider`s to recreate provider-owned `Zero` instances when `onClientStateNotFound` is called
- expand `onClientStateNotFound` to handle `InvalidConnectionRequestLastMutationID` and `InvalidConnectionRequestBaseCookie`

**Details**
- `zero-client` still owns dropping the Replicache DB before recovery continues
- provider-owned React and Solid providers now rotate on `onClientStateNotFound`
- repeated stale-client callbacks are deduped while a rotation is already pending
- `init` is re-run for the replacement instance
- external `zero={...}` usage is unchanged and still requires the caller to replace the instance
- `onUpdateNeeded` behavior is unchanged and still causes page refresh

**Original Issue**

The timeline below shows that we double the load time due to this. It could have been loaded at 4s but ends up settled at 8s with a major refresh flash.

| Time              | What happened                    | Request / signal                    | Result                                  |
|-------------------|----------------------------------|-------------------------------------|-----------------------------------------|
| T+0.000s          | Initial navigation starts         | GET /                               | Browser begins loading CZ               |
| T+2.725s          | Server redirects into app route   | 307 → /z                            | Browser redirects              |
| T+2.918s          | First HTML document is received   | GET /z → 200                        | First app document is available         |
| T+2.934s          | Initial assets begin loading      | CSS, JS, route chunks, fonts        | Client app starts booting               |
| T+3.046s          | First page reaches DOM ready      | DOMContentLoaded                    | First page lifecycle is active          |
| T+3.153s          | App fetches session               | GET /api/auth/get-session → 200     | Session state is loaded                 |
| T+3.430s          | App loads route-specific code     | Team/stacks route chunks            | Target route code is prepared           |
| T+3.847s          | App opens Zero websocket          | wss://sync.../connect → 101         | Zero connection starts                  |
| T+4.080s          | Zero returns stale-client error   | ClientNotFound                      | Local Zero client state is invalid      |
| T+4.084s          | Client triggers hard reload       | GET /z/*redacted*              | Second full document navigation starts  |
| T+4.084s–T+6.367s | Browser waits for replacement HTML| ~2.28s server wait / TTFB           | Likely flash window                     |
| T+6.367s          | Second HTML document arrives      | GET /z/*redacted* → 200        | Reloaded page can start booting         |
| T+6.389s          | Assets load again                 | CSS, JS, route chunks, fonts        | Second app bootstrap starts             |
| T+6.897s          | Second page reaches DOM ready     | DOMContentLoaded                    | Reloaded document is active             |
| T+6.955s          | App opens second Zero websocket   | second websocket connection         | New Zero connection succeeds            |
| T+7.706s+         | Later server/data calls run       | server function POSTs               | Affects data state, not initial flash   |

**Comparisons:**

Here's a comparison before/after with zbugs network slowed to 2 mbps.

**Before:**

https://github.com/user-attachments/assets/6c6f5dcb-56c2-44d5-be5b-bfadadc5c77b

**After:**

https://github.com/user-attachments/assets/7aa1a7cf-952b-42f7-a294-b98d6e41cdd5

### DIFF
--- a/packages/zero-client/src/client/options.ts
+++ b/packages/zero-client/src/client/options.ts
@@ -235,16 +235,24 @@ export type ZeroOptions<
 
   /**
    * `onClientStateNotFound` is called when this client is no longer able
-   * to sync with the zero-cache due to missing synchronization state.  This
+   * to sync with zero-cache due to missing synchronization state. This
    * can be because:
    * - the local persistent synchronization state has been garbage collected.
    *   This can happen if the client has no pending mutations and has not been
    *   used for a while (e.g. the client's tab has been hidden for a long time).
-   * - the zero-cache fails to find the server side synchronization state for
+   * - zero-cache fails to find the server side synchronization state for
    *   this client.
+   * - zero-cache rejects this client's persisted synchronization state,
+   *   requiring the local Replicache database to be reset.
    *
-   * The default behavior is to reload the page (using `location.reload()`).
-   * Provide your own function to prevent the page from reloading automatically.
+   * The current `Zero` instance should be treated as dead and replaced, not
+   * reconnected.
+   *
+   * The default behavior for the Zero client is to reload the page
+   * (using `location.reload()`). The React and SolidJS providers will replace
+   * the client in-place, without a full page reload.
+   *
+   * Provide your own function to prevent this functionality.
    */
   onClientStateNotFound?: (() => void) | undefined;
 
--- a/packages/zero-client/src/client/zero.test.ts
+++ b/packages/zero-client/src/client/zero.test.ts
@@ -15,6 +15,7 @@ import {
   setDeletedClients,
   type DeletedClients,
 } from '../../../replicache/src/deleted-clients.ts';
+import {hasMemStore} from '../../../replicache/src/kv/mem-store.ts';
 import {makeClientV6} from '../../../replicache/src/persist/clients-test-helpers.ts';
 import {
   getClients,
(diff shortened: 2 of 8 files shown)


aboodman:
Do you have a video of the UX impact? seems like there would still be some flashing.

aboodman:
I like the idea of this, but I would like to see the improvement and whether there is still flashing.

packages/zero-client/src/client/options.ts
```diff
@@ -248,6 +248,22 @@ export type ZeroOptions<
...
+   * Replicache database must be reset.
+   *
+   * Zero deletes the current Replicache database before invoking this
+   * callback. The current `Zero` instance should be treated as dead and
+   * replaced, not reconnected.
+   *
+   * The default behavior is to reload the page (using `location.reload()`).
+   * Provide your own function to prevent the page from reloading automatically.
+   */
+  onClientStateResetNeeded?:
```
aboodman: How is this different from the existing `onClientStateNotFound` ?

aboodman: Nevermind, I see it. You're trying to expose the tagged union.

A few thoughts here:

1. Adding the tagged union as a new param to the existing `onClientResetNeeded` is *not* a breaking change. Since it's a callback, the semantics are opposite (existing callback that takes no params can receive new params fine). So you could do that and avoid the new callback.
2. I don't love the idea of exposing the `reason` to dev in the first place. Just feels like internal detail that is not necessary and is more API we have to think about and design.
3. I see that `reason` is being used by your ZeroProvider impl, but I think this could be fixed a different way: I don't think that even `reloadWithReason()` really *needs* `reason` either! All it's used for is to print to the console.

All together, I think it would be better to extend the existing `onClientStateNotFound` like this:

```ts
onClientStateNotFound?: ((opts: {reason: string}) => void) | undefined;
```

Then change `reloadWithReason` to just not use this `reason` enum at all.

0xcadams:
> Do you have a video of the UX impact? seems like there would still be some flashing.

I added a video to the description. With any kind of slower network it's very noticeable. There will be a flash, but no complete refresh of the page which clears any request caches in tanstack query and refetches all JS/CSS which results in a complete roundtrip and white page for >30s.

packages/zero-solid/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "devDependencies": {
     "@solidjs/testing-library": "^0.8.10",
+    "jsdom": "^29.1.1",
```
arv: remove

we use playwright

arv (commented):
I like where this is going.

packages/zero-solid/package.json
0xcadams: I'll look into removing. For whatever reason this PR was blocked on adding jsdom for the tests... Probably my mistake

0xcadams: Screenshot 2026-05-05 at 8 38 07 AM

🤔 vitest seems to want to install this for solid

arv: odd.

🤷🏼

0xcadams: Fixed w/o adding jsdom

packages/zero-client/src/client/options.ts
0xcadams: OK addressed this and simplified the API to stay the same but expand cases that `onClientStateNotFound ` is called.

PR #5904, opened 2026-05-01
fix(zero-cache): gate shared retransform on view-syncer readiness

0xcadams:
Gate shared background retransform on the current `ViewSyncerService` instance reaching first successful pipeline sync.

**Details**
- added explicit lifecycle readiness for shared retransform scheduling in `ConnectionContextManager`
- made `retransformAt` require both:
  - a validated background connection
  - readiness from the current `ViewSyncerService` instance
- set readiness only after the initial `#syncQueryPipelineSet()` succeeds
- clear readiness on pipeline reset, `stop()`, and cleanup

**Original Issue**

`Error: pipelines must be initialized (syncQueryPipelineSet)`

This happened during `SIGTERM` drain of `view-syncer`.

The underlying issue was that a later `ViewSyncerService` instance for the same logical client group could validate a background connection and become eligible for shared auth-maintenance retransform work before that specific service instance had completed its first successful pipeline sync.


grgbkr:
LGTM.

PR #5905, opened 2026-05-05
feat(zero-server): rework handle query/mutate to be object-based

0xcadams:
Reworks the zero-server `handleMutateRequest` / `handleQueryRequest` APIs around a single object argument so server integrations can pass request context explicitly and wire query/mutate handlers the same way.

**Details**
- add object-form request APIs for both helpers with explicit `handler`, `request` or `query` + `body`, `userID`, and `logLevel`
- keep the positional call patterns as deprecated shims so existing callers still work and legacy responses still omit `userID`
- align the query helper shape with the mutate helper so endpoint adapters can be written symmetrically
- update zbugs to use the new object-based calls for both `/mutate` and `/query`
- expand tests to cover object-form request/body inputs, `userID` normalization to `null`, and backwards-compat behavior

**Examples**:

Using a `Request` with `handleMutateRequest`:

```ts
const response = await handleMutateRequest({
  dbProvider,
  handler: (transact, _mutation) =>
    transact((tx, name, args) => {
      const mutator = mustGetMutator(mutators, name);
      return mutator.fn({tx, args, ctx: authData});
    }),
  request,
  userID: authData?.sub,
  logLevel: 'info',
});
```

Using a `Request` with `handleQueryRequest`:

```ts
const response = await handleQueryRequest({
  handler: (name, args) => {
    const query = mustGetQuery(queries, name);
    return query.fn({args, ctx: authData});
  },
  schema,
  request,
  userID: authData?.sub,
  logLevel: 'info',
});
```

Using parsed `query` + `body` with `handleMutateRequest`:

```ts
const response = await handleMutateRequest({
  dbProvider,
  handler: (transact, _mutation) =>
    transact((tx, name, args) => {
      const mutator = mustGetMutator(mutators, name);
      return mutator.fn({tx, args, ctx: authData});
    }),
  query: request.query,
  body: request.body,
  userID: authData?.sub,
  logLevel: 'info',
});
```

Using parsed `query` + `body` with `handleQueryRequest`:

```ts
const response = await handleQueryRequest({
  handler: (name, args) => {
    const query = mustGetQuery(queries, name);
    return query.fn({args, ctx: authData});
  },
  schema,
  query: request.query,
  body: request.body,
  userID: authData?.sub,
  logLevel: 'info',
});
```

### DIFF
--- a/apps/zbugs/api/index.ts
+++ b/apps/zbugs/api/index.ts
@@ -216,8 +216,10 @@ async function queryHandler(
           return query.fn({args, ctx: authData});
         },
         schema,
+        query: request.query,
         body: request.body,
         userID: authData?.sub,
+        logLevel: 'info',
       }),
     );
   });
--- a/apps/zbugs/api/index.ts
+++ b/apps/zbugs/api/index.ts
@@ -168,20 +168,18 @@ async function mutateHandler(
     const postCommitTasks: (() => Promise<void>)[] = [];
     const mutators = createServerMutators(postCommitTasks);
 
-    const response = await handleMutateRequest(
+    const response = await handleMutateRequest({
       dbProvider,
-      (transact, _mutation) =>
+      handler: (transact, _mutation) =>
         transact((tx, name, args) => {
           const mutator = mustGetMutator(mutators, name);
           return mutator.fn({tx, args, ctx: authData});
         }),
-      request.query,
-      request.body,
-      {
-        userID: authData?.sub,
-        logLevel: 'info',
-      },
-    );
+      query: request.query,
+      body: request.body,
+      userID: authData?.sub,
+      logLevel: 'info',
+    });
 
     // we don't yet handle errors here, since Loops emails return 429 very often
     // and we don't want to block the mutation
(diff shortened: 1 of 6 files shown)


packages/zero-server/src/queries/process-queries.ts
```diff
@@ -67,107 +64,203 @@ export function handleTransformRequest<S extends Schema>(
...
+function normalizeQueryRequestInput<S extends Schema>(
+  input: HandleQueryRequestObjectArgs<S>,
+): NormalizedQueryRequestArgs<S> {
+  return 'request' in input
+    ? {
+        type: 'request',
+        handler: input.handler,
+        schema: input.schema,
+        request: input.request,
+        userID: input.userID ?? null,
```
0xcadams, line 201: We allow `undefined` on the new API but collapse to `null` here

apps/zbugs/api/index.ts
```diff
@@ -212,17 +210,17 @@ async function queryHandler(
...
           const query = mustGetQuery(queries, name);
           return query.fn({args, ctx: authData});
         },
         schema,
-        request.body,
-        {
-          userID: authData?.sub,
-        },
-      ),
+        query: request.query,
```
aboodman, line 219: would just passing request here work too? Or is it not right kind of request.

packages/zero-server/src/queries/process-queries.ts
aboodman: Convenient

apps/zbugs/api/index.ts
0xcadams: It's not the right kind of request.

0xcadams: It's tied to Node and not the Fetch standard, and it doesn't look Fastify intends to update that: https://fastify.dev/docs/latest/Reference/Request/

PR #5907, opened 2026-05-05
fix(zero-cache): fix zombie ViewSyncers inflating active-client-groups metric

grgbkr:
When a client disconnects before `initConnection`'s async callback resolves `#initialized`, the ViewSyncer's `run()` stays blocked on `readyState()` forever. The idle-shutdown path (`#checkForShutdownConditionsInLock`) cancels `#stateChanges` but never rejects `#initialized`, so `run()` never exits and the instance accumulates permanently in the ServiceRunner map — only cleaned up on pod drain.

This causes the `active-client-groups` gauge to grow unboundedly while `active-sync-clients` shows the true count.

**Fix:** Reject `#initialized` in the idle-shutdown path (mirroring what `stop()` already does). The resolver's `reject` is idempotent — it's a no-op if already resolved.

### DIFF
--- a/packages/zero-cache/src/services/runner.test.ts
+++ b/packages/zero-cache/src/services/runner.test.ts
@@ -65,4 +65,46 @@ describe('services/runner', () => {
     const s2 = runner.getService('foo');
     expect(s1).not.toBe(s2);
   });
+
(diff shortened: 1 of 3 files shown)


darkgnotic (approved):
Nice find!

PR #5910, opened 2026-05-06
[codex] Preserve original IVM advance errors

aboodman:
## Summary
- poison an IVMSourceBranch after advance() throws so later branch access rethrows the original error
- clear detached source references after the first failed advance to avoid continuing through known-corrupt branch state
- add a regression test that forces a partial advance failure and verifies advance(), getSource(), and forkToHead() preserve the same error

## Why
We periodically get error reports from the field of "Row already exists" from the IVM system.

I saw one of these today in zbugs and it occurred after some other error put the client into error mode. I don't know what error the original one was, but it's clear that errors processing pokes can result in "Row not found" errors after because of the non-transactionality of IVMBranch: we can partially apply a poke, fail, then attempt again when refreshing from the perdag and get the same error.

Preserving the first error keeps the Zero instance in a stable failure mode and leaves the original crash site visible.

### DIFF
--- a/packages/zero-client/src/client/ivm-branch.ts
+++ b/packages/zero-client/src/client/ivm-branch.ts
@@ -46,6 +46,7 @@ import {
 export class IVMSourceBranch {
   readonly #sources: Map<string, MemorySource | undefined>;
   readonly #tables: Record<string, TableSchema>;
+  #advanceError: unknown;
   hash: Hash | undefined;
 
   constructor(
@@ -59,6 +60,8 @@ export class IVMSourceBranch {
   }
 
   getSource(name: string): MemorySource | undefined {
+    this.#throwIfInvalid();
+
     if (this.#sources.has(name)) {
       return this.#sources.get(name);
     }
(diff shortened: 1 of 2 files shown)


tantaman:
> we can partially apply a poke, fail, then attempt again when refreshing from the perdag and get the same error.

~~We don't revert the fork on failure and re-fork when trying again?~~

Nevermind, I see this is the case of updating the main branch..

packages/zero-client/src/client/ivm-branch.ts
```diff
@@ -75,19 +78,33 @@ export class IVMSourceBranch {
...
     );
 
-    applyDiffs(diffs, this);
-    this.hash = newHead;
+    try {
+      applyDiffs(diffs, this);
+      this.hash = newHead;
+    } catch (e) {
+      this.#advanceError = e;
+      this.clear();
```
tantaman, line 105: Why clear?

packages/zero-client/src/client/ivm-branch.ts
tantaman, line 139: do you want to `throwIfInvalid` here too?

packages/zero-client/src/client/ivm-branch.ts
aboodman: Well it's kind of a tough call. The branch is in a weird state at this point. It has received half of an update, then throws. Then `ZeroContext` ends the transaction because of `finally`:

```ts
processChanges(
    expectedHead: Hash | undefined,
    newHead: Hash,
    changes: NoIndexDiff,
  ) {
    this.batchViewUpdates(() => {
      try {
        this.#mainSources.advance(expectedHead, newHead, changes);
      } finally {
        this.#endTransaction();
      }
    });
  }
  ```

So we're displaying half a transaction, we're in a weird inconsistent state. Also the `hash` in `IVMSource` is still the old hash since that is set at the end of `applyDiffs`.

We have no way in IVM to rollback a change so the option of undoing the work we did is not available. We need some way to put the system back into a consistent state. This is the one I chose. WDYT?

packages/zero-client/src/client/ivm-branch.ts
aboodman: good catch - https://github.com/rocicorp/mono/pull/5916

PR #5912, opened 2026-05-06
perf(replicache): use executeRawSync in op-sqlite prepared statements

arv:
## Summary

Switch `OpSQLitePreparedStatement` from the async bridge (`executeRaw`) to the synchronous API (`executeRawSync`) for both reads and writes.

### Problem

Every `get`, `has`, `put`, and `del` in the op-sqlite KV store previously called `executeRaw`, which crosses the JS-to-native async bridge. For serial access patterns this means each key lookup incurs a full async round-trip before the next one can start.

### Change

`firstValue` and `exec` now call `executeRawSync` and return an immediately-resolved `Promise`. The public async interface is unchanged; callers can still `await` as before.

```ts
// Before
async firstValue(params: string[]): Promise<string | undefined> {
  const rows = await this.#db.executeRaw(this.#sql, params);
  return rows[0]?.[0];
}

// After
firstValue(params: string[]): Promise<string | undefined> {
  const rows = this.#db.executeRawSync(this.#sql, params);
  return Promise.resolve(rows[0]?.[0]);
}
```

### Benchmark results (iPhone, 1 000 keys, 10 iterations)

| Pattern | Before | After | Speedup |
|---|---|---|---|
| Serial reads | ~66 ms | ~25 ms | ~2.7× |
| Concurrent reads | ~21 ms | ~10 ms | ~2× |

Benchmarks were measured with the [`expo-replicache`](https://github.com/rocicorp/expo-replicache) benchmark app comparing per-key `executeRaw` vs `executeRawSync`.

### DIFF
--- a/packages/replicache/src/kv/op-sqlite/store.ts
+++ b/packages/replicache/src/kv/op-sqlite/store.ts
@@ -49,13 +49,14 @@ class OpSQLitePreparedStatement implements PreparedStatement {
     this.#sql = sql;
   }
 
-  async firstValue(params: string[]): Promise<string | undefined> {
-    const rows = await this.#db.executeRaw(this.#sql, params);
-    return rows[0]?.[0];
+  firstValue(params: string[]): Promise<string | undefined> {
+    const rows = this.#db.executeRawSync(this.#sql, params);
+    return Promise.resolve(rows[0]?.[0]);
   }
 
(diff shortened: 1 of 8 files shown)


arv:
This PR is a mess. I'll sort it out later

PR #5914, opened 2026-05-06
feat(zql): batch flip-join requests

tantaman:
Uses `WHERE IN (...)` so flip-join fetches BATCH_SIZE keys at once.

This gives Goblin's "school administrator viewing students" query a 6x speedup.

This is also _faster_ than the quicksort variant by 2x 

Finally, we use a min-heap for merging streams so we do not have to visit all streams on every iteration. e.g., `O(logk)` instead of `O(k)` which has a big payoff when there are > 100 streams.

TODO:
1. Split out min-heap to own pr?
2. Put quicksort removal in own pr?
3. Put batching in own pr?


tantaman:
splitting this into smaller PRs

PR #5915, opened 2026-05-06
perf(replicache): Batch writes in SQLiteStore

arv:
This pull request introduces a batching mechanism for SQLite write operations in the `SQLiteWrite` class, optimizing how deletes and upserts are performed.

**SQLite write batching and implementation changes:**

* The `SQLiteWrite` class now extends `WriteImplBase` and batches deletes and upserts, executing them as single statements per operation type in `commit`. This improves efficiency by reducing the number of SQLite statements executed. 
* The prepared statements for `put` and `del` in `setupDatabase` have been updated to accept arrays of values, allowing for batched operations. 

**Testing improvements:**

* Added a new test to verify that `SQLiteWrite` batches deletes and upserts into single statements, ensuring the new batching logic works as expected. 

---

**KV Benchmark Summary (1000 puts+commit):**

The branch version dramatically outperforms released packages:

| Implementation | Avg | Speedup |
|---|---|---|
| **op-sqlite-branch** | 3.96ms | **7.5x faster** than main |
| **expo-sqlite-branch** | 6.30ms | **9x faster** than main |
| op-sqlite-main | 29.87ms | baseline |
| expo-sqlite-main | 56.52ms | slowest |

**Key findings:**
- Branch (current work) shows **massive gains** — likely from the batch commit optimization
- op-sqlite significantly outperforms expo-sqlite across all versions
- Branch versions are more stable, except expo-sqlite-branch has one outlier (23ms spike)
- Main branch (27ms) falls between main and branch, suggesting incremental prior improvements

This validates the batch commit refactoring: ~7-9x speedup on KV operations.


packages/replicache/src/kv/sqlite-store.ts
```diff
@@ -232,48 +237,53 @@ export class SQLiteWrite implements Write {
...
+        deleteKeys.push(entry[0]);
+        this._pending.delete(entry[0]);
+      }
+    }
 
-  async del(key: string): Promise<void> {
-    throwIfTransactionClosed(this);
-    await this.#preparedStatements.del.exec([key]);
-  }
+    const delP =
```
0xcadams, line 259: If this throws and it's not awaited until later does that cause problems? Since upP will unconditionally run

arv: Yeah. Let me add a catch to it

arv: Not a problem. I was thinking of something else.

PR #5917, opened 2026-05-06
feat(zql): add mergeSortedStreams k-way merge utility

tantaman:
On the road to faster flip-join. In support of Goblins to speed up their "district admin viewing students" query (or, in the general case, any query that uses flipped-join).

The existing function that we use to merge streams does O(K) work for each row it emits. As in, each row compares against all other streams.

(K = num streams)

This adds a min-heap based merge so we do O(log K) work per row.

Very useful when a flip-join ends up having hundreds to thousands of unique children as each child is 1 stream.

Standalone for now; used in follow-up work.

Microbench

```
  ┌─────┬───────────────────────┐
  │  K  │ microbench            │
  ├─────┼───────────────────────|
  │  10 │                   tie │
  ├─────┼───────────────────────|
  │  40 │            27% faster │
  ├─────┼───────────────────────|
  │ 100 │            56% faster │
  ├─────┼───────────────────────|
  │ 200 │            70% faster │
  ├─────┼───────────────────────|
  │ 400 │            82% faster │
  └─────┴───────────────────────┘     
```

### DIFF
--- a/packages/zql/src/ivm/memory-source.ts
+++ b/packages/zql/src/ivm/memory-source.ts
@@ -917,3 +917,129 @@ export function stringify(change: SourceChange) {
     typeof v === 'bigint' ? v.toString() : v,
   );
 }
+
+/**
+ * N-way merge of pre-sorted Node streams. Each input stream must yield
+ * Nodes in `compare` order (and may interleave 'yield' values, which are
+ * forwarded as-is). The merged stream yields Nodes in `compare` order.
+ *
+ * Implemented as a binary min-heap keyed by `compare(entry.row, ...)`, so
+ * each emit costs O(log K) (K = number of streams) rather than the O(K)
+ * a linear-scan-of-heads merge would. Matters when FlippedJoin chunks a
+ * large `multiConstraints` IN-list into hundreds of sub-fetches.
+ *
+ * If the merged stream is closed early (e.g. downstream `Take` `break`s
+ * after hitting its limit, prompting JS to call `.return()` on this
+ * generator), the `finally` block propagates `.return()` to each
+ * non-exhausted sub-iterator so the underlying sources (SQLite cursors,
+ * etc.) can run their cleanup. Without this, mid-merge early-termination
+ * leaves cursors open, causing later writes on the same connection to
+ * fail with "database connection is busy executing a query".
+ */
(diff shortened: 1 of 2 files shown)


packages/zql/src/ivm/memory-source.ts
```diff
@@ -917,3 +917,129 @@ export function stringify(change: SourceChange) {
...
+  streams: readonly Stream<Node | 'yield'>[],
+  compare: (a: Node, b: Node) => number,
+): Stream<Node | 'yield'> {
+  const iterators: Iterator<Node | 'yield'>[] = streams.map(s =>
+    s[Symbol.iterator](),
+  );
+  // True while iterators[i] hasn't yet returned `done`. The finally
+  // block uses this to skip already-exhausted streams when propagating
+  // `.return()`.
+  const active: boolean[] = new Array(iterators.length).fill(true);
```
arv, line 949: I though oxlint would flag this pattern?

```suggestion
  const active: boolean[] = Array.from({length: iterators.length}).fill(true);
```

arv (commented):
LGTM

Did you also run this with trace-opts? Would be nice to minimize deopts in this kind of code.

tantaman:
> Did you also run this with trace-opts? Would be nice to minimize deopts in this kind of code.

I did not. Checking this morning.

tantaman:
siftUp, siftDown, and compare never de-opt

pullNext only de-opts 4 times (while being invoked hundreds of thousands of times). I think it de-opts between different benchmark runs.

packages/zql/src/ivm/memory-source.ts
```diff
@@ -917,3 +917,129 @@ export function stringify(change: SourceChange) {
...
+          siftDown(0);
+        }
+      }
+    }
+  } finally {
+    // Close any iterators that aren't already exhausted so their
+    // `finally` blocks (which release cursors / cached statements) run.
+    for (let i = 0; i < iterators.length; i++) {
+      if (active[i]) {
+        iterators[i].return?.();
```
grgbkr, line 1041:
```suggestion
        try {
          iterators[i].return?.();
        } catch (_cleanupError) {
          // catch so other iterators are cleaned up
        }
```

grgbkr (commented):
LGTM

PR #5919, opened 2026-05-06
fix(zero-cache): limit websocket compression to sync clients

aboodman:
## Summary

This keeps `ZERO_WEBSOCKET_COMPRESSION` and `ZERO_WEBSOCKET_COMPRESSION_OPTIONS` scoped to client-facing sync websockets.

Replication/change-streamer websockets no longer negotiate per-message deflate, avoiding the CPU tradeoff on replication-manager -> view-syncer traffic. The replication websocket server still honors `ZERO_WEBSOCKET_MAX_PAYLOAD_BYTES` as an inbound message safety limit.

## Validation

- `npm --workspace=zero-cache run check-format`

Note: PG tests were not run locally because the earlier attempt was blocked by no available container runtime strategy.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/change-streamer-http.ts
+++ b/packages/zero-cache/src/services/change-streamer/change-streamer-http.ts
@@ -62,25 +62,10 @@ export class ChangeStreamerHttpServer extends HttpService {
     backupMonitor: BackupMonitor | null,
   ) {
     super('change-streamer-http-server', lc, opts, async fastify => {
-      const websocketOptions: {perMessageDeflate?: boolean | object} = {};
-      if (config.websocketCompression) {
-        if (config.websocketCompressionOptions) {
-          try {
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/services/change-streamer/change-streamer-http.ts
```diff
@@ -62,25 +62,10 @@ export class ChangeStreamerHttpServer extends HttpService {
...
-          }
-        } else {
-          websocketOptions.perMessageDeflate = true;
-        }
-      }
-
       await fastify.register(websocket, {
-        options: websocketOptions,
+        options: {
+          maxPayload: config.websocketMaxPayloadBytes,
```
grgbkr, line 67: This wasn't being applied to internal rm -> vs traffic before, and I don't think it should be.  This is about protecting against bad clients, not protecting against ourself.

grgbkr:
LGTM

PR #5920, opened 2026-05-06
fix(zero-cache): remove replication websocket max payload

aboodman:
## Summary

Follow-up to #5919: keep the compression split, but do not apply websocket config to replication/change-streamer websockets.

The websocket compression and max-payload options are for untrusted client-facing websocket traffic. Replication-manager -> view-syncer traffic is internal, so the change-streamer websocket registration now goes back to the default `@fastify/websocket` options and `ChangeStreamerHttpServer` no longer accepts the unused config parameter.

## Validation

- `npm --workspace=zero-cache run check-format`


grgbkr:
LGTM

PR #5921, opened 2026-05-07
feat(zql): heap-based merge in #fetchMergeSort

tantaman:
Swap the linear-scan K-way merge for `mergeSortedStreams` (O(log K) per
row vs O(K) per row). K = number of open cursors.

Adds a PERF=1-gated zqlite perf test that exercises the merge-sort
path against a real TableSource so the heap-vs-linear and dedup wins
can be compared against an earlier revision.


packages/zqlite/src/flipped-join-merge.perf.test.ts
arv, line 1: why not a bench?

grgbkr:
LGTM

PR #5923, opened 2026-05-07
fix(replicache): fix putMany rebalancing duplicating entries across adjacent children

grgbkr:
## Problem

`putMany`'s rebalancing logic in `InternalNodeImpl` produced duplicate keys across sibling leaf nodes, causing the BTree `diff` algorithm to report incorrect `add` operations for rows that already existed. This triggered `"Row already exists"` assertion failures in the IVM `MemorySource` during poke processing.

## Root Cause

When two adjacent children (e.g., indices 0 and 1) both needed rebalancing after `putMany`, the old code processed them independently right-to-left:

1. **Child[1]** merged with sibling at `newEntries[0]` (the original child[0] data), producing rebalanced nodes containing child[0]'s entries.
2. **Child[0]** then merged with its next sibling at `newEntries[1]`, which was now the rebalanced result from step 1 — already containing child[0]'s data.

This duplicated child[0]'s entries across multiple leaf nodes, breaking the BTree's sorted+unique invariant. The `diff` algorithm's `diffEntries` merge then produced wrong results because the flattened composite entries were no longer sorted.

## Fix

Adjacent children that need rebalancing are now grouped into contiguous runs. Each group (extended by one sibling on each side for merge context) is flattened and repartitioned as a single unit, preventing any child's data from appearing in two separate merge operations.

The now-unused `putManyMergeAndPartition` helper was removed since its logic is subsumed by the group-based approach.

### DIFF
--- a/packages/replicache/src/sync/patch-putmany-diff.test.ts
+++ b/packages/replicache/src/sync/patch-putmany-diff.test.ts
@@ -1,0 +1,148 @@
+import {LogContext} from '@rocicorp/logger';
+import {describe, expect, test} from 'vitest';
+import type {InternalDiff} from '../btree/node.ts';
+import {TestStore} from '../dag/test-store.ts';
+import {ChainBuilder} from '../db/test-helpers.ts';
+import {newWriteSnapshotDD31} from '../db/write.ts';
+import * as FormatVersion from '../format-version-enum.ts';
+import {deepFreeze} from '../frozen-json.ts';
+import type {PatchOperationInternal} from '../patch-operation.ts';
+import {withWriteNoImplicitCommit} from '../with-transactions.ts';
+import {apply} from './patch.ts';
+
+const formatVersion = FormatVersion.Latest;
+const clientID = 'client-id';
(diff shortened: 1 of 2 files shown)


packages/replicache/src/sync/patch-putmany-diff.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    open: i % 3 !== 0,
+    modified: 1769711510846 + i,
+    created: 1767179995185 + i,
+    creatorID: `usr_${String(i % 100).padStart(4, '0')}`,
+    assigneeID: `usr_${String((i + 37) % 100).padStart(4, '0')}`,
+    description: `Description for entity ${i}. This is some filler text to make the value realistically sized.`,
+    visibility: 'public',
+    projectID: `proj_${String(i % 50).padStart(3, '0')}`,
+    version,
+  };
```
arv, line 31: We really do not need all these columns but 🤷🏼

arv (approved):
LGTM

PR #5925, opened 2026-05-07
feat: Support mutator return values in client and server

arv:
## Summary
This PR adds support for mutators to return JSON values that are propagated back to the client. Previously, mutators could only return `void`. Now they can return any `ReadonlyJSONValue`, with separate values for client-side (optimistic) and server-side (authoritative) results.

## Key Changes

- **Type signature updates**: Changed mutator return type from `Promise<void>` to `Promise<ReadonlyJSONValue | void>` across:
  - `CustomMutatorImpl` in client
  - `MutatorDefinitionFunction` in ZQL
  - `TransactFnCallback` in server
  - Various mutator proxy and registry implementations

- **Client-side changes**:
  - `MutatorResultDetails` now includes optional `data` field to carry return values
  - `MutationTracker` extracts return data from mutation responses and includes it in success results
  - `mutator-proxy.ts` updated to pass return value data to result details
  - `make-replicache-mutators.ts` now returns the mutator result instead of discarding it

- **Server-side changes**:
  - `process-mutations.ts` captures mutator return values and includes them in `MutationResponse.result.data`
  - `push-processor.ts` updated to handle `ReadonlyJSONValue | void` returns
  - `custom.ts` updated with proper type imports

- **ZQL changes**:
  - `mutator-registry.ts` returns the mutator result instead of awaiting and discarding it
  - Type definitions updated to reflect new return type

- **Comprehensive test coverage**:
  - Client tests verify return values appear in client promise results
  - Server tests verify return values appear in push response data
  - Tests confirm client and server can have independent return values
  - Tests validate all JSON scalar types (numbers, strings, booleans, null, arrays, objects)
  - Tests confirm void returns result in undefined data

## Implementation Details

The implementation maintains backward compatibility - mutators that return `void` will have `undefined` data in the result. The client receives optimistic return values immediately when the mutator executes, while the server can return different authoritative values in the push response, allowing for patterns like returning optimistic IDs from the client and database-generated IDs from the server.

https://claude.ai/code/session_01S3wTBrt3VgCypjrt2ScZ83


packages/zero-client/src/client/custom.ts
```diff
@@ -188,13 +189,16 @@ export function makeReplicacheMutator<
   mutator: CustomMutatorImpl<S, TWrappedTransaction>,
   schema: S,
   context: Context,
-): (repTx: WriteTransaction, args: ReadonlyJSONValue) => Promise<void> {
+): (
+  repTx: WriteTransaction,
+  args: ReadonlyJSONValue,
+) => Promise<ReadonlyJSONValue | void> {
   return async (
```
arv: lint will complain

packages/zero-server/src/process-mutations.ts
```diff
@@ -615,11 +615,12 @@ class Transactor<D extends Database<ExtractTransactionType<D>>> {
             mutation.id,
           );
 
+          let returnData: ReadonlyJSONValue | void = undefined;
```
arv: no need to assign undefined here

arv (commented):
Nice

tantaman (commented):
this didn't require any changes to zero-cache?

tantaman:
Errors are persisted and returned over the sync protocol. Successes are ephemeral as implemented here.

I don't recall if we ever made a decision one way or the other as to whether success responses should be persisted along with error responses.

@aboodman ?

arv:
> Errors are persisted and returned over the sync protocol. Successes are ephemeral as implemented here.

Done — success return data is now written to the mutations table inside the same transaction as the LMID update, so it's durable across reconnections just like errors. The result column gets {} for undefined returns and {"data": ...} for non-undefined values.

arv:
@copilot resolve the merge conflicts in this pull request

packages/zero-server/src/process-mutations.ts
```diff
@@ -615,11 +615,18 @@ class Transactor<D extends Database<ExtractTransactionType<D>>> {
             mutation.id,
           );
 
+          let returnData: ReadonlyJSONValue | undefined;
           if (appError === undefined) {
             this.#lc.debug?.(
               `Executing mutator '${mutation.name}' (id=${mutation.id})`,
             );
```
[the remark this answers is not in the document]
arv: This didn't change

packages/zero-server/src/process-mutations.ts
```diff
@@ -615,30 +615,26 @@ class Transactor<D extends Database<ExtractTransactionType<D>>> {
...
+            const returnData = (await cb(
+              dbTx,
+              mutation.name,
+              mutation.args[0],
+            )) as ReadonlyJSONValue | undefined;
+            const successResult: MutationResponse = {
               id: {clientID: mutation.clientID, id: mutation.id},
-              result: {data: returnData},
-            });
+              result: returnData !== undefined ? {data: returnData} : {},
```
arv, line 629: the old one is fine.

JSON.stringify('{a: undefined}') => `{}`

aboodman:
Erik we need the context here. *why* are we doing this? What problem
spurred this change?


a (phone)

arv:
The initial pr was for when I was doing the rest API. I thought I would use mutators return values for that. I ended up reusing the code at a different layer. 

Now margins is asking for this.

packages/zero-client/src/client/custom.ts
arv: Lint is clean on the current head (`pnpm --filter zero-client run lint`), including this `makeReplicacheMutator` wrapper.

packages/zero-server/src/process-mutations.ts
arv: Fixed: this is now `const returnData = (await cb(...)) as ReadonlyJSONValue | undefined;` with no `undefined` initializer.

packages/zero-server/src/process-mutations.ts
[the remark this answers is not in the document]
arv: Not the case on the current head. The `appError` branch returns the error response it persisted (`const mutationResult = makeAppErrorResponse(mutation, appError); await transactionHooks.writeMutationResult(mutationResult); return mutationResult;`), so `persistPreTransactionFailure()` gets the error back. The success branch builds `result: returnData !== undefined ? {data: returnData} : {}`, so void mutators produce `{}` and never `{data: undefined}`.

packages/zero-server/src/process-mutations.test.ts
```diff
@@ -1421,3 +1438,215 @@ describe.each(mutatorInvokers)(
...
+      query: baseQuery,
+      body: makePushBody([makeCustomMutation({name: 'item.noop'})]),
+      userID: null,
+    });
+
+    expect(response).toEqual(
+      makeSuccessResponse(
+        [{id: {clientID: 'cid', id: 1}, result: {data: undefined}}],
+        null,
+      ),
```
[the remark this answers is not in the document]
arv: Already the case: the void-return tests assert `result: {}` (e.g. `makeSuccessResponse([{id: {clientID: 'cid', id: 1}, result: {}}], null)`), and the implementation omits `data` when the mutator returns `undefined`.

packages/zero-client/src/client/mutation-tracker.ts
```diff
@@ -439,7 +437,7 @@ export class MutationTracker {
       // the mutator proxy catches both client and server errors
       entry.resolver.reject(result);
     } else {
-      entry.resolver.resolve(successResultDetails);
+      entry.resolver.resolve(result.data);
```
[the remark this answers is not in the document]
arv: No change needed: `emptyObject` is `Object.freeze({})`, so an lmid-only settle resolves `data` as `undefined`, not `{}`. A returned server value is not lost on reconnect either: the success row is written in the same upstream transaction as the lmid bump, and Replicache fires the mutations watch (`#processMutationResponses`) inside the awaited poke, before `lmidAdvanced` runs, so the row settles the mutation with its `data` first. A regression test covering this path is coming in a follow-up commit.

PR #5930, opened 2026-05-07
fix(zero-cache): classify ConcurrentModificationException as a Rehome

darkgnotic:
This signals the client to reconnect (presumably to the right VS, based on the sticky cookie).

Context:
* https://rocicorp.slack.com/archives/C0AK5KX8HGE/p1778180835614519


0xcadams:
LGTM. I wonder if reconnect on Internal errors is something we should do. They're usually something we messed up in zero-cache.

0xcadams:
Related: https://github.com/rocicorp/mono/pull/5094#discussion_r3205435685

darkgnotic:
> LGTM. I wonder if reconnect on Internal errors is something we should do. They're usually something we messed up in zero-cache.

I think that would be a good idea. A lot of times internal errors are transient.

PR #5931, opened 2026-05-07
feat: added optional query to tx.run() and others

0xcadams:
It felt like this should exist, when using Zero on CZ; I wanted it to exist, and it felt asymmetric and confusing when it didn't.

- Allow conditional/falsy queries in `tx.run`, `zero.run`, `zero.preload`, and server `ZQLDatabase.run`, matching the `useQuery` maybe-query pattern.
- Return `undefined` or a no-op preload handle when a conditional query is disabled instead of trying to execute it.
- Add Vitest `expectTypeOf` coverage for strict vs conditional return types, plus runtime coverage for skipped conditional queries.

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -29,11 +29,13 @@ import {
 } from '../../../shared/src/browser-env.ts';
 import {getDocumentVisibilityWatcher} from '../../../shared/src/document-visible.ts';
 import {getErrorMessage} from '../../../shared/src/error.ts';
+import type {Falsy} from '../../../shared/src/falsy.ts';
 import {h64} from '../../../shared/src/hash.ts';
 import type {ReadonlyJSONValue} from '../../../shared/src/json.ts';
 import {must} from '../../../shared/src/must.ts';
 import {navigator} from '../../../shared/src/navigator.ts';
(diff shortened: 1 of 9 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -939,7 +941,41 @@ export class Zero<
...
+    query: QueryOrQueryRequest<TTable, TInput, TOutput, S, TReturn, C> | Falsy,
+    options?: PreloadOptions,
+  ): {
+    cleanup: () => void;
+    complete: Promise<void>;
+  } {
+    if (!query) {
+      return {
+        cleanup: emptyFunction,
+        complete: promiseVoid,
```
0xcadams, line 975: This is the only more odd API but it feels still more consistent than not allowing conditional queries.

arv:
I'm not sure I like this. The only reason we allow falsy things in useQuery is because the rules of hooks. It is better to get a type error that query is undefined in normal code.

0xcadams:
> I'm not sure I like this. The only reason we allow falsy things in useQuery is because the rules of hooks. It is better to get a type error that query is undefined in normal code.

Yeah that's fair. I left this open because I could go either way, but wanted to point out the asymmetry and get feedback.

I think you're right.

PR #5932, opened 2026-05-08
fix(zero-cache): use pg_class estimates for initial sync progress reporting

grgbkr:
Replace expensive COUNT(*) and SUM(pg_column_size(...)) full table scans in getInitialDownloadState with instant pg_class.reltuples and pg_table_size() lookups. These values are only used for progress reporting status events.

The previous exact queries caused failures on large databases as each scan could take hours on large tables.

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/initial-sync.ts
+++ b/packages/zero-cache/src/services/change-source/pg/initial-sync.ts
@@ -900,7 +900,7 @@ export async function getInitialDownloadState(
   const estimateResult = await sql<
     {totalRows: number; totalBytes: number}[]
   >`SELECT GREATEST(reltuples, 0)::float8 AS "totalRows",
-         pg_relation_size(oid)::float8 AS "totalBytes"
+         pg_table_size(oid)::float8 AS "totalBytes"
     FROM pg_class
     WHERE oid = ${qualifiedName}::regclass`;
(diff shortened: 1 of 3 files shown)


packages/zero-cache/src/services/change-source/pg/initial-sync.ts
```diff
@@ -886,30 +886,37 @@ export async function getInitialDownloadState(
...
-    .unsafe<{totalBytes: bigint}[]>(stmts.getTotalBytes)
-    .execute();
+  // Use pg_class estimates instead of expensive COUNT(*) and
+  // SUM(pg_column_size(...)) full table scans. The exact values are only
+  // used for progress reporting, so estimates are sufficient.
+  const qualifiedName = `${id(spec.schema)}.${id(spec.name)}`;
+  const estimateResult = await sql<
+    {totalRows: number; totalBytes: number}[]
+  >`SELECT GREATEST(reltuples, 0)::float8 AS "totalRows",
+         pg_relation_size(oid)::float8 AS "totalBytes"
```
darkgnotic: I think we'll want to use pg_table_size to include the size of toasted values, since we'll be receiving those in our COPY streams.

darkgnotic (approved):
Thanks @grgbkr !

PR #5933, opened 2026-05-08
feat(zero-cache): add region to litestream config

ericykim:
Expose a `ZERO_LITESTREAM_REGION` to set a region for non-standard AWS Partitions

Context:
When deploying zero-cache in AWS GovCloud (us-gov-west-1) or other non-standard AWS partitions, Litestream S3 backups fail with: `AuthorizationHeaderMalformed: the region 'us-east-1' is wrong; expecting 'us-gov-west-1'`


aboodman:
cool, thanks!

PR #5940, opened 2026-05-08
[codex] Revert cap EXISTS wiring

aboodman:
## Summary

This reverts `ce9b6d14e02633c59ac5ad35527157b11f9bc4fd` (`feat(zql): bring back cap!`) and keeps a focused regression test for the prod failure mode.

The reverted commit made non-flipped `EXISTS` child pipelines connect to their source without an order, then replaced `Take` with `Cap` for those children. That breaks when the non-flipped child query itself contains a flipped filter branch: `applyFilterWithFlips` constructs `UnionFanIn`, and `UnionFanIn` requires sorted input.

## Regression Test

Adds a unit test for this shape:

```text
issue
  WHERE EXISTS comments
    WHERE comment.text = "public"
       OR EXISTS owner WHERE owner.role = "teacher"  // flipped branch
```

Before the revert, the test fails with:

```text
UnionFanIn requires sorted input
```

After the revert, it passes.

## Validation

```bash
npm test -- --project=zql packages/zql/src/ivm/cap.push.test.ts -t 'non-flipped EXISTS child with flipped OR branch keeps sorted input for UnionFanIn'
```

Result:

```text
Test Files  1 passed (1)
Tests       1 passed | 10 skipped (11)
```


grgbkr:
LGTM

tantaman:
hmm, I guess we would have to swap `Cap` to `Take` if `exists` is inside a `UFI`

PR #5943, opened 2026-05-08
feat(cap): wire cap, do not use it in a UFI path

tantaman:


packages/zql/src/ivm/cap.ts
```diff
@@ -88,56 +88,36 @@ export class Cap implements Operator {
     assert(!req.start, 'Cap does not support start');
     assert(!req.reverse, 'Cap does not support reverse');
 
-    if (
```
tantaman, line 91: idk why this change was made in https://github.com/rocicorp/mono/commit/ad038d44d9522322530290a9f7cc058ad2a6e99b#diff-4e39bf99f2add265725b9dd6ca98b911bace7753ef76eaa01d6c5259a4c0bc96

but reverting it with the revert

packages/zql/src/ivm/cap.ts
```diff
@@ -88,56 +88,36 @@ export class Cap implements Operator {
...
+      'Cap fetch: constraint must match partition key when partitioned',
+    );
+
+    const capStateKey = getCapStateKey(this.#partitionKey, req.constraint);
+    const capState = this.#storage.get(capStateKey);
+    if (!capState) {
+      yield* this.#initialFetch(req);
       return;
     }
-    // There is a partition key, but the fetch is not constrained or constrained
```
tantaman, line 119: this should never happen for cap.

packages/zql/src/ivm/cap.ts
tantaman: oh, because `bring back cap` had the builder change + this change. Got it.

PR #5944, opened 2026-05-08
chore(zero-cache): handle NULL current_query() and log unparseable ddl events

darkgnotic:
For debugging a unexpected parse error when handling a ddl event message:

```
TypeError: Expected string at context.query. Got null: Expected string at context.query. Got null
```

Context:
* https://rocicorp.slack.com/archives/C0B2JKKK9QC/p1778261548602499


arv (commented):
LGTM

PR #5945, opened 2026-05-09
feat(zql): fix types for args to be optional when undefined

0xcadams:
Custom query and mutator execution functions now treat `args` as optional when the args type already allows `undefined`.

This lets callers write:

```ts
queries.teams.fn({ctx: user});
mutators.teams.fn({tx, ctx: user});
```

instead of having to pass:

```ts
queries.teams.fn({args: undefined, ctx: user});
mutators.teams.fn({args: undefined, tx, ctx: user});
```

## Details

- Adds conditional execution option types for custom queries and mutators.
- Keeps `args` required when the args type does not include `undefined`.
- Preserves existing `ctx` behavior: `ctx` is only optional for unknown context.
- Preserves existing mutator behavior: `tx` is always required.
- Adds type coverage for omitted `args`, explicit `args: undefined`, optional validator inputs, and required-args rejection.


arv (commented):
LGTM

PR #5950, opened 2026-05-11
chore(zero-cache): loosen invariant in shadow sync

tantaman:
https://rocicorp.slack.com/archives/C0AK5KX8HGE/p1778482696395909?thread_ts=1778344567.202889&cid=C0AK5KX8HGE

Shadow sync would throw if a table that was synced could not be queried by ZQL. Prod just silently ignores this case.


darkgnotic (commented):
LGTM

PR #5952, opened 2026-05-12
chore(zero-cache): make a TestDB options object for easier configurability

darkgnotic:
purely testing refactoring


packages/zero-cache/src/test/db.ts
```diff
@@ -36,6 +36,11 @@ const CONNECTION_URI = mustInject('pgConnectionString');
 
 export type OnNoticeFn = (n: postgres.Notice) => void;
 
+export type TestDBOptions = {
+  onNotice?: OnNoticeFn;
```
arv, line 40: the general rule is to also allow undefined. makes code simpler

arv (commented):
LGTM

PR #5954, opened 2026-05-12
feat(zero): add typed init helpers

0xcadams:
- Recreates #5404.
- Adds typed `initZero`, `initZeroReact`, and `initZeroSolid` helpers for schema/context-bound APIs.
- Updates zbugs to use local typed Zero helpers/hooks and removes `@rocicorp/zero` module augmentation.


apps/zbugs/server/server-mutators.ts
```diff
@@ -13,13 +7,19 @@ import {
...
+  defineMutator,
+  defineMutators,
+  type Transaction,
+} from '../shared/zero.ts';
 import {notify} from './notify.ts';
+import type {ServerTransaction} from './zero.ts';
 
 export type PostCommitTask = () => Promise<void>;
 
 function asServerTransaction(tx: Transaction): ServerTransaction {
```
aboodman: This smells to me, but might be a different problem.

aboodman:
excited to bring this back. please don't land until i have a chance to review carefully. let's take this one slowly.

0xcadams:
> excited to bring this back. please don't land until i have a chance to review carefully. let's take this one slowly.

Sounds good. It's still wip and I have some ideas for cleaning up the types but not sure how it'll feel yet.

I will tag you when it's ready.

apps/zbugs/shared/zero.ts
```diff
@@ -0,0 +1,19 @@
...
+  postgres.TransactionSql
+>({schema});
+
+export const {Zero, defineMutator, defineMutators, defineQuery, defineQueries} =
+  zero;
+
+export type ZeroTypes = typeof zero;
+export type Row = ZeroTypes['~']['$row'];
+export type Transaction = ZeroTypes['~']['$transaction'];
+export type ServerTransaction = ZeroTypes['~']['$serverTransaction'];
```
0xcadams, line 18: @aboodman @arv I added these phantom types. I think I like them. Curious your thoughts. The alternative is a lot of manual type definitions. I think this gets back some of the goodness from the global types but without the magic of the global types.

apps/zbugs/shared/zero.ts
```diff
@@ -0,0 +1,19 @@
+import {initZero} from '@rocicorp/zero';
+import type postgres from 'postgres';
+import type {AuthData} from './auth.ts';
+import {schema} from './schema.ts';
+
+export const zero = initZero<
```
aboodman, line 6: More of a pattern thing, but I don't think we should name the return value `zero`, as it's not an instance of `Zero`. Instead directly destructure inline to the value on line 12.

apps/zbugs/shared/zero.ts
```diff
@@ -0,0 +1,19 @@
...
+  typeof schema,
+  AuthData | undefined,
+  postgres.TransactionSql
+>({schema});
+
+export const {Zero, defineMutator, defineMutators, defineQuery, defineQueries} =
+  zero;
+
+export type ZeroTypes = typeof zero;
+export type Row = ZeroTypes['~']['$row'];
```
aboodman, line 16: I think it's a little 'clever' (in a not good way) to put the types on `initRetVal["~"]`.

Is it that hard to say:

```ts
type Row = ZeroRow<typeof schema>
...
```

What does it look like to do it without any helpers?

apps/zbugs/shared/zero-hooks.ts
```diff
@@ -0,0 +1,10 @@
+import {wrapZeroReact} from '@rocicorp/zero/react';
```
aboodman, line 1: Rather than having separate facade files for the hooks and core, I was hoping to have a single entrypoint in zero-react that returns everything, including the stuff from the low-level ts library:

```ts
import {initZero} from "@rocicorp/zero-react";

const {ZeroProvider, useZero, Zero, ...} = initZero({schema});
```

apps/zbugs/shared/zero.ts
aboodman: If that's undesirable, then I think my next favorite thing is to return a `types` object from `initZero()` which at runtime is undefined, but which has properties having all the types:

```ts
const {types} = initZero({schema})

// at runtime, this is actually undefined
assertUndefined(types);

// but typesystem thinks it has all these fields
type Types = typeof types;
type Row = Types["row"];

...
```

apps/zbugs/shared/zero.ts
```diff
@@ -0,0 +1,19 @@
+import {initZero} from '@rocicorp/zero';
+import type postgres from 'postgres';
+import type {AuthData} from './auth.ts';
+import {schema} from './schema.ts';
+
+export const zero = initZero<
+  typeof schema,
+  AuthData | undefined,
+  postgres.TransactionSql
```
aboodman, line 9: I do not understand how we can pass a raw postgres.js type here. I would have expected some wrapper.

apps/zbugs/shared/zero-hooks.ts
aboodman: I wonder if there should be a separate `initZero` for server-side, so that we don't have to have things like `ServerTransaction` in shared client code.

So basically there'd be an `initZero` for all of these places:

* `@rocicorp/zero`
* `@rocicorp/zero/react`
* `@rocicorp/zero/solid`
* `@rocicorp/zero/server`

apps/zbugs/shared/zero.ts
0xcadams: Makes sense to me. Those helpers like `ZeroRow<typeof schema>` work under the hood by accessing the phantom types. But we don't need to have users access those directly. I will take a pass with this.

apps/zbugs/shared/zero-hooks.ts
0xcadams: On the single entrypoint - I landed on this `wrap` pattern because it reduced the number of times you need to define type params. Your example above would need to look like:

```tsx
import {initZero} from "@rocicorp/zero-react";

const {ZeroProvider, useZero, Zero, ...} = initZero<
  typeof schema,
  AuthData | undefined
>({schema});

// or

const {ZeroProvider, useZero, Zero, ...} = initZero<
  typeof schema,
  AuthData | undefined
>();
```

It ended up with less boilerplate to make it composable.

0xcadams: Of course we can separate these out, but that's the tradeoff. Composability means you put the type params only once, standalone means no dependencies on each other but you redefine type params in multiple places.

apps/zbugs/shared/zero.ts
0xcadams: That's the underlying `wrappedTransaction` type. When you say "wrapper" what were you thinking?

apps/zbugs/shared/zero-hooks.ts
0xcadams: On separate `initZero` for server - I agree. 

A related point: we could have the `@rocicorp/zero/server` return a server-typed `defineMutator`.

It's closer to how people use `defineMutator` anyways. We have these `Transaction` types that you can narrow, but then usually you end up having `server-mutators.ts` and *every* mutator in that needs to do the type narrowing individually, by asserting `tx.location === 'server'`.

apps/zbugs/shared/zero.ts
0xcadams: Sounds good - this was used for the composability problem below: https://github.com/rocicorp/mono/pull/5954#discussion_r3415617589

PR #5959, opened 2026-05-13
chore(zero-cache): add performance benchmark suite

Karavil:
## AI-generated draft

This draft PR was prepared by an AI coding agent. It is intentionally still draft, but the bar is still normal review quality: the motivation, benchmark scenarios, decision criteria, limits, risks, and follow-up PR links are included here.

## Why This Exists

The worst version of this stack is not "performance work without perfect benchmarks." The worst version is performance work where every PR sounds plausible, every diff is locally reasonable, and no reviewer has a shared way to tell whether pressure moved from one part of `zero-cache` to another.

That is how regressions become institutional memory instead of testable facts: six months later, someone can accidentally grow poke frames back to 800 KiB, increase ACK chatter, or make flow-control padding theoretical again, and the review has no repeatable scenario to point at.

This PR gives the rest of the stack a measurement surface before the production PRs ask for trust.

```text
Without this PR
---------------
idea -> production diff -> reviewer intuition -> "seems faster?"

With this PR
------------
idea -> scenario -> main number -> PR number -> explicit tradeoff
```

The suite is not claiming to be production. It is claiming something narrower and more useful for review: if a PR says it reduces a specific source of pressure, there should be a scenario where that pressure is visible.

## Clean PR Set

The original Capy branches were a linear development stack. The benchmark-only followups are folded into this PR so reviewers do not have to review fake stack layers. The active PRs are clean draft branches against `main`.

Base for the cleaned branches: `rocicorp/mono@65f1461f761936288a11fde0585d100432ce9aef`.

| PR | Branch | Role | How to read it |
|---|---|---|---|
| [#5959](https://github.com/rocicorp/mono/pull/5959) | `capy/zero-load-ws-harness` | Benchmark suite | Establishes the measuring tools |
| [#5960](https://github.com/rocicorp/mono/pull/5960) | `capy/byte-aware-poke-flush` | Production poke flush change | Uses `perf:poke` evidence from this PR |
| [#5963](https://github.com/rocicorp/mono/pull/5963) | `capy/cumulative-ack-ws-streams` | Production ACK batching | Uses `perf:ack` evidence from this PR |
| [#5964](https://github.com/rocicorp/mono/pull/5964) | `capy/scale-fanout-flow-control` | Production fanout release timing | Uses branch-local fanout evidence because it exercises the new `Broadcast` timer API |

Superseded by this PR:

| Closed PR | Why it is no longer separate |
|---|---|
[...]

PR #5960, opened 2026-05-13
perf(zero-cache): flush poke parts by byte budget

Karavil:
## AI-generated draft

This draft PR was prepared by an AI coding agent. It is intentionally still draft, but it is a clean production PR against `main`; benchmark scaffolding lives in [#5959](https://github.com/rocicorp/mono/pull/5959).

## Why This Is Worth It

The current flush rule asks one question: "have we collected 100 row/config patches yet?" That is easy to reason about, but it ignores the part a browser, WebSocket buffer, and JS runtime actually feel: bytes.

The scary case is not the average poke. It is one large transaction landing in one subscribed tab:

```text
100 rows, 8 KiB each
--------------------
current rule:  count rows to 100
result:        one ~810 KiB pokePart frame

client side:
large frame -> allocate -> parse JSON -> dispatch patch batch
            -> event loop is busy while every other tab task waits
```

This PR changes the control unit from "about 100 patches" to "about 128 KiB or 100 patches, whichever comes first." The wire protocol does not change. Only the cut points change.

```text
Before
------
patch patch patch ... patch
|-------------------------|
        one large pokePart

After
-----
patch patch patch ... patch
|-----|-----|-----|-----|--
  smaller pokeParts with the same pokePart shape
```

The improvement is not compression. The improvement is turning one tail-latency-shaped frame into several smaller frames so WebSocket framing, JSON parse, dispatch, and backpressure get multiple scheduling boundaries instead of one large synchronous-looking lump.

## Clean PR Set

| PR | Role | Relationship |
|---|---|---|
| [#5959](https://github.com/rocicorp/mono/pull/5959) | Benchmark suite | Provides `perf:poke` evidence |
| [#5960](https://github.com/rocicorp/mono/pull/5960) | This PR | Production byte-aware poke flushing |
| [#5963](https://github.com/rocicorp/mono/pull/5963) | Production cumulative ACKs | Independent |
| [#5964](https://github.com/rocicorp/mono/pull/5964) | Production fanout release timing | Independent |

## What Changed

- Adds `pokePartFlush.maxBytes` and `pokePartFlush.maxRows` config.
- Keeps `maxRows = 100` as the old pathological-object guard.
- Adds `maxBytes = 128 * 1024` as the new target byte budget.
- Flushes when either the byte target or row target is reached.
- Batches patch insertion with `addPatches()` so hot paths do not await per patch.
- Pre-encodes the outbound pokePart once it is ready to send.
- Keeps the sync protocol unchanged: clients still receive normal `pokeStart`, `pokePart`, and `pokeEnd` messages.

## Benchmark Headline

Benchmark harness: [#5959](https://github.com/rocicorp/mono/pull/5959), `npm --workspace=zero-cache run perf:poke:full`.

For the workload count-based flushing handles worst, the max frame drops from 810.2 KiB to 129.7 KiB. Total encoded bytes stay essentially the same.

```text
100 rows/tx, 8 KiB rows
-----------------------
main:     1 pokePart per tx, max frame 810.2 KiB
this PR: ~7 pokeParts per tx, max frame 129.7 KiB

volume:       same order, about 791 MiB either way
distribution: one huge frame becomes several smaller frames
tradeoff:     more WebSocket messages and dispatch events
```

| Scenario | Main / fixed-count batching | This PR / byte-aware batching | What changed |
|---|---:|---:|---|
| 10 rows/tx, small 96 B rows | 3,000 messages, 1,000 pokeParts, 2.05 MiB encoded, 1.98 KiB max msg, p95 0.083 ms, CPU 16.45 ms | 3,000 messages, 1,000 pokeParts, 2.05 MiB encoded, 1.98 KiB max msg, p95 0.103 ms, CPU 17.63 ms | No real chunking change; overhead is tiny |
| 10 rows/tx, large 8 KiB rows | 3,000 messages, 1,000 pokeParts, 79.3 MiB encoded, 81.0 KiB max msg, p95 0.627 ms, CPU 157.55 ms | 3,000 messages, 1,000 pokeParts, 79.3 MiB encoded, 81.0 KiB max msg, p95 0.591 ms, CPU 153.02 ms | Same chunking; no regression |
| 100 rows/tx, small 96 B rows | 3,000 messages, 1,000 pokeParts, 19.2 MiB encoded, 19.6 KiB max msg, p95 0.440 ms, CPU 93.55 ms | 3,000 messages, 1,000 pokeParts, 19.2 MiB encoded, 19.6 KiB max msg, p95 0.489 ms, CPU 101.28 ms | No chunking change; small CPU/p95 cost |
| 100 rows/tx, large 8 KiB rows | 3,000 messages, 1,000 pokeParts, 791.3 MiB encoded, 810.2 KiB max msg, p95 0.993 ms, CPU 452.39 ms | 9,000 messages, 7,000 pokeParts, 791.5 MiB encoded, 129.7 KiB max msg, p95 1.101 ms, CPU 480.33 ms | Max message -84.0%; messages +200%; CPU +6.2% |

That last row is the whole argument for the PR: it buys a much smaller worst frame in exchange for more frames. The +200% message count is real; the bet is that several ~130 KiB frames are kinder to buffers, parse work, and dispatch scheduling than one ~810 KiB frame. Review should focus on whether that is the right tradeoff for large-row subscriber experience.

- Cannot change the sync protocol. The outbound message is still a normal `pokePart`; only the grouping changes.
- Cannot remove the old guard. `maxRows = 100` remains, so pathological object count is still bounded.
- Cannot promise a hard 128 KiB cap. The implementation estimates bytes before flushing, so a frame can exceed the target by the patch that crosses it.
- Cannot grow a global pre-encoding cache. Pre-encoded payloads are keyed by the downstream message object and disappear with that object lifetime.
- Cannot send patches for clients that should not see them. The existing base-version filtering still happens before a patch is added to the pending body.
- Cannot make one huge row small. A single row larger than the byte budget can still produce a large frame; the threshold only prevents groups of large rows from accumulating into one frame.

- Large-row workloads send more messages and pokeParts. That can increase WebSocket framing overhead, event-loop scheduling, and client dispatch overhead.
- This benchmark does not measure browser parse time. That is the missing client-side confirmation for the theory that smaller frames are easier on the tab.
- Byte estimation is intentionally approximate because final JSON shape and framing can differ from the estimate.
- The 128 KiB default is a tunable runtime tradeoff, chosen as a conservative target far below the measured 810 KiB tail frame, not as a protocol law.
- Smaller chunks can alter backpressure timing. That is part of the intended behavior, but it deserves rollout attention.

## Validation

- `npm --workspace=zero-cache run test:no-pg -- client-handler.test.ts connection.test.ts zero-config.test.ts`

`client-handler.test.ts` includes the new byte-threshold flush case and the retained max-row guard case.

### DIFF
--- a/packages/zero-cache/src/config/zero-config.test.ts
+++ b/packages/zero-cache/src/config/zero-config.test.ts
@@ -476,6 +476,19 @@ test('zero-cache --help', () => {
                                                                                                                                                                                                    
                                                                         A password is optional in development mode but required in production mode.                                                
                                                                                                                                                                                                    
+     --poke-part-flush-max-bytes number                                 default: 131072                                                                                                            
+       ZERO_POKE_PART_FLUSH_MAX_BYTES env                                                                                                                                                          
+                                                                        Target maximum JSON bytes per pokePart before the view-syncer flushes to WebSocket.                                        
+                                                                                                                                                                                                   
+                                                                        This bounds large row pokes without changing the sync protocol. Actual messages can exceed                                 
+                                                                        this target by one patch because rows are estimated before flushing. Default: 128KB.                                       
+                                                                                                                                                                                                   
+     --poke-part-flush-max-rows number                                  default: 100                                                                                                               
+       ZERO_POKE_PART_FLUSH_MAX_ROWS env                                                                                                                                                           
+                                                                        Maximum row/config patches per pokePart before flushing.                                                                   
+                                                                                                                                                                                                   
+                                                                        This remains as a pathological-object guard in addition to the byte target.                                                
+                                                                                                                                                                                                   
      --websocket-compression boolean                                    default: false                                                                                                             
        ZERO_WEBSOCKET_COMPRESSION env                                                                                                                                                              
                                                                         Enable WebSocket per-message deflate compression.                                                                          
--- a/packages/zero-cache/src/config/zero-config.ts
+++ b/packages/zero-cache/src/config/zero-config.ts
@@ -734,6 +734,26 @@ export const zeroOptions = {
     ],
   },
 
+  pokePartFlush: {
+    maxBytes: {
+      type: v.number().default(128 * 1024),
+      desc: [
+        'Target maximum JSON bytes per pokePart before the view-syncer flushes to WebSocket.',
+        '',
+        'This bounds large row pokes without changing the sync protocol. Actual messages can exceed',
+        'this target by one patch because rows are estimated before flushing. Default: 128KB.',
+      ],
+    },
+    maxRows: {
+      type: v.number().default(100),
+      desc: [
+        'Maximum row/config patches per pokePart before flushing.',
+        '',
+        'This remains as a pathological-object guard in addition to the byte target.',
+      ],
+    },
+  },
+
   websocketCompression: {
     type: v.boolean().default(false),
     desc: [
--- a/packages/zero-cache/src/services/view-syncer/client-handler.test.ts
+++ b/packages/zero-cache/src/services/view-syncer/client-handler.test.ts
@@ -11,6 +11,7 @@ import type {
 import {Subscription} from '../../types/subscription.ts';
 import {
   ClientHandler,
+  DEFAULT_POKE_PART_FLUSH_BYTES,
   ensureSafeJSON,
   startPoke,
   type Patch,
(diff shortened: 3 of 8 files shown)


Karavil:
Closing this draft for now because the current 1 RM / 16 view-syncer e2e throughput benchmarks do not show a convincing production-shaped win. Keeping the benchmark harness open and working backwards from those scenarios to find changes that actually move throughput.

PR #5963, opened 2026-05-13
perf(zero-cache): batch change-streamer ACKs

Karavil:
## AI-generated draft

This draft PR was prepared by an AI coding agent. It is intentionally still draft, but it is a clean production PR against `main`; the ACK benchmark harness lives in [#5959](https://github.com/rocicorp/mono/pull/5959).

## Why This Is Worth It

The internal change-streamer WebSocket stream currently pays one ACK frame for every consumed message. That is correct, but it means ACK traffic scales with message count even when the receiver is processing a normal contiguous pipeline.

The worst case is not that each ACK is large. Each ACK is tiny. The problem is the shape under throughput:

```text
10,000 streamed messages/sec
----------------------------
old behavior: 10,000 data frames + 10,000 ACK frames

Every ACK competes for:
- WebSocket framing
- JSON parse/stringify
- event-loop turns
- sender-side pending-message cleanup
```

This PR changes only the ACK watermark for the opt-in internal change-streamer path. Instead of saying "message 1 is consumed, message 2 is consumed, message 3 is consumed," the receiver can say "everything through 3 is consumed."

```text
Before
------
server -> msg 1 -> client
server <- ack 1 <- client
server -> msg 2 -> client
server <- ack 2 <- client
server -> msg 3 -> client
server <- ack 3 <- client

After
-----
server -> msg 1 -> client
server -> msg 2 -> client
server -> msg 3 -> client
server <- ack 3 <- client
```

The key property: `ack 3` means all messages up through 3 were consumed. It does not acknowledge gaps.

## Clean PR Set

| PR | Role | Relationship |
|---|---|---|
| [#5959](https://github.com/rocicorp/mono/pull/5959) | Benchmark suite | Provides `perf:ack` evidence |
| [#5960](https://github.com/rocicorp/mono/pull/5960) | Production byte-aware poke flushing | Independent |
| [#5963](https://github.com/rocicorp/mono/pull/5963) | This PR | Production cumulative ACKs |
| [#5964](https://github.com/rocicorp/mono/pull/5964) | Production fanout release timing | Independent |

## What Changed

- `streamIn(..., {ack: "cumulative"})` can batch ACKs.
- The change-streamer HTTP client opts into cumulative ACKs for internal streams.
- Generic `streamIn` callers keep per-message ACKs unless they explicitly opt in.
- The sender accepts monotonic cumulative ACKs and marks all pending messages up to that ID as consumed.
- The receiver only advances the ACK watermark across contiguous consumed messages.
- Pending ACKs flush on cleanup so a short stream is not left unacknowledged.
- Backwards ACKs are treated as protocol errors and close the connection.

## Gap Semantics

This is the correctness core of the PR.

```text
received IDs:  1  2  3  4  5
consumed IDs:  y  y  n  y  y
ackable ID:    2

No ack 5 is sent yet.
Message 3 is still the hole in the consumed prefix.
When 3 is consumed, the watermark can advance to 5 in one ACK.
```

There is no new timeout for a gap. If message 3 never becomes consumed, the sender keeps the later messages pending just like it would keep an unacked message pending today. The PR reduces normal-case ACK chatter; it does not weaken the consumed-before-release contract.

## Benchmark Headline

Benchmark harness: [#5959](https://github.com/rocicorp/mono/pull/5959), `npm --workspace=zero-cache run perf:ack`.

Smoke scenarios reduce ACK frames by about 53% to 87%. The biggest wins show up in bursty streams because many messages become one contiguous consumed prefix.

[#5959](https://github.com/rocicorp/mono/pull/5959) also includes an idealized simulation table with larger ACK reductions. This PR uses the actual implementation numbers below, which are intentionally more conservative because they include batching thresholds, cleanup flushing, and real stream behavior.

| Scenario | Main / per-message ACK | This PR / cumulative ACK | ACK traffic delta | p95 latency delta | Decision |
|---|---:|---:|---:|---:|---|
| steady, 1 msg every 1 ms, small 96 B payload | 1,000 ACKs, 10.6 KiB, p95 0.494 ms | 466 ACKs, 4.96 KiB, p95 0.594 ms | -53.4% messages, -53.4% bytes | +0.100 ms | GO |
| steady, 1 msg every 1 ms, medium 1 KiB payload | 1,000 ACKs, 10.6 KiB, p95 0.636 ms | 461 ACKs, 4.91 KiB, p95 0.549 ms | -53.9% messages, -53.9% bytes | -0.087 ms | GO |
| burst, 10 msgs every 10 ms, small 96 B payload | 1,000 ACKs, 10.6 KiB, p95 2.415 ms | 200 ACKs, 2.13 KiB, p95 1.805 ms | -80.0% messages, -80.0% bytes | -0.610 ms | GO |
| burst, 10 msgs every 10 ms, medium 1 KiB payload | 1,000 ACKs, 10.6 KiB, p95 2.472 ms | 200 ACKs, 2.13 KiB, p95 1.492 ms | -80.0% messages, -80.0% bytes | -0.980 ms | GO |
| burst, 100 msgs every 100 ms, small 96 B payload | 1,000 ACKs, 10.6 KiB, p95 8.177 ms | 130 ACKs, 1.38 KiB, p95 5.717 ms | -87.0% messages, -87.0% bytes | -2.460 ms | GO |
| burst, 100 msgs every 100 ms, medium 1 KiB payload | 1,000 ACKs, 10.6 KiB, p95 8.346 ms | 130 ACKs, 1.38 KiB, p95 5.890 ms | -87.0% messages, -87.0% bytes | -2.456 ms | GO |

The important read is not "ACK bytes dominate the system." They probably do not. The important read is that this removes fixed per-message control-plane work from a hot internal stream without changing the streamed application messages.

- Cannot acknowledge an unconsumed gap. The receiver only sends the highest contiguous consumed ID.
- Cannot move the ACK watermark backwards. A backwards ACK closes the connection.
- Cannot acknowledge a future message after normal startup. The sender rejects ACKs beyond what it has sent.
- Cannot affect generic stream users by default. Cumulative ACKs are opt-in.
- Cannot strand a short stream just below the batch threshold. Cleanup flushes the highest contiguous ACK that is actually ackable; it does not jump across a gap.
- Cannot change the data frame payload. The streamed message shape is unchanged; this PR changes ACK semantics for the opted-in path.

- Mixed-version rollout is the biggest operational risk. New sender with old receiver is naturally okay because per-message ACKs are monotonic. Old sender with new cumulative receiver can be unsafe because the old pipelined sender expects exact per-message ACK IDs. That is why cumulative mode is opt-in and why rollout should keep the internal change-streamer sender/receiver on compatible code.
- A bug in contiguous-ID tracking could release messages before they are actually consumed. The tests focus on batching, cleanup flush, gaps, and backwards ACK rejection because that is the scary correctness surface.
- Delayed ACKs can keep more messages pending on the sender for a few milliseconds.
- The benchmark is still synthetic. Longer/full-mode reruns are useful before treating the exact p95 deltas as capacity evidence.

## Validation

- `npm --workspace=zero-cache run test:no-pg -- streams.test.ts`

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/change-streamer-http.ts
+++ b/packages/zero-cache/src/services/change-streamer/change-streamer-http.ts
@@ -238,7 +238,7 @@ export class ChangeStreamerHttpClient implements ChangeStreamer {
     const params = getParams(ctx);
     const ws = new WebSocket(uri + `?${params.toString()}`);
 
-    return streamIn(this.#lc, ws, downstreamSchema);
+    return streamIn(this.#lc, ws, downstreamSchema, {ack: 'cumulative'});
   }
 }
 
--- a/packages/zero-cache/src/types/streams.test.ts
+++ b/packages/zero-cache/src/types/streams.test.ts
@@ -6,6 +6,7 @@ import Fastify, {type FastifyInstance} from 'fastify';
 import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest';
 import WebSocket from 'ws';
 import {unreachable} from '../../../shared/src/asserts.ts';
+import type {JSONValue} from '../../../shared/src/bigint-json.ts';
 import {
   createSilentLogContext,
   TestLogSink,
@@ -219,6 +220,7 @@ describe('streams with internal acks', () => {
   let consumed: Queue<Message>;
   let cleanedUp: Promise<Message[]>;
   let cleanup: (m: Message[]) => void;
+  let ackMessages: number[];
   let port: number;
 
   let ws: WebSocket;
@@ -231,6 +233,7 @@ describe('streams with internal acks', () => {
     cleanup = resolve;
 
     consumed = new Queue();
+    ackMessages = [];
     producer = Subscription.create({
       consumed: m => consumed.enqueue(m),
       cleanup: resolve,
@@ -240,6 +243,16 @@ describe('streams with internal acks', () => {
     server = Fastify();
     await server.register(websocket);
     server.get('/', {websocket: true}, ws => {
+      ws.on('message', data => {
+        const text = data.toString();
+        if (!text.startsWith('{"ack":')) {
+          return;
+        }
+        const {ack} = JSON.parse(text) as {ack?: unknown};
+        if (typeof ack === 'number') {
+          ackMessages.push(ack);
+        }
+      });
       void streamOut(lc, producer, ws);
       void streamOutStringified(lc, stringifiedProducer, ws);
     });
@@ -260,11 +273,9 @@ describe('streams with internal acks', () => {
     ws = new WebSocket(`http://localhost:${port}/`);
     return {
       ws,
-      consumer: (await streamIn(
-        lc,
-        ws,
-        messageSchema,
-      )) as Subscription<Message>,
+      consumer: (await streamIn(lc, ws, messageSchema, {
+        ack: 'cumulative',
+      })) as Subscription<Message>,
     };
   }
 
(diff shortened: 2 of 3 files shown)


Karavil:
Closing this draft for now because the current 1 RM / 16 view-syncer e2e throughput benchmarks do not show a convincing production-shaped win. Keeping the benchmark harness open and working backwards from those scenarios to find changes that actually move throughput.

PR #5964, opened 2026-05-13
perf(zero-cache): release fanout flow control precisely

Karavil:
## AI-generated draft

This draft PR was prepared by an AI coding agent. It is intentionally still draft, but it is a clean production PR against `main`; benchmark-only stack layers have been removed.

## Why This Is Worth It

Flow control is supposed to keep replication from outrunning downstream subscribers. The current policy is already the right shape: once a majority of subscribers have ACKed a flushed change, wait a little longer for stragglers, then let replication continue.

The bug is timing. The code says "wait for majority plus padding," but the release check is tied to a 1 second progress-monitor tick. That makes a 100 ms padding target mostly theoretical.

For a reviewer, the user-facing version is: a healthy majority can be ready, but replication still waits for the next monitor tick. In a subscribed view, that is the kind of stale window that feels like "I wrote something, why is everyone else still behind?"

```text
Before
------
t=000ms  majority ACKs
t=100ms  desired release point
t=1000ms progress monitor notices and releases

After
-----
t=000ms  majority ACKs
t=100ms  Broadcast timer releases
```

This PR moves the timer to the object that actually observes majority completion: `Broadcast`.

## Clean PR Set

| PR | Role | Relationship |
|---|---|---|
| [#5959](https://github.com/rocicorp/mono/pull/5959) | Benchmark suite | Related measurement work |
| [#5960](https://github.com/rocicorp/mono/pull/5960) | Production byte-aware poke flushing | Independent |
| [#5963](https://github.com/rocicorp/mono/pull/5963) | Production cumulative ACKs | Independent |
| [#5964](https://github.com/rocicorp/mono/pull/5964) | This PR | Production precise majority-padding fanout release |

## What Changed

- `Broadcast` can receive flow-control options.
- Once a majority completes, `Broadcast` arms a padding timer directly.
- If another subscriber completes after majority, the timer resets so padding is measured from the latest completion, preserving the existing intent.
- If all subscribers complete, `Broadcast` resolves immediately and clears the timer.
- Negative padding still disables early release.
- Default consensus padding changes from `1s` to `0.1s`.

```text
Broadcast state machine
-----------------------
send change
  |
  v
wait for majority
  |
  v
majority complete
  |
  v
start/reset padding timer
  |
  +-- all subscribers done
  |      |
  |      v
  |   resolve Broadcast.done -> replication can continue
  |
  +-- padding expires with minority still pending
         |
         v
      resolve Broadcast.done -> replication can continue
```

## Benchmark Headline

The fanout benchmark was run against this PR's development branch because it exercises the new `Broadcast` timer API introduced here. The production PR keeps that benchmark scaffolding out of the diff so reviewers can focus on runtime behavior.

Command used:

```sh
npm --workspace=zero-cache run perf:fanout
```

The config-only column is the important control. It shows that changing the default from 1s to 100ms is not enough while the release path is still polled at 1s. The mechanism change is load-bearing.

| Scenario | Main / polling 1s monitor, 1s padding | Config-only polling 1s monitor, 100ms padding | This PR / precise 100ms padding | Delta vs main |
|---|---:|---:|---:|---|
| 4 subscribers, 1 slow subscriber | p95 wait 2001.2 ms, 0.7 flushes/s, max lag 4 | p95 wait 1001.2 ms, 1.0 flushes/s, max lag 4 | p95 wait 101.4 ms, 9.8 flushes/s, max lag 4 | p95 -1899.8 ms (-94.9%); flush cadence about 13.1x |
| 8 subscribers, 1 slow subscriber | p95 wait 2001.0 ms, 0.7 flushes/s, max lag 4 | p95 wait 1000.9 ms, 1.0 flushes/s, max lag 4 | p95 wait 101.1 ms, 9.9 flushes/s, max lag 4 | p95 -1899.9 ms (-94.9%); flush cadence about 13.2x |
| 16 subscribers, 1 slow subscriber | p95 wait 1001.6 ms, 1.0 flushes/s, max lag 4 | p95 wait 1000.6 ms, 1.0 flushes/s, max lag 4 | p95 wait 201.4 ms, 7.4 flushes/s, max lag 4 | p95 -800.3 ms (-79.9%); flush cadence about 7.4x |

In this benchmark, `max lag 4` means the slow subscriber remains about four benchmark changes behind. The PR improves majority release timing without increasing that measured lag in these scenarios.

The 16-subscriber row starts from a lower main p95 than the 4- and 8-subscriber rows, so the absolute delta is smaller. The control still tells the same story: config-only 100 ms padding does not buy the intended win until the release mechanism moves out of the 1 second polling cadence.

- Cannot release before majority. The timer is not armed until `completed.length >= floor(subscriberCount / 2) + 1`.
- Cannot drop slow subscribers. It resolves the flow-control wait; subscribers still exist and continue processing.
- Cannot change the majority rule. The rule remains `floor(subscriberCount / 2) + 1`.
- Cannot cut the padding short after post-majority progress. The timer resets when another subscriber completes.
- Cannot leave stale release callbacks after everyone finishes. The timer is cleared when `Broadcast` resolves.
- Cannot remove the emergency disable path. Negative padding still disables early release.
- Cannot double-release if the progress monitor and direct timer race. Both paths go through the same idempotent `#setDone()` path.

- The default changes from 1s to 100ms, a 10x tightening. The mechanism and default could be split if reviewers want a more conservative rollout.
- Slow minority subscribers can lag more in production because the majority is no longer held hostage to the 1 second polling cadence. The benchmark did not show worse max lag, but production subscriber mix can be different.
- Releasing more often can increase upstream/send pressure in high-fanout rooms.
- Timer lifecycle bugs would be bad here, so the tests focus on arming, resetting, clearing, disabling, and all-subscriber completion.
- The original reason for the 1s default is not proven in this PR. The new 100ms value is configurable, so production signals can push it back without reverting the mechanism.

## Validation

- `npm --workspace=zero-cache run test:no-pg -- broadcast.test.ts zero-config.test.ts`

### DIFF
--- a/packages/zero-cache/src/config/zero-config.test.ts
+++ b/packages/zero-cache/src/config/zero-config.test.ts
@@ -406,11 +406,11 @@ test('zero-cache --help', () => {
                                                                         rather, it protects the system when the upstream throughput exceeds the downstream                                         
                                                                         throughput.                                                                                                                
                                                                                                                                                                                                    
-     --change-streamer-flow-control-consensus-padding-seconds number    default: 1                                                                                                                 
+     --change-streamer-flow-control-consensus-padding-seconds number    default: 0.1                                                                                                               
        ZERO_CHANGE_STREAMER_FLOW_CONTROL_CONSENSUS_PADDING_SECONDS env                                                                                                                             
-                                                                        During periodic flow control checks (every 64kb), the amount of time to wait after the                                     
-                                                                        majority of subscribers have acked, after which replication will continue even if                                          
-                                                                        some subscribers have yet to ack. (Note that this is not a timeout for the entire send,                                    
+                                                                        During flow control flushes, the amount of time to wait after the majority of                                              
+                                                                        subscribers have acked, after which replication will continue even if some subscribers                                     
+                                                                        have yet to ack. (Note that this is not a timeout for the entire send,                                                     
                                                                         but a timeout that starts after the majority of receivers have acked.)                                                     
                                                                                                                                                                                                    
                                                                         This allows a bounded amount of time for backlogged subscribers to catch up on each flush                                  
--- a/packages/zero-cache/src/config/zero-config.ts
+++ b/packages/zero-cache/src/config/zero-config.ts
@@ -651,11 +651,11 @@ export const zeroOptions = {
     },
 
     flowControlConsensusPaddingSeconds: {
-      type: v.number().default(1),
+      type: v.number().default(0.1),
       desc: [
-        `During periodic flow control checks (every 64kb), the amount of time to wait after the`,
-        `majority of subscribers have acked, after which replication will continue even if`,
-        `some subscribers have yet to ack. (Note that this is not a timeout for the {italic entire} send,`,
+        `During flow control flushes, the amount of time to wait after the majority of`,
+        `subscribers have acked, after which replication will continue even if some subscribers`,
+        `have yet to ack. (Note that this is not a timeout for the {italic entire} send,`,
         `but a timeout that starts {italic after} the majority of receivers have acked.)`,
         ``,
         `This allows a bounded amount of time for backlogged subscribers to catch up on each flush`,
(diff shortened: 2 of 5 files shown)


Karavil:
Closing this draft for now because the current 1 RM / 16 view-syncer e2e throughput benchmarks do not show a convincing production-shaped win. Keeping #5970 open and working backwards from the shared e2e scenarios to find changes that actually move throughput.

PR #5968, opened 2026-05-13
perf(zero-cache): batch change-log inserts

Karavil:

## Why This Is Worth It

When the replication manager is healthy but the view-syncer is behind, the worst on-call experience is watching the bridge burn time on bookkeeping instead of data movement. A large transaction currently asks Postgres to accept one `changeLog` row per message. That makes the storer pay per-row statement overhead even though the rows already belong to one ordered transaction.

This PR keeps the transaction semantics exactly where they are, but changes the storage shape from "one INSERT per change" to "small ordered INSERT batches".

```
Before
  RM tx: begin data data data ... commit
          |    |    |    |        |
          v    v    v    v        v
        INSERT INSERT INSERT ... INSERT   => 20,400 INSERT statements

After
  RM tx: begin data data data ... commit
          \_________________________/
                  batched rows          => 400 INSERT statements
```

## Clean PR Set

```
main
 |-- #5968 batch changeLog inserts            (this PR)
 |-- #5972 head-indexed queues
 |-- #5969 cache DML SQL plans
```

| PR | Role | Link |
| --- | --- | --- |

```text
1 replication manager
  -> ChangeProcessor
  -> Storer/Postgres changeLog
  -> Forwarder
```

Command shape:

```bash
ZERO_RM_VS_FULL=1 \
ZERO_RM_VS_DURATION_MS=1500 \
ZERO_RM_VS_SUBSCRIBERS=16 \
```

## This PR's Read

The healthy rows are the relevant e2e signal for this PR. With 1 RM and 16 view-syncer-side subscribers, batching improves the single-row flood by +12.7% and the medium batch by +3.5%. The large-row case is flat because the run is dominated by 8 KiB payload serialization and subscriber fanout, not Postgres statement count.

The slow/reconnect large-row cell is -7.4%. I am not using that as evidence that this PR helps that scenario. It is the row reviewers should scrutinize: if batching were actually causing a slowdown there, the symptom would be storer drain time growing or changeLog rows appearing out of order. The component benchmark below is the direct evidence that the write-amplification part of the path got smaller.

## Benchmark

Command:

```bash
npm --workspace=zero-cache run perf:storer-changelog
```

| --- | ---: | ---: | ---: |
| 200 tx x 100 data changes | 9,104 changes/s | 57,052 changes/s | 6.27x faster |
| Wall time | 2240.8 ms | 357.6 ms | -84.0% |
| changeLog INSERT statements | 20,400 | 400 | -98.04% |

## What Changed

- Buffer non-schema `changeLog` rows inside the existing storer transaction.
- Flush buffered rows in batches of 100, and always flush at commit.
- Flush any pending batch before schema metadata tracking so DDL-adjacent ordering stays explicit.
- Clear pending rows on rollback.
- Add a PG regression test proving a 250-row transaction persists contiguous `begin/data/commit` positions.
- Add `perf:storer-changelog` to compare one-row INSERTs with batched INSERTs.

## Why This Is Safe

The important invariant is not "one SQL statement per change"; it is "the changeLog has the same rows, in the same order, committed or rolled back with the same transaction." This keeps that invariant.

Worst case: a batch INSERT fails halfway through. Because the batch runs through the same transaction pool as the old single-row inserts, Postgres rolls back the transaction unit instead of exposing a partial `changeLog`. A rollback message clears the pending batch before aborting, so buffered rows cannot leak into the next transaction.

The scary ordering case is schema metadata. This PR flushes pending data rows before a schema-change row and its metadata statements, so the ordering remains:

```
data rows already seen
  -> flush batch
schema change row
  -> metadata writes
later data rows
```

## Expected Risks

- Bigger SQL statements can stress parameter limits if the batch is too large. The batch is capped at 100 rows.
- A bad batching bug would show up as missing, duplicated, or reordered `pos` values. The new PG test checks all positions in a 252-row transaction.
- This does not change the downstream protocol or subscriber behavior; the blast radius is the storer's write shape.

## Validation

```bash
npm --workspace=zero-cache run test:pg17 -- storer.pg.test.ts
npm --workspace=zero-cache run perf:storer-changelog
npx oxfmt --check packages/zero-cache/src/services/change-streamer/storer.ts packages/zero-cache/src/services/change-streamer/storer.pg.test.ts packages/zero-cache/bench/storer-changelog.ts packages/zero-cache/package.json
git diff --check
```

### DIFF
--- a/packages/zero-cache/bench/storer-changelog.ts
+++ b/packages/zero-cache/bench/storer-changelog.ts
@@ -1,0 +1,179 @@
+/* oxlint-disable no-console */
+import {performance} from 'node:perf_hooks';
+import {PostgreSqlContainer} from '@testcontainers/postgresql';
+import postgres from 'postgres';
+import {createSilentLogContext} from '../../shared/src/logging-test-utils.ts';
+import {
+  ensureReplicationConfig,
+  setupCDCTables,
+} from '../src/services/change-streamer/schema/tables.ts';
+import {postgresTypeConfig, type PostgresDB} from '../src/types/pg.ts';
+import {cdcSchema, type ShardID} from '../src/types/shards.ts';
+
+type Mode = 'single' | 'batched';
+
+type ChangeLogInsert = {
+  watermark: string;
+  precommit: string | null;
+  pos: number;
+  change: string;
+};
+
+type Result = {
+  readonly mode: Mode;
+  readonly elapsedMs: number;
+  readonly changes: number;
+  readonly changesPerSec: number;
+  readonly insertStatements: number;
+  readonly statementReductionPct: number;
+};
+
+type Summary = {
+  readonly name: 'storer-changelog-insert';
+  readonly generatedAt: string;
+  readonly txCount: number;
+  readonly changesPerTx: number;
+  readonly batchSize: number;
+  readonly results: Result[];
+};
+
+const lc = createSilentLogContext();
+const shard: ShardID = {appID: 'bench', shardNum: 0};
+const schema = cdcSchema(shard);
+
+const txCount = envInt('ZERO_STORER_CHANGELOG_TX', 200);
+const changesPerTx = envInt('ZERO_STORER_CHANGELOG_CHANGES_PER_TX', 100);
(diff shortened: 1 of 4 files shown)


Karavil:

PR #5969, opened 2026-05-13
perf(zero-cache): cache DML SQL plans

Karavil:

## Why This Is Worth It

The replicator applies many rows with the same DML shape: same table, same columns, different values. The hot path currently rebuilds SQL strings for each insert/update/delete even when the statement shape is identical to the last thousand rows.

This PR caches those DML SQL plans by table and column shape, then binds values by column name.

```
Before
  row 1 -> build SQL -> bind -> run
  row 2 -> build SQL -> bind -> run
  row 3 -> build SQL -> bind -> run

After
  row 1 -> build SQL -> cache plan -> bind -> run
  row 2 -> cache hit -------------> bind -> run
  row 3 -> cache hit -------------> bind -> run
```

## Clean PR Set

```
main
 |-- #5972 head-indexed queues
 |-- #5969 cache DML SQL plans                  (this PR)
```

| PR | Role | Link |
| --- | --- | --- |

```text
1 replication manager
  -> ChangeProcessor
  -> Storer/Postgres changeLog
  -> Forwarder
```

Command shape:

```bash
ZERO_RM_VS_FULL=1 \
ZERO_RM_VS_DURATION_MS=1500 \
ZERO_RM_VS_SUBSCRIBERS=16 \
```

## This PR's Read

The e2e matrix says this change is mostly neutral in the full RM -> VS bridge, with a small healthy-medium gain (+3.5%) and no large-row movement. That is expected: this PR removes repeated SQL string construction inside `ChangeProcessor`, but the mixed e2e harness also includes Postgres changeLog writes, forwarding, subscriber ACKs, and reconnect timing.

The sharper question for this PR is not "does the whole bridge get faster every time?" It is "can repeated-shape replica writes reuse SQL without changing bindings?" The component benchmark below isolates that path and shows the stable repeated-shape case dropping 30.2%. The safety section then explains why a cache hit cannot bind the wrong values.

## Benchmark

Command:

```bash
npm --workspace=zero-cache run bench -- change-processor.bench.ts
```

| Scenario | `main` / uncached | This PR / cached | Delta |
| --- | ---: | ---: | ---: |
| 1k stable-shape inserts | 3.05 ms/iter | 2.13 ms/iter | -30.2% |
| 1k inserts + updates + deletes | 7.74 ms/iter | 5.81 ms/iter | -24.9% |

## What Changed

- Add a small `DmlSqlPlanCache` for insert/upsert, update, and delete statement shapes.
- Cache by table plus column-shape key.
- Keep a "last plan per table" fast path for the common repeated-shape case.
- Bind values by the cached column list, not by current object property order.
- Clear the DML plan cache whenever table specs reload.
- Add a regression test that reorders object properties between repeated DML calls.
- Add a `change-processor.bench.ts` benchmark comparing cached and uncached processors.

## Why This Is Safe

The dangerous failure mode is not "cache missed"; it is "cache hit with the wrong bindings." This PR avoids that by separating SQL shape from value lookup. The cached plan owns the column order, and value arrays are built by reading those column names from the row/key objects.

Worst case: schema changes while a cached plan still exists. The cache is cleared on table-spec reload, so the next DML call rebuilds the SQL plan against the new table shape. Cache keys are length-prefixed (`table`, `column count`, `column names`) to avoid accidental collisions.

Object property order is explicitly tested: the same table gets rows with `{id, shard, alpha, beta}` and `{beta, shard, id, alpha}`. The cached plan still binds the right values to the right columns.

## Expected Risks

- This touches the replicator write path, so correctness beats the 25-30% benchmark win.
- If a rare DML shape appears once, the cache adds minor map work. The common case is repeated shapes in a transaction.
- Schema reload invalidation is the key review point.

## Validation

```bash
npm --workspace=zero-cache run test:no-pg -- change-processor.test.ts
npm --workspace=zero-cache run bench -- change-processor.bench.ts
npx oxfmt --check packages/zero-cache/src/services/replicator/change-processor.ts packages/zero-cache/src/services/replicator/change-processor.test.ts packages/zero-cache/src/services/replicator/change-processor.bench.ts
git diff --check
```

### DIFF
--- a/packages/zero-cache/src/services/replicator/change-processor.bench.ts
+++ b/packages/zero-cache/src/services/replicator/change-processor.bench.ts
@@ -1,0 +1,136 @@
+import {bench, describe, use} from '../../../../shared/src/bench.ts';
+import {createSilentLogContext} from '../../../../shared/src/logging-test-utils.ts';
+import {Database} from '../../../../zqlite/src/db.ts';
+import {StatementRunner} from '../../db/statements.ts';
+import type {DataOrSchemaChange} from '../change-source/protocol/current/data.ts';
+import type {ChangeStreamData} from '../change-source/protocol/current/downstream.ts';
+import {ChangeProcessor} from './change-processor.ts';
+import {initReplicationState} from './schema/replication-state.ts';
+import {ReplicationMessages} from './test-utils.ts';
+
+const ROWS = 1_000;
+const lc = createSilentLogContext();
+const messages = new ReplicationMessages({foo: ['id', 'shard']});
+
+describe('change-processor DML apply', () => {
+  const stableInsertChanges = transaction(
+    '10',
+    Array.from({length: ROWS}, (_, i) =>
+      messages.insert('foo', {
+        id: i,
+        shard: i % 10,
+        alpha: i * 2,
+        beta: i * 3,
+      }),
+    ),
+  );
+  const stableInsertUncached = setupBenchmark(false);
+  const stableInsertCached = setupBenchmark(true);
+
+  bench(
+    '1k stable-shape inserts (uncached plans)',
+    () => {
+      use(apply(stableInsertUncached, stableInsertChanges));
+    },
+    {min_cpu_time: 5e8, min_samples: 20},
+  );
+
+  bench(
+    '1k stable-shape inserts (cached plans)',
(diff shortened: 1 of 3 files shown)


Karavil:

PR #5970, opened 2026-05-13
perf(zero-cache): flow-control catchup backlog

Karavil:
```text
RM
`- ChangeStreamer
   |- Storer      : replays changeDB entries after watermark W
   |- Forwarder   : receives new live commits
   `- Subscriber  : bridges replay and live delivery for each VS
                     |
                     v
                    VS (view-syncer / replicator)
                     |
                     v
              browser / app clients
```

**On reconnect, the VS sends watermark `W`.** `Storer` replays historical `changeDB` entries after `W`. While that replay is running, `Forwarder` can still receive new commits. `Subscriber` is the bridge between those two phases: it has to preserve order across the replay/live boundary, and it is the component that decides when a VS is actually caught up enough to resume serving downstream clients.

**This PR makes "caught up" truthful under load; it is not a row-processing speedup.** The old behavior was plausible when the reconnect backlog was small. During catchup, live `send()` calls went into `Subscriber`'s private backlog and returned. When replay finished, `setCaughtUp()` dumped that backlog into the downstream `Subscription` pending queue without awaiting each send, then reported caught up. For a short backlog, reported catchup and actual catchup were close enough that this looked fine.

**That same handoff becomes dangerous when reconnect overlaps real write load.** If the RM is still receiving `1000 tx/s` while a VS spends `10s` catching up, and each transaction fans out to 3 downstream messages, each VS accumulates `30,000` buffered live messages. With 16 VSs reconnecting, that is `480,000` buffered live messages. In the old path, the RM could report "caught up" after enqueueing those `480,000` messages, not after the VS had consumed them. So the system looked done while it still had seconds of hidden drain work queued below it.

```text
Old handoff

Storer replay ---> Subscriber private backlog
                              |
                              `-- setCaughtUp()
                                  |- push buffered live changes into downstream pending queue
                                  `- report "caught up" immediately

New handoff

Storer replay ---> CatchupBacklog
                       |
                       `-- flushWith(#sendChange) in order
                           |- enqueue(change) -> receipt
                           |- receipt resolves after #sendChange finishes
                           `- only then report "caught up"
```

**The new behavior restores backpressure instead of moving it off to the side.** `CatchupBacklog` owns the queue mechanics. `enqueue()` returns a receipt for each buffered live change. `flushWith()` drains in order through the normal downstream send path, keeps including entries that arrive while the flush is already running, and rejects pending receipts if downstream delivery fails. `Subscriber` no longer threads resolver arrays directly. `setCaughtUp()` waits for the flush, and `sendStatus` does not report caught up until the last buffered live change has actually finished `#sendChange`. The storer read transaction is already closed before this drain runs, so we do not hold a replay read transaction open while waiting for downstream consumption.

**The benchmark to focus on is `16-vs-10s-1000tps`, because it captures the failure mode directly.** The harness models 16 VSs spending 10 seconds catching up while the RM keeps taking `1000 tx/s`. It does not literally sleep for 10 seconds; instead it derives the live backlog as `1000 tx/s * 10s * 3 downstream messages/tx = 30,000 messages per VS`, then adds the 10-second catchup window to both reported and actual catchup times. That gives `480,000` buffered live messages across the 16 VSs. In the table below, `report ms` is when the RM says the VS is caught up, and `actual ms` is when the buffered live backlog has actually finished sending.

**In that scenario, the old path reports about 522 ms earlier, but that earlier report is false.** The VS still has almost `480,000` downstream messages left to drain. The new path reports later because it waits for real completion, but it reaches actual catchup about `5.38s` sooner because it never creates the giant downstream pending queue in the first place.

| mode | reported catchup | actual catchup | false caught-up window | pending at report | max aggregate pending |
| --- | ---: | ---: | ---: | ---: | ---: |
| fire-and-forget handoff | 11,209.1 ms | 17,115.4 ms | 5,906.3 ms | 479,984 | 480,000 |
| flow-controlled handoff | 11,730.9 ms | 11,730.9 ms | 0.0 ms | 0 | 16 |

**The broader matrix tells the same story.** This change is not about making individual rows cheaper. It is about making completion mean consumed, bounding downstream pending work, preserving backpressure during reconnect, and cutting off the tail-risk that shows up as huge pending queues, memory bursts, and GC churn.

```text
scenario          | subs |   tx/s | catchup |  msgs/sub |  payload | mode                    |  report ms |  actual ms |  false ms |        msg/s | pending@report | max agg pending |   max/sub
small-reconnect   |    1 |      - |    0.0s |    10,000 |    192 B | fire-and-forget handoff |        7.2 |       33.9 |       26.6 |       295399 |          9,999 |          10,000 |    10,000
small-reconnect   |    1 |      - |    0.0s |    10,000 |    192 B | flow-controlled handoff |       21.6 |       21.6 |        0.0 |       463915 |              0 |               1 |         1
baseline-100k     |    1 |      - |    0.0s |   100,000 |    192 B | fire-and-forget handoff |      110.6 |     3748.4 |     3637.8 |        26678 |         99,999 |         100,000 |   100,000
baseline-100k     |    1 |      - |    0.0s |   100,000 |    192 B | flow-controlled handoff |      248.8 |      248.8 |        0.0 |       401901 |              0 |               1 |         1
large-row-burst   |    1 |      - |    0.0s |    25,000 |    4 KiB | fire-and-forget handoff |      121.6 |      468.6 |      347.0 |        53350 |         24,999 |          25,000 |    25,000
large-row-burst   |    1 |      - |    0.0s |    25,000 |    4 KiB | flow-controlled handoff |      185.8 |      185.8 |        0.0 |       134517 |              0 |               1 |         1
slow-downstream   |    1 |      - |    0.0s |    20,000 |    192 B | fire-and-forget handoff |       17.3 |      623.2 |      606.0 |        32090 |         19,999 |          20,000 |    20,000
slow-downstream   |    1 |      - |    0.0s |    20,000 |    192 B | flow-controlled handoff |      629.5 |      629.6 |        0.0 |        31768 |              0 |               1 |         1
16-vs-reconnect   |   16 |      - |    0.0s |    10,000 |    192 B | fire-and-forget handoff |      144.0 |      384.9 |      240.9 |       415646 |        159,984 |         160,000 |    10,000
16-vs-reconnect   |   16 |      - |    0.0s |    10,000 |    192 B | flow-controlled handoff |      329.2 |      329.2 |        0.0 |       486065 |              0 |              16 |         1
16-vs-outage-load |   16 |      - |    0.0s |    25,000 |    512 B | fire-and-forget handoff |      627.6 |     4604.9 |     3977.3 |        86863 |        399,984 |         400,000 |    25,000
16-vs-outage-load |   16 |      - |    0.0s |    25,000 |    512 B | flow-controlled handoff |     1384.3 |     1384.3 |        0.0 |       288958 |              0 |              16 |         1
16-vs-10s-1000tps |   16 |  1,000 |   10.0s |    30,000 |    512 B | fire-and-forget handoff |    11209.1 |    17115.4 |     5906.3 |        67459 |        479,984 |         480,000 |    30,000
16-vs-10s-1000tps |   16 |  1,000 |   10.0s |    30,000 |    512 B | flow-controlled handoff |    11730.9 |    11730.9 |        0.0 |       277316 |              0 |              16 |         1
```

**The memory probe is useful because it shows where the danger moved.** Both versions still need the private backlog while catchup is in progress. The difference is the handoff. In `16-vs-10s-1000tps`, the old path then materializes another `480,000` downstream queued entries and about `95.7 MB` of extra heap (`~209` extra bytes per queued entry). The new path hands off at most one downstream queued entry per VS (`16` total) and adds effectively `0.0 MB` of extra heap.

```text
scenario: 16-vs-10s-1000tps
backlog entries: 480,000
private backlog heap: 245.4 MB
old handoff queued entries: 480,000
old handoff extra heap: 95.7 MB
new handoff queued entries: 16
new handoff extra heap: 0.0 MB
old extra bytes / queued entry: 209
```

**The code shape is intentionally small and explicit.** `CatchupBacklog` owns queue behavior in one place: `enqueue()` returns receipts, `flushWith()` drains in order and keeps pulling entries that arrive during the drain, and downstream failures reject any unresolved receipts. That leaves `Subscriber` simpler: it no longer manages parallel resolver arrays or pretends a backlog is done once it has been handed to someone else.

**Validation ran on the focused surfaces plus the new perf harness.**

```bash
npm --workspace=zero-cache run lint -- ./bench/catchup-backlog ./src/services/change-streamer/catchup-backlog.ts ./src/services/change-streamer/catchup-backlog.test.ts ./src/services/change-streamer/subscriber.ts
npm --workspace=zero-cache run test:no-pg -- subscriber.test.ts catchup-backlog.test.ts
npm --workspace=zero-cache run check-types -- --pretty false
npm --workspace=zero-cache run perf:catchup-backlog
npm --workspace=zero-cache run perf:catchup-backlog:memory
```

**This PR is intentionally independent in the stack.** It is the correctness/backpressure piece.

```text
main
 |
 |-- #5970 catchup backlog handoff (this PR)
 |
 `-- #5976 https://github.com/rocicorp/mono/pull/5976
       |
       `-- #5977 https://github.com/rocicorp/mono/pull/5977
             |
             `-- #5978 https://github.com/rocicorp/mono/pull/5978
```

_AI-generated draft PR body. The PR is intentionally draft while reviewers check the benchmark interpretation; the code and benchmark claims above were validated against this branch._


arv:
Superceeded by #6186

@tantaman

tantaman:
hah, didn't realize this was in the queue.

PR #5971, opened 2026-05-13
perf(zero-cache): suppress no-op CVR row writes

Karavil:

## Why This Is Worth It

The view-syncer often sees rows whose effective CVR row record did not change. Today those rows can still enter the pending persistence map, only to be thrown away later during flush after another deep equality check. That is extra map churn and flush work on the path where the view-syncer is already trying to catch up.

This PR moves that no-op decision earlier.

```
Before
  row update -> build row record -> pendingRowRecordUpdates[100k]
             -> flush -> compare with existing -> drop unchanged row

After
  row update -> build row record -> compare with existing
             -> unchanged? clear pending update immediately
```

The user-facing patch path still runs. This only suppresses a persistence write that would have been dropped at flush time.

## Clean PR Set

```
main
 |-- #5972 head-indexed queues
 |-- #5971 suppress no-op CVR row writes        (this PR)
 |-- #5969 cache DML SQL plans
```

| PR | Role | Link |
| --- | --- | --- |

```text
1 replication manager
  -> ChangeProcessor
  -> Storer/Postgres changeLog
  -> Forwarder
```

Command shape:

```bash
ZERO_RM_VS_FULL=1 \
ZERO_RM_VS_DURATION_MS=1500 \
ZERO_RM_VS_SUBSCRIBERS=16 \
```

## This PR's Read

The shared e2e matrix is a regression check for this PR, not the main evidence of improvement. `perf:rm-vs-load:full` exercises the RM -> storer -> forwarder -> subscriber bridge; this change lives one layer deeper in view-syncer CVR persistence after rows have already reached the view-syncer.

That is why the e2e cells are basically flat/noisy. The useful claim is narrower: when a view-syncer reconstructs a CVR row record and it deep-equals the existing one, we can avoid enqueueing a persistence write that flush would later drop anyway. The component benchmark below isolates that exact no-op CVR case.

## Benchmark

Command:

```bash
npm --workspace=zero-cache run perf:cvr-noop
```

| --- | ---: | ---: | ---: |
| 100k unchanged CVR row records | 1,156,039 rows/s / 86.5 ms | 5,081,796 rows/s / 19.7 ms | 4.40x faster |
| Rows queued for persistence | 100,000 | 0 | removed |
| Rows cleared as no-op | 0 | 100,000 | explicit early suppression |

## What Changed

- Add `CVRStore.clearRowRecordUpdate()` for cancelling a pending row-record write.
- In `CVRQueryDrivenUpdater.received()`, compare the reconstructed row record with the existing CVR row record.
- If the row record is structurally identical, clear the pending persistence update instead of enqueueing it.
- Preserve normal persistence for changed rows, deleted rows, newly referenced rows, and forced updates.
- Add tests for unchanged rows, changed rows, deleted rows, and changed-then-restored rows.
- Add `perf:cvr-noop` to measure the old enqueue-then-drop shape versus early suppression.

## Why This Is Safe

The safety boundary is strict: only rows that deep-equal the existing CVR row are suppressed. If any field changes, the normal persistence path runs.

Worst case: a row changes and then changes back in the same advancement. That is exactly the scenario that can trick optimistic no-op logic. The test suite covers it. The changed leg still persists normally; the restored leg is suppressed only once the reconstructed row record is identical to what the CVR already has.

Importantly, this does not suppress row patches. The view-syncer can still emit the client-visible patch stream. The optimization only avoids writing a CVR row record that the flush path would later recognize as unchanged and drop.

## Expected Risks

- Moving the equality check earlier could hide a forced write if force state is not cleared with the pending update. `clearRowRecordUpdate()` clears both the pending row update and the force marker.
- Deep equality still has a cost, but the old flush path already performed the same kind of comparison after queueing. This PR avoids the map churn and later flush work.
- The change is scoped to CVR row-record persistence; query/client/desire config flushes are untouched.

## Validation

```bash
npm --workspace=zero-cache run test:no-pg -- cvr.test.ts
npm --workspace=zero-cache run perf:cvr-noop
npx oxfmt --check packages/zero-cache/src/services/view-syncer/cvr.ts packages/zero-cache/src/services/view-syncer/cvr-store.ts packages/zero-cache/src/services/view-syncer/cvr.test.ts packages/zero-cache/bench/cvr-noop.ts packages/zero-cache/package.json
git diff --check
```

### DIFF
--- a/packages/zero-cache/bench/cvr-noop.ts
+++ b/packages/zero-cache/bench/cvr-noop.ts
@@ -1,0 +1,158 @@
+/* oxlint-disable no-console */
+import {performance} from 'node:perf_hooks';
+import {deepEqual} from '../../shared/src/json.ts';
+import type {ReadonlyJSONValue} from '../../shared/src/json.ts';
+import type {
+  RowID,
+  RowRecord,
+} from '../src/services/view-syncer/schema/types.ts';
+
+type Result = {
+  readonly mode: string;
+  readonly rows: number;
+  readonly queuedRows: number;
+  readonly clearedRows: number;
+  readonly elapsedMs: number;
+  readonly rowsPerSec: number;
+};
+
+type Summary = {
+  readonly name: 'cvr-noop-row-suppression';
+  readonly generatedAt: string;
+  readonly rows: number;
+  readonly results: Result[];
+};
+
+const rows = envInt('ZERO_CVR_NOOP_ROWS', 100_000);
+const existing = Array.from({length: rows}, (_, i) => makeRowRecord(i));
+const candidates = existing.map(row => ({...row}));
+
+const results = [
+  runBaseline(existing, candidates),
+  runEarlySuppression(existing, candidates),
+];
+
+for (const result of results) {
+  console.log(
+    `${result.mode}: ${formatRate(result.rowsPerSec)} rows/s | ` +
+      `${result.elapsedMs.toFixed(1)} ms | queued=${result.queuedRows} | ` +
+      `cleared=${result.clearedRows}`,
+  );
+}
+
+const summary: Summary = {
+  name: 'cvr-noop-row-suppression',
+  generatedAt: new Date().toISOString(),
+  rows,
+  results,
+};
+console.log(JSON.stringify(summary));
+
+function runBaseline(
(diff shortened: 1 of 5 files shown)


Karavil:

PR #5972, opened 2026-05-13
perf(zero-cache): use head-indexed queues

Karavil:

## Why This Is Worth It

When the view-syncer falls behind, the worst thing a queue can do is make every dequeue copy the rest of the backlog. That is what `Array.shift()` does. The larger the backlog gets, the more each "remove one item" operation becomes an accidental array compaction job.

This PR changes the hot FIFO internals in `Queue` and `Subscription` to use a head index. Dequeue becomes "read slot, advance pointer"; occasional compaction happens only after enough head slots are dead.

```
Before: shift()
  [a b c d e f g] -> pop a -> [b c d e f g]
                         copies the tail every time

After: head index
  [a b c d e f g], head=0
  [_ b c d e f g], head=1
  [_ _ c d e f g], head=2
       compact later, not on every pop
```

## Clean PR Set

```
main
 |-- #5972 head-indexed queues                  (this PR)
 |-- #5969 cache DML SQL plans
```

| PR | Role | Link |
| --- | --- | --- |

```text
1 replication manager
  -> ChangeProcessor
  -> Storer/Postgres changeLog
  -> Forwarder
```

Command shape:

```bash
ZERO_RM_VS_FULL=1 \
ZERO_RM_VS_DURATION_MS=1500 \
ZERO_RM_VS_SUBSCRIBERS=16 \
```

## This PR's Read

The e2e suite gives this PR a sanity check under real-ish RM/VS pressure: healthy `single-row-flood` is +7.2%, and slow/reconnect `single-row-flood` is +2.5%. The larger rows are basically flat because queue mechanics are not the dominant cost once payload serialization and subscriber pacing take over.

The worst case this PR is trying to prevent is not visible as a tiny per-row improvement. It is the pathological backlog shape where every dequeue copies the remaining array. The component benchmark below creates that backlog directly and shows why the data-structure change is worth reviewing even when the mixed e2e rows are modest.

## Benchmark

Command:

```bash
npm --workspace=zero-cache run perf:queue
```

| --- | ---: | ---: | ---: |
| `Queue` FIFO drain | 77,258 ops/s / 2588.7 ms | 12,494,469 ops/s / 16.0 ms | 161.7x faster |
| `Subscription` queued drain | 78,231 ops/s / 2556.5 ms | 2,919,381 ops/s / 68.5 ms | 37.3x faster |

## What Changed

- Replace `Array.shift()` in `shared/Queue` with a head-indexed internal array.
- Replace `Array.shift()` in `zero-cache` `Subscription` with the same bounded FIFO shape.
- Preserve the public APIs and cleanup semantics.
- Compact consumed head slots when the dead prefix is large enough.
- Add tests for timeout cleanup, delete-after-dequeue, cancellation, `end()`, and pipeline cleanup.
- Add `perf:queue` to compare old shift-style FIFO drains against the new shape.

## Why This Is Safe

The behavior is still FIFO. The representation changed from "the next item is always at index 0" to "the next item is at `head`." That is a mechanical data-structure change, not a protocol or lifecycle change.

Worst case: the head index advances forever and keeps references alive. The implementation resets the array when it empties and compacts when the dead prefix is both larger than 1024 and more than half the array. That bounds stale slots without making every dequeue pay for compaction.

The cases most likely to break are the non-happy paths: timeout consumers, deletes from the queued region, cancellation while messages are in flight, and `end()` sent after head advancement. Those are covered by focused tests.

## Expected Risks

- Off-by-one errors around compaction are the main risk. The tests exercise empty queues, sparse deleted slots, and active consumers.
- A consumer cleanup regression would be worse than a throughput regression. Existing cleanup paths still receive the unconsumed entries; tests cover cancel/end behavior.
- This PR touches shared infrastructure, so review should focus on semantics first and benchmark numbers second.

## Validation

```bash
npx vitest run packages/shared/src/queue.test.ts packages/zero-cache/src/types/subscription.test.ts
npm --workspace=zero-cache run perf:queue
npx oxfmt --check packages/shared/src/queue.ts packages/shared/src/queue.test.ts packages/zero-cache/src/types/subscription.ts packages/zero-cache/src/types/subscription.test.ts packages/zero-cache/bench/queue-overhead.ts packages/zero-cache/package.json
git diff --check
```

### DIFF
--- a/packages/shared/src/queue.test.ts
+++ b/packages/shared/src/queue.test.ts
@@ -219,4 +219,52 @@ describe('Queue', () => {
     queue.enqueue(3);
     await promise;
   });
+
+  test('dequeue and drain keep FIFO order after many dequeues', async () => {
+    const queue = new Queue<number>();
+    for (let i = 0; i < 2048; i++) {
+      queue.enqueue(i);
+    }
+
+    for (let i = 0; i < 1024; i++) {
+      expect(await queue.dequeue()).toBe(i);
+    }
+    queue.enqueue(2048);
+    queue.enqueue(2049);
+
+    expect(queue.size()).toBe(1026);
+    expect(queue.drain()).toEqual(
+      Array.from({length: 1026}, (_, i) => i + 1024),
+    );
+    expect(queue.size()).toBe(0);
+  });
+
+  test('delete and drain ignore already dequeued head entries', async () => {
+    const queue = new Queue<string>();
+    queue.enqueue('a');
+    queue.enqueue('b');
+    queue.enqueue('c');
+    queue.enqueue('d');
+
+    expect(await queue.dequeue()).toBe('a');
+    expect(queue.delete('c')).toBe(1);
+
+    expect(queue.size()).toBe(2);
+    expect(queue.drain()).toEqual(['b', 'd']);
+    expect(queue.size()).toBe(0);
+  });
+
+  test('timed out consumers do not block later enqueues', async () => {
+    const queue = new Queue<string>();
+    const timedOut = Promise.all(
+      Array.from({length: 8}, () => queue.dequeue('timed out', 5)),
+    );
+
+    expect(await timedOut).toEqual(Array.from({length: 8}).fill('timed out'));
+
+    queue.enqueue('live');
+    expect(queue.size()).toBe(1);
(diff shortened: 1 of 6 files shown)


Karavil:

PR #5985, opened 2026-05-14
fix(zero): retry on OoOMutation

tantaman:
https://rocicorp.slack.com/archives/C08AHHFQMMY/p1778251833425319

### DIFF
--- a/packages/zero-protocol/src/error.ts
+++ b/packages/zero-protocol/src/error.ts
@@ -13,7 +13,6 @@ const basicErrorKindSchema = v.literalUnion(
   ErrorKind.InvalidConnectionRequestLastMutationID,
   ErrorKind.InvalidConnectionRequestClientDeleted,
   ErrorKind.InvalidMessage,
-  ErrorKind.InvalidPush,
   ErrorKind.MutationRateLimited,
   ErrorKind.MutationFailed,
   ErrorKind.Unauthorized,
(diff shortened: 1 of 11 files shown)


aboodman:
This seems like it will change the behavior of other error cases we don't want to change.

And in general I understand now why this doesn't retry - typically if a client sends a mutation out of order it's a protocol failure and will not recover just by trying harder. This is an odd case where it will.

Hrm.

tantaman:
> And in general I understand now why this doesn't retry

Like why it doesn't retry today? Because it is classified as a terminal error.

> typically if a client sends a mutation out of order it's a protocol failure and will not recover just by trying harder

I think out of order mutation can happen in a number of cases, no? Like failover between view-syncers.

tantaman:
> This seems like it will change the behavior of other error cases we don't want to change.



That's valid. I'll add a reason to only retry OOO

tantaman:
Updated to only retry on OoO Mutations by adding a `reason` which is wire compatible with old protocol versions.

IMO OoO mutations are expected when:
- client is fighting between two view syncers (we've seen this in production before in a different context: "[cvr concurrently modified](https://rocicorp.slack.com/archives/C013XFG80JC/p1778189991508839?thread_ts=1778182430.166939&cid=C013XFG80JC)")
- view syncer is re-deployed so client re-connects to a new view syncer with mutations that were applied but not yet acked.

packages/zero-protocol/src/error.ts
```diff
@@ -29,6 +28,19 @@ const basicErrorBodySchema = v.object({
   origin: v.literalUnion(ErrorOrigin.Server, ErrorOrigin.ZeroCache).optional(),
 });
 
+const invalidPushBodySchema = v.object({
```
0xcadams, line 31: We don't need to expand the protocol for this. It already exists in `PushFailed` w/ reason: `OutOfOrderMutation`

packages/zero-protocol/src/protocol-version.ts
```diff
@@ -57,7 +57,8 @@ import {assert} from '../../shared/src/asserts.ts';
 // -- version 49 adds `scalar` to CorrelatedSubqueryCondition, removes scalarSubquery
 // -- version 50 adds OTEL headers to push and query messages
 // -- version 51 changes inspector metrics fields
-export const PROTOCOL_VERSION = 51;
+// -- version 52 adds optional `reason` field to InvalidPush error body
+export const PROTOCOL_VERSION = 52;
```
0xcadams, line 61: We shouldn't need a version bump

0xcadams (commented):
Looks like some of the error handling is inconsistent here. I think I caused that. Server handles the OOO mutation w/ PushFailed, so we probably want to align those.

PR #5986, opened 2026-05-14
perf(shared): O(n²) → O(1) amortized Queue.dequeue()

grgbkr:
### Problem

CPU profiling of the replication manager under heavy transaction load revealed that **`Queue.dequeue()` consumed up to 85% of CPU time** (19–52s out of 61s profiles). The root cause: `Queue` uses an array as a FIFO, calling `Array.shift()` to dequeue.

V8 optimizes `Array.shift()` to O(1) via backing-store pointer adjustment for arrays under ~10k elements, but falls back to O(n) element copying above that threshold. Under heavy replication load, the storer queue routinely grows to tens or hundreds of thousands of changes, well past this threshold, causing **O(n²) total drain cost**.

This directly impacts throughput in the `Storer.#processQueue()` loop, which `await`s `dequeue()` on every change in the replication stream.

### Fix

Introduce a `RingBuffer<T>` class — a circular array with O(1) `push`/`shift` — and refactor `Queue` to use it as its backing store.

**`RingBuffer` (`ring-buffer.ts`):**
- Circular array with `#head` pointer and `#size` counter
- `push()` — O(1) amortized (doubles capacity when full, in-place)
- `shift()` — O(1) always (advance head pointer, no copying)
- In-place `#grow()` — extends array length, relocates only wrapped-around elements
- In-place `#shrink()` — moves elements down and truncates when utilization < 25%
- `drain()` / `delete()` for bulk operations

**`Queue` (`queue.ts`):**
- Replaced manual array + cursor management with `RingBuffer<Consumer>` and `RingBuffer<Produced>`
- Simplified timeout handler (uses `RingBuffer.delete()` by identity instead of `indexOf` + `find` + `splice`)
- No API changes — drop-in replacement

### Performance Results

Profiled processing the same transaction on CloudZero, with and without the fix:

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| **Queue.dequeue() CPU** | 19.9s (31%) | 0.03s | **Eliminated** |
| **Active CPU time** | 62.8s | 33.4s | **−47%** |
| **Idle headroom** | 1.0s (1.5%) | 28.5s (46%) | **System no longer CPU-bound** |
| writev (WebSocket sends) | 8.0s | 2.7s | −66% |
| GC pressure | 6.0s | 3.6s | −39% |
| JSON stringify | 2.2s | 0.7s | −70% |
| Promise/async overhead | 3.3s | 3.0s | −9% |

The CPU went from **98.5% saturated** to **54% utilized** under the same workload. The downstream improvements (writev, GC, stringify) are a cascading effect — without the dequeue bottleneck, messages flow through without backlog, reducing contention and memory pressure across the entire pipeline.

Ring buffer microbenchmark (100k items push + shift):
- `Array.shift()`: ~1150ms
- Allocating resize: ~12ms
- In-place ring buffer: ~1ms

### Changes

- **`packages/shared/src/ring-buffer.ts`** [NEW] — Generic FIFO ring buffer with O(1) push/shift, in-place grow/shrink
- **`packages/shared/src/ring-buffer.test.ts`** [NEW] — 10 tests: FIFO ordering, growth, wrap-around, drain, delete, interleaved ops, performance regression
- **`packages/shared/src/queue.ts`** — Refactored to delegate to `RingBuffer` instead of managing arrays directly
- **`packages/shared/src/queue.test.ts`** — Added performance regression test (100k dequeue in <200ms)

[control1.cpuprofile](https://github.com/user-attachments/files/27816665/control1.cpuprofile)
[test1.cpuprofile](https://github.com/user-attachments/files/27816666/test1.cpuprofile)

yellow is test with this change
green is control
image

### DIFF
--- a/packages/shared/src/queue.ts
+++ b/packages/shared/src/queue.ts
@@ -1,5 +1,6 @@
 import {resolver, type Resolver} from '@rocicorp/resolver';
 import {assert} from './asserts.ts';
+import {RingBuffer} from './ring-buffer.ts';
 
 /**
  * A Queue allows the consumers to await (possibly future) values,
@@ -7,9 +8,9 @@ import {assert} from './asserts.ts';
  */
 export class Queue<T> {
   // Consumers waiting for entries to be produced.
-  readonly #consumers: Consumer<T>[] = [];
+  readonly #consumers = new RingBuffer<Consumer<T>>();
   // Produced entries waiting to be consumed.
-  readonly #produced: Produced<T>[] = [];
+  readonly #produced = new RingBuffer<Produced<T>>();
 
   enqueue(value: T): void {
     const consumer = this.#consumers.shift();
@@ -43,12 +44,14 @@ export class Queue<T> {
   delete(value: T): number {
     assert(value !== undefined, 'Queue delete value must not be undefined');
 
+    // Drain, filter, and re-push. This is O(n) but delete() is rare.
+    const items = this.#produced.drain();
     let count = 0;
-    for (let i = this.#produced.length - 1; i >= 0; i--) {
-      const p = this.#produced[i];
+    for (const p of items) {
       if (p.value === value) {
-        this.#produced.splice(i, 1);
         count++;
+      } else {
+        this.#produced.push(p);
       }
     }
     return count;
@@ -66,17 +69,17 @@ export class Queue<T> {
       return produced.value ?? Promise.reject(produced.rejection);
     }
     const r = resolver<T>();
+    const consumer: Consumer<T> = {resolver: r, timeoutID: undefined};
     const timeoutID =
       timeoutValue === undefined
         ? undefined
         : setTimeout(() => {
-            const i = this.#consumers.findIndex(c => c.resolver === r);
-            if (i >= 0) {
-              const [consumer] = this.#consumers.splice(i, 1);
+            if (this.#consumers.delete(consumer) > 0) {
               consumer.resolver.resolve(timeoutValue);
             }
           }, timeoutMs);
-    this.#consumers.push({resolver: r, timeoutID});
+    consumer.timeoutID = timeoutID;
+    this.#consumers.push(consumer);
     return r.promise;
   }
 
@@ -95,13 +98,7 @@ export class Queue<T> {
    * ```
    */
   drain(): (T | undefined)[] {
-    const ret: (T | undefined)[] = [];
-    for (const p of this.#produced) {
-      ret.push(p.value);
-    }
-    this.#produced.length = 0;
-
-    return ret;
+    return this.#produced.drain().map(p => p.value);
   }
 
   /**
@@ -111,7 +108,7 @@ export class Queue<T> {
    *          handed to the consumer and the Queue's size remains 0.
    */
   size(): number {
-    return this.#produced.length;
+    return this.#produced.size;
   }
 
   asAsyncIterable(cleanup = NOOP): AsyncIterable<T> {
--- a/packages/shared/src/ring-buffer.ts
+++ b/packages/shared/src/ring-buffer.ts
@@ -1,0 +1,143 @@
+const DEFAULT_INITIAL_CAPACITY = 16;
+const MIN_CAPACITY = 16;
+
+/**
+ * A FIFO ring buffer backed by a circular array. Provides:
+ *
+ * - `push()` — O(1) amortized (O(n) on resize)
+ * - `shift()` — O(1)
+ * - `size`    — O(1)
+ * - `drain()` — O(n)
+ *
+ * The buffer grows (doubles) when full and shrinks (halves) when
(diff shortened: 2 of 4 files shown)


arv:
I think another optimization would be to use a cyclic queue. I still think amortized shrinking is wanted though but it would reduce the memory usage

arv:
TIL: v8 optimized shift for small arrays 🤯

packages/shared/src/ring-buffer.ts
```diff
@@ -0,0 +1,116 @@
...
+    this.#buffer = Array.from({length: this.#capacity});
+    return result;
+  }
+
+  /**
+   * Copies elements into a new buffer of the given capacity,
+   * linearizing the circular layout.
+   */
+  #resize(newCapacity: number): void {
+    const newBuffer: (T | undefined)[] = Array.from({length: newCapacity});
```
darkgnotic: This version reduces temporary memory usage and copies fewer elements.
(It grows the existing array and only copies the wraparound elements before head)

Not sure if it makes a difference; maybe a benchmark would tell:

```ts
const oldLen = this.#buffer.length;
const newLen = oldLen * 2;
this.#buffer.length = newLen;
this.#capacity = newLen;

// Copy all wrapped around elements (before #head)
// into the newly allocated space.
for (let i = 0; i < this.#head; i++) {
  this.#buffer[oldLen + i] = this.#buffer[i];
  this.#buffer[i] = undefined;  // ensure gc when the element is removed
}
```

packages/zero/Dockerfile
```diff
@@ -1,13 +1,20 @@
 FROM golang:1.23 AS litestream
 
+# Install cross-compilation toolchain for arm64 CGO builds
```
darkgnotic: I don't think you mean to include this file in the PR?

packages/shared/src/ring-buffer.ts
darkgnotic: ah, i see that you size down too ... that function is only for sizing up  🤔 .
not sure if it's worth the optimization to have different functions for sizing up and sizing down.

packages/zero/Dockerfile
grgbkr: need this to get a build i can deploy to cloudzero for benchmarking

packages/shared/src/ring-buffer.ts
darkgnotic: Same-array shrink algo:

```ts
const oldLen = this.#buffer.length;
const newLen = this.#buffer.length / 2;

const left = Math.max(newLen, this.#head);
const right = Math.min(this.#head + this.#size, oldLen);

for (let i = left ; i < right; i++) {
  this.#buffer[i - newLen] = this.#buffer[i];
}
if (this.#head >= newLen) {
  this.#head -= newLen;
}
this.#buffer.length = newLen;
this.#capacity = newLen;
```

grgbkr: nice idea, yeah this got us another 10x on the microbenchmark

packages/shared/src/queue.ts
```diff
@@ -43,12 +44,14 @@ export class Queue<T> {
   delete(value: T): number {
     assert(value !== undefined, 'Queue delete value must not be undefined');
 
+    // Drain, filter, and re-push. This is O(n) but delete() is rare.
+    const items = this.#produced.drain();
```
arv, line 48: This can be O(1) as well. drain could return the underlying storage as an iterable and the original buffer can get new storage.

see comment on drain

packages/shared/src/ring-buffer.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+export class RingBuffer<T> {
+  #buffer: (T | undefined)[];
+  #head = 0;
+  #size = 0;
+  #capacity: number;
+
+  constructor(initialCapacity = DEFAULT_INITIAL_CAPACITY) {
+    this.#capacity = Math.max(MIN_CAPACITY, initialCapacity);
+    this.#buffer = Array.from({length: initialCapacity});
```
arv, line 23: This should use `this.#capacity`

packages/shared/src/ring-buffer.ts
```diff
@@ -0,0 +1,143 @@
...
+    this.#buffer = newBuffer;
+    this.#head = 0;
+    this.#size = newSize;
+    return count;
+  }
+
+  /**
+   * Removes and returns all elements in FIFO order, resetting the buffer.
+   */
+  drain(): T[] {
```
arv, line 92: How about returning an Iterable<T> instead. Then it could reuse the current storage and reset the own storage.

Something like:


```ts
const iter = new DrainIterator(this.#buffer, this.#size, this.#head, this.#capacity);
this.#buffer = ...
this.#size = 0;
this.#head = 0;
return iter;
```

arv (commented):
One actual bug and one suggestion

PR #5987, opened 2026-05-14
Bound change-streamer retries

aboodman:
## What

Limits the change-streamer to five consecutive in-process replication loop failures before surfacing an unrecoverable error. The HTTP wrapper now propagates a failed change-streamer run so the worker exits instead of stopping quietly.

## Why

If user resets database while the replication-manager is running, it stays Ready and retries forever. Bounded retries let fast transient failures recover in process, while repeated failures hand recovery to the process/Kubernetes restart path so startup initialization can run again.


packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
```diff
@@ -384,21 +374,11 @@ class ChangeStreamerImpl implements ChangeStreamerService {
         this.#storer.run().catch(e => stream.changes.cancel(e));
 
         this.#stream = stream;
-        if (
-          this.#state.resetBackoff() >
-          REPLICATION_STATUS_ERROR_DELAY_THRESHOLD_MS
-        ) {
-          // After recovering from a backoff for which a replication status
-          // error was published, publish an OK status
-          this.#replicationStatusPublisher.publish(
```
darkgnotic, line 393: Please don't get rid of the status publishing logic. It serves an important purpose for letting the user know of issues and we haven't decided on a replacement for it yet.

I'm not sure if we can drive everything with logs because the status events can be big (e.g. they encode the entire schema).

PR #5989, opened 2026-05-15
fix(zero-cache): Abort on missing ChangeDB CDC table

aboodman:
If tables go missing while we are running, there is no sense trying to recover.

We need to abort so that we can re-initialize the schema.

See https://app.incident.io/rocicorp/incidents/222


darkgnotic:
Thank you!

Can you also add the equivalent logic to the CVRPurger so that the view-syncers also recover:

https://github.com/rocicorp/mono/blob/e5bf8bd4f64dadb203a04015778484e05650a739/packages/zero-cache/src/services/view-syncer/cvr-purger.ts#L95

darkgnotic:
If you're up for a more general solution that works for both the change-streamer and the cvr-purger, you can modify the backoff logic here:

https://github.com/rocicorp/mono/blob/e5bf8bd4f64dadb203a04015778484e05650a739/packages/zero-cache/src/services/running-state.ts#L153

To recognize this PostgresError as another type of terminal error (in addition to AbortError and UnrecoverableError).

aboodman:
done

PR #5990, opened 2026-05-15
fix(inspector): use actual query plan generated by table source

tantaman:
Use the plan from the query that was prepared by table source itself rather than re-preparing. This will prevent future divergences if table-source ends up inlining parameters to get better plans. E.g., `LIKE ?` vs `LIKE '%inlined'` produce different plans. The former being a worse plan than the latter.

Additionally, stop injecting the random `sdf` value which created divergent plans than what were actually used.


arv (commented):
LGTM

PR #5992, opened 2026-05-15
chore(zql): penalize flip in the planner's cost model

tantaman:
## Summary

  The planner's cost model under-counts the runtime cost of
  `FlippedJoin` because it doesn't model two IVM-level costs
  that are invisible to SQLite's `scanstatus`:

  1. **Eager child load** — `FlippedJoin.fetch` reads ALL
  children into an array before any parent work, so even with
  a downstream `LIMIT` every child pays IVM cost (generator
  yields, debug accounting, btree-set inserts).
  2. **Chunk priming** — `mergeSortedStreams` opens every
  chunk to seed its heap before yielding the first row; each
  open runs an IN-list SQL to its first match.

  Concretely this means the cost model treats SQLite row scans
   and IVM row processing as if they cost the same — but in
  practice IVM is ~100× slower per row. For limited queries
  with large child cardinality this leads the planner to pick
  a flipped plan when semi would short-circuit far earlier.

  This change adds `child.scanEst *
  FLIP_IVM_PER_CHILD_OVERHEAD` (constant `3`) to the
  flipped-join cost, gated on `parent.limit !== undefined &&
  child.scanEst > getMultiConstraintChunkSize()`. The gate
  matters:

  - **No downstream limit** → semi has to scan everything
  anyway, so flipped's eager-load isn't wasted. Don't
  penalize.
  - **Single chunk** (`child.scanEst ≤ 256`) → no
  `mergeSortedStreams` priming happens, so no IVM tax. Don't
  penalize.

  This preserves flipped wins on full-scan queries and on
  small child sets, while penalizing the case it's actually
  wrong for: limited `TAKE` queries with huge child
  cardinality.

  ## Measured impact

  Benchmarked on a 211 GB zbugs replica via
  `zql-benchmarks/src/zbugs-profile.ts`. The pathological
  query is `project=gatewaycore AND open=true AND
  whereExists(label=api-gateway) AND
  whereExists(label=async-processing) ORDER BY modified DESC
  LIMIT 50`:

  | | before | after |
  |---|---:|---:|
  | Plan picked | flipped + semi (cost 401k) | **semi + semi 
  (cost 1.24M)** |
  | Rows scanned by SQLite | 1,098,099 | **53,712** (20×
  fewer) |
  | Rows read into JS | 475,051 | **48,397** (10× fewer) |
  | Wall time | **65 s** | **7.4 s** (8.8× faster) |

  Other zbugs benchmark variants (with assignee, single-label,
   no-label) are unchanged in plan choice and row counts —
  their costs don't trip the gate.

  ## Cost model details

  For a flipped join with parent `P`, child `C`, and chunk
  size `K`:

  cost = C.cost
       + ceil(C.scanEst / K) * P.startupCost          //
  per-chunk prepare
       + C.scanEst * (P.cost + P.scanEst)             //
  per-child seek
       + (P.limit !== undefined && C.scanEst > K
           ? C.scanEst * 3                            // ← new
   IVM-overhead term
           : 0)

  The constant `3` is calibrated against the observed gap:
  pre-change flipped at 401k vs semi-semi at 1.24M for the v6
  query; multiplying the 416k-row child scan by 3 (≈ 1.25M
  added) tips the choice. Other calibrations tried:
  - `0` — baseline, v6 picks the bad plan (65 s).
  - `1`, `2` — v6 still picks flipped (semi cost not reached).
  - `3` — v6 picks semi-semi. Used.
  - Higher values aren't needed and start over-penalizing
  legitimate flipped wins.

  ## Test impact

  Existing `planner-join.test.ts` chunk-boundary tests now
  exercise a connection with `limit=50` so the gate fires and
  the IVM term is exercised. The helper `ovhFor(n, chunkSize)`
   makes assertions read symmetrically across the `n ≤
  chunkSize` / `n > chunkSize` cases.

  ### Integration tests

  All `chinook/planner.pg.test.ts` (5/5) and
  `pagila/planner-exec.pg.test.ts` (18/18) pass unchanged.

  `chinook/planner-exec.pg.test.ts` is 25/26. The one
  regression is **correlation-only**, not a plan-choice
  regression:

  - Test: `'extreme selectivity - artist to album to long
  tracks'` (indexed variant)
  - Picked plan: still optimal (`within-optimal: 1.00x`,
  passes)
  - Spearman correlation: 0.20 (threshold 0.35, -42.9%
  headroom)

  The cost model's *ranking* of all-flipped plans against semi
   alternatives drifts below this test's correlation
  threshold, but the planner still picks the actually-optimal
  plan. The base-DB threshold for this same test is already
  `0.15`, so loosening the indexed threshold to `0.20` would
  be consistent with the existing tolerance. Per the project's
   CI margin convention (~30% headroom for cost-model
  thresholds), the threshold was tighter than the model's
  underlying noise floor; this PR just exposes it.

  ## Depends on

  This PR can be reviewed independently but its impact is best
   measured **on top of** the `Debug.rowVended` O(N²) fix
  (separate PR — branch `mlaw/fix-debug-rowvended-quadratic`).
   Without that fix the v6 query takes 260 s instead of 65 s
  and the 7.4 s post-fix number can't be reproduced.

  ## Test plan

  - [ ] `npm --workspace=zql run test -- planner` (90 tests,
  all pass)
  - [ ] `npm --workspace=zql-integration-tests run test --
  --project='*pg-18*' src/chinook/planner-exec.pg.test.ts`
  (25/26 — known soft fail on `extreme selectivity`
  correlation)
  - [ ] `npm --workspace=zql-integration-tests run test --
  --project='*pg-18*' src/pagila/planner-exec.pg.test.ts`
  (18/18)
  - [ ] `npm --workspace=zql-integration-tests run test --
  --project='*pg-18*' src/chinook/planner.pg.test.ts` (5/5)
  - [ ] Multi-pg sweep: rerun the same with
  `--project='*pg-15*'`, `'*pg-16*'`, `'*pg-17*'` before merge
  - [ ] Decide on the `extreme selectivity` indexed
  correlation threshold: leave the test failing as a known
  follow-up, or relax `0.35 → 0.20` in this PR

  ## Follow-ups

  - The `extreme selectivity` indexed correlation threshold
  above.
  - Calibration: the constant `3` is fitted to one workload.
  If other workloads surface where flipped is the right choice
   but gets penalized, consider scaling the overhead by
  `chunks` instead of `child.scanEst`, or making it
  proportional to the existing SQL cost rather than a flat
  constant.
  - Stat4 has accurate per-label cardinality (verified —
  `lbl_proj_000_15` and `lbl_proj_000_45` both at 416k rows,
  planner gets ~393k estimates), so the cost model is
  correctly informed; this PR is about cost *formula*, not
  *stats*.


tantaman:
flip is faster now that we batch it and does not require a penalty

PR #5995, opened 2026-05-17
refactor(otel-proxy): replace @vercel/node types with local interfaces

tantaman:
Wdyt? Worth it?

The handler only used @vercel/node for type-only imports of VercelRequest and VercelResponse, but the package shipped vulnerable build-tooling transitives (undici, path-to-regexp, @vercel/python-analysis, @vercel/static-config, @vercel/build-utils) that npm audit flagged. None were reachable in the deployed serverless function — Vercel's platform invokes the handler with its own runtime types, not the bundled ones — but they inflated the audit count.

Replaces the imports with a minimal local ProxyRequest/ProxyResponse pair matching the handler's actual surface. vercel deploy still works since it uses the platform CLI, not @vercel/node.

Clears 6 npm audit advisories (39 → 33; high 27 → 23, moderate 12 → 10).

### DIFF
--- a/apps/otel-proxy/api/index.test.ts
+++ b/apps/otel-proxy/api/index.test.ts
@@ -1,6 +1,5 @@
-import type {VercelRequest, VercelResponse} from '@vercel/node';
 import {describe, it, expect, vi, beforeEach} from 'vitest';
-import handler from './index';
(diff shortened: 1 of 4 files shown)


arv (approved):
LGTM

This makes sense to me.

PR #5998, opened 2026-05-17
chore: Pin GitHub Actions to specific versions

arv:
Update GitHub Actions workflows to use pinned versions for better stability and predictability in builds. This change ensures that the workflows do not break due to updates in the actions used.


0xcadams:
LGTM

PR #6007, opened 2026-05-18
chore: Add a script that detects vulnerabilities in package-lock.json

arv:
You can run this from any repository with a package-lock.json file, and it will check for vulnerabilities in the dependencies listed in that file.

### DIFF
--- a/tools/verify-package-deps/src/package-lock-commit-deps.ts
+++ b/tools/verify-package-deps/src/package-lock-commit-deps.ts
@@ -1,0 +1,1664 @@
+#!/usr/bin/env node
+
+/* eslint-disable no-console */
+import {execFile as execFileCallback} from 'node:child_process';
+import {writeFile} from 'node:fs/promises';
+import {join} from 'node:path';
+import {promisify} from 'node:util';
+
+const execFile = promisify(execFileCallback);
+
+const DEFAULT_LOCKFILES = ['package-lock.json'];
+const DEPENDENCY_FIELDS = [
+  'dependencies',
+  'devDependencies',
+  'optionalDependencies',
+  'peerDependencies',
+] as const;
+const NOTION_API_BASE_URL = 'https://api.notion.com/v1';
+const NOTION_API_VERSION = '2022-06-28';
+const OSV_BATCH_QUERY_URL = 'https://api.osv.dev/v1/querybatch';
+const GITHUB_ADVISORY_BASE_URL = 'https://github.com/advisories';
+const NVD_CVE_BASE_URL = 'https://nvd.nist.gov/vuln/detail';
+const SEMVER_VERSION_REGEX = /^v?\d+\.\d+\.\d+(?:[-+][0-9A-Za-z.-]+)?$/;
+const LEADING_V_REGEX = /^v/;
+const GITHUB_HTTPS_REMOTE_REGEX =
+  /^https:\/\/github\.com\/([^/]+\/[^/]+?)(\.git)?$/;
+const GITHUB_SSH_REMOTE_REGEX = /^git@github\.com:([^/]+\/[^/]+?)(\.git)?$/;
+
+const options = parseArgs(process.argv.slice(2));
+const [repoRoot, githubCommitBaseUrl] = await Promise.all([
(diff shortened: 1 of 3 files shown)


tools/verify-package-deps/src/package-lock-commit-deps.ts
```diff
@@ -0,0 +1,1664 @@
...
+        name: packageVersion.dependency,
+      },
+      version: packageVersion.version,
+    })),
+  };
+
+  const response = await fetch(OSV_BATCH_QUERY_URL, {
+    method: 'POST',
+    headers: {'Content-Type': 'application/json'},
+    body: JSON.stringify(body),
```
tantaman, line 803: is there a limit to how many we can sedn in a batch at once?

tools/verify-package-deps/src/package-lock-commit-deps.ts
```diff
@@ -0,0 +1,1664 @@
...
+}
+
+async function getFileAtRevision(
+  revision: string,
+  path: string,
+): Promise<string | undefined> {
+  try {
+    return await runGit(['show', `${revision}:${path}`]);
+  } catch {
+    return undefined;
```
tantaman, line 395: you don't want to at least log the error?

tools/verify-package-deps/src/package-lock-commit-deps.ts
```diff
@@ -0,0 +1,1664 @@
...
+  if (
+    value.startsWith('link:') ||
+    value.startsWith('workspace:') ||
+    value.startsWith('file:') ||
+    value.startsWith('git+') ||
+    value.startsWith('github:') ||
+    value.startsWith('http://') ||
+    value.startsWith('https://')
+  ) {
+    return undefined;
```
tantaman, line 584: we should log about these so we can investigate out of band.

tools/verify-package-deps/src/package-lock-commit-deps.ts
```diff
@@ -0,0 +1,1664 @@
...
+  }
+
+  return SEMVER_VERSION_REGEX.test(value)
+    ? value.replace(LEADING_V_REGEX, '')
+    : undefined;
```
tantaman, line 589: why undefined if it doesn't match semver?

tools/verify-package-deps/src/package-lock-commit-deps.ts
```diff
@@ -0,0 +1,1664 @@
...
+      deduped.set(key, change);
+    }
+  }
+
+  return [...deduped.values()];
+}
+
+function parseResolvedVersion(value: string | undefined): string | undefined {
+  if (!value) {
+    return undefined;
```
tantaman, line 572: in what cases can `value` be missing?

tantaman (commented):
I think the script needs to check transitive dependencies. If we query OSV with Vite 5.1, does it tell us about vulns only in Vite or in Vite + its deps? If only in Vite, we need to do the transitive dep walk ourselves.

What about checking lockfile hash integrity? E.g., compare the hash we have in the lockfile against what the npm registry says.

tantaman:
What about pairing with OSV-Scanner? https://github.com/google/osv-scanner

Does this workflow work, so we can leverage everything osv-scanner already does?

1. `git log --since=... -- package-lock.json` to get the commits of interest.
2. For each commit: `git show <sha>:path/to/package-lock.json > /tmp/snapshot.json`, then `osv-scanner  --lockfile=package-lock.json:/tmp/snapshot.json --format=json`.
3. Diff the JSON results across commits to compute the timeline / presence ranges.
4. Emit the HTML/Markdown/Notion report.

arv:
> I think the script needs to check transitive dependencies. If we query OSV with Vite 5.1, does it tell us about vulns only in Vite or in Vite + its deps? If only in Vite, we need to do the transitive dep walk ourselves.

It operates on the package-lock.json which already has all the transitive dependencies.

> What about checking lockfile hash integrity? E.g., compare the hash we have in the lockfile against what the npm registry says.

I would think npm install checks that but I can explore.

arv:
I'm going to delete this now.

PR #6010, opened 2026-05-18
chore(pnpm): add phantom TS dependencies

tantaman:
Many of our packages have phantom dependencies which prevents updating to pnpm. pnpm requires all packages to explicitly list the deps they use.

This adds TS everywhere it is used.


tantaman:
@arv started this already

PR #6012, opened 2026-05-18
ci: harden gh actions

0xcadams:
- Cleans up GitHub Actions workflow structure.
- Separates build and deploy steps for the SST deploy workflows.
- Tightens CI job permissions and checkout behavior.
- Updates benchmark workflows so PR jobs only produce artifacts.


0xcadams:
Splitting up

PR #6014, opened 2026-05-19
chore(ci): remove legacy assistant workflow

0xcadams:
Removes an unused assistant workflow from the repository. This keeps the active workflow set focused on current automation.


arv (commented):
LGTM

PR #6016, opened 2026-05-19
chore(ci): update benchmark workflows

0xcadams:
Updates benchmark workflow structure so measurement and publishing are handled separately. Also refreshes runner selection and related Playwright setup.


arv (commented):
LGTM

PR #6017, opened 2026-05-19
chore(ci): update file size workflows

0xcadams:
Updates file size and bundle size workflows to pass generated outputs between jobs. This keeps measurement and publishing steps more clearly separated.


arv (commented):
LGTM

PR #6019, opened 2026-05-19
chore(ci): update mirror workflow

0xcadams:
Updates the branch mirror workflow implementation. The new flow uses a GH app instead of a personal GH token.

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -4,22 +4,33 @@ on:
     branches:
       - 'main'
 
+permissions:
+  contents: read
+
 jobs:
   mirror_job:
     runs-on: ubuntu-latest
     name: Mirror main branch to sandbox and terabugs branch
     steps:
-      - name: Mirror main to sandbox
-        id: mirror-sandbox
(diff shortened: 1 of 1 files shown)


.github/workflows/mirror.yml
```diff
@@ -4,22 +4,33 @@ on:
...
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: |
+          for branch in sandbox terabugs; do
+            gh api \
+              --method PATCH \
+              "/repos/$GITHUB_REPOSITORY/git/refs/heads/$branch" \
+              -f "sha=$GITHUB_SHA" \
+              -F force=true
+          done
```
arv, line 36: what does this do?

arv (commented):
LGTM

.github/workflows/mirror.yml
0xcadams: It does the same thing that google mirror workflow did, but without a PAT. Mirrors those sandbox/terabugs branches to the same commit as main.

PR #6024, opened 2026-05-19
ci: fix permissions and pnpm install

0xcadams:
- install pnpm before setup-node in SST deploy workflows so pnpm cache/install steps can run
- grant reusable Bencher caller jobs the contents/checks/pull-requests permissions requested by nested jobs
- apply the same fixes to PR Bencher callers and the Gigabugs SST workflow

### DIFF
--- a/.github/workflows/perf-v2.js.yml
+++ b/.github/workflows/perf-v2.js.yml
@@ -21,6 +21,8 @@ jobs:
         with:
           persist-credentials: false
       - uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6
+        with:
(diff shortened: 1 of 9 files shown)


.github/workflows/perf-v2.js.yml
```diff
@@ -21,6 +21,8 @@ jobs:
         with:
           persist-credentials: false
       - uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6
+        with:
```
arv, line 24: why is this needed?

0xcadams: Yeah it's running on the same runner and so conflicting actions were failing since they modify a central pnpm dir

PR #6025, opened 2026-05-19
chore(ci): remove npx usage

0xcadams:
- Replace CI `npx` usage with lockfile-backed local binaries via `pnpm exec`.
- Run SST deploy commands from `prod/sst` with the workspace-installed `sst` binary.
- Remove CI-invoked `npx` from package scripts, add explicit `tsx` dev dependencies, and update the lockfile.


.github/workflows/js.yml
```diff
@@ -59,7 +59,8 @@ jobs:
         with:
           node-version: 24.x
           cache: 'pnpm'
-      - run: npx oxfmt@^0.45.0 --check
+      - run: pnpm install --frozen-lockfile
```
arv, line 62: this is a bit unfortunate.

Using npx allowed us to skip the install.

arv (commented):
LGTM

PR #6026, opened 2026-05-20
fix(zero-cache): run ddl event triggers as definer

Karavil:
## What we're seeing

We're getting failures when a custom database role, let's call it `foo`, tries to manage its own schema. `foo` owns `foo_private`, so creating a table there should be an ordinary operation, while its lack of access to Zero's internal shard schema is intentional.

```sql
CREATE TABLE foo_private.audit_log (
  id TEXT PRIMARY KEY
);
```

Zero's DDL hooks are installed for the whole database, so PostgreSQL wakes them up for this statement even though `foo_private.audit_log` is not published. Today those hooks run as the caller, which means they enter Zero's internal schema as `foo` and hit the permission boundary we wanted `foo` to have.

```text
foo creates foo_private.audit_log
  |
  v
PostgreSQL wakes Zero's DDL hooks
  |
  |  current_user = foo
  v
Zero's internal shard schema
  |
  v
permission denied, the whole CREATE TABLE rolls back
```

Giving `foo` access to Zero's schema would make the error disappear, but it would also undo that separation. Zero should do its small amount of internal bookkeeping using Zero's established function owner, then return control to `foo` so PostgreSQL can finish the original statement.

## What this changes

This PR makes the two outer hooks, `emit_ddl_start()` and `emit_ddl_end()`, run as `SECURITY DEFINER` with `search_path` fixed to `pg_catalog, pg_temp`. The hooks briefly use their existing owner's privileges, while the application's DDL still runs as `foo`, and the table it creates is still owned by `foo`.

On existing shards, migration v24 replaces the hooks in place and preserves their current owner. On new shards, they are owned by the role that sets the shard up. There is no hardcoded Zero actor role, since creating and administering one would be a separate operator contract that differs across PostgreSQL providers.

The same migration revokes direct `PUBLIC` execution on the two privileged entrypoints and revokes `PUBLIC` creation in Zero's helper schema. Those ACLs become security relevant in this migration because this is the first time the hooks can run with their owner's privileges. Event triggers keep calling the functions by OID, so restricted roles do not need direct `EXECUTE` permission for their DDL to work.

`update_schemas()` remains `SECURITY INVOKER`, and the two definer entrypoints take no arguments from the caller and use no dynamic SQL.

## Tests

The behavior test creates an unpublished table as a restricted role that has no access to Zero's schema, internal tables, function owner, or definer entrypoints, then confirms the DDL succeeds and the table still belongs to that role. The catalog and migration tests cover the function owner, definer flag, fixed search path, public ACLs, and the v23 to v24 upgrade.

The two focused test files pass with 37 tests on each supported version of PostgreSQL, 15 through 18. Formatting, lint, and type checking also pass.

## Versioning note

The staged `arv/lightweight-ddl-start-phase2` work also claims v24. If this lands first, that migration needs to become v25 and preserve these definer, search path, and ACL changes when it recreates the hooks.

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/schema/ddl.pg.test.ts
+++ b/packages/zero-cache/src/services/change-source/pg/schema/ddl.pg.test.ts
@@ -54,11 +54,12 @@ describe('change-source/tables/ddl', () => {
   ];
   const EXPECTED_NO_SHARD_PRIVILEGES = [
     {
+      ddlEntrypointExecute: false,
       schemaUsage: false,
       schemaCreate: false,
       publishedSchemaSelect: false,
       publishedSchemaUpdate: false,
-      memberOfShardOwner: false,
+      memberOfFunctionOwner: false,
     },
   ];
 
@@ -134,9 +135,9 @@ describe('change-source/tables/ddl', () => {
     CREATE PUBLICATION nonzeropub FOR TABLE pub.foo, pub.boo;
     `;
 
-  test('event trigger functions use the shard owner and locked search path', async () => {
-    const [{shardOwner}] = await upstream<{shardOwner: string}[]>`
-      SELECT current_user AS "shardOwner"
+  test('event trigger functions lock down their privileged execution context', async () => {
+    const [{functionOwner}] = await upstream<{functionOwner: string}[]>`
+      SELECT current_user AS "functionOwner"
     `;
 
     const eventTriggers = await upstream<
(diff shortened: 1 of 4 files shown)


Karavil:
@darkgnotic, I rebased this and took another pass after we hit the original failure with a restricted database role managing its own schema. The role is allowed to create tables there and intentionally has no access to Zero's internal schema, but Zero's database wide DDL hooks wake up as that role and abort the otherwise valid DDL when they try to enter the shard schema.

The PR now makes only the two outer hooks `SECURITY DEFINER`, adds the v24 migration for existing shards, and has a regression test that creates an unpublished table through a restricted role, proves the role gets no Zero grants, and checks that the table is still owned by that role afterward. The focused PostgreSQL suite is 37/37.

I also called out the v24 collision with `arv/lightweight-ddl-start-phase2` in the description. Would appreciate a look when you get a chance.

PR #6036, opened 2026-05-20
chore: Update package manager references from npm to pnpm

arv:
Change all references in configuration files and scripts from npm to pnpm to standardize the package management process.


tantaman (commented):
lgtm

PR #6037, opened 2026-05-20
fix(zql): include singular flag in query hash

arv:
.one() sets limit: 1 in the AST and singular: true in the format. .limit(1) sets only limit: 1. Previously both produced the same hash, causing them to share a view and return the wrong shape.

Add tests covering .one() vs .limit(1) and queries that differ only in their relationship.

https://discord.com/channels/830183651022471199/1288232858795769917/1506476345662636092

### DIFF
--- a/packages/zql/src/query/query-impl.ast.test.ts
+++ b/packages/zql/src/query/query-impl.ast.test.ts
@@ -2353,3 +2353,27 @@ describe('whereExists with scalar option', () => {
     });
   });
 });
+
+function hash(q: AnyQuery) {
+  return asQueryInternals(q).hash();
+}
+
+describe('query hash uniqueness', () => {
+  test('.one() and .limit(1) produce different hashes', () => {
+    const issueQuery = newQuery(schema, 'issue');
+    const withOne = issueQuery.one();
+    const withLimit = issueQuery.limit(1);
+
+    // Both produce the same AST (limit: 1) but .one() also sets singular: true
+    // in the format, so they must hash differently.
+    expect(hash(withOne)).not.toBe(hash(withLimit));
+  });
+
+  test('queries with different relationships produce different hashes', () => {
+    const issueQuery = newQuery(schema, 'issue');
+    const withComments = issueQuery.related('comments');
+    const withOwner = issueQuery.related('owner');
+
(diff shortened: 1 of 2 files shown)


tantaman:
Ouch. Seems like the longer term fix would be to bake this into the AST as it impacts actual query results.

alamothe:
Thanks for the fix — but I think it only covers the top-level case and still collides on nested relationships.

The ViewStore cache key in #6065 (and current main) is:

```ts
const hash = qi.hash() + (qi.format.singular ? 't' : 'f') + zero.clientID;
```
That only reads the parent format.singular. But the one() vs limit(1) ambiguity exists at every nesting level, and the AST can't disambiguate it: CorrelatedSubquery in the AST has no cardinality/singular field — one() just sets limit: 1 on the subquery and stores singular: true in format.relationships[…]. So related('x', q => q.one()) and related('x', q => q.limit(1)) produce an identical AST hash, differing only in the nested format.

Concrete repro (same parent, nested differs):
```ts
const a = z.query.issue.where('id', '=', id)
  .related('owner', q => q.one());      // format.relationships.owner.singular = true

const b = z.query.issue.where('id', '=', id)
  .related('owner', q => q.limit(1));   // format.relationships.owner.singular = false
```
For both: parent format.singular === false, and hashOfAST is identical (owner subquery is limit: 1 either way). So the #6065 key resolves to hash + 'f' + clientID for both → they share one ViewWrapper. Whichever materializes first wins, and the other query gets the wrong shape for owner ([row] instead of row, or vice-versa) — the same bug as the original report, one level down.

Fix would be to fold the whole format into the key (recursing through relationships), not just the top-level singular. Something like:
```ts
function serializeFormat(f) {
  return JSON.stringify({
    singular: f.singular,
    relationships: Object.fromEntries(
      Object.entries(f.relationships)
        .sort(([a], [b]) => a.localeCompare(b))
        .map(([k, v]) => [k, serializeFormat(v)]),
    ),
  });
}
const hash = `${qi.hash()}:${serializeFormat(qi.format)}:${zero.clientID}`;
```
@arv

arv:
You are correct. We will need to include more of the format or the whole format.

PR #6040, opened 2026-05-20
chore(docker): use pnpm to gate transitive install scripts

tantaman:
`npm install -g` of zero's tarball runs install/postinstall scripts for every transitive dep. Switch to `pnpm add` with
`pnpm.onlyBuiltDependencies` in /opt/app/package.json; pnpm v10+ ignores dependency lifecycle scripts by default and only runs them for allowlisted packages.

Allowlist: `@rocicorp/zero-sqlite3` (its `install` runs prebuild-install to fetch the native binary, which isn't in its tarball) and `esbuild` (via tsx; included defensively until runtime confirms tsx is unused).

Install is local to /opt/app with `node-linker=hoisted` so the existing zero-sqlite3 symlink and config.yml copy paths still resolve; PATH is extended so CMD drops `npx`.

### DIFF
--- a/packages/zero/Dockerfile
+++ b/packages/zero/Dockerfile
@@ -27,10 +27,11 @@ RUN test -n "$ZERO_VERSION"
 RUN apk add --update curl
 RUN apk add --no-cache readline readline-dev
 
-# Install pnpm: its onlyBuiltDependencies allowlist (v10+) blocks lifecycle
-# scripts from transitive deps unless explicitly opted in, shrinking the
-# postinstall supply-chain surface to packages we name below.
-RUN npm install -g pnpm@10
+# Install pnpm via corepack (bundled with Node). pnpm's onlyBuiltDependencies
+# allowlist (v10+) blocks lifecycle scripts from transitive deps unless
+# explicitly opted in, shrinking the postinstall supply-chain surface to
+# packages we name below.
+RUN corepack enable && corepack prepare pnpm@10 --activate
 
 RUN mkdir -p /opt/app
 WORKDIR /opt/app
--- a/packages/zero/Dockerfile
+++ b/packages/zero/Dockerfile
@@ -36,10 +36,6 @@ RUN corepack enable && corepack prepare pnpm@10 --activate
 RUN mkdir -p /opt/app
 WORKDIR /opt/app
 
-# Flat node_modules so downstream paths (zero-sqlite3 symlink, config.yml) stay
-# resolvable without walking pnpm's virtual store.
-RUN printf 'node-linker=hoisted\n' > .npmrc
-
 # Allowlist for lifecycle scripts. zero-sqlite3 ships no native binary in its
 # tarball (see its package.json `files`) — its `install` script runs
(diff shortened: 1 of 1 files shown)


packages/zero/Dockerfile
```diff
@@ -27,30 +27,60 @@ RUN test -n "$ZERO_VERSION"
 RUN apk add --update curl
 RUN apk add --no-cache readline readline-dev
 
+# Install pnpm: its onlyBuiltDependencies allowlist (v10+) blocks lifecycle
+# scripts from transitive deps unless explicitly opted in, shrinking the
+# postinstall supply-chain surface to packages we name below.
+RUN npm install -g pnpm@10
```
arv: can you use corepack?

packages/zero/Dockerfile
```diff
@@ -27,30 +27,60 @@ RUN test -n "$ZERO_VERSION"
...
+
 RUN mkdir -p /opt/app
 WORKDIR /opt/app
 
+# Flat node_modules so downstream paths (zero-sqlite3 symlink, config.yml) stay
+# resolvable without walking pnpm's virtual store.
+RUN printf 'node-linker=hoisted\n' > .npmrc
```
arv: Do we really need this? I got all our deps in the right place as part of the transition to pnpm

packages/zero/Dockerfile
```diff
@@ -27,30 +27,60 @@ RUN test -n "$ZERO_VERSION"
...
+
+# Allowlist for lifecycle scripts. zero-sqlite3 ships no native binary in its
+# tarball (see its package.json `files`) — its `install` script runs
+# prebuild-install to fetch one, so it must be allowed. esbuild is pulled in
+# transitively via tsx and downloads a platform binary in postinstall; keep it
+# allowlisted unless runtime testing confirms tsx is never invoked.
+RUN cat > package.json <<'EOF'
```
arv, line 44: This should be in in the yaml file now I believe.

packages/zero/Dockerfile
tantaman: Fixed

packages/zero/Dockerfile
tantaman: Removed

packages/zero/Dockerfile
tantaman: pnpm v10's `pnpm.onlyBuiltDependencies` is read from the install root's
`package.json`  (`/opt/app/package.json` here) and the workspace yaml's
`allowBuilds:` only governs `pnpm install` inside this monorepo.

When I removed the HEREDOC:

```
Ignored build scripts: @rocicorp/zero-sqlite3@1.0.18, protobufjs@7.6.0.
Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
```

→ native binary never fetched (`build/Release/` missing) → `zero-cache`
crashes on startup with `Error: Could not locate the bindings file`.

If you know of a way to make pnpm honor the workspace allowlist for a
non-workspace install, happy to switch.

arv: we are using pnpm 11 

If we are picking up 10 then we need to figure out why

PR #5997, opened 2026-05-17
chore: Migrate from npm to pnpm as package manager

arv:
## Summary
This PR migrates the monorepo from npm to pnpm as the primary package manager. This includes updating all CI/CD workflows, documentation, package.json files, and workspace configuration.

### Package Manager Configuration
- Updated root `package.json` to specify `pnpm@11.1.2` as the package manager
- Removed npm-style `workspaces` array from root `package.json`
- Created new `pnpm-workspace.yaml` with workspace configuration and build allowlist for native dependencies
- Added `.npmrc` with `link-workspace-packages=true` for proper workspace linking

### Workspace Dependencies
- Updated all internal workspace dependencies across 40+ package.json files from `"0.0.0"` to `"workspace:*"` protocol
- This applies to all packages including: zero-cache, zero-client, zero-server, zql, zqlite, shared, otel, and others

### CI/CD Workflows
- Added `pnpm/action-setup@v4` step to all GitHub Actions workflows (js.yml, api-snapshot.yml, bundle-sizes.js.yml, perf-smoke.yml, perf-v2.js.yml, bencher-benchmarks.yml, bencher-benchmarks-pr.yml, canary-release.yml, reusable-benchmark.yml, reusable-file-sizes.yml, sst-*.yml)
- Updated Node.js setup to use `cache: 'pnpm'` instead of `cache: 'npm'`
- Replaced `npm ci` with `pnpm install --frozen-lockfile`
- Replaced `npm install` with `pnpm install`
- Updated all `npm run` commands to `pnpm run`
- Updated workspace-specific commands from `npm --workspace=` to `pnpm --filter`

### Documentation
- Updated AGENTS.md with pnpm equivalents for all npm commands
- Changed examples from `npm install`, `npm run`, `npm --workspace=` to pnpm equivalents

### Build Scripts
- Updated package.json scripts in zero-cache, zero-client, zql-benchmarks, replicache-perf, and others to use `pnpm run` instead of `npm run`
- Updated release.ts build script to use pnpm commands

## Implementation Details
- The migration maintains full feature parity with npm workspaces
- All workspace dependencies now use the `workspace:*` protocol for proper monorepo resolution
- The pnpm-workspace.yaml includes an allowlist for packages that require native builds (prisma, esbuild, ssh2, etc.)
- Lock file changed from package-lock.json to pnpm-lock.yaml

https://claude.ai/code/session_01D686NfZE6RZreuFbsYLywT

PR #6043, opened 2026-05-21
fix(zero): restore flat out/ layout by setting preserveModulesRoot

arv:
Since the pnpm migration (#5997), Rollup's preserveModules emitted transitive chunks under out/packages/<pkg>/... instead of out/<pkg>/..., because under pnpm's isolated linker the resolved module paths run through .pnpm/ symlinks and Rollup's implicit common-ancestor root ends up at mono/ rather than mono/packages/. This split .js chunks away from their sibling .d.ts files (tsc still emits to the flat tree) and broke imports like @rocicorp/zero/server/adapters/pg at runtime.

Pinning preserveModulesRoot to packages/ makes chunk paths install-state independent and restores the layout shipped through 1.5.0.


0xcadams (commented):
LGTM thanks!

PR #6046, opened 2026-05-21
chore(zero): externalize peer deps and guard against bundling

arv:
Adapter entries (./server/adapters/pg, ./react, ./solid) imported pg, react, and solid-js, but they weren't in dependencies or peerDependencies, so Rolldown followed the imports and inlined them (along with pg-pool, pg-protocol, pg-types, postgres-*, split2, xtend, etc.) into out/node_modules/.pnpm/. Consumers ended up with duplicate copies that break Pool instanceof checks and React's single-instance rule.

Declare pg, drizzle-orm, react, and solid-js as optional peerDependencies so getExternalFromPackageJSON marks them external. Add @opentelemetry/semantic-conventions to dependencies — it's used directly by zero-cache and is an internal runtime dep, not a peer.

Add a post-build assertion that fails if out/node_modules exists, so a future undeclared import that gets inlined breaks the build with a pointer at the fix instead of silently shipping in the tarball.


packages/zero/package.json
```diff
@@ -193,17 +194,33 @@
   },
   "peerDependencies": {
     "@op-engineering/op-sqlite": ">=15",
+    "drizzle-orm": "^0.45.2",
```
0xcadams, line 197: Oh nice I've wanted to do this for a while

0xcadams (commented):
LGTM!

PR #6047, opened 2026-05-21
fix(zero-cache): reconnect if no response from PG

tantaman:
Fix for this incident: https://rocicorp.slack.com/archives/C013XFG80JC/p1779331755395539

tldr: we do not detect when the inbound half of the replication stream goes silent. This updates to detect that and reconnect.

- related: #6042


darkgnotic (approved):
Thank you @tantaman !

PR #6062, opened 2026-05-26
chore(zero-client): Update PROTOCOL_VERSION to 51 in tests

arv:
This pull request updates the protocol version used in the zero client, ensuring consistency across tests and storage naming. The changes are minor and primarily involve incrementing the protocol version number from 50 to 51.


arv:
It is unclear how this ever passed on the CI.

I think we should maybe turn off turbo caching :'(

PR #6063, opened 2026-05-26
chore: Update API snapshot

arv:


arv:
This should have prevented the original from landing.

PR #6064, opened 2026-05-26
Revert "fix(zql): include singular flag in query hash"

arv:
Reverts rocicorp/mono#6037

This makes the hash different on client and server.


arv:
It is fine to have one() and limit(1) share the pipeline. The caching just needs to be fixed in the view manager

PR #6076, opened 2026-05-26
perf(zero-cache): use ring buffers for subscription queues

Karavil:
When a subscription accumulates a deep message backlog, draining its array queues can spend seconds removing entries from the front. Use the existing shared `RingBuffer` for pending messages and consumers.

The ring buffer gains tail access for coalescing and snapshot/clear operations for cleanup. Cleanup callbacks still observe the pending queue before it is cleared; cancellation releases extra buffer capacity. The existing `Queue.drain()` capacity reuse, string termination marker, consumption results, and abort-signal behavior are preserved.

## Performance

The committed `subscription.bench.ts` checks every delivered value and verifies that the subscription drains completely. It separates a persistent subscription with bounded queue depth from a deliberately preloaded backlog, and includes delivery to an already-waiting consumer.

```sh
pnpm --filter zero-cache run bench subscription
```

Median wall time on macOS arm64 / Node 22.22.0. Both versions ran the same benchmark, with a 2-second minimum measurement budget and at least 10 samples per case. The baseline is unmodified `Subscription` from upstream `b760ebbd12e1513cd24248ac3291f64e69b073d6`. Copy the benchmark into a checkout of that revision to reproduce the baseline.

**Bounded streaming: 100,000 total messages through one subscription.**

| Delivery pattern | Upstream | This PR | Change |
| --- | ---: | ---: | ---: |
| Window of 1 | 18.114 ms | 14.799 ms | -18.3% |
| Window of 64 | 16.705 ms | 15.338 ms | -8.2% |
| Window of 1,024 | 18.400 ms | 15.075 ms | -18.1% |
| Consumer waits before each push | 21.494 ms | 19.838 ms | -7.7% |

**Preload-and-drain stress cases: all messages queued before consumption.** These bypass producer backpressure and do not establish an application or end-to-end replication speedup.

| Preloaded messages | Upstream | This PR | Change |
| ---: | ---: | ---: | ---: |
| 1,000 | 0.165 ms | 0.133 ms | -18.9% |
| 10,000 | 1.904 ms | 1.601 ms | -15.9% |
| 20,000 | 65.171 ms | 3.491 ms | -94.6% |
| 100,000 | 2808.077 ms | 20.062 ms | -99.3% |

V8 can trim some arrays from the front, but [falls back to copying when the backing store cannot move](https://github.com/nodejs/node/blob/v22.22.0/deps/v8/src/objects/elements.cc#L2325-L2339). [Large backing stores are one case where it cannot move](https://github.com/nodejs/node/blob/v22.22.0/deps/v8/src/heap/heap.cc#L3367-L3392). The stress-case increase is consistent with that path; its threshold and magnitude depend on the runtime and GC state.

An independent check outside Vitest used direct monotonic timers, alternating baseline/PR execution order, and assertions on every value, consumption count, cleanup, and abort signal. It reproduced the deep-backlog effect on Node 22.22.0, 24.11.1, and 26.0.0. No production queue-depth measurements or current end-to-end catchup timings are claimed.

## Validation

- 83 focused tests pass on Node 22: queue/ring-buffer behavior, wrapped tail replacement, snapshots, cleanup ordering, coalescing after wraparound, subscription termination, subscriber, and forwarder.
- All eight committed benchmark cases pass against both implementations on Node 22.
- Both affected packages pass lint and typechecking; changed files pass formatting and `git diff --check`.
- The broader shared/node and zero-cache/no-pg run passed 2,496 tests, with four failures in `vfs-watermark-poller.test.ts`. Running the same suite with unmodified upstream source reproduced those exact four failures (2,490 passed). Those failures are outside this change.

### DIFF
--- a/packages/shared/src/head-indexed-queue.test.ts
+++ b/packages/shared/src/head-indexed-queue.test.ts
@@ -1,0 +1,55 @@
+import {describe, expect, test} from 'vitest';
+import {HeadIndexedQueue} from './head-indexed-queue.ts';
+
+describe('HeadIndexedQueue', () => {
+  test('keeps FIFO order across many shifts without exposing compacted slots', () => {
+    const queue = new HeadIndexedQueue<number>();
+    for (let i = 0; i < 2050; i++) {
+      queue.push(i);
+    }
+
+    for (let i = 0; i < 1500; i++) {
+      expect(queue.shift()).toBe(i);
+    }
+    queue.push(2050);
+
+    expect(queue.size).toBe(551);
+    expect(queue.peek()).toBe(1500);
+    expect(queue.last()).toBe(2050);
+    expect(queue.toArray()).toEqual(
+      Array.from({length: 551}, (_, i) => i + 1500),
+    );
+  });
+
+  test('delete and replace operations skip holes left in the active window', () => {
+    const queue = new HeadIndexedQueue<string>();
+    queue.push('a');
+    queue.push('b');
+    queue.push('c');
+    queue.push('d');
+
+    expect(queue.deleteFirst('a')).toBe(true);
+    expect(queue.deleteMatching(value => value === 'c')).toBe(1);
+    queue.replaceLast('D');
+
+    expect(queue.size).toBe(2);
+    expect(queue.peek()).toBe('b');
+    expect(queue.last()).toBe('D');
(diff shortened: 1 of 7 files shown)


gerardatkonvo:
Really looking forward to see this merged :)

PR #6077, opened 2026-05-26
perf(zero-cache): batch v6 websocket change frames

Karavil:
## summary

Preserves the v6 `Downstream` payload shape while allowing websocket batch frames and `subscribeBatched()`. Incremental sync flushes the worker batch after a received websocket frame.

ACK semantics stay per-message in this PR.

## delta vs previous PR

Previous PR: hot queue helpers.

Scenario: `medium-wide-batch-pressure`, 15s, 1 RM stream consumer, websocket transport, v6 protocol.

| metric | before | after | delta |
| --- | ---: | ---: | ---: |
| tx/s | 1,652.3 | 2,099.4 | +27.1% |
| rows/s | 33,046.2 | 41,987.6 | +27.1% |
| fanout msg/s | 51,531.8 | 68,585.4 | +33.1% |
| websocket messages | 807,248 | 557 | 99.9% lower |
| websocket ACKs | 807,248 | 1,046,498 | higher because ACKs are still per-message |
| reconnect catchup to join | 1,658.3 ms | 275.1 ms | 83.4% lower |

## why this is safe

The child messages remain normal v6 `Downstream` payloads. The batch frame is a transport envelope only. Old clients that do not request batch frames keep the one-message stream shape.

## validation

```text
pnpm exec vitest --project='*no-pg*' run src/types/streams.test.ts src/services/change-streamer/subscriber.test.ts src/services/replicator/worker-message-batcher.test.ts
pnpm exec vitest --project='*15*' run src/services/change-streamer/change-streamer-http.pg.test.ts src/services/change-streamer/change-streamer-service.pg.test.ts src/services/change-streamer/storer.pg.test.ts
pnpm run check-types
```

## related perf workstream PRs

Benchmarked with the harness from #6070. This implementation PR intentionally does not include the benchmark harness files.

1. #6070: benchmark harness
2. #6071: storer batching
3. #6072: change processor batching
4. #6073: write-worker batching
5. #6074: serving local apply
6. #6075: RM fanout flow control
7. #6076: hot queue helpers
8. #6077: v6 websocket batch frames
9. #6078: cumulative ACK negotiation

### DIFF
--- a/packages/shared/src/head-indexed-queue.test.ts
+++ b/packages/shared/src/head-indexed-queue.test.ts
@@ -1,0 +1,55 @@
+import {describe, expect, test} from 'vitest';
+import {HeadIndexedQueue} from './head-indexed-queue.ts';
+
+describe('HeadIndexedQueue', () => {
+  test('keeps FIFO order across many shifts without exposing compacted slots', () => {
+    const queue = new HeadIndexedQueue<number>();
+    for (let i = 0; i < 2050; i++) {
+      queue.push(i);
+    }
+
+    for (let i = 0; i < 1500; i++) {
+      expect(queue.shift()).toBe(i);
+    }
+    queue.push(2050);
+
+    expect(queue.size).toBe(551);
+    expect(queue.peek()).toBe(1500);
+    expect(queue.last()).toBe(2050);
+    expect(queue.toArray()).toEqual(
+      Array.from({length: 551}, (_, i) => i + 1500),
+    );
+  });
+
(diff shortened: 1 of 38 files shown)


Karavil:
closing this for now because the current branch cannot be represented as an atomic upstream-main PR without pulling in prerequisite work. we'll reopen the v6 websocket batch-frame change after the earlier perf PRs it depends on have landed, or if we get an upstream base branch for a true stack.

PR #6082, opened 2026-05-28
chore(ci): update action pins

0xcadams:
- update GitHub Action pins to exact resolved versions
- move stale action pins to the latest vetted tags at least 14 days old
- keep zizmor action compatible by pinning zizmor v1.24.1


arv (approved):
LGTM

PR #6087, opened 2026-05-29
feat: Add column codecs for ZQL schema

arv:
This PR introduces a **codec** system for Zero schemas: a column can be stored, synced, and compared as a JSON-native value while exposing a decoded, application-friendly type (e.g. `Date`) at runtime.

## Schema & types

- New `Codec<Encoded, Decoded>` type and a `.codec()` method on `ColumnBuilder`. The column keeps its underlying JSON value type for storage; query results decode to `Decoded` and writes/`where` comparisons encode back.
  ```ts
  const event = table('event').columns({
    id: string(),
    at: number().codec<Date>({
      decode: n => new Date(n),
      encode: d => d.getTime(),
    }),
  }).primaryKey('id');
  ```
- The decoded type flows through the user-facing types automatically (`SchemaValueToTSType` → row results, `InsertValue`/`UpdateValue`, and `where` filter types). `Codec` / `SchemaValueWithCodec` are exported from `@rocicorp/zero`.

## Write path (encode)

Codec columns are encoded back to their stored JSON form before anything downstream sees them:
- CRUD `insert`/`upsert`/`update`/`delete` args (shared `makeTableCRUD` → `encodeRow`), covering both client and server (`zero-server`) mutators.
- `where('col', op, value)` literals and the expression-builder `cmp(...)` form, plus `IN`/`NOT IN` arrays (`encodeFilterValue`, which keys off the operator). The AST therefore always carries encoded literals, so this is consistent on the server with no extra work.

## Read path (decode) — raw row attached to the entry

Decoding happens in the view layer (`view-apply-change.ts` / `applyChange`):

- Entries store the **decoded (app) values** directly (`decodeRowFields`).
- The original **encoded `Row` is attached to each entry** via an internal `encodedRowSymbol` back-pointer. Binary search, ordering, and row identity all read through that back-pointer (`entry[encodedRowSymbol] ?? entry`), so sorting/identity stay on the encoded values while consumers see decoded ones.
- Because entries are already decoded, the framework bindings are simple: React `useQuery` passes the (immutable) snapshot straight through and Solid reads the store directly — no decode projection or special cloning step.

## Fast path (no extra cost without codecs)

Codec-free schemas are zero-copy: `decodeRowFields` / `encodeRow` return the input row unchanged and no `encodedRowSymbol` is attached, so non-codec entries and snapshots are byte-for-byte what they were before.

## Invariant

Stored / synced / IVM values remain JSON, so sorting, identity, the wire protocol, and IndexedDB are unaffected; the codec must be a stable bijection at the encode/decode boundary.

## Testing

Unit and type tests for `.codec()`, encode (CRUD + `where`/expression builder), decode (ArrayView data + listeners), and end-to-end coverage that `run()` and `materialize()` return decoded values.


packages/zql/src/mutate/crud.ts
```diff
@@ -169,21 +170,29 @@ export function makeTransactionMutate<TSchema extends Schema>(
...
+  const {name, columns} = tableSchema;
   return Object.fromEntries(
     CRUD_KINDS.map(kind => [
       kind,
-      (value: unknown) => executor(tableName, kind, value),
+      (value: unknown) =>
+        executor(
+          name,
+          kind,
+          encodeRow(value as Record<string, unknown>, columns),
```
arv, line 194: I think we can "lift up" columnsHaveCodecs to here so we do not have to check for every row.

packages/zql/src/ivm/codec.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+  return (result ?? row) as T;
+}
+
+/**
+ * Encodes a single value for `column`, or returns it unchanged if the column
+ * has no codec or the value is `null`/`undefined`.
+ */
+export function encodeValue(
```
arv: This is only used in tests. Remove.

packages/zql/src/ivm/codec.ts
```diff
@@ -0,0 +1,147 @@
...
+  row: T,
+  columns: Record<string, SchemaValue>,
+): T {
+  if (!columnsHaveCodecs(columns)) {
+    return row;
+  }
+  let result: Record<string, unknown> | undefined;
+  for (const key in row) {
+    const value = row[key];
+    const codec = isNullish(value) ? undefined : getCodec(columns[key]);
```
arv: This can also be lifted and done once?

packages/zql/src/ivm/array-view.ts
```diff
@@ -102,7 +114,15 @@ export class ArrayView<V extends View> implements Output, TypedView<V> {
...
-    return this.#root[''] as V;
+    const raw = this.#root[''] as V;
+    if (!this.#hasCodecs) {
+      return raw;
+    }
+    if (this.#decodedFrom !== this.#root) {
+      this.#decoded = decodeView(raw as View, this.#schema, this.#format) as V;
+      this.#decodedFrom = this.#root;
+    }
+    return this.#decoded as V;
```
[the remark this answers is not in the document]
arv: Fixed in 30d781b. `decodeView` now takes an optional `DecodeCache` (a `WeakMap` keyed by the raw object), and `ArrayView` holds a persistent one. Because `applyChange` is immutable with structural sharing, unchanged rows/subtrees keep their raw-object identity, so the cache returns the same decoded entry/list for them — decoded identity is now stable across updates (no `new Date(...)` churn for unchanged rows). Added an `array-view` test asserting unchanged decoded rows are `===` across a change. Solid keeps decoding without a cache since its store is reconciled in place.

---

packages/zql/src/query/expression.ts
```diff
@@ -22,6 +26,36 @@ export type ParameterReference = {
...
+    column === undefined ||
+    isParameterReference(value)
+  ) {
+    return value;
+  }
+  const codec = getCodec(column);
+  if (!codec) {
+    return value;
+  }
+  if (Array.isArray(value)) {
```
tantaman: we should add a test that will catch this if/when we add multi dimensional arrays.

packages/zql/src/ivm/codec.ts
```diff
@@ -0,0 +1,168 @@
...
+  schema: SourceSchema,
+  format: Format,
+  cache?: DecodeCache,
+): Entry {
+  const cached = cache?.get(entry);
+  if (cached !== undefined) {
+    return cached as Entry;
+  }
+  const result: Record<string, unknown> = {};
+  for (const key in entry) {
```
tantaman: do we care about the id and rc symbols?

arv: I restructured this so this is no longer relevant.

The difference is that ArrayView now uses the decoded (app) types. It has a reference to the raw Row and that is used when comparing in binary search etc. This prevents extra copies and it simplifies React/Solid.

packages/zql/src/query/expression.ts
arv: I changed it to check if this is `IN` or `NOT IN` instead.

aboodman:
@arv is there any additional background on this feature? I remember Chase working on something similar back in the day and we ran into a few hard design problems. Or was this contributed by a user? Or just new from whole cloth? Help me connect the dots here.

arv:
Over the last few weeks it came up twice in Discord. One was asking if we would accept a PR. 

This is an old open rfe and something I've been thinking on, on and off. With Claude things like these are too easy to explore and it started off like that here as well. I just had to fine tune a few things to my likings.

aboodman:
@0xcadams I feel like we explored something like this and hit some roadblocks. Do you remember?

aboodman:
If the `decode()` function is not pure, it will have same problem as described here: https://bugs.rocicorp.dev/p/zero/issue/3465#comment-jrBHt1uohlTRFf6Kxrqjg.

We could document that it must be pure.

aboodman:
Perhaps it is also possible to implement runtime checking to enforce that is is pure/detect impurity.

packages/zql/src/ivm/codec.ts
```diff
@@ -0,0 +1,130 @@
...
+  for (const name in columns) {
+    if (getCodec(columns[name]) !== undefined) {
+      return true;
+    }
+  }
+  return false;
+}
+
+/**
+ * Returns a copy of `row` with each codec column's `decode` applied. Used on
```
aboodman, line 67: Does not technically always return a copy.

aboodman (commented):
How does it interact with custom mutators? Say I want a `Date` column on one of my rows. I would expect to be able to pass `Date` as an arg to mutators, but i don't think that would work because we need to send that arg over the wire as JSON.

I think this is similar to the problem we faced when trying to use the codec features of zod.

aboodman:
I think there is leverage to make this work somehow though. Since we have the codec we have all the information necessary - we can encode before sending over wire and decode before sending into mutator.

Other things to think through:

- Same for queries I assume?
- How does this relate to the codec feature of validator libraries like zod? We tried to do something similar by that path and hit the problem that standard schema doesn't expose that api. But even if it did, this approach here seems more general since it also applies to the read path. Do we eventually want to enable both?

arv:
> How does it interact with custom mutators? Say I want a `Date` column on one of my rows. I would expect to be able to pass `Date` as an arg to mutators, but i don't think that would work because we need to send that arg over the wire as JSON.
> 
> I think this is similar to the problem we faced when trying to use the codec features of zod.

As defined in this PR as of a today there is no mapping from/to args. I think the logical choice is to allow a codec instead of a standard validator.


```ts
defineMutator({
  decode: ({id, when}: {id: number; when: number}): {id: number; when: Date} => ({
   id,
   when: new Date(when),
  }),
  encode: ({id, when}: {id: number; when: Date}): {id: number; when: number} => ({
    id,
    when: when.getTime(),
  })
},  async ({tx, args: {id, when}, ctx: authData}) => {
  assert(when instanceof Date);
  ...
})
```

arv:
> * How does this relate to the codec feature of validator libraries like zod? We tried to do something similar by that path and hit the problem that standard schema doesn't expose that api. But even if it did, this approach here seems more general since it also applies to the read path. Do we eventually want to enable both?

It uses a type compatible type with Zod codecs.

The type is very minimal:

```ts
export type Codec<Encoded, Decoded> = {
  decode: (value: Encoded) => Decoded;
  encode: (value: Decoded) => Encoded;
};
```

and it is easy to adopt by Valibot, Valita etc.

PR #6088, opened 2026-05-29
fix(zql): range filters (< <= > >=) should use UTF-8 order to match ORDER BY / SQLite

sravan27:
## Summary

`createPredicateImpl` (`packages/zql/src/builder/filter.ts`) evaluates `<` / `<=` / `>` / `>=` with raw JS operators. For strings that compares by **UTF-16 code unit**. But the same `MemorySource` sorts with `compareValues` → `compareUTF8` (**UTF-8 / code-point order**, *"the default on SQLite and we need to match it"* per `ivm/data.ts`) for `ORDER BY`.

The two disagree for any non-BMP character (emoji, supplementary CJK), so a client-side range filter can include/exclude a row that `ORDER BY` would place on the *other* side of the bound — a silent `WHERE` vs `ORDER BY` inconsistency.

## Example

`'｡'` (U+FF61) vs `'😀'` (U+1F600): by UTF-16 code unit `'｡' > '😀'`, but by code point / UTF-8 `'｡' < '😀'` (matching SQLite). So `where('col', '<', '😀')` wrongly **excludes** a row with `col = '｡'` that `orderBy('col', 'asc')` ranks below it.

## Fix

Route the range comparisons through `compareValues`, so filtering uses the same UTF-8 ordering as `ORDER BY` and SQLite. Numeric and boolean comparisons are unchanged (`compareValues` compares those numerically). Adds a regression test in `filter.test.ts`.

Same theme as #6083 (LIKE dotall): making the client-side evaluator match SQLite.

### DIFF
--- a/packages/zql/src/builder/filter.ts
+++ b/packages/zql/src/builder/filter.ts
@@ -115,18 +115,21 @@ function createPredicateImpl(
       return lhs => lhs === rhs;
     case '!=':
       return lhs => lhs !== rhs;
-    // Use compareValues (UTF-8 / code-point order for strings) so range
-    // comparisons match ORDER BY and SQLite. Raw JS `<`/`>` compare strings by
-    // UTF-16 code unit, which disagrees for non-BMP characters (e.g. emoji),
-    // wrongly including/excluding rows relative to the sort order.
+    // Use compareValues for same-type comparisons so string ranges match ORDER
+    // BY and SQLite. Preserve the existing JS mixed-type behavior for
+    // compatibility instead of letting compareValues throw.
     case '<':
-      return lhs => compareValues(lhs, rhs) < 0;
+      return lhs =>
+        typeof lhs === typeof rhs ? compareValues(lhs, rhs) < 0 : lhs < rhs;
     case '<=':
-      return lhs => compareValues(lhs, rhs) <= 0;
+      return lhs =>
+        typeof lhs === typeof rhs ? compareValues(lhs, rhs) <= 0 : lhs <= rhs;
     case '>':
-      return lhs => compareValues(lhs, rhs) > 0;
+      return lhs =>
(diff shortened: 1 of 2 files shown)


arv (approved):
LGTM

Thanks

packages/zql/src/builder/filter.ts
```diff
@@ -114,14 +115,21 @@ function createPredicateImpl(
       return lhs => lhs === rhs;
     case '!=':
       return lhs => lhs !== rhs;
+    // Use compareValues for same-type comparisons so string ranges match ORDER
+    // BY and SQLite. Preserve the existing JS mixed-type behavior for
+    // compatibility instead of letting compareValues throw.
     case '<':
-      return lhs => lhs < rhs;
+      return lhs =>
+        typeof lhs === typeof rhs ? compareValues(lhs, rhs) < 0 : lhs < rhs;
```
arv: We should not allow this. I know it makes the tests fail at this point but we should fix the tests. Sorry.

arv (changes requested):

sravan27:
Thanks @arv — done. Dropped the mixed-type fallback so all four range ops use `compareValues` unconditionally (UTF-8 / code-point order, matching ORDER BY and SQLite). Mixed-type ordered comparisons now throw instead of silently falling back to JS semantics.

For the tests I split the old basic-operators property test into two:
- **equality** (`=` / `!=`) across types — still uses `===` / `!==`, so cross-type comparisons stay valid; and
- **ordered** (`<` `<=` `>` `>=`) over same-typed numerics, where `compareValues` matches JS ordering.

String UTF-8 ordering stays covered by the dedicated `string range comparisons use UTF-8…` test. Ready for another look 🙏

arv (approved):
LGTM

PR #6090, opened 2026-05-30
feat(zero-svelte): add Svelte integration package

thomasmol:
Adds the zero-svelte package with context helpers, Query class, ViewStore, and Svelte-friendly bindings. Includes tests and Svelte plugin configuration.

Fixes https://bugs.rocicorp.dev/issue/3592

Just trying to get this working. Tested this package locally in a SvelteKit project and works OK so far


arv:
Thanks. This looks pretty great. I will have to review it in greater details.

FYI @stolinski

thomasmol:
Thank you! Let me know if it needs changes. It is indeed heavily based on https://github.com/stolinski/zero-svelte

packages/zero-svelte/src/zero.ts
arv, line 1: I don't think this file is needed.

packages/zero-svelte/src/bindings.ts
arv, line 1: this file is not needed

thomasmol: I followed the `zero-react` and `zero-solid` packages here, which both have `bindings.ts` and `zero.ts` re-export files. We can remove them, but that spreads a lot more direct imports across the package and makes Svelte differ from the other bindings. Also might have to rename `zero-client.ts` to `zero.ts` be in consistent with the react and solid packages.

thomasmol:
FYI for anyone looking for Svelte support with latest version of Zero, there is a workaround:

Add `context.ts`, `query.svelte.ts` and `query.svelte.ts` of this PR directly in your project in your lib folder (e.g. `/src/lib/zero/...`). You'll have to change the imports a bit to e.g. `@rocicorp/zero/bindings` and `@rocicorp/zero`. Create  a `client.ts` like you would with https://github.com/stolinski/zero-svelte, and it should all work as expected.

PR #6098, opened 2026-06-03
zqlite: adopt @rocicorp/zero-sqlite3 1.1.1 (Unicode ILIKE, self-contained)

arv:
Fixes https://bugs.rocicorp.dev/p/zero/issue/3043

## What

Bump `@rocicorp/zero-sqlite3` to **1.1.1**, which ships Unicode-aware `lower()`/`upper()` built from an embedded case-mapping table (generated from Node's `String.prototype.toLowerCase`/`toUpperCase`, including the Greek final-sigma rule). zqlite compiles `ILIKE` to `lower(col) LIKE lower(pattern)`, so this is what makes `ILIKE` genuinely case-insensitive beyond ASCII — e.g. `'MÜLLER' ILIKE 'müller'` now matches.

The production wiring for this (`PRAGMA case_sensitive_like = ON`, the `lower(...) LIKE lower(...)` compilation in `query-builder.ts`) already landed separately. This PR is the dependency bump plus the tests that lock the behavior in.

## Why 1.1.1 and not the ICU build

Earlier attempts linked SQLite's ICU extension. That turned out to be unshippable across Linux distros — distro static ICU archives aren't `-fPIC`, dynamic ICU couples the binary to a specific `libicu` soname (glibc/musl/version skew), QEMU arm64 crashed, and ICU pulls in ~28 MB of data. **1.1.1 drops ICU entirely** in favor of a self-contained Unicode table, so the prebuilt binaries are portable with **no extra runtime dependency** (no `apk add icu-libs`, no soname caveat) on every platform including Windows and Alpine.

## Changes

- **Bump the floor** `^1.0.18 → ^1.1.1` (`zqlite`, `zero-cache`, `zero`, `replicache`) + lockfile.
- **New behavioral test** in `query-builder.test.ts`: runs the actual compiled `ILIKE` SQL against a Unicode row (`'MÜLLER' ILIKE 'müller'`). Fails on an ASCII-only `lower()`, passes with the Unicode `lower()` in 1.1.1.
- **New `ilike-parity.test.ts`** (using `test.for`): asserts the client-side IVM `ILIKE` matcher and the zqlite SQL `ILIKE` agree on the same inputs, so the two sides of the query engine stay in sync.
- **`pnpm-workspace.yaml`:** exclude `@rocicorp/zero-sqlite3` from `minimumReleaseAge` by package name. The version-pinned form (`@rocicorp/zero-sqlite3@1.1.1`) is not honored by pnpm's frozen-lockfile supply-chain check, so a freshly published bump otherwise trips `ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION` in CI. Name-only matches and avoids per-version churn for our own first-party package.

### DIFF
--- a/packages/zqlite/src/ilike-parity.test.ts
+++ b/packages/zqlite/src/ilike-parity.test.ts
@@ -1,4 +1,4 @@
-import {expect, test} from 'vitest';
+import {beforeAll, test} from 'vitest';
 import {createSilentLogContext} from '../../shared/src/logging-test-utils.ts';
 import {getLikePredicate} from '../../zql/src/builder/like.ts';
 import {Database} from './db.ts';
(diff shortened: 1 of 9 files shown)


packages/zqlite/src/ilike-parity.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  [String.raw`100\%`, '100x'], // no match (literal %)
+  [String.raw`a\_b`, 'a_b'],
+  [String.raw`a\_b`, 'axb'], // no match (literal _)
+
+  // ß is the case where folding (ß->ss) and lowering differ. BOTH backends
+  // lower() rather than fold, so neither matches "ss" — and they still agree.
+  ['straße', 'STRASSE'],
+];
+
+test('client-side ILIKE and zqlite ILIKE agree (Unicode, wildcards, escapes)', () => {
```
arv: use test.for

arv: Done — switched to `test.for`, so each (pattern, input) pair is its own test and failures isolate to the specific case.

PR #6101, opened 2026-06-05
bump zero-sqlite3 to 1.1.2

aboodman:
## Summary

Bumps all workspace references to `@rocicorp/zero-sqlite3` from `^1.1.1` to `^1.1.2` and refreshes `pnpm-lock.yaml` so installs resolve the patched native package.

## Why

`@rocicorp/zero-sqlite3@1.1.1` can fail for users installing/building Zero with newer Node versions because it may fall back to a source build with an incomplete tarball. The `1.1.2` patch includes the upstream fix, preventing future users from hitting that install failure.

## Validation

- `pnpm why @rocicorp/zero-sqlite3` resolves a single version: `1.1.2`
- `pnpm run format`
- `pnpm run lint` on Node `24.11.1`
- `pnpm run check-types` on Node `24.11.1`

### DIFF
--- a/packages/replicache/package.json
+++ b/packages/replicache/package.json
@@ -80,7 +80,7 @@
   },
   "devDependencies": {
     "@op-engineering/op-sqlite": ">=15",
-    "@rocicorp/zero-sqlite3": "^1.1.1",
+    "@rocicorp/zero-sqlite3": "^1.1.2",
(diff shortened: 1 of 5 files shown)


0xcadams (approved):
LGTM

arv (commented):
LGTM

PR #6105, opened 2026-06-09
fix(ci): restrict release workflow refs

0xcadams:
## Summary

- Remove the manual `workflow_dispatch` `ref` input from the Release workflow.
- For manual releases, pass the workflow-selected `github.ref_name` to `release.ts`.
- Keep release privilege gating in the GitHub `Release` environment.

## Why

A free-form `workflow_dispatch` ref let any user who could dispatch the workflow choose what code the privileged release job built. That path was risky because the release job has `contents: write`, `id-token: write`, `pull-requests: write`, Docker Hub credentials, npm trusted publishing/OIDC capability, and cosign signing. Once a caller-selected ref reached the release flow, package lifecycle scripts during `pnpm install` or release code from that ref could run in that credentialed context and potentially exfiltrate credentials or publish malicious npm/Docker artifacts. The later ref syntax validation did not establish trust in the ref.

This change removes arbitrary ref selection from workflow inputs. Manual releases now use the ref selected by the workflow dispatch UI, and the `Release` environment enforces approval and allowed ref policy before environment secrets are available.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,10 +12,6 @@ on:
         options:
           - canary
           - stable
-      ref:
-        description: Branch, tag, or commit SHA to build the release from
-        required: true
-        type: string
       dry_run:
         description: Build but skip git push, npm publish, and Docker push
         required: false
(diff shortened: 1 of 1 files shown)


arv:
How do you control what to release? Does this always pick main?

0xcadams:
> How do you control what to release? Does this always pick main?

You can choose the branch/tag to run off of here:

Screenshot 2026-06-09 at 12 57 15 PM

PR #6106, opened 2026-06-09
fix(zero-cache): filter rows to client schema

Karavil:
## Summary

This changes row patch serialization so zero-cache projects row payloads to the configured client schema before sending `pokePart.rowsPatch` to clients.

The cache still keeps the full replica row available for IVM, predicates, permissions, ordering, joins, and catchup computation. The projection only happens at the final server to client poke boundary.

I also added a defensive zero-client projection when merging incoming pokes. That protects newer clients during rolling upgrades if they receive full row payloads from an older cache.

## Security note

This is not a malicious-client security boundary. zero-cache still validates a client schema against the replicated Zero table specs, so a column that is replicated and supported by Zero can still be requested by a client that declares it.

Apps should continue to keep sensitive columns out of the replicated/public Zero surface, for example by excluding them from the publication or syncing a sanitized table/view.

This change should be read as honoring the accepted client schema at the poke/storage boundary for honest clients, plus reducing network bytes and client storage. It prevents columns omitted from that accepted client schema from being sent or stored, but it does not replace reducing the set of columns Zero can replicate in the first place.

## Compatibility

For new cache versions, omitted client schema columns are not sent over the wire. Tables absent from the client schema are omitted from row patches.

For new client versions talking to older caches, unknown columns are dropped before Replicache patch values are produced.

Primary key fields are retained in projected put payloads even if a malformed schema leaves them out of `columns`.

## Testing

* `pnpm --filter zero-cache run test:no-pg -- src/services/view-syncer/client-handler.test.ts`
* `pnpm --filter zero-client exec vitest run --browser.enabled=false src/client/zero-poke-handler.test.ts`
* `pnpm --filter zero-cache run check-types`
* `pnpm --filter zero-client run check-types`
* `pnpm --filter zero-cache run lint`
* `pnpm --filter zero-client run lint`
* `pnpm --filter zero-cache run check-format`
* `pnpm --filter zero-client run check-format`

### DIFF
--- a/packages/zero-client/src/client/zero-poke-handler.ts
+++ b/packages/zero-client/src/client/zero-poke-handler.ts
@@ -7,6 +7,7 @@ import type {
 import type {PatchOperation} from '../../../replicache/src/patch-operation.ts';
 import type {ClientID} from '../../../replicache/src/sync/ids.ts';
 import {unreachable} from '../../../shared/src/asserts.ts';
+import type {Row} from '../../../zero-protocol/src/data.ts';
 import type {MutationPatch} from '../../../zero-protocol/src/mutations-patch.ts';
 import type {
   PokeEndBody,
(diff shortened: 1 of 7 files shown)


packages/zero-types/src/name-mapper.ts
```diff
@@ -58,12 +59,13 @@ export class NameMapper {
...
       return row;
     }
-    const clientRow: Record<string, V> = {};
-    for (const col in row) {
-      // Note: columns with unknown names simply pass through.
-      clientRow[columns[col] ?? col] = row[col];
-    }
-    return clientRow;
+    // Note: columns with unknown names simply pass through.
+    return Object.fromEntries(
```
arv: Use mapEntries from shared/objects.ts

packages/zero-client/src/client/zero-poke-handler.ts
```diff
@@ -365,39 +366,123 @@ function rowsPatchOpToReplicachePatchOp(
...
+  return projectServerKeys(serverPrimaryKeyNames(tableSchema), row);
+}
+
+function projectServerKeys(columns: readonly string[], row: Row): Row {
+  return Object.fromEntries(
+    columns
+      .filter(column => Object.hasOwn(row, column))
+      .map(column => [column, row[column]]),
+  ) as Row;
+}
```
arv: too many allocations

packages/zero-client/src/client/zero-poke-handler.ts
```diff
@@ -365,39 +366,123 @@ function rowsPatchOpToReplicachePatchOp(
...
-          : undefined,
-        constrain: serverToClient.columns(op.tableName, op.constrain),
       };
+    }
     default:
       unreachable(op);
   }
 }
+
+function projectServerRow(
```
arv: Is this section really correct. I thought the client doesn't do any mapping from client to server and vice versa (except for in the inspector).

arv:
In general this seems like the right way to approach this.

@aboodman thought there was something subtle about this so I will let him speak to that.

aboodman:
If it is not a security boundary I really don't want to do this because I don't want to give people a false sense of security. Prefer to do it the right way with column permissions or something that actually enforces something.

Karavil:
> If it is not a security boundary I really don't want to do this because I don't want to give people a false sense of security. Prefer to do it the right way with column permissions or something that actually enforces something.

Agreed tbh. I assumed there was an easy way to know which columns are syncable but there doesn't seem be since zero-cache isn't aware of the canonical schema.

aboodman:
I think the right thing is for the api server to tell zero-cache what the syncable columns are. But that's so close to projections/column permissions that it should be considered at same time as an alternative.

PR #6108, opened 2026-06-10
chore(build): add greeting comment at the end of the build script

arv:


arv:
Sorry for the noise. Testing that auto assigning reviewers works as expected.

It works.

PR #6109, opened 2026-06-10
Support forwarding request headers

tjenkinson:
This adds 2 new options
- `mutate-allowed-request-headers`
- `query-allowed-request-headers`

which are an allow list of headers to forward from the incoming request upstream.

We need to record the IP address of the caller, which arrives on an `x-forwarded-for` header. We can't provide this in the client as it needs to come from the load balancer server side. I think allowing any header to be forwarded is a useful feature but open to other suggestions :)

Also I spotted a lot of `Object.assign`, including when checking headers, and replaced them with a safe version that can't pollute the prototype. Similar issue in places where we're setting a property directly on an object. If the property is `__proto__` and the object isn't one with a null prototype that could cause an issue. Let me know if you'd like this breaking out into a separate PR. Didn't for now because the headers logic also uses it.

Found a related bug here in the bug tracker: https://bugs.rocicorp.dev/p/zero/issue/3928

### DIFF
--- a/packages/shared/src/build-nested-object-path.ts
+++ b/packages/shared/src/build-nested-object-path.ts
@@ -39,7 +39,7 @@ export function buildNestedObjectPath<
   let current = target as Record<string, unknown>;
   for (let i = 0; i < parts.length - 1; i++) {
     const part = parts[i];
-    let next = current[part];
+    let next = Object.hasOwn(current, part) ? current[part] : undefined;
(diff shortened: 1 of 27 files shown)


0xcadams:
Could you please slim this down to the bare minimum changes? Without the `Object.assign` changes please and a separate PR for those to make it easier to review.

PR #6110, opened 2026-06-10
fix(zero-cache): verify actual backup state before purging change-log

arv:
The BackupMonitor previously trusted the litestream_replica_progress
metric as proof that a watermark was durably backed up before purging
the change-log up to it. During the 2026-06-10 incident (INC-684), a
wedged litestream replicate process kept advancing that metric without
writing anything to S3; the change-log was purged past the real backup,
leaving view-syncers unable to restore (watermark < minWatermark) and
crash-looping until manual intervention.

The cleanup watermark is now only advanced after verifying the claimed
backup time against the actual backup state, by listing snapshots and
WAL segments at the backup destination via the bundled litestream CLI.
If verification fails or the destination is stale, the purge is
conservatively skipped, a rate-limited warning is logged, and a new
replica.purge_blocked counter is incremented for alerting.

Addresses the TODO in backup-monitor.ts about verifying backups
directly rather than trusting litestream metrics.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/backup-monitor.ts
+++ b/packages/zero-cache/src/services/change-streamer/backup-monitor.ts
@@ -1,4 +1,5 @@
 import type {LogContext} from '@rocicorp/logger';
+import {resolver} from '@rocicorp/resolver';
 import parsePrometheusTextFormat from 'parse-prometheus-text-format';
 import {must} from '../../../../shared/src/must.ts';
 import {promiseVoid} from '../../../../shared/src/resolved-promises.ts';
@@ -8,7 +9,7 @@ import {
   getOrCreateGauge,
 } from '../../observability/metrics.ts';
 import {Subscription} from '../../types/subscription.ts';
-import {RunningState} from '../running-state.ts';
+import {RunningState, UnrecoverableError} from '../running-state.ts';
 import type {Service} from '../service.ts';
 import type {ChangeStreamerService} from './change-streamer.ts';
 import type {SnapshotMessage} from './snapshot.ts';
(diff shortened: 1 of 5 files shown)


packages/zero-cache/src/services/change-streamer/backup-monitor.ts
```diff
@@ -12,6 +16,22 @@ import type {SnapshotMessage} from './snapshot.ts';
 export const CHECK_INTERVAL_MS = 60_000;
 const MIN_CLEANUP_DELAY_MS = 30_000;
 
+/**
+ * Allowance for clock skew between the machine reporting litestream metrics
+ * and the timestamps reported by the backup destination (e.g. S3).
+ */
+export const BACKUP_VERIFICATION_SLACK_MS = 60_000;
```
darkgnotic, line 24: Perhaps this should be slightly larger than CHECK_INTERVAL_MS, since that's the amount of time it can take before we check the latest backup.

packages/zero-cache/src/services/change-streamer/backup-monitor.ts
```diff
@@ -223,32 +261,101 @@ export class BackupMonitor implements Service {
...
+    const lastUpload = must(this.#lastVerifiedUploadTime);
+    const verifiedWatermark = this.#maxWatermarkUpTo(
+      Math.min(
+        latestCleanupTime,
+        lastUpload.getTime() + BACKUP_VERIFICATION_SLACK_MS,
+      ),
+    );
+    if (verifiedWatermark.length === 0) {
+      this.#purgesBlocked.add(1, {reason: 'backup-stale'});
+      this.#lc.warn?.(
```
darkgnotic, line 356: If s3 writes are disappearing without litestream it's actually a serious problem, as it can result in corrupt backups, since litestream will think it has backed something up, when in reality it hasn't, and the format of the wal backups are such that some, but not all, gaps are detectable.

I think it might be safer to just kill the server here (provided we have given the logic enough slack, per the previous comment).

Then, investigating the source of the black holing should be the next priority.  🙏

darkgnotic (commented):
Thank you for investigating this @arv !

arv:
I took another stab at this taking your feedback into consideration.

Please take another look.

darkgnotic (approved):

PR #6113, opened 2026-06-11
chore(devcontainer): v0 of devcontainer for mono

grgbkr:
- Does not address how to run `apps/zbugs/docker/docker-compose.yml` for local dev.  This might be how we want to address that https://app.notion.com/p/replicache/Devcontainer-Docker-Compose-Sibling-Services-37c3bed8954580ee90f4c9dc68008c85
- Would be great to figure out how to share with cloudzero and other repos.

### DIFF
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Do toolchain setup outside the mounted repo/workspace.
+cd "$HOME"
+
+sudo corepack enable pnpm
+
+# Configure pnpm to use the correct store directory
+pnpm config set store-dir "$HOME/.local/share/pnpm/store"
+
+# Remove npm and npx symlinks to force pnpm usage
+sudo rm -rf \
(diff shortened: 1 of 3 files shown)


arv:
Why not #6111?

.devcontainer/post-create.sh
```diff
@@ -0,0 +1,16 @@
...
+set -euo pipefail
+
+# Do toolchain setup outside the mounted repo/workspace.
+cd "$HOME"
+
+sudo corepack enable pnpm
+
+# Configure pnpm to use the correct store directory
+pnpm config set store-dir "$HOME/.local/share/pnpm/store"
+
```
arv, line 11: So no signed commits on non cloudzero repos?

Seems fine with me but I don't understand why we are not consistent?

arv (changes requested):

arv:
Closing this since I'm actively working on this in the other PR.

.devcontainer/post-create.sh
grgbkr: i may just have missed that we are moving to signed commits.  im for it.

PR #6120, opened 2026-06-12
chore: replace Docker-in-Docker with sibling Compose services in dev container

arv:
## What

Removes Docker-in-Docker from the dev container (added in #6116) and runs Postgres as **sibling Compose containers** instead.

## Why

The dev container is the sandbox for normal development and AI-agent work. Docker-in-Docker requires privileged mode and turns the workspace into a nested container-orchestration environment — code/agents inside can create arbitrary containers, pull images, run root, and keep state outside Git review. That materially weakens the sandbox (see Greg's design doc, Notion "Devcontainer + Docker Compose Sibling Services").

The only reason DinD was added was so the `zero-cache` pg tests (testcontainers) could run inside the container. We can get that without giving the workspace any Docker control.

## How

- **Default profile** is now Compose-based: the `dev` workspace plus static `pg15`–`pg18` services, configured identically to the testcontainers setup (`wal_level=logical`, 100 slots/senders, UTC for 15/16, UTC+1 for 17/18), data on tmpfs so a restart fully resets them.
- **`pg-container-setup.ts`** reads `TEST_PG_<major>` (e.g. `TEST_PG_16=postgres://…`) and uses it instead of starting a container; unset (host, CI) it falls back to testcontainers exactly as before. Mirrors the existing `CUSTOM_PG` hook.
- **zbugs profile** (`.devcontainer/zbugs/`) layers the existing `apps/zbugs/docker/docker-compose.yml` on as siblings via Compose `include:`, with `ZERO_UPSTREAM_DB` preset to the service name. No `db-up` needed inside.
- Drops the `docker` feature + `docker-in-docker` from the config and lock files.

The workspace has **no Docker daemon and no host Docker socket**; it reaches the siblings over the Compose network by service name only.

## Verification

- All four pg versions come up healthy with correct timezones (incl. pg18's relocated volume path).
- zero-cache pg suites pass against the siblings via `TEST_PG_*` (including replication-slot / logical-replication tests), and still pass via testcontainers when the vars are unset.
- Both compose profiles validate (`docker compose config`); zbugs profile resolves all 7 services with 6 health-gated deps.
- `lint`, `format`, `check-types` clean.

## Notes for reviewers

- The pg siblings live in the **default** profile (not a separate one) specifically so agents can run `pnpm test` out of the box — which is what #6116 was solving. Static siblings don't violate the no-Docker-control invariant. Easy to split into their own profile if you'd prefer a leaner default.
- Different vitest projects now share one instance per pg version; test DB names are unique per package today (same exposure `CUSTOM_PG` already has).
- This is the in-repo half of a broader rollout — new repos use the new [rocicorp/devcontainer-templates](https://github.com/rocicorp/devcontainer-templates).


tantaman (commented):
fancy. lgtm

PR #6121, opened 2026-06-12
fix(zqlite): compile NULL cursor bounds into null-safe start constraints

YevheniiKotyrlo:
> Rebased onto #6189, which fixed the leading-key range comparison and the sargable leading bound for NULL cursor values. This PR is now the remaining delta.

## What

- `nullableAwareEquality` gets the same runtime NULL branch #6189 gave the range comparison: a NULL bound value compiles to `IS` instead of `=`. Without it, a cursor anchored inside a NULL **tie-break** group compiles `(a = ? AND b = NULL AND id > ?)` — never true — so the rest of the group is silently skipped, and a `basis: 'at'` anchor with any NULL field is unreachable. When nothing else follows the anchor, the window hydrates empty and a later synced UPDATE hits `Take`'s `Bound should be set` assert, which is still on `main` — the tie-break repro in `table-source.test.ts` is red on current `main`.
- `mapLiteDataTypeToZqlSchemaValue` populates `optional: nullableUpstream(liteDataType)`. Replica specs currently drop the `|NOT_NULL` attribute the replica records, so every column reads as required no matter what the app schema declares, and the `optional`-gated forms — the `(col IS NULL OR col < ?)` backward-walk admission, #6184's nullable-leading-column skip — can never engage server-side. With a truthful flag, a backward walk below a non-NULL bound stops silently dropping the NULL group that sorts before it.

## Why

Production stack: a list window sorted on nullable columns (all NULL in the data), cursor anchored in the NULL region. Before #6189 the leading-key lowering hydrated it empty and one UPDATE killed the whole client group with `Bound should be set` (stack traces in the standalone repro). After #6189, the same failure remains reachable through the tie-break shape above, and backward walks still lose the NULL-sorted head of the list — both need this delta.

## Proof

- `query-builder.test.ts` — NULL-bound truth-table snapshots (forward / reverse / descending / `basis: 'at'`) and the optional-column forms.
- `table-source.test.ts` — `fetching across a NULL-sorted cursor region` against real SQLite with a `makeComparator` oracle; the tie-break case fails on stock `main` and passes here.
- `lite-tables.test.ts` — `optional` mirrors each column's `|NOT_NULL` attribute.
- zqlite 183/183, `lite-tables` 25/25 on the merged branch; `tsc`, `oxfmt`, and `oxlint` clean on the changed files (0 errors; the warnings on those files pre-date this change).
- End-to-end: the standalone reproduction (real Postgres → zero-cache → Zero clients, https://github.com/YevheniiKotyrlo/zero-view-syncer-bound-should-be-set) — its reverse-walk probe isolates the mapper half: 1 row instead of 4 until `optional` is truthful.

Sibling: #6122 (the `Take` empty-window edit branch — the assert the tie-break shape still reaches).


aboodman:
There's a semantics question here. `start(null)` could just do what not having a start does, and that's tempting. but there's a question of what `.order('foo', 'desc').start(null)` should mean. Do you get all rows or none? If `start(null)` is the same as removing the start then you should get all. If `start(null)` means to treat the null as an entry in the stream then you should get none.

aboodman:
Wait nevermind, the nullability we're talking about here is on the column, not the row entirely.

YevheniiKotyrlo:
@tantaman — thanks for taking this (and #6122). One heads-up so the CI here doesn't read the wrong way: the GitHub Actions runs are all sitting in `action_required` (the external-contributor approval gate), so the `JS` suite hasn't actually run. The only checks reporting are the two Vercel preview deploys — which fail purely because they need Rocicorp-team authorization — and Socket. If it's useful, approving the workflow runs would surface the real suite in-thread.

Locally on `edf748c0` it's green: `query-builder.test.ts` 20/20 (the four NULL-bound snapshots are red on stock), zqlite vitest 178 pass, and the zero-cache `no-pg` failure set is byte-identical to `main` on the same machine (pre-existing Windows temp-file issues, zero new). The branch is 17 commits behind `main` but merges clean — no conflicts.

(On the `start(null)` note — agreed, it's the column's nullability, not the cursor row; the declared-descending case is pinned in `query-builder.test.ts` if that's useful to see.) Happy to adjust anything.

YevheniiKotyrlo:
Updated the branch: merged `main` and gave #6184's `sargableLeadingStartBound` [the same runtime guard](https://github.com/YevheniiKotyrlo/mono/commit/0d0ba12bf07a081016ec9a8392bbe31c7f5c8d36) the other helpers get here — with metadata absent (what replica specs are today), a NULL leading bound compiled to `"a" >= NULL AND (…)`, never true, which failed five of this PR's tests on the merge ref. Also carried the now-populated `optional` flags into the new `lite-tables` expectation. zqlite 179/179, lite-tables 25/25 on the merge; the #6184 tests pass unchanged.

tantaman:
Merging this once tests pass. Thanks again.

YevheniiKotyrlo:
> Merging this once tests pass. Thanks again.

Thank you for creating such an awesome framework! It really enables me to build fast, ship fast, and provides an excellent DX!

tantaman (approved):

PR #6122, opened 2026-06-12
fix(zql): handle an edit pushed onto an empty Take window

YevheniiKotyrlo:
## What

Makes `Take.#pushEditChange` tolerate an undefined bound (an empty take window), the way the add-, remove-, and child-change handlers already do, instead of asserting `Bound should be set` and crashing the zero-cache view-syncer.

## Why

A cursor query builds `Skip → Take`. `Skip` forwards an edit to its output whenever the old AND new rows both sort past its bound (`maybeSplitAndPushEditChange`). When the take's current page holds no rows it stores `{ size: 0, bound: undefined }`, so a forwarded edit reaches `#pushEditChange` with `takeState.bound === undefined`. The add-, remove-, and child-change branches all handle an undefined bound; only the edit branch asserts:

```ts
assert(takeState.bound, 'Bound should be set');
```

On the server this is fatal — the assert propagates out of the view-syncer, which stops the whole client group and closes every connection with an `Internal` ProtocolError:

```
Error: Bound should be set
    at assert (out/shared/src/asserts.js:3:16)
    at #pushEditChange (out/zql/src/ivm/take.js:240:3)
    at Take.push (out/zql/src/ivm/take.js:130:31)
    at maybeSplitAndPushEditChange (out/zql/src/ivm/maybe-split-and-push-edit-change.js:11:51)
    at Skip.push (out/zql/src/ivm/skip.js:58:11)
    at MeasurePushOperator.push (out/zql/src/query/measure-push-operator.js:30:23)
```

A self-consistent source can't drive this state through the pipeline (an empty take implies no rows past the skip bound — the remove branch backfills from the source before the window drains), but zero-cache ships a source whose hydration disagrees with the IVM comparator: the zqlite compilation of a `.start()` bound with a NULL sort value returns no rows while rows past the bound exist (fixed in #6121). With that skew, any synced UPDATE of a row past such a cursor kills the view-syncer — I have production stack traces of exactly this from a list window anchored on NULL-sorted rows. The fix is correct independently of the producer: the operator's input contract doesn't guarantee Skip/Take agreement, and the other three branches already tolerate the state.

## Changes

- `Take.#pushEditChange`: when `takeState.bound === undefined`, surface the new row as an add (the window has room — `size 0 < limit`) and return — the same outcome the add-change path produces for a row entering an empty window. This narrows `takeState.bound` for the rest of the method, so the standalone assert is no longer needed. A no-op (like the remove branch) would be wrong here: it would drop a row that should become visible.
- Tests in `take.push.test.ts` driving the operator directly (the same direct construction `take.fetch.test.ts` uses): hydrate a `Take(limit 5)` over an empty source, push an edit whose rows sort past the absent bound, assert exactly one add change and no throw; and the partitioned variant — a partition whose window hydrated empty takes the same add path, and the established bound routes a follow-up sort-stable edit through the normal edit branch. The harness path can't reach the state (its source is self-consistent), which is documented in the tests.

## Proof

- Red on stock: with the bare assert, the new test fails with the verbatim production error (`Error: Bound should be set`); with the fix it passes, and the full zql suite is green (1302 tests). `tsc`, `oxfmt`, and `oxlint` are clean on the changed files.
- End-to-end: a standalone reproduction (real Postgres → zero-cache → synced-query server → Zero clients, `@rocicorp/zero@1.6.2`; cursor anchored on a NULL-sorted row; one SQL UPDATE) shows the stock view-syncer dying with this exact error — killing only the client group that owns the poisoned query while a sibling group keeps receiving updates — and the patched build surfacing the update as an add instead: https://github.com/YevheniiKotyrlo/zero-view-syncer-bound-should-be-set
- The repro's CI runs the comparison as three lights: the same correct-behavior spec is red against stock (the bug-named tests fail while a harness-sanity test passes) and green with the fixes applied, plus a fix-isolation matrix proving each patch resolves exactly its bug across four install variants with no regressions. With ONLY this change applied (the matrix's take-only leg) the view-syncer survives the update and the forward window goes 0 -> 1 as the row surfaces. The crash alone reproduces in one command (`bun run bug:crash`); the clients use synced queries exclusively and assert that legacy ad-hoc queries stay disabled.

Sibling PR: #6121 (the zqlite NULL-bound start-constraint compilation that produces the empty-window skew in production).


aboodman:
@grgbkr can you please review this and #6121

aboodman:
Change my mind - @tantaman can you please review this and #6121 .

YevheniiKotyrlo:
@tantaman — thanks for picking this up (and #6121). Quick CI note so the checks here don't read the wrong way: the Actions runs are all in `action_required` (external-contributor approval), so the `JS` suite hasn't run — the only things reporting are the two Vercel preview deploys (failing only because they need Rocicorp-team authorization) and Socket. Approving the workflow runs would surface the real suite here.

Locally on `bca686be` it's green: `take.push.test.ts` 106/106 (the new empty-window case is red on stock with the verbatim `Bound should be set`), full zql suite 1302 pass. The branch is 17 commits behind `main` but merges clean. This is a small defensive change (+135/−1) matching what the add/remove/child branches already do for an empty window — happy to adjust anything.

Sibling: #6121 (the zqlite NULL-bound skew that produces the empty window in production).

tantaman:
Taking a look. take _should_ always have state that is filled by initial fetch. Still trying to wrap my head around how the state can ever end up `undefined`

tantaman:
Going to try an e2e test that includes `fetch` hydrating `take` before a push. I do agree that in isolation (e.g., in a unit test) this case we have here is possible. I think it should be impossible in the full operator graph though.

tantaman:
ok, I was able to repro. Thanks for taking the time to find this.

tantaman:
Interesting, this seems to only be possible on schema drift. E.g., where zero things a column is required but upstream has set it to null.

tantaman:
trying to decide between:

a. apply this fix
b. fail loudly that this case means Zero disagrees with PG about the shape of the data

Leaning towards b. The only question is if this makes certain classes of migrations harder. 

Walking through a case:
- Step 1: Relax col A to be nullable in the DB
- Step 2: all app code should be updated to expect `null` before anyone ever writes `null` to the db, if not many bad things happen


so...

(b) seems correct.

---

Taking this fix would mean that `push` and `fetch` are allowed to disagree which is a dangerous precedent.

I'm going to fix the error messaging.

YevheniiKotyrlo:
The disagreement here isn't client-vs-PG schema drift though — it's zero-cache's own replica introspection. [`mapLiteDataTypeToZqlSchemaValue`](https://github.com/rocicorp/mono/blob/cb014857f6dbce50983a83612bfad5c4e0643739/packages/zero-cache/src/types/lite.ts#L116-L120) drops the `|NOT_NULL` attribute, so on the server every column reads as required no matter what the app schema declares. The standalone repro [declares its sort column `.optional()`](https://github.com/YevheniiKotyrlo/zero-view-syncer-bound-should-be-set/blob/6ff9c9c0fcf8e8e0e9c0ab5bd016ef9b503e2637/schema.ts#L16) and still dies — re-verified on 1.7.0. (Your #6184 nullable-leading-column guard reads the same flag, so it can't engage server-side either.)

So (b) alone isn't satisfiable today — there's no app-side step to take; sorting by any nullable column is enough. #6121 fixes that producer (the mapper + the NULL-bound truth table). With it in, agreed (b) is the right take-side policy — happy to close this in favor of your messaging change once #6121 lands.

For the message: it'll fire beyond cursors (take's own maintenance fetches hit the same lowering — a `.limit()` window over a NULL-sorted head loses its backfill on DELETE, then the next edit hits this assert), and today it kills the whole client group and re-arms on rehydrate — so naming the query/table/column is what makes it diagnosable.

tantaman:
Just saw your message @YevheniiKotyrlo . Still parsing through it but, during that time, I found this issue which was also a contributing factor to the take problem: https://github.com/rocicorp/mono/pull/6189

YevheniiKotyrlo:
#6189 closes the leading-key half. Two residues, both still live on `main`:

1. A NULL in a **tie-break** key: [`nullableAwareEquality` has no runtime branch](https://github.com/rocicorp/mono/blob/c6f4c9712e45c50b652bba65bf3a0a62757587a8/packages/zqlite/src/query-builder.ts#L291-L303), so a cursor anchored inside a NULL group compiles `(a = ? AND b = NULL AND …)` and misses the rest of the group — and when nothing else follows the anchor, the window hydrates empty and [this assert still fires](https://github.com/rocicorp/mono/blob/c6f4c9712e45c50b652bba65bf3a0a62757587a8/packages/zql/src/ivm/take.ts#L448) (verified on `main`: comparator says one row past the anchor, hydration `[]`, one edit → `Bound should be set`). Same shape hits `basis: 'at'` anchors.
2. The replica `optional` derivation you flagged — also what lets the `<` form admit the NULL group on backward walks.

I've rebased #6121 onto #6189 down to exactly that delta (the equality branch + [`optional: nullableUpstream(...)`](https://github.com/YevheniiKotyrlo/mono/blob/e42cdaa49162a7e0ce8636accbada3ebb1f9a0ae/packages/zero-cache/src/types/lite.ts#L118-L122) + tests; [the tie-break case in `table-source.test.ts`](https://github.com/YevheniiKotyrlo/mono/blob/e42cdaa49162a7e0ce8636accbada3ebb1f9a0ae/packages/zqlite/src/table-source.test.ts#L325-L390) is red on current `main`). zqlite 183/183, `lite-tables` 25/25 on the merge. And since #6188 closed with the take branch unresolved: happy to repoint this PR at fail-closed (the drop + the direct-operator tests) if that's the direction you want.

tantaman:
I think we can close this out now based on the latest PR that landed?

PR #6123, opened 2026-06-15
Fix/zero cache preserve changelog boundary

aboodman:


packages/zero-cache/src/services/change-streamer/storer.ts
```diff
@@ -257,8 +257,17 @@ export class Storer implements Service {
...
       const [{deleted}] = await sql<{deleted: bigint}[]>`
-        WITH purged AS (
+        -- The backup watermark can be ahead of the durable changeLog if the
+        -- storer is behind but the backup replica has consumed forwarded
+        -- changes. Preserve the latest committed changeLog transaction as the
+        -- catchup boundary instead of assuming the backup watermark exists.
+        WITH keep AS (
+          SELECT max(watermark) AS watermark
+          FROM ${this.#cdc('changeLog')}
+          WHERE change->>'tag' = 'commit'
```
darkgnotic: IIUC, you do not need the 'tag' = 'commit' condition, because transactions are committed all or nothing. So preserving the max should be sufficient.

darkgnotic (approved):
Nice investigation!!

packages/zero-cache/src/services/change-streamer/storer.ts
aboodman: Agree, fixed.

PR #6127, opened 2026-06-15
perf(zql): collapse MemorySource fetch generator stack

arv:
For the common fetch shape — no active overlay, no `start`, and no connection filters — collapse the five chained generators (generateWithOverlay → generateWithStart → skipYields → generateWithConstraint) into a single loop: walk rows from scanStart, wrap each as a Node, and stop once the constraint no longer matches (rows are sorted by the constraint key first, so matches are contiguous). This is the shape of a join child-lookup, a plain scan, and a source fetch beneath a downstream Filter. When overlay/start/filter are present it falls through to the unchanged generator chain.

Perf — memory-ivm-deopt, this change alone vs baseline:
  MemorySource scan, 1 key   342 µs -> 168 µs  (2.0x)
  Filter fetch (1000)        468 µs -> 312 µs  (1.5x)
  Join fetch 1000 -> 20      3.70 ms -> 2.67 ms (1.4x)

Validated: zql/ivm suite (792) and full zql suite (1300) pass. Adds a Docker-free bench config (vitest.config.bench-mem.ts) to run the in-memory benchmarks without a container runtime.

### DIFF
--- a/packages/zql-benchmarks/vitest.config.bench-mem.ts
+++ b/packages/zql-benchmarks/vitest.config.bench-mem.ts
@@ -1,0 +1,45 @@
+import {mergeConfig} from 'vitest/config';
+import {benchConfig} from '../shared/src/tool/vitest-config.ts';
+import {ChangeIndex} from '../zql/src/ivm/change-index.ts';
+import {ChangeType} from '../zql/src/ivm/change-type.ts';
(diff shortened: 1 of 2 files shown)


tantaman:
Why'd you end up closing?

arv:
I thought there might be a more general purpose approach to it. I explored that but it turned out to not bear fruit. I'll reopen.

arv:
PTAL

PR #6128, opened 2026-06-15
chore(zero-cache): log qid and name on failure

tantaman:
For incidents where a query fails, this will let us know _which_ query so we can ask the customer and debug.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
+++ b/packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
@@ -421,7 +421,7 @@ export class PipelineDriver {
     queryID: string,
     query: AST,
     timer: Timer,
-    queryName: string | undefined = undefined,
+    queryName?: string,
   ): Iterable<RowChange | 'yield'> {
     return this.#trackRowSetSignatures(
       this.#addQueryImpl(transformationHash, queryID, query, timer, queryName),
@@ -433,7 +433,7 @@ export class PipelineDriver {
     queryID: string,
     query: AST,
     timer: Timer,
-    queryName: string | undefined = undefined,
+    queryName?: string,
   ): Iterable<RowChange | 'yield'> {
     assert(
       this.initialized(),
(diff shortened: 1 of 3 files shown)


packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
```diff
@@ -421,6 +433,7 @@ export class PipelineDriver {
     queryID: string,
     query: AST,
     timer: Timer,
+    queryName: string | undefined = undefined,
```
arv: FYI


```suggestion
    queryName?: string,
```

is the same

packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
```diff
@@ -459,7 +473,7 @@ export class PipelineDriver {
           createStorage: () => this.#createStorage(),
           decorateSourceInput: (input: SourceInput, _queryID: string): Input =>
             new MeasurePushOperator(
-              input,
+              new QueryFailureLoggingOperator(this.#lc, input, queryInfo),
```
arv: Maybe pass the parts instead of an object for less object allocations. Does the object buy us anything?

packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
```diff
@@ -878,18 +901,37 @@ export class PipelineDriver {
...
+  readonly #logQueryFailure:
+    | ((queryID: string, error: unknown) => void)
+    | undefined;
 
   constructor(
     primaryKeys: Map<string, PrimaryKey>,
     tableSpecs: Map<string, LiteAndZqlSpec>,
+    logQueryFailure:
+      | ((queryID: string, error: unknown) => void)
+      | undefined = undefined,
```
arv: use optional here too for simplicity

arv (approved):
LGTM with nits/suggestions

arv (commented):
Thanks

Still LGTM

PR #6131, opened 2026-06-16
Add SQLite dialect support to compiler and SQL formatting

tantaman:
I want to beef up the fuzzer. Fuzzing against SQLite will reduce the footprint of all these tests.

----

### Motivation

- Enable emitting SQLite-compatible SQL from the existing compiler and reuse existing PG logic where possible.
- Provide internal formatting for SQLite so tests can compare concrete SQL text/params like the PG path.

### Description

- Added a `Dialect` field to `Spec` and a `compileSQLite` wrapper that calls `compile(..., 'sqlite')` to produce SQLite SQL.
- Implemented dialect-specific behavior throughout the compiler: `toJSON` now emits `json_object`/`json_group_array` for SQLite and `row_to_json`/`json_agg` for Postgres, `any` uses `json_each` for `IN` lists on SQLite, and `comparison` maps `ILIKE`/`NOT ILIKE` to `LOWER(...) LIKE LOWER(...)` for SQLite.
- Adjusted ordering/nulls handling with `nulls` helper to omit `NULLS FIRST/LAST` for SQLite and made `selectIdent` and other value conversion helpers dialect-aware.
- Added `resultKeys` to compute JSON output keys and updated call sites to pass `spec` and keys into `toJSON`.
- Extended SQL formatting utilities in `sql.ts` with `formatSqliteInternalConvert`, a `SQLiteConvertFormat` class, and updated `ReusingFormat` to support positional `?` placeholders for SQLite internal conversion.
- Tests updated to import `compileSQLite` and `formatSqliteInternalConvert` and three new SQLite integration-style output tests were added.

### Testing

- Ran the `vitest` suite for the modified package and the updated `packages/z2s/src/compiler.output.test.ts` tests executed successfully.
- Added tests: `compileSQLite emits full SQL for a basic query`, `compileSQLite emits full SQL for a related query`, and `compileSQLite emits full SQL for a singular scalar subquery`, and all three passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a318f4ad6748324944f22b193386f44)

### DIFF
--- a/packages/z2s/src/compiler.output.test.ts
+++ b/packages/z2s/src/compiler.output.test.ts
@@ -13,6 +13,7 @@ import type {ServerSchema} from '../../zero-types/src/server-schema.ts';
 import {
   any,
   compile,
+  compileSQLite,
   distinctFrom,
   limit,
   makeCorrelator,
@@ -22,7 +23,7 @@ import {
   simple,
   type Spec,
 } from './compiler.ts';
-import {formatPgInternalConvert} from './sql.ts';
+import {formatPgInternalConvert, formatSqliteInternalConvert} from './sql.ts';
 
 // Tests the output of basic primitives.
 // Top-level things like `SELECT` are tested by actually executing the SQL as inspecting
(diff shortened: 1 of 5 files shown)


tantaman:
thinking I'll stick with PG on second though given we support reading from PG via ZQL.

arv:
OK. I thought it would be nice to speed things up... Maybe PGLite for the fuzzer?

PR #6133, opened 2026-06-16
test(zql): pin ZQLite↔Postgres text semantics and document the ILIKE Unicode-folding divergence

tantaman:
## Why this test exists

ZQL queries are evaluated by **two different engines** that are supposed to be semantically identical:

- **ZQLite** — queries compiled to SQLite and run against the local replica (plus the in-memory IVM matcher on the client).
- **Compiled ZQL → SQL** — queries compiled by `z2s` and run directly on **Postgres**.

If those two disagree on something as fundamental as text comparison, ordering, `LIKE`/`ILIKE`, or `NULL` logic, a client can see different rows than the server authorized — silently. This PR adds an **oracle corpus** (`packages/zql-integration-tests/src/text-semantics.pg.test.ts`) that runs the same query through both paths over a deliberately nasty set of text values (null, empty, ASCII, mixed case, `%`/`_`/`\`, newline, combining vs composed accents, Greek, Cyrillic, emoji) and asserts the results match. It's a regression net for text semantics drift between SQLite and Postgres.

## What we found

While building the corpus, the two engines **disagree on case-insensitive matching (`ILIKE`) for non-ASCII text**:

| query (data: `Ωmega`, `ωmega`) | ZQLite | Postgres `COLLATE "C"` |
| --- | --- | --- |
| `value ILIKE 'ωmega'` | matches **both** | matches only `ωmega` |

Same story for Cyrillic (`ПРИВЕТ`/`привет`) and Latin-1 accents (`CAFÉ`/`café`). ASCII-only `ILIKE` agrees, which is why it went unnoticed.

## Why it happens

Case folding is done differently on each side:

- **ZQLite** compiles `ILIKE` to `lower(col) LIKE lower(pat)` using `@rocicorp/zero-sqlite3`'s **ICU-backed `lower()`**, which folds the *full Unicode range* (`Ω`→`ω`) independent of any collation. The in-memory IVM matcher folds the same way via JS `String.toLowerCase()` — so client and replica agree with each other (pinned by `zqlite/src/ilike-parity.test.ts`).
- **`z2s`** emits a **bare `ILIKE`**, so Postgres folds case according to the **column collation**. Under `COLLATE "C"` that's ASCII-only, so non-ASCII letters aren't folded.

The divergence is therefore **collation-dependent, not an operator bug**. The PR proves this with a control table: the identical `ILIKE` queries that diverge under `COLLATE "C"` *agree* under the database's default Unicode-folding collation (`en_US.utf8`), because Postgres then folds case the same way ZQLite always does. Verified deterministic across Postgres 15/16/17/18.

## What the test does about it (and the fix going forward)

This PR **documents and pins the current behavior** rather than hiding it:

- `text semantics oracle corpus` — parity cases that must always match (ordering uses `compareUTF8`, mirroring SQLite/`COLLATE "C"` byte order, so ordering already agrees).
- `known divergence: ILIKE folds Unicode in ZQLite but not under PG COLLATE "C"` — asserts the **exact, differing** id-sets each engine returns today, and asserts they differ. If the gap is ever closed, the pinned Postgres result changes and **this block goes red on purpose** — the cue to delete it and promote the cases into the parity corpus.
- `collation control: ILIKE agrees under default (Unicode-folding) collation` — locks in that the divergence is purely the `COLLATE "C"` choice.

The real fix is a separate decision (tracked in follow-up issues): make the two engines fold case the same way regardless of collation — e.g. have `z2s` lower-case per the column collation to match the replica, or have ZQLite respect collation. Until that lands, this corpus makes the divergence visible and prevents accidental further drift.

## Verification

- 19 tests pass on Postgres 15, 16, 17, and 18.
- `format`, `lint` (0 errors), and `check-types` all clean.

---
[Original Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a31982759488324bba2a2bc6d7013cc)

### DIFF
--- a/packages/zql-integration-tests/src/text-semantics.pg.test.ts
+++ b/packages/zql-integration-tests/src/text-semantics.pg.test.ts
@@ -3,6 +3,23 @@
  *
  * These tests pin cases where ZQLite/SQLite and compiled ZQL SQL on
  * PostgreSQL must agree for text comparison, LIKE/ILIKE, and NULL logic.
+ *
+ * It also pins one *known divergence*: case-insensitive matching (ILIKE) folds
+ * case differently on the two sides.
+ *  - ZQLite compiles ILIKE to `lower(col) LIKE lower(pat)` using
+ *    @rocicorp/zero-sqlite3's ICU-backed `lower()`, which folds the *full
+ *    Unicode* range (Ω→ω) regardless of any column collation. The in-memory IVM
+ *    matcher folds the same way via JS `String.toLowerCase()` (see
+ *    zqlite/src/ilike-parity.test.ts and zql/src/builder/like.ts).
+ *  - Compiled ZQL SQL emits a bare `ILIKE`, so Postgres folds case according to
+ *    the *column collation*. Under `COLLATE "C"` that is ASCII-only, so non-ASCII
+ *    letters are not folded.
+ *
+ * The result is that ILIKE over non-ASCII text agrees under a Unicode-folding
+ * collation (the default `en_US.utf8`) but diverges under `COLLATE "C"`. The
+ * `textItemC` table pins the divergence; the `textItemDefault` table pins that
+ * it disappears under the default collation. See the "known divergence" and
+ * "collation control" describe blocks below.
  */
 import {afterAll, beforeAll, describe, expect, test} from 'vitest';
 import {testLogConfig} from '../../otel/src/test-log-config.ts';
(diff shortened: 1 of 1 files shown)


arv:
We should also compare this with our JS impl of LIKE and ILIKE that is used on the client.

packages/zql-integration-tests/src/text-semantics.pg.test.ts
```diff
@@ -0,0 +1,323 @@
...
+  {id: '06-underscore', value: 'a_z'},
+  {id: '07-backslash', value: 'a\\z'},
+  {id: '08-newline', value: 'a\nb'},
+  {id: '09-omega', value: 'Ωmega'},
+  {id: '10-emoji', value: 'emoji😀x'},
+  {id: '11-combining', value: 'é'},
+  {id: '12-composed', value: 'é'},
+  {id: '13-z', value: 'z'},
+  {id: '14-zz', value: 'zz'},
+  // Non-ASCII cased rows for the ILIKE folding cases below.
```
arv, line 112: there is also the odd case with the greek final sigma

https://github.com/rocicorp/zero-sqlite3/commit/7b9bfb578418ba21464b1eb3a35c94a311255621#diff-0d0431f911eed43fcbe5be222c3b2e626e71df4003d4f85e59d2decf6b7cbcd8R59-R70

arv (commented):
Maybe we should change our "collator". It is sad but these semantic differences are warts.

arv (commented):
LGTM

PR #6136, opened 2026-06-17
test(zql-integration-tests): coverage-driven differential fuzzer for the ZQL/IVM engine

tantaman:
## What

A **coverage-driven differential fuzzer** for the ZQL/IVM engine, living in `packages/zql-integration-tests/src/chinook/fuzz/`. It replaces the old purely-random `chinook-fuzz-hydration.pg.test.ts` (removed here) with a structured, measured, reproducible generator stack.

**Oracle = Postgres via z2s.** Every generated query is checked three ways and must agree: the IVM **memory** view, the IVM **sqlite** view, and the **Postgres** result (z2s-compiled). The existing harness (`helpers/runner.ts`) supplies the oracle, parity comparison, and push protocol wholesale — this PR adds the generator + driver on top.

## How it's organized

Each layer is **deterministic in a printed seed** (the repro key) and any divergence is **auto-minimized** to a readable ZQL repro by a structure-aware shrinker.

**Generation layers**
- **L0** — bounded-exhaustive skeletons (depth ≤ 2), every shape lowered + hydrated.
- **L1** — a greedy **pairwise covering array** over `filter × exists × order × limit`, asserted to **100% pairwise** coverage, lowered onto every root *and* nested child collection.
- **L2 swarm** — random feature masks (bugs that surface only when a feature is *absent*).
- **L3 mutation** — corpus skeleton + one random twist ("simple + one twist", where minimal repros live).
- **L4 random tail** — deep random shapes (depth ≤ 4) past the enumerator's reach, behind a static **cost gate**.

**Correctness surfaces beyond plain hydration**
- **Four-phase push protocol** — remove-all / add-back / edit-to-random / edit-to-match on root + deepest leaf (incl. EXISTS-gate tables), parity re-checked **after every step**.
- **Decorated (top-N) push** — the `order`/`limit` × push cross-product the other lanes miss.
- **Flip-invariance** — every `2^k` flip assignment (semi-join vs `FlippedJoin`) of an EXISTS-bearing query must agree with the oracle (the plan choice z2s ignores).
- **Metamorphic relations** — oracle-free IVM-vs-IVM invariants (redundant conjunct, AND reorder, large limit, start-before-first).
- **Scalar subqueries** — production mirror: the original `scalar:true` AST runs through z2s (`col = (SELECT … LIMIT 1)`) while the IVM runs the **pre-resolved** AST via `resolveSimpleScalarSubqueries` — the exact transform zero-cache's pipeline-driver applies.
- **Random-yield interleave** — both IVM sources wrapped so a random fraction of fetch/push items is preceded by a `'yield'`, perturbing cooperative scheduling (the reentrancy axis inherited from the old fuzzer).
- **Regression corpus replay** — committed minimized repros are re-run as permanent guards.

## Test lanes

- **Per-PR (mini fixture, ~70s pg-16):** `chinook-fuzz-backbone` (L0 / L1 / push / flip / decorated-push / random-yield), `chinook-fuzz-sweep` (L2 / L3 / L4), `chinook-fuzz-scalar`, `chinook-fuzz-regressions`.
- **Nightly / on-demand (`CHINOOK_SCALE=1`, full chinook):** `chinook-fuzz-scale` — random tail + swarm + yield-tail at scale under the cost gate.
- **No-PG units (~3s):** `fuzz.test.ts` — 35 fast generator/coverage/resolver checks.

## Inert axes (documented, dropped)
z2s does not compile `start` (keyset paging) and mono ZQL has no `select`/projection, so both are dropped from generation (Postgres would silently ignore them and every case would diverge spuriously). Re-add `start` if z2s gains keyset support.

## Results
**No engine divergences found** across any layer. The fuzzer is green per-PR and at chinook scale.

## Test plan
```bash
# fast no-PG units
pnpm exec vitest run --project='*no-pg*' fuzz.test
# per-PR PG lanes
pnpm exec vitest run --project='*pg-16*' chinook-fuzz-backbone chinook-fuzz-sweep chinook-fuzz-scalar chinook-fuzz-regressions
# nightly scale lane
CHINOOK_SCALE=1 pnpm exec vitest run --project='*pg-16*' chinook-fuzz-scale
```


packages/zql-integration-tests/src/chinook/fuzz/rng.ts
arv, line 1: move to shared?

arv (commented):
LGTM

PR #6140, opened 2026-06-18
chore(zero-cache): interface for the backup monitor

tantaman:
We will be swapping to a litestream v5 backup monitor. This pulls out an interface so we can use either v3 or v5 while we migrate to v5.

I renamed the old `backup-monitor` to `litestream-backup-monitor` which shows up as a remove and an add which makes this diff so big :(


arv (commented):
LGTM

PR #6141, opened 2026-06-18
fix(zero): repair release Docker image build

0xcadams:
## Summary

- Add `protobufjs` to the Zero Docker image pnpm `allowBuilds` config because it is pulled in transitively by OpenTelemetry gRPC exporters and runs a postinstall script.
- Pin the Dockerfile pnpm version to the repo package manager version, `11.5.3`.
- Fix the Docker publish retry loop so the final failed attempt throws instead of reporting release success.
- Add a simple JS CI job that packs the local Zero package and verifies the Docker image builds on `linux/amd64` without pushing.

## What Happened

The `zero/v1.7.0-canary.2` release pushed the git tag and published npm, then the Docker build failed during `pnpm add @rocicorp/zero@1.7.0-canary.2` with `ERR_PNPM_IGNORED_BUILDS` for `protobufjs@7.6.4`. The Dockerfile creates a minimal pnpm workspace config and did not include the root repo allowlist entry for `protobufjs`.

The release script then retried Docker three times, but the retry condition used `i < 3` inside a `for (let i = 0; i < 3; i++)` loop, so the final failure was also swallowed after sleeping. The script printed success even though no image was pushed, and the later `cosign sign rocicorp/zero:1.7.0-canary.2` step failed with `MANIFEST_UNKNOWN` because the tag did not exist in Docker Hub.

### DIFF
--- a/.github/workflows/js.yml
+++ b/.github/workflows/js.yml
@@ -109,6 +109,44 @@ jobs:
       - run: pnpm install --frozen-lockfile
       - run: pnpm run check-types
 
+  zero-docker:
+    name: Build Zero Docker Image
+    runs-on: ubuntu-latest
+    needs: [syncpack, verify-deps]
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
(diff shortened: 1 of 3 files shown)


arv (commented):
LGTM

PR #6142, opened 2026-06-19
perf(zero-cache): batch changeLog inserts into multi-row INSERTs

arv:
## What
Persist upstream transactions to the CDC `changeLog` in **multi-row INSERT batches**
(`change.logBatchSize`, default **2000**, `hidden`) instead of one INSERT per change.

## Why (INC-810)
Zero replication lag on a customer's Supabase cluster grew to ~13 min in a staircase
pattern; the lag was ~100% upstream. Root cause: the change-streamer's `Storer` persisted
one single-row INSERT into `changeLog` per change — a 20k–30k-statement upstream
transaction became 20k–30k serial INSERTs taking 17–33s. Statements were already
pipelined, so the residual cost is per-statement server-side work; multi-row INSERTs
eliminate most of it.

## How
- Per-change hot loop accumulates rows into `PendingTransaction.batch`, flushed as
  multi-row `INSERT INTO changeLog ${sql(rows)}`.
- Flush at `changeLogBatchSize` (awaiting the previous flush, pipeline depth ~2) and at
  commit before the `replicationState` UPDATE. Schema changes flush-first, then write the
  schema row + backfill metadata as one unit.
- `MAX_CHANGELOG_INSERT_ROWS = 5000` caps rows/statement under PG's 65535 bind-param limit.
  `changeLogBatchSize` clamped `>= 1` (1 = disabled).

## Benchmark (new storer-bench.pg.test.ts, ZERO_BENCH=1; local pg-16)
| Changes | unbatched | batched | speedup |
|---|---|---|---|
| 20,000 | 1234 ms, 126.6 MB | 94 ms, 15.9 MB | 13.1× |
| 50,000 | 3395 ms, 325.2 MB | 257 ms | 13.2× |

## Test plan
check-types/lint/format clean; storer.pg (88) + change-streamer-service.pg (104) pass
across pg-15..18; zero-config `--help` snapshot green; bench skips without ZERO_BENCH.


arv (commented):
## Review: batch changeLog inserts into multi-row INSERTs

Reviewed the full `storer.ts` change against the `TransactionPool` semantics, the `changeLog` schema, and the config/integration wiring. **The core mechanism is correct.** The one substantive item is a test-coverage gap; everything else is confirmation or nits.

### Why the batching is correct

- **Single-worker FIFO pool guarantees order.** The tx `TransactionPool` is created with no `initialWorkers`/`maxWorkers`, so it defaults to a single worker draining tasks in submission order over one connection. The multiple `process()` calls issued by `#flushChangeLog` (and the schema-change path) therefore execute in order. `await prevFlush` is purely backpressure/pipeline-depth bounding — it is **not** what guarantees ordering, which is the robust design.
- **Row identity is `PRIMARY KEY (watermark, pos)`** and every row carries its own `pos`, so the order of rows *within* a multi-row INSERT is irrelevant. Collapsing N single-row INSERTs into one is value-preserving.
- **Lock ordering preserved** (`SELECT replicationState FOR UPDATE` → `INSERT changeLog …` → `UPDATE replicationState`): the `FOR UPDATE` is queued at `begin`, flushes in the middle, the `UPDATE` at commit — FIFO keeps it intact.
- **Schema changes** flush the buffered batch first, then write the schema row + backfill metadata as one `process()` unit, preserving stream order.
- **Rollback** pushes its entry to `tx.batch` but never flushes it before `pool.abort()`, so aborted txns leave nothing in `changeLog`.
- **Bind-param cap** is right: `MAX_CHANGELOG_INSERT_ROWS = 5000` × 4 cols = 20k params, well under PG's 65535; oversized configured batches are split rather than failing.

### Verified by deeper tracing

- **Full config→Storer wiring is complete and compiler-enforced.** `change.logBatchSize` → `change-streamer.ts` opts → `initializeStreamer`/`ChangeStreamerImpl` (`TuningOptions = StorerOptions & {…}`) → `new Storer(..., opts)`. Because `changeLogBatchSize` is a required field, TS forces every construction site to supply it — no call site missed.
- **No `change`-column encoding regression.** The prod change-DB connection uses `{sendStringAsJson: true}` (`server/change-streamer.ts`), which the bench mirrors. postgres.js serializes each `change` value through the same per-value hook in multi-row `sql(rows)` as in single-row `sql(entry)`.
- **Heterogeneous `precommit` within a batch is safe** — all `ChangeLogRow`s share the identical key set `{watermark, precommit, pos, change}`, so postgres.js column inference from `rows[0]` is stable; `null` vs string is just a per-row bound value.
- **No false ACK; error-surfacing point unchanged.** `process()` resolves (never rejects) even after `fail()`, and `done()` rejects on worker failure — so a failed flush surfaces only at the commit-time `await tx.pool.done()`, before the ACK. The old `await processed` every 100 statements also never rejected, so the per-batch await does not change where errors are detected.
- Edge cases walked and confirmed: schema-only txn, consecutive schema changes, batch boundary landing exactly at commit (no double-insert), `>=` off-by-one (push→check→flush includes the triggering row), and the empty-batch guard preventing `sql([])`.

### Suggestion (not a blocker): cover the batching in CI

Both test opts use `changeLogBatchSize: 2000` with small transactions, so **nothing in CI exercises the batching itself** — neither the mid-transaction flush branch (`tx.batch.length >= this.#changeLogBatchSize`) nor the `MAX_CHANGELOG_INSERT_ROWS` slice loop is hit outside the opt-in (`ZERO_BENCH=1`) bench. A small `.pg.test.ts` case with `changeLogBatchSize: 2` (or `3`), feeding a handful of data changes with one schema change interleaved and asserting `changeLog` row count + `pos` ordering, would lock in the cross-flush ordering invariant the whole change rests on.

### Minor nits
- `changeLogBatchSize = 1` ("disabled") becomes one-INSERT-per-change but awaits the previous flush on *every* change (depth-2), so it's slightly more serial than the prior every-100 await — "disable batching" doesn't restore the old pipelining exactly. Harmless.
- `logBatchSize` has no `.min(1)` on the schema; it relies on the runtime `Math.max(1, …)` clamp. A `v.number().min(1)` would surface a misconfigured `0`/negative instead of silently clamping.
- Bench heap-delta figures are indicative only (`process.memoryUsage().heapUsed` without `global.gc()`); the hard gate `batchedMs <= unbatchedMs` is robust given the ~13× headroom.

CI is fully green (Check Types, Lint, Oxfmt, Test PG 15–18 × shards, API Snapshot, Docker build); the PR is just `behind` `main` and needs a rebase. Nice win — the per-statement server-side cost was clearly the right thing to attack.


---

packages/zero-cache/src/services/change-streamer/storer.ts
```diff
@@ -358,6 +387,27 @@ export class Storer implements Service {
...
+  #flushChangeLog(tx: PendingTransaction): Promise<unknown> | undefined {
+    const {batch} = tx;
+    if (batch.length === 0) {
+      return tx.lastFlush;
+    }
+    tx.batch = [];
+    for (let i = 0; i < batch.length; i += MAX_CHANGELOG_INSERT_ROWS) {
+      const rows = batch.slice(i, i + MAX_CHANGELOG_INSERT_ROWS);
+      tx.lastFlush = tx.pool.process(sql => [
+        sql`INSERT INTO ${this.#cdc('changeLog')} ${sql(rows)}`,
```
darkgnotic: Very cool!

One gotcha with using a plain batched `INSERT` is that, because postgres.js uses prepared statements, we can end up running out of memory in the postgres DB because it caches all of these prepared statements of different lengths.

The solution that I arrived at for a similar scenario (writing the batch of CVR row records) is to use `json_to_recordset()`, which actually ended up being faster than multi-insert statements:

https://github.com/rocicorp/mono/pull/3511

(My previous attempt to address this was to pin batch sizes to powers of 2, which was an improvement, but ultimately not as good as the single json_to_recordset() approach: https://github.com/rocicorp/mono/pull/3395)

Can you give the json_to_recordset() approach a go? Hopefully it will perform reasonably well in the bench mark.

PR #6144, opened 2026-06-19
feat(zero-cache): forward incoming request headers to API

tjenkinson:
This adds 2 new options
- `mutate-allowed-request-headers`
- `query-allowed-request-headers`

which are an allow list of headers to forward from the incoming request upstream.

We need to record the IP address of the caller, which arrives on an `x-forwarded-for` header. We can't provide this in the client as it needs to come from the load balancer server side. I think allowing any header to be forwarded is a useful feature but open to other suggestions :)

Found a related bug here in the bug tracker: https://bugs.rocicorp.dev/p/zero/issue/3928

Extracted from https://github.com/rocicorp/mono/pull/6109


0xcadams (approved):
LGTM - thanks for the contribution!

PR #6145, opened 2026-06-19
Add `safeSet` / `safeAssign` to protect against prototype pollution

tjenkinson:
**This is stacked on https://github.com/rocicorp/mono/pull/6144 but I can't change the base branch here**

I spotted a lot of `Object.assign`, including when checking headers, and replaced them with a safe version that can't pollute the prototype. Similar issue in places where we're setting a property directly on an object. If the property is `__proto__` and the object isn't one with a null prototype that could cause an issue.

Extracted from https://github.com/rocicorp/mono/pull/6109

### DIFF
--- a/packages/shared/src/build-nested-object-path.ts
+++ b/packages/shared/src/build-nested-object-path.ts
@@ -1,4 +1,5 @@
 import type {DeepMerge} from './deep-merge.ts';
+import {safeSet} from './objects.ts';
 
 /**
  * Helper function to build nested object structure from a dot-separated key path.
@@ -38,15 +39,15 @@ export function buildNestedObjectPath<
   let current = target as Record<string, unknown>;
   for (let i = 0; i < parts.length - 1; i++) {
     const part = parts[i];
-    let next = current[part];
+    let next = Object.hasOwn(current, part) ? current[part] : undefined;
     if (next === undefined) {
       next = {};
-      current[part] = next;
+      safeSet(current, part, next);
     }
     current = next as Record<string, unknown>;
   }
   // oxlint-disable-next-line typescript/no-non-null-assertion
-  current[parts.at(-1)!] = value;
+  safeSet(current, parts.at(-1)!, value);
   return target as DeepMerge<T, BuildNested<Path, Separator, Value>>;
 }
 
--- a/packages/shared/src/deep-clone.ts
+++ b/packages/shared/src/deep-clone.ts
@@ -1,5 +1,6 @@
 import {hasOwn} from './has-own.ts';
 import type {JSONValue, ReadonlyJSONValue} from './json.ts';
+import {safeSet} from './objects.ts';
 
 export function deepClone(value: ReadonlyJSONValue): JSONValue {
   const seen: Array<ReadonlyJSONValue> = [];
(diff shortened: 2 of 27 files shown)


0xcadams (commented):
I'm not following the benefit here. We have an allow-list that is controlled on the zero-cache side for the headers. So that would filter a bad `__proto__` header or whatever the bad thing is that is being sent. Can you expand on the problem that this is this solving? Maybe I'm missing it.

packages/shared/src/build-nested-object-path.ts
```diff
@@ -38,15 +39,15 @@ export function buildNestedObjectPath<
   let current = target as Record<string, unknown>;
   for (let i = 0; i < parts.length - 1; i++) {
     const part = parts[i];
-    let next = current[part];
+    let next = Object.hasOwn(current, part) ? current[part] : undefined;
```
tjenkinson, line 42: this one prevents traversing and setting things on the `constructor` property, which would mutate the global object

tjenkinson: Unless `constructor` is an actual property that was set and not being inherited from the prototype

If you're happy with this PR let me know and I'll add a test specifically for this and fix the conflicts

arv:
This is a very subtle change and I do not think we can accept it. A few comments.

`Object.assign` only copies own enumerable properties. `__proto__` is very special. It is non enumerable by default. That is why `Object.assign` is safe in general.

Using objects as maps is always problematic. It is always better to use a `Map` where possible (performance of Map is good) or at least `Object.create(null)`.

When it comes to user facing APIs we tend to use own enumerable properties for convenience. (`Object.keys`/`Object.entries` etc). Allowing enumerable non own properties would be convenient but it is a big change with not enough value at this point.

arv:
The general guideline is to not use keys named `__proto__`. If there are APIs that allows this to come from user data we should fix those.

I do not want to change this at this point in time.

arv:
JSON.parse creates enumerable own data properties.

arv:
The other option is to document that tables, columns, relations named `__proto__` are not supported.

I think we would need to figure out where, if anywhere, we want to support own enumerable keys named `__proto__`.

I do not want to add this extra code check everywhere we use objects. `Object.defineProperty` is a lot slower than `[[Set]]`

One way to deal with this is to use branded types. I haven't thought this through fully but we could tag the input types and then make it an error to blindly use Object.assign for this type:

```ts
type Dict<T> = Record<string, T> & {[brand]: true}

function parseDict<T>(v: Record<string, T>) {
  return v as Dict<T>;
}

function objectAssign<T>(o: Dict<T>, v: Record<string, T>): never;
function objectAssign<T>(o: Record<string, T>, v: Record<string, T>): ...;
```

arv:
I think we are better off starting with identifying which APIs should support `__proto__`.

- Table names
- Column names
- Relationship names
- Queries
- Mutators

But I'm concerned about SolidJS (and others) use of `Proxy.set`.

I think we should try one (column names) and see how it turns out.

arv:
> > But I'm concerned about SolidJS (and others) use of Proxy.set
> 
> You mean they'd provide a proxy into zero? The clone function writes to a new none proxy object so don't think there'd be any change in behaviour there

SolidJS uses a view that uses `[[Set]]` to know about the changes. We need to ensure that we keep using set/assign in the view but maybe we can make the view use `Object.create(null)` to prevent `__proto__` from being problematic.

PR #6147, opened 2026-06-20
fix(zero-server): schema-qualify casts to non-public Postgres types

jschaf:
Server-side CRUD/query SQL cast values to their Postgres column type (e.g. `$1::text::"some_enum"`). For enum and other user-defined types the generated cast used only the bare type name, which Postgres resolves through `search_path`. A table in a non-public schema whose enum columns use enum types in that same schema therefore failed with `type "..." does not exist` under the default search_path.

Root cause: `getServerSchema` joined `pg_namespace` but never recorded the type's namespace, so `ServerColumnSchema` kept only `typname`; the SQL generator then rendered the cast as a single bare/quoted identifier.

Fix:
- Record the type's namespace (element type's namespace for arrays) as `ServerColumnSchema.typeSchema`, collected from the catalog. Also match the type by name AND `udt_schema` so same-named types in different schemas no longer resolve ambiguously.
- When generating casts, schema-qualify user-defined types that live outside `public`/`pg_catalog` (e.g. `$1::text::"app"."some_enum"`), using `escapePostgresIdentifier` for both segments. Public and built-in types stay byte-compatible with prior output.

### DIFF
--- a/packages/z2s/src/sql.test.ts
+++ b/packages/z2s/src/sql.test.ts
@@ -980,3 +980,152 @@ describe('string arg packing', () => {
     });
   });
 });
+
+describe('schema-qualified user-defined types', () => {
+  test('enum in a non-public schema is schema-qualified', () => {
+    expect(
+      formatPgInternalConvert(
+        sql`INSERT INTO "app"."doc_review_item" ("kind") VALUES (${sqlConvertColumnArg(
+          {
+            isArray: false,
+            isEnum: true,
+            type: 'doc_review_item_kind',
+            typeSchema: 'app',
+          },
+          'edge_case',
+          false,
+          false,
+        )})`,
+      ),
+    ).toMatchInlineSnapshot(`
+      {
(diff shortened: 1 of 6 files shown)


0xcadams:
@tantaman maybe you're the best person to take a look at this?

jschaf:
As a meta note, I have a few more fixes queued up. I'm happy to adjust how or who I send them to. Fixes:

- Recursively resolve Postgres domain types to get the underlying base type to use for the decoder.
- Support single-process mode with a flag. Cuts cold start time from 1.8 seconds to 0.5 seconds.

I have a fair bit of experience wrangling Postgres types, having written my own ORM: https://github.com/jschaf/pggen. Here's a fun [query](https://github.com/jschaf/pggen/blob/main/internal/pg/query.sql#L81) to resolve composite types

0xcadams:
Very cool. Any help is great. Thanks for the offer. If you open separate PRs, we'll (eventually) get eyes on them and route them to the best person.

PR #6148, opened 2026-06-20
fix: support drizzle-orm 1.0-rc+ prepareQuery signature in the drizzle adapter

typedrat:
## Problem

`zero-server`'s drizzle adapter calls `PgSession.prepareQuery` with the **pre-1.0 positional signature**:

```ts
// packages/zero-server/src/adapters/drizzle.ts
this.wrappedTransaction._.session.prepareQuery(
  {sql, params},
  undefined,   // fields
  undefined,   // name
  false,       // isResponseInArrayMode
);
```

`drizzle-orm` changed that signature in the 1.0 RC line:

| version | `prepareQuery(...)` |
|---|---|
| `<= 1.0.0-beta` | `(query, fields, name, isResponseInArrayMode, ...)` |
| `>= 1.0.0-rc.1` | `(query, mode: 'arrays' \| 'objects' \| 'raw', name, mapper?, ...)` |

So on a newer drizzle, the `undefined` in slot 2 is the **`mode`**, not `fields`. drizzle then returns rows as **arrays** instead of objects. `getServerSchema` introspects `information_schema` via this `query()` method and reads `row.dataType` — which is `undefined` on an array row — so **every** server-mutator transaction fails at the first `.toLowerCase()`:

```
Cannot read properties of undefined (reading 'toLowerCase')
```

This makes `@rocicorp/zero`'s custom-mutator push completely unusable with `drizzle-orm@1.0.0-rc.*` (and any 1.0 release). It bites at runtime only — the call still type-checks because `zero` pins `drizzle-orm@^0.45`.

## Fix

Keep the old signature as the typed path (zero is built against `drizzle-orm@^0.45`), and detect a 1.0-rc+ drizzle at runtime to call the new object-mode signature:

```ts
const prepared = drizzleUsesModeArg(session.prepareQuery)
  ? (session.prepareQuery as unknown as ModePrepareQuery)(query, 'objects', undefined, undefined)
  : session.prepareQuery(query, undefined, undefined, false);
```

Detection is **empirical** — no version string, no `package.json` read. The only difference that matters is the second parameter: the new API names it `mode`, the old one names it `fields`. A small `detect-drizzle-version` module reads that name off `Function.prototype.toString(prepareQuery)`, with a parameter-count fallback if the source is minified (the old signature has more leading params than the new one). On any failure it falls back to the old signature — i.e. no behavior change for the pinned `drizzle-orm@0.45`.

This drops the `semver` dependency entirely (it was only added for this), and isolates the detection in its own module so it's unit-testable without exposing it on the adapter's public surface.

## Notes / validation

- No behavior change for `drizzle-orm@0.45`; the new branch only runs on `>= 1.0.0-rc.1`.
- Detection logic is covered by `detect-drizzle-version.test.ts`: both real signatures (matching drizzle's actual parameter names) plus both minified arity-fallback paths.
- Validated locally: `oxfmt --check` clean, `oxlint` clean (0/0), `tsc` clean, and the adapter tests pass. I wasn't set up to run the full monorepo suite locally (the `pnpm` install hook runs `prisma generate`, which fails in my environment), so I'm leaning on CI for the rest — happy to adjust.
- Discovered while running `@rocicorp/zero@1.6.1` against `drizzle-orm@1.0.0-rc.3`. I'd previously shipped this as a local `pnpm patch`; this is the upstream version.

Let me know if you'd rather just bump the `drizzle-orm` dev/peer range to 1.0 and switch unconditionally instead of supporting both — happy to take it whichever direction you prefer.

### DIFF
--- a/packages/zero-server/src/adapters/detect-drizzle-version.test.ts
+++ b/packages/zero-server/src/adapters/detect-drizzle-version.test.ts
@@ -1,0 +1,70 @@
+import {describe, expect, test} from 'vitest';
+
+import {detectUsesModeArg} from './detect-drizzle-version.ts';
+
+describe('detectUsesModeArg', () => {
+  test('detects the new (>= rc.1) signature by its `mode` parameter', () => {
+    // Mirrors `drizzle-orm@>=1.0.0-rc.1` `PgSession.prepareQuery`. Parameter
+    // names must match drizzle's exactly — detection reads them off the source.
+    function prepareQuery(
+      query: unknown,
+      mode: 'arrays' | 'objects' | 'raw',
+      name: unknown,
+      mapper?: unknown,
+      queryMetadata?: unknown,
+      cacheConfig?: unknown,
+    ) {
+      void [query, mode, name, mapper, queryMetadata, cacheConfig];
+    }
+    expect(detectUsesModeArg(prepareQuery)).toBe(true);
+  });
+
+  test('detects the old (<= beta / 0.45) signature by its `fields` parameter', () => {
+    // Mirrors `drizzle-orm@^0.45` `PgSession.prepareQuery`. Parameter names
+    // must match drizzle's exactly — detection reads them off the source.
+    function prepareQuery(
+      query: unknown,
+      fields: unknown,
+      name: unknown,
+      isResponseInArrayMode: boolean,
+      customResultMapper?: unknown,
+      queryMetadata?: unknown,
+      cacheConfig?: unknown,
+    ) {
+      void [
+        query,
+        fields,
+        name,
+        isResponseInArrayMode,
+        customResultMapper,
+        queryMetadata,
+        cacheConfig,
+      ];
+    }
+    expect(detectUsesModeArg(prepareQuery)).toBe(false);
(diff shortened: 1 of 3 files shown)


packages/zero-server/package.json
```diff
@@ -19,13 +19,15 @@
     "check-fmt": "oxfmt --check ."
   },
   "dependencies": {
-    "@rocicorp/logger": "^6.0.0"
+    "@rocicorp/logger": "^6.0.0",
+    "semver": "^7.5.4"
```
0xcadams: Can we do this without a new dependency? Is there a property we can detect at runtime and not the package json?

0xcadams (changes requested):
Thanks for the contribution. We shouldn't need to add dependencies for this. I can pick this up or if you want to do the research to find a simpler way to do 1.0 detection.

typedrat:
Okay, this is a little messier but avoids the dependency entirely.

0xcadams:
Thanks for your contribution! I made some changes but couldn't push to your branch so opened a separate PR.

PR #6149, opened 2026-06-23
fix(zero-cache): fix logging of catchup wait time

arv:
The method is not bound to the logger, so we need to call it as a method and not as an extracted function. This was causing the logging to fail and throw an error.


darkgnotic (commented):
LGTM

PR #6150, opened 2026-06-23
chore: update @rocicorp/logger to version 6.1.0

arv:
This is to get the type checking of non bound methods


arv:
I verified that this caught the error that was fixed in #6149

PR #6152, opened 2026-06-23
chore: extra safety before swapping to litestreamv5

tantaman:
Checks if the litestream v5 executable is set before trying to use a v5 backup monitor.


arv (approved):
LGTM

PR #6154, opened 2026-06-23
fix(zero-server): support drizzle rc prepareQuery

0xcadams:
Closes #6148.

## Summary

Updates the Zero Drizzle adapter to support both `drizzle-orm@0.45.x` and the `drizzle-orm@1.0.0-rc` `prepareQuery` signature.

Drizzle changed `PgSession.prepareQuery` from the old positional form:

```ts
prepareQuery(query, fields, name, isResponseInArrayMode)
```

to the RC form:

```ts
prepareQuery(query, mode, name, mapper)
```

Zero was still calling the old form. With Drizzle RC, that leaves `mode` unset and causes raw query results to come back in array mode instead of object mode.

## Changes

- Removes direct internal Drizzle type imports from the adapter.
- Adds small local structural types for the Drizzle database/session/transaction shape Zero needs.
- Detects the `prepareQuery` signature with `session.prepareQuery.length`.
- Uses the old call form for `<1`.
- Uses the RC call form with `'objects'` mode for `>=1.0.0-rc`.

## Verification

Tested against the repo's current `drizzle-orm@0.45.2`:

```sh
pnpm --filter zero-server run format
pnpm --filter zero-server run lint
pnpm --filter zero-server run check-types
pnpm --filter zero-server run test:pg17 -- src/adapters/adapters.pg.test.ts
```

Also temporarily verified against `drizzle-orm@1.0.0-rc.3` with a workspace override and a temporary runtime smoke test covering:

- `drizzle-orm/node-postgres`
- `drizzle-orm/postgres-js`
- RC constructor form: `drizzle({client})`
- `zeroDrizzle(...).run(...)`
- `tx.dbTransaction.query(...)`
- object-row results from the new `prepareQuery(query, 'objects', ...)` path

The temporary override and smoke test were removed afterward; no dependency-file changes are included.

## Attribution

Based on the original work by @typedrat in #6148.

### DIFF
--- a/packages/zero-server/src/adapters/drizzle.ts
+++ b/packages/zero-server/src/adapters/drizzle.ts
@@ -1,9 +1,3 @@
-import type {
-  PgDatabase,
-  PgQueryResultHKT,
-  PgTransaction,
-} from 'drizzle-orm/pg-core';
-import type {ExtractTablesWithRelations} from 'drizzle-orm/relations';
 import type {AST} from '../../../zero-protocol/src/ast.ts';
 import type {Format} from '../../../zero-types/src/format.ts';
 import type {Schema} from '../../../zero-types/src/schema.ts';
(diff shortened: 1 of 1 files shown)


packages/zero-server/src/adapters/drizzle.ts
```diff
@@ -92,12 +123,22 @@ class DrizzleInternalTransaction<
...
-    const prepared = this.wrappedTransaction._.session.prepareQuery(
-      {sql, params},
-      undefined,
-      undefined,
-      false,
-    );
+    const {session} = this.wrappedTransaction._;
+    const query = {sql, params};
+    const prepared =
+      session.prepareQuery.length < 7
```
arv, line 129: it is kind of scary to rely on the length here. Is there no other signal we can use?

arv (commented):
LGTM

PR #6155, opened 2026-06-23
perf(zero-cache): copy largest tables first during initial sync

0xcadams:
Sort initial-sync table copy work by estimated `totalBytes` descending.

## Why

After a logical replication slot invalidation, Zero may need to rebuild the replica with a full initial sync. For large datasets, the copy phase can dominate recovery time.

Initial sync currently submits table copy tasks in publication order. That can leave the largest tables until later, causing worker tail time when smaller table copies finish and one expensive table remains.

Submitting larger estimated tables first is a small, safe scheduling improvement that should reduce tail latency without changing replica contents or sync semantics.


aboodman:
Don't they all run in parallel? It's hard to imagine the order they are
launched in making any difference.

PR #6157, opened 2026-06-25
feat(zero-cache): calculate how far behind a view syncer is on serving clients

tantaman:
### What We Want To Measure

  For each active client group, we want to know:

  > How long has there been replica data ready locally that this ViewSyncer has not finished serving to clients yet?

  That time is `sync.serving-lag`.

  ---

  ## The Pieces

  ### Replica Ready State

  When the local replica finishes applying a change, we record:

  ```ts
  {
    watermark: '04',
    replicaReadyTimeMs: 123456789,
  }

  Think of watermark as “version number” and replicaReadyTimeMs as “when that version became readable.”

  ### ViewSyncer State

  Each active ViewSyncer exposes:

  {
    createdAtMs,
    servedVersion,
  }

  - createdAtMs: ignore replica changes that were already ready before this ViewSyncer existed.
  - servedVersion: the latest watermark this ViewSyncer has fully served through pokeEnd.

  ———

  ## Flow

  1. Replication applies a change
      - IncrementalSyncer processes upstream changes into the local SQLite replica.
      - When a new watermark is ready, it sends:

```js
  {
    state: 'version-ready',
    watermark,
    replicaReadyTimeMs: Date.now(),
  }
```

  2. Notifier forwards the message
      - The replicator notifier sends this to subscribers.
      - If notifications coalesce, it keeps the newest watermark but preserves the earliest ready time so lag is not
        underreported.

  3. Syncer records ready states
      - Syncer receives the notification.
      - If it has both watermark and replicaReadyTimeMs, it appends it to #replicaReadyStates.
      - If there are no active ViewSyncers, it clears the list immediately.
      - Otherwise it caps the list at 10_000 entries.

  4. ViewSyncers serve changes
      - ViewSyncers advance pipelines, flush CVRs, and send pokeEnd.
      - Only after pokeEnd completes do they update servedVersion.

  5. OTel gauge computes lag
      - When metrics collection runs, the sync.serving-lag gauge calls computeMaxServingLagMs.

  ———

  ## Computation

  For each active ViewSyncer:

  1. Find the first replica-ready state after createdAtMs.
  2. Find the first replica-ready state with watermark > servedVersion.
  3. Use the later of those two positions.
  4. Lag is:

  now - replicaReadyTimeMs

  The metric reports the max lag across all active ViewSyncers.

### DIFF
--- a/packages/zero-cache/src/services/replicator/incremental-sync.ts
+++ b/packages/zero-cache/src/services/replicator/incremental-sync.ts
@@ -77,7 +77,12 @@ export class IncrementalSyncer {
       await this.#worker.getSubscriptionState();
 
     // Notify any waiting subscribers that the replica is ready to be read.
-    void this.#notifier.notifySubscribers();
+    // This initial notification intentionally omits replicaReadyTimeMs because
+    // it represents already-current state, not newly-unserved work.
+    void this.#notifier.notifySubscribers({
+      state: 'version-ready',
+      watermark: initialWatermark,
+    });
 
     while (this.#state.shouldRun()) {
       const {replicaVersion, watermark} =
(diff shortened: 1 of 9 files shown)


tantaman:
Revising this to compute percentiles. Max is not helpful signal. E.g., imagine if only 1 client is wedged but all other 1k clients are fine.

PR #6159, opened 2026-06-25
feat(ci): add human/approval gate to block agent self-merges

arv:
Adds a required-status-check workflow that publishes a `human/approval` check. The check is green when the PR author is a roci.dev human (org OWNER/MEMBER, not an agent) OR a roci.dev human other than the author has an active approving review.

Because agents run under a distinct bot identity, they cannot author a green PR or approve one, so they can never merge their own work once the check is required in the branch ruleset. Human-authored PRs pass immediately, so post-commit review is unaffected.

Mirrors the existing codeowners-approval workflow's Checks-API pattern.

### DIFF
--- a/.github/workflows/human-approval.yml
+++ b/.github/workflows/human-approval.yml
@@ -1,0 +1,190 @@
+name: Human Approval
+
+# Gate that prevents agent (bot) identities from landing changes unilaterally.
+#
+# Publishes a `human/approval` check that is green when EITHER:
+#   - the PR author is a @roci.dev human (org OWNER/MEMBER, not an agent), OR
+#   - a @roci.dev human other than the author has an active approving review.
+#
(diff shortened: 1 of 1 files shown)


.github/workflows/human-approval.yml
```diff
@@ -0,0 +1,190 @@
...
+              core.setFailed('No pull_request payload.');
+              return;
+            }
+
+            const CHECK_NAME = 'human/approval';
+
+            // Agent / bot identities that must never satisfy the gate on their
+            // own. `type === 'Bot'` covers GitHub App identities (e.g.
+            // claude[bot]); add any dedicated machine-account logins here.
+            const AGENT_LOGINS = new Set(
```
arv, line 54: Lets not hard code these. Basically, any PR that is not authored by @roci.dev should be flagged

PR #6160, opened 2026-06-26
bench: added pg benchmarks for replication

0xcadams:
## Summary

Adds a PostgreSQL benchmark suite for `zero-cache` that can be run with:

`pnpm --filter zero-cache run bench:pg`

The suite uses generated PG fixtures so we can compare release branches more consistently.

## What Changed

- Adds `bench:pg` / `bench:pg:bmf` scripts and a PG-specific Vitest benchmark config.
- Adds shared PG benchmark fixture helpers for deterministic multi-table fixture data.
- Adds initial sync throughput benchmark from generated PG fixture into SQLite.
- Adds change-streamer `Storer` throughput benchmarks for single-transaction and sustained-stream workloads.
- Adds end-to-end logical replication throughput benchmark from upstream PG writes through change-streamer into the SQLite replica.
- Extends the shared benchmark recorder so manual async benchmarks can emit the same formatted and JSON benchmark output as existing mitata benchmarks.


arv (commented):
LGTM

PR #6161, opened 2026-06-26
feat(release): also publish Docker image to GHCR

arv:
Publish the zero Docker image to ghcr.io/rocicorp/zero alongside the existing Docker Hub image. A single buildx build pushes both tags, so SBOM and provenance attestations apply to both registries.

Authenticate to GHCR with the ephemeral, run-scoped GITHUB_TOKEN (packages: write) instead of a stored PAT, and extend cosign keyless signing (OIDC) to the GHCR image. make-latest.js retags and verifies the `latest` digest on GHCR as well.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,6 +33,7 @@ jobs:
     permissions:
       contents: write # Required to push release tags.
       id-token: write # Required for cosign keyless signing.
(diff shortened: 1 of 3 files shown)


0xcadams (commented):
LGTM, nice

PR #6162, opened 2026-06-26
refactor(release): split build and Docker Hub publish into separate jobs

arv:
Follow-up to #6161 (GHCR publishing), now rebased onto `main` — this PR contains **only** the build/publish job split.

## Goal

Reduce the blast radius of the **long-lived `DOCKERHUB_TOKEN`** by isolating it to a minimal publish job. It's the only stored, long-lived secret in the release workflow — npm publish (OIDC trusted publishing), the git tag push (`GITHUB_TOKEN`), the GHCR push (`GITHUB_TOKEN`), and cosign (OIDC keyless) all use ephemeral, run-scoped credentials already.

## How

Split the single `release` job into two:

- **`build` — "Build & stage to GHCR"** (`contents: write`, `id-token: write`, `packages: write`; **no `DOCKERHUB_TOKEN`**). Runs all the heavy, untrusted code: `pnpm install`, the workspace build, and the Dockerfile build (which execute dependency lifecycle scripts). Builds, npm-publishes, git-tags, and pushes the image to **GHCR only**. Signs the GHCR image.
- **`publish` — "Promote to Docker Hub"** (`needs: build`; **only `DOCKERHUB_TOKEN`** + `id-token` for signing + `packages: read`). Runs no checkout/install/build. Promotes the staged GHCR image to Docker Hub via `docker buildx imagetools create` (copying the multi-arch manifest + SBOM/provenance attestations verbatim), signs the Docker Hub image, and comments on the PR.

`release.ts`'s `pushDocker` now pushes to GHCR only; Docker Hub publishing becomes the promote step. The two jobs are chained via `needs:` and run automatically in one workflow run — the split is purely for credential isolation, not an extra manual step.

This keeps the sensitive Docker Hub credential away from all the dependency/Dockerfile code execution — it only ever touches a job running `imagetools` + `cosign`. No build-from-tarball / Dockerfile changes were needed because the Docker build still pulls from npm after npm publish, all within the build job.

## Heads-up for branch protection

The required status check names change from `Release` to `Build & stage to GHCR` and `Promote to Docker Hub`. If branch protection requires the old `Release` check, it'll need updating.

> [!NOTE]
> Not validated end-to-end (no Actions run available in the authoring environment). The `dry_run` workflow_dispatch input is the safest way to smoke-test before a real release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,16 +25,23 @@ concurrency:
   cancel-in-progress: ${{ github.event_name != 'workflow_dispatch' }}
 
 jobs:
-  release:
-    name: Release
+  # Build job: runs all the heavy, untrusted code (pnpm install, the workspace
+  # build, and the Dockerfile build, which execute dependency lifecycle scripts).
+  # It deliberately does NOT have the long-lived Docker Hub credential. It uses
+  # only ephemeral, run-scoped credentials: OIDC for npm provenance + cosign, and
+  # GITHUB_TOKEN for the git tag push and the GHCR image push. The image is staged
+  # to GHCR; the publish job promotes it to Docker Hub.
+  build:
+    name: Build & stage to GHCR
     if: github.event_name == 'workflow_dispatch' || (github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'release-canary'))
     runs-on: ubuntu-latest
     environment: Release
+    outputs:
+      version: ${{ steps.release.outputs.version }}
     permissions:
       contents: write # Required to push release tags.
-      id-token: write # Required for cosign keyless signing.
-      packages: write # Required to push images to GHCR (ghcr.io).
-      pull-requests: write # Required to comment on canary release PRs.
+      id-token: write # Required for npm provenance and cosign keyless signing.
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -25,16 +25,23 @@ concurrency:
...
+  # build, and the Dockerfile build, which execute dependency lifecycle scripts).
+  # It deliberately does NOT have the long-lived Docker Hub credential. It uses
+  # only ephemeral, run-scoped credentials: OIDC for npm provenance + cosign, and
+  # GITHUB_TOKEN for the git tag push and the GHCR image push. The image is staged
+  # to GHCR; the publish job promotes it to Docker Hub.
+  build:
+    name: Build & stage to GHCR
     if: github.event_name == 'workflow_dispatch' || (github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'release-canary'))
     runs-on: ubuntu-latest
     environment: Release
```
0xcadams, line 38: This job has the environment still? I'm confused the comment below is saying something else.

> Publish job: the ONLY job that holds the long-lived Docker Hub credential. It

arv: Ignoring now that you refactored all of this

PR #6165, opened 2026-06-26
feat: add encryptRow/decryptRow hooks for end-to-end row encryption

omerchief:
## Summary

Adds optional, **inert-by-default** row-encryption hooks to Zero so an application can encrypt designated columns at rest and on the sync wire. zero-cache continues to replicate and fan out rows byte-for-byte — it never sees plaintext or holds a key. There are three seams, and when the corresponding hook is unset the existing behavior is completely unchanged.

### Client — decrypt on sync read

`new Zero({ decryptRow })` (`packages/zero-client/src/client/options.ts`, `zero.ts`). `PokeHandler` applies `decryptRow` to `put` values and `update` merges as rows are converted to Replicache patches, just before they land in the local store. `del`/`clear` are untouched.

```ts
const z = new Zero({
  ...opts,
  decryptRow: (tableName, row) => decryptColumns(tableName, row),
});
```

### Server — encrypt on write, decrypt on read

`zeroDrizzle(schema, db, { encryptRow, decryptRow })` (`packages/zero-server/src/adapters/drizzle.ts`).

- **Write**: `encryptRow` runs in the `CRUDMutatorFactory` executor on `insert`/`upsert`/`update`. `delete` is skipped — its args are a bare primary key and must not be transformed.
- **Read**: `decryptRow` runs in `DrizzleInternalTransaction.runQuery`, walking `related` subqueries recursively so nested rows are decrypted under their own table name before the parent.

```ts
const zql = zeroDrizzle(schema, drizzleDb, {
  encryptRow: (tableName, row) => encryptColumns(tableName, row),
  decryptRow: (tableName, row) => decryptColumns(tableName, row),
});
```

`RowCrypto` (`(tableName: string, row: Row) => Row | Promise<Row>`) and `ZeroServerCryptoOptions` are exported from `@rocicorp/zero/server/adapters/drizzle`; `DecryptRow` is exported from the client options.

## Notable change

`mergePokes` and `rowsPatchOpToReplicachePatchOp` become `async` to await the (possibly async) decrypt hook. All new parameters are optional, so existing call sites are unaffected; the `mergePokes` unit tests are updated to `await` and the cookie-gap test now asserts via `.rejects.toThrow()`.

## Why

We run these exact transforms in production today via a `patch-package` patch over the published `@rocicorp/zero` bundle. Upstreaming the hooks removes a brittle patch that silently re-points at compiled output on every release. The design is intentionally minimal: hooks are opt-in, the encrypt/decrypt boundary is the application's responsibility, and Zero stays key-agnostic.

## Validation — please read

This change was **ported from our production patch** (against the v1.6.2 compiled output) onto current `main` by hand. It has **not** been run through this repo's `oxfmt` / `oxlint --type-aware` / `tsc` / test toolchain locally, so I've opened it as a **draft** and am relying on CI to confirm formatting, types, and tests. Happy to adjust naming, types, doc comments, or split client/server into separate PRs — whatever fits the project's conventions.

### DIFF
--- a/packages/zero-client/src/client/options.ts
+++ b/packages/zero-client/src/client/options.ts
@@ -1,6 +1,7 @@
 import type {LogLevel, LogSink} from '@rocicorp/logger';
 import type {StoreProvider} from '../../../replicache/src/kv/store.ts';
 import * as v from '../../../shared/src/valita.ts';
+import type {Row} from '../../../zero-protocol/src/data.ts';
 import type {
   BaseDefaultContext,
   BaseDefaultSchema,
@@ -10,6 +11,13 @@ import type {
 import type {AnyMutatorRegistry} from '../../../zql/src/mutate/mutator-registry.ts';
 import type {CustomMutatorDefs} from './custom.ts';
 import {UpdateNeededReasonType} from './update-needed-reason-type.ts';
+
+/**
+ * Async hook applied to each server row before it is written to the local
+ * store, e.g. to decrypt end-to-end encrypted columns. The returned row is
+ * stored in place of the original.
+ */
+export type DecryptRow = (tableName: string, row: Row) => Row | Promise<Row>;
 
 /**
  * Configuration for {@linkcode Zero}.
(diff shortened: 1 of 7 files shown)


aboodman:
I think we should do this with the more general `codec` feature.

omerchief:
My apologies Aaron, forgot to clean this up

PR #6166, opened 2026-06-27
chore(zero): Bump version to 1.8

aboodman:
Now that we've created the 1.7 branch head should be 1.8 so that the canaries from main are correct.


0xcadams (commented):
LGTM thanks

PR #6176, opened 2026-06-29
ci(zero): use pinned skopeo container

0xcadams:
Replace `apt-get install skopeo` in the credentialed Zero Docker release job with the pinned official Skopeo container image.


arv (commented):
LGTM

PR #6177, opened 2026-06-30
ci(zero): make npm promotion manual

0xcadams:
- Remove npm dist-tag promotion from the Zero promote workflow so CI no longer depends on `NPM_TOKEN`.
- Keep Docker `latest` promotion/signing and git `latest` tag promotion in CI.
- Add a `$GITHUB_STEP_SUMMARY` section with concrete local npm commands to move `latest`, remove `staging`, and verify tags.
- Update Zero release docs to describe the manual npm follow-up.


arv (commented):
LGTM

PR #6179, opened 2026-06-30
chore: e2e throughput measurements

tantaman:
Measure full E2E throughput from PG -> ReplicationManager -> ViewSyncer -> Client

```
pnpm --filter zero-throughput start -- \
    --profile relational \
    --users 100 \
    --queries-per-user 3 \
    --rows-per-query 50 \
    --write-rate 25 \
    --duration-ms 60000 \
    --zero-num-sync-workers 1 \
    --output results/relational-100u-3q-25wps-1syncer.json
```

Note that the "WPS" here are writes that impact every query being viewed. Not writes that are un-observed by queries. We should measure the latter too.

## Param Sweep

Screenshot 2026-06-30 at 10 21 20 PM

### DIFF
--- a/ZERO_THROUGHPUT_DESIGN.md
+++ b/ZERO_THROUGHPUT_DESIGN.md
@@ -1,0 +1,296 @@
+# Zero E2E Throughput Benchmark Design
+
+## Purpose
+
+We need a repeatable benchmark that answers how much write throughput Zero can
+sustain before connected clients fall behind.
+
+The benchmark should measure the full path:
+
+1. PostgreSQL accepts committed writes.
+2. The replicator observes and applies WAL changes.
+3. The view-syncer evaluates standing queries and produces diffs.
+4. Synthetic clients receive and apply query results over the Zero protocol.
+
+The primary output is the maximum sustainable PostgreSQL write rate for a given
+application profile, user count, and standing-query mix.
+
+## Questions
+
+- At a fixed user count and query mix, what writes/sec can Zero sustain while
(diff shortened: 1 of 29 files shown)


arv:
I like the idea of this but maybe it does not belong in the mono repo?

Any advice on how to review this?

tantaman:
> I like the idea of this but maybe it does not belong in the mono repo?

Where do you suggest?

For reviewing, I think checking that the queries look representative for the use case and indices are set up correctly in the DB. This is codex authored so I've been running it, asking Codex questions about each result, looking at query plans (it writes them each run) to make sure it isn't doing something dumb for the test.

tantaman:
The `relational` test turns out to be a worst case test for our engine in that every push requires a take re-fetch.

Note that the WPS is also worst case because it assumes every query sees every write which is unrealistic in practice but very good for a throughput test as it gives us the floor.

tantaman:
One big win we got:
1. [Moved the sagrable term out of](https://github.com/rocicorp/mono/pull/6184) `fetch` for `take` so SQLite does the right thing. O(N) -> O(logN) lookups now

Other questions:
1. New strategy for leap-frogging? changeLog2 endlessly backs up and we never shed load once it does
2. RM -> VS fanout is horribly slow. 10 VS is not much better than 3 VS

See https://app.notion.com/p/replicache/Proactive-workload-profiling-3913bed8954580d4a0e1e32db7c2b7a5?source=copy_link for more notes

PR #6182, opened 2026-07-01
chore(zero-cache): simple covering algo + log impact

tantaman:
# Query Covering Shadow Mode

  “Covering” detects when all results of one query are already contained in another running query.

  If query **B** covers query **A**, then every row A can return is also returned by B. In the current implementation this is used only for shadow logging; it
  does not change hydration behavior.

  ## Runtime Shape

  For each hydration batch, view-syncer builds an index of currently running queries keyed by normalized root:

  ```ts
  (schema, table, alias)

  For each newly hydrated query, we only compare against running queries with the same root. After each successful hydration in the batch, the new query is
  added to the index so later queries in the same batch can be covered by earlier ones.

  The summary log is:

  query coverage shadow summary

  with scalar context fields:

  queryCoverageMode = shadow
  hydrationPath = add | hydrate-unchanged
  totalHydratedQueries
  coveredHydratedQueries
  uncoveredHydratedQueries
  appID
  shardNum
  clientGroupID

  If there is at least one hit, the log also includes one representative covered/covering query hash pair.

  ## Conservative By Design

  The algorithm only returns covered when it can prove coverage from AST shape. Unsupported or ambiguous cases return false.

  It does not read rows, query SQLite, or inspect client state. It only compares transformed ASTs.

  ## Covered Examples

  ### Unfiltered Query Covers Filtered Query

  A: issues.where('id', '1')
  B: issues

  B covers A because B returns all issues.

  ### Equality Covered By IN

  A: issues.where('id', '1')
  B: issues.where('id', 'IN', ['1', '2', '3'])

  ### Stronger Conjunction Covered By Weaker Filter

  A: issues.where('status', 'open').where('owner', 'alice')
  B: issues.where('status', 'open')

  A’s result is a subset of B’s result.

  ### Numeric Range Implication

  A: issues.where('priority', '>', 5)
  B: issues.where('priority', '>', 3)

  Also similar cases like:

  A: issues.where('priority', '<', 5)
  B: issues.where('priority', '<=', 5)

  ### OR Coverage

  A: issues.where('type', 'bug')

  B: issues.where(({or}) =>
    or(
      q => q.where('type', 'bug'),
      q => q.where('type', 'feature'),
    ),
  )

  A is covered by one branch of B.

  For an OR query to be covered, every branch must be covered:

  A: issues.where(({or}) =>
    or(
      q => q.where('id', '1'),
      q => q.where('id', '2'),
    ),
  )

  B: issues.where('id', 'IN', ['1', '2', '3'])

  ## Limits And Starts

  Unlimited queries cover limited queries:

  A: issues.where('status', 'open').limit(10)
  B: issues.where('status', 'open')

  Same-basis windows are covered when the covering limit is larger:

  A: issues.where('status', 'open').start(cursor).limit(10)
  B: issues.where('status', 'open').start(cursor).limit(20)

  Same-basis means:

  same root
  same where semantics
  same orderBy
  same start
  B.limit >= A.limit

  A broader limited query does not necessarily cover a narrower limited query:

  A: issues.where('status', 'open').limit(10)
  B: issues.limit(100)

  This is not treated as covered because B’s limit may be consumed by rows that A would filter out.

  The algorithm also does not reason about relative cursor positions:

  A: issues.where('status', 'open').start(cursor10).limit(10)
  B: issues.where('status', 'open').start(cursor0).limit(100)

  This may be covered semantically, but is not detected today.

  ## Related Queries

  Related query coverage is recursive.

  A: issues
    .where('status', 'open')
    .related('comments', q => q.where('text', 'hello'))

  B: issues
    .related('comments', q => q)

  B covers A because:

  - B covers the root issue set.
  - B’s related comments query covers A’s narrower related comments query.

  Relationship edges must match by:

  correlation
  hidden
  system
  subquery alias

  ## EXISTS

  A: issues.whereExists('comments', q => q.where('text', 'hello'))
  B: issues.whereExists('comments', q => q)

  B covers A because any issue with a matching “hello” comment also has some comment.

  ## NOT EXISTS

  For NOT EXISTS, implication reverses inside the subquery.

  A: issues.whereNotExists('comments', q => q)
  B: issues.whereNotExists('comments', q => q.where('text', 'hello'))

  A is covered by B because if an issue has no comments at all, then it also has no comments whose text is hello.

  ## Flip

  flip does not affect semantic coverage.

  A flipped correlated subquery condition and an unflipped equivalent condition can cover each other.

  ## Not Detected Today

  These may be semantically valid in some cases, but the current algorithm intentionally does not infer them.

  ### Singleton IN As Equality

  A: issues.where('id', 'IN', ['1'])
  B: issues.where('id', '1')

  ### LIKE / ILIKE / NOT IN Reasoning

  A: issues.where('title', 'hello')
  B: issues.where('title', 'LIKE', 'h%')

  ### Cross-Root Coverage

  A: comments.where('id', '1')
  B: issues.related('comments', q => q)

  Different root table queries are not compared.

  ### Relative Window Coverage

  A: issues.start(cursor10).limit(10)
  B: issues.start(cursor0).limit(100)

  The algorithm only handles exact same start.

  ## Athena Metric

  Covered hydration percent can be computed from the summary logs:

  SELECT
    SUM(CAST(context.coveredHydratedQueries AS bigint)) AS covered,
    SUM(CAST(context.totalHydratedQueries AS bigint)) AS total,
    100.0 * SUM(CAST(context.coveredHydratedQueries AS double))
      / NULLIF(SUM(CAST(context.totalHydratedQueries AS double)), 0) AS covered_pct
  FROM logs
  WHERE message = 'query coverage shadow summary'
    AND context.queryCoverageMode = 'shadow';

### DIFF
--- a/packages/zero-cache/src/config/zero-config.test.ts
+++ b/packages/zero-cache/src/config/zero-config.test.ts
@@ -248,6 +248,15 @@ test('zero-cache --help', () => {
                                                                                                                                                                                                    
                                                                         You can disable the planner if it is picking bad strategies.                                                               
                                                                                                                                                                                                    
+     --enable-query-covering boolean                                    default: true                                                                                                              
+       ZERO_ENABLE_QUERY_COVERING env                                                                                                                                                              
+                                                                        Enable shadow-mode query covering detection during query hydration.                                                        
+                                                                                                                                                                                                   
+                                                                        When enabled, view-syncers compare newly hydrated queries against running                                                  
+                                                                        queries with the same root table and log aggregate coverage stats.                                                         
+                                                                                                                                                                                                   
+                                                                        You can disable this if covering detection adds too much CPU overhead.                                                     
+                                                                                                                                                                                                   
      --yield-threshold-ms number                                        default: 10                                                                                                                
        ZERO_YIELD_THRESHOLD_MS env                                                                                                                                                                 
                                                                         The maximum amount of time in milliseconds that a sync worker will                                                         
@@ -773,6 +782,17 @@ test.each([['isok'], ['has_underscores'], ['1'], ['123']])(
     expect(config.app.id).toBe(appID);
   },
 );
+
+test('--enable-query-covering can be disabled', () => {
+  const {config} = parseOptionsAdvanced(zeroOptions, {
+    argv: ['--enable-query-covering', 'false'],
+    envNamePrefix: 'ZERO_',
+    allowUnknown: false,
+    allowPartial: true,
+  });
+
+  expect(config.enableQueryCovering).toBe(false);
+});
 
 test('--shard-id disallowed', () => {
   const logger = {info: vi.fn()};
--- a/packages/zero-cache/src/config/zero-config.ts
+++ b/packages/zero-cache/src/config/zero-config.ts
@@ -488,6 +488,18 @@ export const zeroOptions = {
     ],
   },
 
+  enableQueryCovering: {
+    type: v.boolean().default(true),
+    desc: [
+      `Enable shadow-mode query covering detection during query hydration.`,
+      ``,
+      `When enabled, view-syncers compare newly hydrated queries against running`,
+      `queries with the same root table and log aggregate coverage stats.`,
+      ``,
+      `You can disable this if covering detection adds too much CPU overhead.`,
+    ],
+  },
+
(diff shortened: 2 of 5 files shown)


arv:
This is cool. 

How about inclusive/exclusive start?

PR #6185, opened 2026-07-02
fix(zero): preserve __proto__ as user data

arv:
Define schema and mutator records with own data properties so `__proto__` table, column, and relationship names are not routed through the legacy prototype setter.

Use null-prototype ZQL meta entries so rows and relationships named `__proto__` survive view updates without changing the entry prototype.

Based on #6145 by @tjenkinson

### DIFF
--- a/packages/shared/src/objects.ts
+++ b/packages/shared/src/objects.ts
@@ -7,7 +7,6 @@ export function mapValues<T extends Record<string, unknown>, U>(
   };
 }
 
-
 export function defineOwnDataProperty<T>(
   object: object,
   key: string,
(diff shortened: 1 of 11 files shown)


packages/shared/src/objects.ts
```diff
@@ -1,38 +1,67 @@
...
+  if (key === '__proto__') {
+    // Shadow the `__proto__` property on the object to prevent prototype
+    // pollution.
+    Object.defineProperty(object, key, {
+      value,
+      writable: true,
+      enumerable: true,
+      configurable: true,
+    });
+    // fall through to the normal assignment to ensure that potentional Proxy
```
tjenkinson, line 35: Great idea

tjenkinson (commented):
thanks @arv this looks great!

I think https://github.com/rocicorp/mono/blob/d379d93cbeb5e51cb40ce413439940904f7b5b57/packages/shared/src/build-nested-object-path.ts#L41-L44 still needs similar treatment?

Opened a PR for that here: https://github.com/rocicorp/mono/pull/6200

arv:
> thanks @arv this looks great!
> 
> I think
> 
> https://github.com/rocicorp/mono/blob/d379d93cbeb5e51cb40ce413439940904f7b5b57/packages/shared/src/build-nested-object-path.ts#L41-L44
> 
> still needs similar treatment?
> Opened a PR for that here: #6200

I think you are right... Looking.

PR #6186, opened 2026-07-02
chore(zero-cache): cap the subscriber queue during catchup

tantaman:
active, post-catchup subscribers that was already the expectation.

The chain was:

```
 forwardWithFlowControl() -> Broadcast -> Subscriber.send() -> #sendChange() -> Subscription.push().result
```

  That push().result resolves when the downstream message is consumed, coalesced, or canceled. So forwardWithFlowControl() was designed to occasionally stop
  the replication loop until subscribers had actually made progress.

  The bug was that catchup subscribers bypassed that contract.

  Before this change, if #backlog was non-null, Subscriber.send() just did:

  this.#backlog.push(change);

  and returned immediately. So from Broadcast’s perspective, that subscriber had “completed” the send even though the change was only sitting in an unbounded
  in-memory backlog. Then setCaughtUp() dumped the backlog into downstream with void #sendChange(...), also not waiting on the downstream subscription.

  So the original expectation was valid, but only for the non-catchup path. Catchup created a blind spot:

  - active/live subscriber: send() supplied backpressure
  - catching-up subscriber: send() acknowledged enqueue into local backlog, not downstream progress
  - catchup completion: backlog flush ignored flow control and could move the memory pressure into the downstream queue

  This PR makes catchup participate in the same basic backpressure story: while catching up, send() can block on backlog capacity, and once catchup completes,
  draining into downstream is bounded instead of firehosing the whole backlog.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/forwarder.test.ts
+++ b/packages/zero-cache/src/services/change-streamer/forwarder.test.ts
@@ -7,6 +7,10 @@ import {createSubscriber} from './test-utils.ts';
 
 const json = BigIntJSON.stringify;
 
+function nextEventLoopTurn() {
+  return new Promise<void>(resolve => setImmediate(resolve));
+}
+
 describe('change-streamer/forwarder', () => {
   const messages = new ReplicationMessages({issues: 'id'});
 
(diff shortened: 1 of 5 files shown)


packages/zero-cache/src/services/change-streamer/forwarder.test.ts
```diff
@@ -10,6 +10,42 @@ const json = BigIntJSON.stringify;
 describe('change-streamer/forwarder', () => {
   const messages = new ReplicationMessages({issues: 'id'});
 
+  test('flow control waits on catching-up subscriber backlog', async () => {
```
tantaman, line 17: - Current branch: passes.
- Pre-fix base a99e63093: fails in forwarder.test.ts with expected true to be false, meaning forwardWithFlowControl() had already released before catchup drained.

PR #6188, opened 2026-07-02
feat(zql): fail loudly when PG and Zero disagree

tantaman:
Some users still get "take bound" errors. 

- https://github.com/rocicorp/mono/pull/6122 attempts to address it

That fix would allow the push to flow but cause `push` and `fetch` to no longer agree.

The root of the problem is that PG and Zero disagree on the shape of the data. E.g., Zero says "this field cannot be null" but there is null data in it. In that case the `start` comparator returns the wrong data because it created comparison operations assuming the inability of `null` to be present.

### DIFF
--- a/packages/zql-integration-tests/src/chinook/fuzz/driver.ts
+++ b/packages/zql-integration-tests/src/chinook/fuzz/driver.ts
@@ -35,6 +35,7 @@ import type {AnyQuery} from '../../../../zql/src/query/query.ts';
 import {mapResultToClientNames} from '../../../../zqlite/src/test/source-factory.ts';
 import {type Delegates, runAndCompare} from '../../helpers/runner.ts';
 import {schema} from '../schema.ts';
+import {pkOf, tables} from './axes.ts';
 import type {CostModel} from './cost.ts';
 import {
   applyLimit,
@@ -50,7 +51,7 @@ import {Coverage} from './coverage.ts';
 import {flipAssignments, flippableExistsCount, setFlips} from './flip.ts';
 import type {Data} from './literals.ts';
 import {mutate} from './mutate.ts';
-import {type Mutation, pushForSkeleton} from './push.ts';
+import {fourPhase, type Mutation, pushForSkeleton} from './push.ts';
 import type {Regression} from './regressions.ts';
 import {rng} from './rng.ts';
 import {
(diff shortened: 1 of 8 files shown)


packages/zql-integration-tests/src/chinook/fuzz/driver.ts
```diff
@@ -508,6 +562,37 @@ export async function checkPushWalk(
   return {total, failures};
 }
 
+/**
+ * **Start push/refetch sweep:** `start` is deliberately excluded from the PG-oracle axes
+ * because z2s currently ignores it. This lane checks the property directly instead:
```
tantaman, line 567: oh, this is bad 😂. Putting up a separate PR for this bug.

packages/zql/src/ivm/take.ts
```diff
@@ -445,7 +445,14 @@ export class Take implements Operator {
       return;
     }
 
-    assert(takeState.bound, 'Bound should be set');
+    if (takeState.bound === undefined) {
```
tantaman, line 448: this shouldn't have changed.

YevheniiKotyrlo:
Fail-closed plus the push/refetch lane seems right — it ends the group-crash without licensing push/fetch divergence. Happy to close #6122 in favor of this. Verified end-to-end: the standalone repro now runs this PR's take semantics as a [fifth matrix leg](https://github.com/YevheniiKotyrlo/zero-view-syncer-bound-should-be-set/commit/6b87778) — on 1.7.0 the update is dropped, the window stays at 0 rows, the assert never fires, and the sibling client group keeps syncing.

One heads-up on the new table-source test: it pins the broken NULL-bound lowering itself. I ran it on this PR merged with #6121 and both expectations flip — hydration returns `b`, and the push arrives as a plain edit through the normal bound-row path. A direct operator-level push (the construction #6122's tests use) pins the fail-closed branch independently of the producer; it still exercises the branch on the merged pair.

And since the return is silent, a debug log naming the table/column/bound would keep the state diagnosable — in production the disagreement doesn't need mismatched app schemas: the replica specs [carry no nullability at all](https://github.com/rocicorp/mono/blob/cb014857f6dbce50983a83612bfad5c4e0643739/packages/zero-cache/src/types/lite.ts#L116-L120), so any NULL-anchored window hits this, and it stays silently empty until #6121 lands.

tantaman:
- this has been superseded by https://github.com/rocicorp/mono/pull/6189

tantaman:
> the replica specs [carry no nullability at all](https://github.com/rocicorp/mono/blob/cb014857f6dbce50983a83612bfad5c4e0643739/packages/zero-cache/src/types/lite.ts#L116-L120)

That seems problematic :/

PR #6189, opened 2026-07-02
fix(z2s): z2s did not compile `start` | queryBuilder incorrectly handled `null` start fields

tantaman:
Two big bugs:

1. issue.start(foo) did not compile in z2s (which we use for mutators and our oracle)
2. table source incorrectly compared against `null`  when the cursor was null.  

This expands the fuzzer to include `start` testing which apparently it never included. The expansion of the fuzzer found problem (2). Which is convenient because it was the root cause of the take bound not set issue which I was concurrently working.

### DIFF
--- a/packages/zqlite/src/query-builder.ts
+++ b/packages/zqlite/src/query-builder.ts
@@ -306,6 +306,10 @@ function nullableAwareRangeComparison(
   operator: '>' | '<',
   columnType: SchemaValue,
 ): SQLQuery {
+  if (value === null) {
+    return operator === '>' ? sql`${sql.ident(field)} IS NOT NULL` : sql`FALSE`;
+  }
+
   // For non-nullable columns, skip IS NULL checks to avoid breaking
   // SQLite's MULTI-INDEX OR optimization, which falls back to a full
   // table scan when any OR branch involves NULL.
@@ -328,6 +332,10 @@ function sargableLeadingStartBound(
   operator: '>' | '<',
   columnType: SchemaValue,
 ): SQLQuery | undefined {
+  if (value === null) {
+    return undefined;
+  }
+
   if (columnType.optional === true) {
     return undefined;
   }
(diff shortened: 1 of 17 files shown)


packages/zqlite/src/query-builder.ts
```diff
@@ -306,6 +306,10 @@ function nullableAwareRangeComparison(
   operator: '>' | '<',
   columnType: SchemaValue,
 ): SQLQuery {
+  if (value === null) {
+    return operator === '>' ? sql`${sql.ident(field)} IS NOT NULL` : sql`FALSE`;
```
tantaman, line 310: this means "nulls are the least value there is"

```
> null -> is not null
< null -> false
```

tantaman: If SQLite does not return rows in exactly the order ZQL's comparators expect, then we get `take bound not found` errors. This fixes the bug where `null` in cursor values incorrectly compared.

packages/zqlite/src/query-builder.ts
```diff
@@ -392,7 +403,7 @@ function gatherStartConstraints(
       } else {
         const [jField] = order[j];
         const columnType = columnTypes[jField];
-        const value = toSQLiteType(from[jField], columnType.type);
+        const value = toSQLiteType(from[jField] ?? null, columnType.type);
```
tantaman, line 406: if a user passes an object missing fields to `start`... the only sane choice is to treat them as `null`.

e.g.,

```
issues.after({}).orderBy('created')
```

^^-- cursor should have had `created` in it. We'll treat it as `null` if it is missing.

arv: Can we not throw here instead?

tantaman: I think it is valid to pass `undefined`, no? I see your point though.. "make the user explicitly use null" vs "allow absence to mean null"

PR #6192, opened 2026-07-02
ci: parallelize JS workflow jobs

0xcadams:
Screenshot 2026-07-03 at 10 33 30 AM
Got these down to ~30s 

- Run fast JS preflight checks in parallel: syncpack, verify-deps, format, and lint.
- Gate expensive jobs on preflight so tests do not run when fmt/lint/package checks fail.
- Let tests, PG tests, typecheck, and Docker build run in parallel after preflight.
- Replace setup-node's shared pnpm cache with explicit scoped pnpm-store caches.
- Use smaller root-tools installs for syncpack, format, and lint.
- Add `--ignore-scripts` for filtered preflight installs where lifecycle scripts are unnecessary.


arv (approved):
LGTM

PR #6193, opened 2026-07-03
fix: postgres@3.4.7 crash

arv:
When the socket disconnects it becomes null. The upstream code was not checking for this and was crashing when it tried to call `socket.write` on a null socket. This change adds a check for null before calling `socket.write`.


arv:
We should upstream this if it fixes the bug. I'm setting a reminder for myself.

PR #6194, opened 2026-07-03
fix(zero-cache): skip backfill for update_schemas defaults

arv:
Stacked on #6222, which adds the optional `newColumns` field to the DDL event schema; this PR teaches the upstream trigger functions to emit it and uses it.

Teach DDL events to report published columns created in the current transaction, then use that metadata when deciding whether an added column needs backfill. This lets the manual update_schemas() path replicate simple defaults directly when invoked in the same transaction as the schema change.

Keep backfilling columns that are only newly published, since those may hold arbitrary existing values, and add end-to-mid coverage for both paths.

``https://app.notion.com/p/replicache/Apply-backfill-less-defaults-optimization-to-update_schemas-path-3843bed8954580209a05fa5be881a0f2``?source=copy_link

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
+++ b/packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
@@ -23,6 +23,19 @@ export const ddlEventSchema = triggerEvent.extend({
   version: v.literal(PROTOCOL_VERSION),
   schema: publishedSchema,
   event: v.object({tag: v.string()}),
+  // Maps the OID of each published table to the `attnum`s of published
+  // columns that were created in the (upstream) transaction that emitted
+  // the event. Such columns are guaranteed to have the column default in
+  // all pre-existing rows, and can thus be replicated without backfill if
+  // the default value itself is replicable. Tables whose publication
+  // entries (e.g. column lists) were modified in the same transaction are
+  // excluded, since a newly *published* (as opposed to newly *created*)
+  // column may hold arbitrary values in existing rows.
+  //
+  // The field is absent in messages from older versions of the upstream
+  // functions (in which case backfill decisions fall back to the command
+  // tag heuristic), and `null` when there are no such columns.
+  newColumns: v.record(v.array(v.number())).nullable().optional(),
 });
 
 /**
--- a/packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
+++ b/packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
@@ -333,6 +333,12 @@ $$ LANGUAGE plpgsql;
 -- same transaction as the schema change statement(s); among other things,
 -- this allows columns added with replicable defaults (e.g. constants) to
 -- be replicated without a backfill.
+--
+-- Note that it must be invoked *before* any subsequent DML on the altered
+-- tables: since this hook emits the schema change at the point of the
+-- call (rather than at each DDL statement, as event triggers do), row
+-- changes between a column's creation and the call reference a column
+-- that the replica does not yet know about, and fail replication.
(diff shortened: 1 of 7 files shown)


packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
```diff
@@ -209,13 +209,47 @@ RETURNS void AS $$
...
-  
+
   IF prev_schema_specs::text != schema_specs::text THEN
     UPDATE ${schema}."publishedSchema" SET current = schema_specs;
+
+    -- Report the published columns that were created in the current
+    -- transaction (i.e. pg_attribute rows inserted by this transaction).
+    -- All pre-existing rows of such columns are guaranteed to contain the
+    -- column default, allowing the zero-cache to replicate the default
+    -- directly (i.e. without backfill) if the default value is replicable.
```
tantaman: idk this just seems so suss to me.

Here is one case it breaks:

https://github.com/rocicorp/mono/pull/6229

arv: Confirmed — your test from #6229 fails on this branch as described (the event came out as `dflt: '2'` with no backfill, stamping `2` over rows that hold `1`). It's cherry-picked into this branch now, along with a fix: b8bc0c2.

The root cause is that the *current* default is not what pre-existing rows hold — Postgres stamps the ADD-time default into `pg_attribute.attmissingval`, and later `SET DEFAULT`s only affect new rows. So "created in this transaction" (however it's detected) is not a sufficient condition for skipping backfill.

The fix makes the trigger report each created column's `attmissingval` (new optional `missingValues` field on the DDL event), and the backfill is skipped only when the current default is a literal that provably evaluates to that value (`defaultValueMatches()` — conservative allowlist mirroring `mapPostgresToLiteDefault`; any mismatch, volatile/expression default, non-scalar, or out-of-safe-range number falls back to backfill). Your case now backfills and the existing skip cases (`DEFAULT true`, plus a new text-default case) still skip.

---

tantaman: I think there's also a case where a DML inside the migration transaction can break this. E.g., inserting a row with a column set to a different value than the default in the same transaction that the column was added.

arv: Verified with an end-to-mid test (`ADD "dml" INT4 DEFAULT 7; INSERT ... (id, dml) VALUES (100, 42); update_schemas()` in one transaction) — it breaks, but not silently, and not because of the backfill skip:

The INSERT precedes the DDL message in the WAL (the manual hook emits at the point of the *call*, not at each DDL statement like event triggers do), so the row change reaches the replica before the column exists and the change-processor throws `Unknown column dml in table your.new_table` (`processInsert → liteRow`). This happens on the shipped always-backfill path too — the pipeline fails before any add-column/backfill is processed — so it's a pre-existing constraint of `update_schemas()`, independent of this PR. Backfill wouldn't rescue it. I've documented the constraint on the hook (9af113c): it must be called before any subsequent DML on the altered tables.

Worth noting the flip side: this loud failure is what closes the door on silent divergence generally. Any sequence that puts non-default values into a new column before `update_schemas()` emits row changes *carrying that column* (decoded tuples include all columns post-ADD, named in the statement or not), which fail against a replica that lacks it. Combined with the `attmissingval` match, the skip is either correct or replication has already failed loudly before the stamp is applied. DML *before* the ADD (or after the call) remains fine — those rows read `attmissingval` upstream, which is exactly what gets stamped.

If you can construct a sequence that diverges *silently* under the skip, that would definitely kill the approach — but I couldn't find one.

---

arv:
Superseded by #6593, which includes this change rebased onto main and extends it to columns created and published in the same transaction.

PR #6195, opened 2026-07-03
feat(zero-cache): accept context-only ddlStart events (DDL protocol v2, phase 1 of 2)

arv:
## Problem

Every DDL command matching the event trigger `TAGS` emits a `ddlStart` event containing a **full snapshot of the published schema**, even when the command does not affect the published schema at all. For databases with large schemas this bloats the WAL — and everything downstream of it (archiving, physical replicas, the change-streamer, logs) — with redundant multi-hundred-KB messages.

The pathological case is `REFRESH MATERIALIZED VIEW CONCURRENTLY`: its internal `CREATE`/`ALTER`/`DROP TABLE` sub-commands fire the event triggers on every refresh. A customer running a 5-minutely refresh emits ~850 full schema snapshots per day that encode no information (observed during INC-910 on the Goblins Production 2 stack).

## Change (phase 1 of 2)

Per review, the fix is split into two releases so that each release can be rolled back to the previous one:

**This PR (phase 1)** makes the zero-cache *accept* protocol-v2 context-only `ddlStart` events without emitting them:

- `ddl.ts`: the event parser accepts `version: 1 | 2`, and `ddlStart.schema` / `previousSchema` are optional. `PROTOCOL_VERSION` stays at **1** and the trigger SQL is **unchanged from main** — triggers still emit full v1 messages, so rolling back this release is safe.
- `change-source.ts`: schema-less `ddlStart` events are recorded for command-tag context and otherwise ignored; `#makeSchemaChanges` takes the next schema explicitly.

**Phase 2 (follow-up, staged on `arv/lightweight-ddl-start-phase2`)**: once this release is stable and rollback past it is no longer needed, bump `PROTOCOL_VERSION` to 2 and upgrade the triggers (shard schema v24) to emit context-only `ddlStart` events (~200 bytes) when the published schema is unchanged. That's when the WAL win materializes.

## Testing

- New test: parsing of v2 context-only `ddlStart` events (what phase 2 will emit); the legacy v1 parse test is retained.
- New test: `REFRESH MATERIALIZED VIEW CONCURRENTLY` emits only no-op `ddlStart` events and no `ddlUpdate` events, pinning the INC-910 scenario (to be tightened to context-only in phase 2).
- Ran `ddl`, `init`, `shard`, `change-source`, and `change-source.end-to-mid` pg suites against a local Postgres 16: green.
- `check-types`, `lint`, `format` clean.


packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
```diff
@@ -11,28 +11,37 @@ import {publishedSchema, publishedSchemaQuery} from './published.ts';
 //
 // Increment this when changing the format of the contents of the "ddl" events.
 // This will allow old / incompatible code to detect the change and abort.
-export const PROTOCOL_VERSION = 1;
+//
+// v2: "ddlStart" events that are not associated with a schema change are
```
aboodman: Why send the `ddlStart` event at all in the case there is no schema change we care about? that would avoid need to rev the protocol.

arv:
Re: [@aboodman's question](https://github.com/rocicorp/mono/pull/6195#discussion_r3521250180) on why we emit `ddlStart` at all when there's no schema change (answering here — inline reply is blocked by a pending review):

The `ddlStart` sequence is load-bearing for command-tag attribution, which decides backfill-vs-skip. Postgres doesn't expose the *enclosing* command's tag to a nested event trigger, so with nested DDL (`start(A) → start(B) → end(B) → end(A)`), a schema change detected at `start(B)` was caused by A but fires with `TG_TAG = B`. The zero-cache recovers A by looking at the *previous* event in the stream (the `effectiveTag` logic in `#handleDdlMessage`) — which only works if no-op `ddlStart`s are emitted.

Dropping them entirely also breaks the non-nested case subtly: `#lastReplicationEvent` can go stale. If a previous statement's last emitted event was a change-carrying `ddlStart` (its `ddl_end` was a no-op and emitted nothing), a later unrelated `ddlUpdate` would inherit that stale tag — e.g. an `ALTER PUBLICATION` (needs backfill) attributed to an old `CREATE TABLE` (skips backfill) → missing rows. Old consumers would hit that silently, which seemed worse than a rev: the `version` field makes incompatible consumers fail loudly instead.

The rev is also unavoidable for a different reason: v1 consumers require `schema` on every event, so even a slimmed `ddlStart` fails their parse. The context-only event is ~200 bytes, so the WAL win is the same either way.

Happy to go the other way if you'd rather redesign attribution to not need the start events, but that's a bigger change than this PR.

---

darkgnotic (commented):
In general I like the change; the `ddlStart` event's purpose is _usually_ only needed to determine the current command, in order to provide the context for whether a potential subsequent schema change needs a backfill. So making `prevSchema` and `schema` optional for that message makes sense.

However, it is quite risky to bump the protocol version ... once the trigger is migrated to version 2, we can't roll back the server if memory serves me correctly ... the previous release will error and halt replication progress if it sees a v2 message. I don't know what we'd do if there's a bug in the release that requires a rollback ...

(In retrospect I don't think the version field was well thought out given that we always want to support one version of rollback.  😓 ) 

Can we instead split this change in two phases/releases?

* First release: change the optionality of the fields in the ddlStart message, and add corresponding the handling of absent fields in `change-source.ts`, but do not yet change the trigger logic ... in this phase it still always sets the fields that a previous zero-cache version expects.
* Next release: once the first release is stable, and we know we won't need to rollback to a release before that, then we land the logic in the trigger that actually makes the fields optional.

Although this adds one release cycle to the fix, I think it is worth the safety (and allows us to fulfill our contract) of being able to roll back one release.

Would this work?

arv:
That makes sense. 

I was very wary of the protocol version change and the rollback issue.

I will split this and set a reminder.

arv:
Split done — this PR is now **phase 1 only**: the parser accepts v2 context-only `ddlStart` events (optional `schema`/`previousSchema`, `version: 1 | 2`) and `change-source.ts` handles their absence, but `PROTOCOL_VERSION` stays at 1 and the trigger SQL is unchanged from `main`, so this release remains rollback safe.

The trigger-side change (protocol bump + v24 trigger migration) is staged on [`arv/lightweight-ddl-start-phase2`](https://github.com/rocicorp/mono/tree/arv/lightweight-ddl-start-phase2) to land as a follow-up PR once this release is stable.

(Note: the phase-1 commit may take a moment to appear — pushing through a flaky connection.)

---

packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
```diff
@@ -13,15 +13,24 @@ import {publishedSchema, publishedSchemaQuery} from './published.ts';
...
 const triggerEvent = v.object({
   context: v.object({query: v.string()}).rest(v.string()),
 });
 
-// All DDL events contain a snapshot of the current tables and indexes that
-// are published / relevant to the shard.
 export const ddlEventSchema = triggerEvent.extend({
-  version: v.literal(PROTOCOL_VERSION),
-  schema: publishedSchema,
+  version: versionSchema,
```
darkgnotic, line 33: Just to confirm the rollout plan, there next RM will accept both v1 and v2, which is rollback safe, and then we'll bump it fully to v2 with the trigger change. At that point, the rollback safe RM will accept it, but any previous RM (that only understands v1) will fail to parse these events.

I think this is correct, just confirming the plan.

darkgnotic (approved):

arv:
Re: [@darkgnotic's rollout-plan question](https://github.com/rocicorp/mono/pull/6195#discussion_r3573755526) (inline reply blocked by a pending review, answering here):

Confirmed — that's exactly the plan. This release's RM accepts both v1 and v2 while still emitting v1, so it can be rolled back to the previous release. Phase 2 (staged on [`arv/lightweight-ddl-start-phase2`](https://github.com/rocicorp/mono/tree/arv/lightweight-ddl-start-phase2)) then upgrades the triggers to emit v2: rolling back one release to this RM stays safe, but anything older (v1-only) will fail to parse v2 events and would need a resync.

---

PR #6196, opened 2026-07-03
fix(zero-cache): bump same-hash query rehydrates

tantaman:
We hit a case where a query may not exist in the PipelineDriver but exist in the CVR. In this case, we need to hydrate the query. The catch is the row set could change even at the same CVR version.

Our row-set-signature mechanism handled this for all other cases except this one. This case the pipeline is not running so we have no two signatures to compare. The only thing we can do is to bump the cvr version and run the query.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/view-syncer.ts
+++ b/packages/zero-cache/src/services/view-syncer/view-syncer.ts
@@ -341,6 +341,15 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
       'persistent non-zero values indicate non-deterministic query execution ' +
       '(e.g. Cap operator picking different N-row subsets).',
   );
+  readonly #sameHashRehydrationVersionBumps = getOrCreateCounter(
+    'sync',
+    'query.same-hash-rehydrations-forced-bump',
+    'Number of times query-set reconciliation forced a configVersion bump ' +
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -2106,12 +2115,38 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
         queryID => this.#pipelines.rowSetSignature(queryID),
       );
 
-      // For queries being re-executed solely due to rowSetSignature drift
-      // (not a transformationHash change), trackQueries does not bump
-      // configVersion. Force a bump so the row diff produced by received()
-      // gets propagated to the client via a poke. Must happen before
-      // startPoke so the pokers see the final cookie version.
-      if (addQueries.some(q => driftedQueryIDs.has(q.id))) {
+      const sameHashRehydratedQueryIDs = addQueries
```
tantaman, line 2118: the issue here is that:
a. the query is in the CVR
b. the pipeline driver never saw the query

In this case the row set signature is persisted but that is useless to us here because the pipeline driver is not running the query so we don't know the new signature!

This change makes us hydrate the missing query at a new cvr version and diff rather than just hydrating and skipping the diff.

PR #6200, opened 2026-07-04
fix(zero): preserve `__proto__` in `buildNestedObjectPath`

tjenkinson:
related to https://github.com/rocicorp/mono/pull/6185


arv:
Thanks. This is actually dead code. I'm just going to delete it instead.

PR #6202, opened 2026-07-06
fix(zqlite): bound the StatementCache with LRU eviction

arv:
StatementCache.return() previously pushed statements back into the cache with no size check, so every distinct SQL text permanently retained a native sqlite3_stmt handle. TableSource.fetch builds SQL from the query AST (a client-influenced key space) and TableSources live for the view-syncer lifetime, so distinct query shapes accumulated native statement handles without bound.

Make the cache an LRU with a configurable maxSize (default 1000):

- Map insertion order doubles as recency order; entries are re-inserted on every get hit and return, so the first entry is the LRU.
- return() evicts least recently used statements once the bound is exceeded. Evicted statements are released to GC, which finalizes the underlying native handle.
- Only idle statements are evicted; in-flight statements are checked out of the cache by get() and are never touched.
- drop(n) is reimplemented on the same eviction path.

The default bound leaves headroom for the replication change-processor, which prepares several statements per table (plus partial-column update variants) through a single StatementRunner cache.

### DIFF
--- a/packages/zqlite/src/internal/statement-cache.ts
+++ b/packages/zqlite/src/internal/statement-cache.ts
@@ -6,6 +6,20 @@ export type CachedStatement = {
   sql: string;
   statement: Statement;
 };
+
+/**
+ * The default maximum number of statements retained by a StatementCache.
+ *
+ * Each cached statement holds a native `sqlite3_stmt` handle (prepared
+ * bytecode + SQLite-side memory), so the cache must be bounded. The bound
+ * needs enough headroom for callers with a large working set of distinct
+ * SQL — e.g. the replication change-processor prepares several statements
+ * per table (with additional variants for partial-column updates) through
+ * a single cache — while still capping growth from unbounded key spaces
+ * such as client-influenced query shapes.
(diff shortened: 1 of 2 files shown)


packages/zqlite/src/internal/statement-cache.ts
```diff
@@ -6,6 +6,20 @@ export type CachedStatement = {
...
+ *
+ * Each cached statement holds a native `sqlite3_stmt` handle (prepared
+ * bytecode + SQLite-side memory), so the cache must be bounded. The bound
+ * needs enough headroom for callers with a large working set of distinct
+ * SQL — e.g. the replication change-processor prepares several statements
+ * per table (with additional variants for partial-column updates) through
+ * a single cache — while still capping growth from unbounded key spaces
+ * such as client-influenced query shapes.
+ */
+export const DEFAULT_MAX_CACHED_STATEMENTS = 1_000;
```
tantaman, line 21: seems reasonable.

tantaman (approved):
lgtm.

PR #6206, opened 2026-07-07
perf(zero-cache): smarter load shedding

tantaman:
This PR adds smarter load shedding to zero-cache pipeline advancement. When an advance
  looks like it will cost materially more than rehydrating the query pipelines, we abort
  advancement and reset instead of continuing to burn CPU and hold WAL resources. It also
  catches pathological single-change pushes, while allowing advances that are already mostly
  complete to finish.

  Why: large or skewed transactions can make incremental advancement slower than a fresh
  hydrate and can create runaway backlog/WAL pressure. The new logic makes that tradeoff
  explicit and bounded.

  Also added more realistic throughput models in apps/zero-throughput to validate the
  behavior under production-like read/write patterns, larger samples, and more representative
  workload shapes.

The big change against the old model:

We now project total cost of pushing the backlog of changes and bail if the projection is higher than hydration.

---

Tested against various throughput profiles. All saw improved latency.

CleanShot 2026-07-07 at 13 20 10

More realistic load tests:

CleanShot 2026-07-07 at 15 23 22

CleanShot 2026-07-07 at 17 07 07


packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
```diff
@@ -900,50 +977,58 @@ export class PipelineDriver {
...
+            if (!tableSource) {
+              // no pipelines read from this table, so no need to process the change
+              continue;
+            }
+            const primaryKey = mustGetPrimaryKey(this.#primaryKeys, table);
+            let editOldRow: Row | undefined = undefined;
+            for (const prevValue of prevValues) {
+              if (
+                nextValue &&
+                deepEqual(
```
arv, line 996: We should we use the **comparator** here instead of creating a new object and deep equal that.

arv: I guess this is old code, indentation changed...

apps/zero-throughput/realistic-throughput-models.md
arv, line 1: remove plan md?

arv (commented):
LGTM

PR #6210, opened 2026-07-08
feat(zero-cache): added replication slot metrics

0xcadams:
## Summary

Adds minimal Postgres replication slot health metrics for the active Zero replication slot:

- `zero.replication.slot_health`
- `zero.replication.slot_retained_wal_bytes`
- `zero.replication.slot_safe_wal_bytes`

These help diagnose slot invalidation incidents by showing whether the active slot is healthy/missing/lost and how much WAL it is retaining.


tantaman (commented):
👏

PR #6212, opened 2026-07-08
chore: remove quick-check

tantaman:
was seen to be too slow in production although locally it added about 1 second per gig.

removes the quickcheck additions from https://github.com/rocicorp/mono/pull/6174/changes


0xcadams (commented):
LGTM

arv (commented):
LGTM

PR #6215, opened 2026-07-08
fix(zero-cache): log sqlite corruption diagnostics

tantaman:
we see corrupt replicas sometimes. On RM and ViewSyncer. Trying to gather some information about the db files when this happens.

### DIFF
--- a/packages/zero-cache/src/server/otel-log-sink.ts
+++ b/packages/zero-cache/src/server/otel-log-sink.ts
@@ -3,17 +3,26 @@ import {
   SeverityNumber,
   type AnyValueMap,
   type Logger,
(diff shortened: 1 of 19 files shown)


packages/zero-cache/src/server/otel-log-sink.ts
```diff
@@ -40,6 +49,10 @@ export class OtelLogSink implements LogSink {
     }
     this.#logger.emit(payload);
   }
+
+  async flush(): Promise<void> {
+    await this.#loggerProvider.forceFlush?.();
```
tantaman, line 54: why force flush?

PR #6220, opened 2026-07-09
fix(zero-cache): reset half-open pg sockets after inactivity

cesara:
A proxy can keep the client-facing TCP socket alive after its backend is gone, so no 'close'/'error' event ever fires and statements like BEGIN / COMMIT — which postgres.js awaits with no timeout and which are not covered by the TransactionPool's statementResponseTimeout — hang forever (see porsager/postgres#1089). This wedged the change-streamer's storer in production: back pressure was applied and never released because the drain loop was stuck on a silent socket.

Supply a custom socket factory to postgres.js that resets the connection after 2 minutes without wire activity (ZERO_PG_SOCKET_INACTIVITY_TIMEOUT overrides; 0 disables). The reset rejects all in-flight queries through the existing error path, letting the pool fail and recover. Idle pool connections are closed cleanly via idle_timeout before the reset timer can fire.

Notes:
- postgres.js does not call connect() on custom sockets and reads socket.host for the TLS SNI servername, so the factory handles both.
- Wire activity resets the timer, so streaming ops (e.g. COPY) are safe; statements that compute silently for longer (e.g. index builds) need the env override.

### DIFF
--- a/packages/zero-cache/src/types/pg.ts
+++ b/packages/zero-cache/src/types/pg.ts
@@ -1,3 +1,4 @@
+import {Socket} from 'node:net';
 import {
   PG_CONFIGURATION_LIMIT_EXCEEDED,
   PG_CONNECTION_DOES_NOT_EXIST,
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/types/pg.ts
```diff
@@ -329,6 +330,61 @@ export type PostgresTransaction = postgres.TransactionSql<{
...
+ *
+ * Exported for testing.
+ */
+export function inactivityTimeoutSocket(
+  timeoutMs = SOCKET_INACTIVITY_TIMEOUT_MS,
+) {
+  return (options: SocketFactoryOptions): Socket => {
+    const socket = new Socket();
+    if (timeoutMs > 0) {
+      socket.setTimeout(timeoutMs, () => socket.resetAndDestroy());
```
darkgnotic: Can you pass the `lc: LogContext` into this method and log a `warn()` message before calling `socket.resetAndDestroy()`? This way we'll have some way to understand (and confirm) that this is working.

darkgnotic (dismissed):
Thank you @cesara! I am excited and hopeful that we finally get to the bottom of these weird postgres hangs.

packages/zero-cache/src/types/pg.ts
```diff
@@ -329,6 +330,71 @@ export type PostgresTransaction = postgres.TransactionSql<{
...
+  lc: LogContext,
+  timeoutMs = SOCKET_INACTIVITY_TIMEOUT_MS,
+) {
+  return (options: SocketFactoryOptions): Socket => {
+    const socket = new Socket();
+    const target = options.path
+      ? options.path
+      : `${options.host[0]}:${options.port[0]}`;
+    if (timeoutMs > 0) {
+      socket.setTimeout(timeoutMs, () => {
```
tantaman, line 377: I was expecting to see something that resets this timeout when something happens 🤔

What prevents this from firing when there is activity?

PR #6221, opened 2026-07-09
fix(zero-cache): make the socket inactivity watchdog survive TLS upgrade

cesara:
Auditing against postgres.js 3.4.7 revealed that socket.setTimeout() registers its callback as a 'timeout' event listener, and secure() in postgres.js calls removeAllListeners() on the raw socket before wrapping it with tls.connect() — silently disarming the watchdog on every TLS connection, i.e. exactly the deployments (PlanetScale et al) this guard is for. Plaintext connections were unaffected, which is why tests passed.

Replace setTimeout() with an interval that samples the raw socket's bytesRead+bytesWritten: intervals are not socket listeners, so the TLS upgrade cannot remove them, and encrypted traffic still flows through the raw socket, advancing its counters. A half-open socket is now reset after one to two timeout periods of inactivity. The interval is unref'ed and self-clears once the socket is destroyed.

Adds a regression test that simulates the TLS upgrade's removeAllListeners() and asserts the reset still fires.

### DIFF
--- a/packages/zero-cache/src/types/pg.test.ts
+++ b/packages/zero-cache/src/types/pg.test.ts
@@ -607,7 +607,6 @@ describe('inactivityTimeoutSocket', () => {
 
     // postgres.js reads socket.host as the SNI servername in secure().
     expect(socket).toMatchObject({host: '127.0.0.1', port});
-    expect(socket.timeout).toBe(120_000);
   });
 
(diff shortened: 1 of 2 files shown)


cesara:
tested against hello-zero-solid and planetscale TLS

PR #6222, opened 2026-07-10
feat(zero-cache): add optional newColumns field to DDL events

arv:
Split out of #6194: declare the optional `newColumns` field on DDL events before anything emits it.

`newColumns` maps the OID of each published table to the `attnum`s of published columns created in the (upstream) transaction that emitted the event. #6194 will teach the upstream trigger functions to populate the field and use it to skip backfill for columns added with replicable defaults via the manual `update_schemas()` path.

Landing the reader-side declaration first, per the usual two-phase protocol rollout, lets this sit for a release so that every version in the rollback window understands the field before the trigger functions start emitting it.

Also adds a parse test covering messages with `newColumns` present, `null`, and absent.

`https://app.notion.com/p/replicache/Apply-backfill-less-defaults-optimization-to-update_schemas-path-3843bed8954580209a05fa5be881a0f2`

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
+++ b/packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
@@ -23,6 +23,19 @@ export const ddlEventSchema = triggerEvent.extend({
   version: v.literal(PROTOCOL_VERSION),
   schema: publishedSchema,
   event: v.object({tag: v.string()}),
+  // Maps the OID of each published table to the `attnum`s of published
+  // columns that were created in the (upstream) transaction that emitted
+  // the event. Such columns are guaranteed to have the column default in
+  // all pre-existing rows, and can thus be replicated without backfill if
+  // the default value itself is replicable. Tables whose publication
+  // entries (e.g. column lists) were modified in the same transaction are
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/services/change-source/pg/schema/ddl.ts
```diff
@@ -23,6 +23,19 @@ export const ddlEventSchema = triggerEvent.extend({
   version: v.literal(PROTOCOL_VERSION),
   schema: publishedSchema,
   event: v.object({tag: v.string()}),
+  // Maps the OID of each published table to the `attnum`s of published
+  // columns that were created in the (upstream) transaction that emitted
+  // the event. Such columns are guaranteed to have the column default in
+  // all pre-existing rows, and can thus be replicated without backfill if
```
tantaman, line 29: Is the flow:
- receive new column
- apply `add column foo with default bar` to sqlite

and then we have the correct state because SQLite applied the defaults?

arv: This is implemented in https://github.com/rocicorp/mono/pull/6194. 

This PR just adds the field to the protocol so that we can roll back without breaking things.

#6194 computes the columns that was added in the transaction.

PR #6225, opened 2026-07-11
test(zero-cache): reproduce compound index reorder

tantaman:
Was spending more time trying to figure out the SQLite DB corruption and discovered that we do not preserve index order during certain migrations.

--

SET NOT NULL changes the synthetic SQLite column type, causing the change processor to replace the column and rebuild its indexes. When the changed column is not last in a compound index, the temporary-column substitution appends it to the end and reverses the indexed column order. Add an intentionally failing regression test that expects the original order to be preserved.

---

CleanShot 2026-07-10 at 22 19 23@2x


packages/zero-cache/src/services/replicator/change-processor.test.ts
```diff
@@ -3589,6 +3589,49 @@ describe('replicator/change-processor', () => {
...
+    backupProcessor.processMessage(lc, [
+      'commit',
+      messages.commit(),
+      {watermark: '03'},
+    ]);
+
+    const index = must(
+      listIndexes(backupReplica).find(({name}) => name === 'foo_tenant_id_id'),
+    );
+    expect(Object.keys(index.columns)).toEqual(['tenant_id', 'id']);
```
tantaman, line 3632: this currently fails -- demonstrating the bug.

packages/zero-cache/src/services/replicator/change-processor.ts
```diff
@@ -647,33 +647,64 @@ class TransactionProcessor {
...
-        `);
-      for (const idx of indexes) {
-        // Re-create the indexes to reference the new column.
-        idx.columns[tmpName] = idx.columns[oldName];
-        delete idx.columns[oldName];
-        stmts.push(createLiteIndexStatement(idx));
-      }
+      const tmpTable = `tmp.${table}`;
+      const newColumns = Object.fromEntries(
+        Object.entries(tableSpec.columns).map(([column, spec]) => [
```
arv: use mapEntries from objects.ts

packages/zero-cache/src/services/change-source/pg/schema-change-differential.pg.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    streamed = replicaFile.connect(lc);
+    const processor = createChangeProcessor(streamed);
+
+    await upstream`
+      ALTER TABLE foo ALTER COLUMN tenant_id SET NOT NULL
+    `;
+    await applyNextTransaction(lc, processor, changes);
+
+    const rebuilt = new Database(lc, ':memory:');
```
arv: no `using` support?

arv (dismissed):
LGTM

Very nice.

PR #6226, opened 2026-07-11
perf(zbugs): hoist list Row out of ListPage to stop remounts

arv:
Row was defined inside ListPage, giving it a new component identity every render, so React remounted the whole row subtree churning DOM the virtualizer measures and anchors. Move it to module scope, pass its closure values (issue, top, sortField, permalinkID, projectName, listContext, isLoggedIn) as props, and wrap in memo so unchanged rows skip re-rendering. Passing `top` as a number (not a fresh style object) keeps memo effective.


tantaman (commented):
makes sense

PR #6227, opened 2026-07-11
feat(zbugs): migrate to zero-virtual 0.6

arv:
Move off TanStack Virtual to the 0.6 flow-layout API (items + spaceBefore/ spaceAfter + rowAttributes), and rename use-wouter-permalink-state → use-wouter-scroll-state for scroll restoration.


tantaman (commented):
lgtm

PR #6232, opened 2026-07-14
fix(zero-client): avoid unhandled rejection when connect times out

steobrien:
We’ve sporadically seen this fire over the last month for various users.

I suspect this is some transient networking type issue, and not very serious, but it is causing us exception noise.

This implementation mirrors a few other callsites I saw like https://github.com/rocicorp/mono/blob/7da45c5589d550c22b6775f38afdee40fef3847f/packages/zero-cache/src/services/change-streamer/storer.ts#L520

(Note: we do forward connection errors to Sentry as recommended here: https://zero.rocicorp.dev/docs/connection?ref=localfirstnews.com#error, and this isn't coming through that way. It's arriving as an unhandled promise rejection.)

### DIFF
--- a/packages/zero-client/src/client/zero.test.ts
+++ b/packages/zero-client/src/client/zero.test.ts
@@ -73,6 +73,7 @@ import {
 import {createBuilder} from '../../../zql/src/query/create-builder.ts';
 import type {Row} from '../../../zql/src/query/query.ts';
 import {nanoid} from '../util/nanoid.ts';
+import {ActiveClientsManager} from './active-clients-manager.ts';
 import {ClientErrorKind} from './client-error-kind.ts';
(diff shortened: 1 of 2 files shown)


aboodman:
Thanks for the contribution. Two questions:

1. what is the message from the unhandled rejection you are seeing?
2. i'm always nervous about just catching these and swallowing them. what do you think about logging it to error or warning? (I guess it depends on which message you're seeing often).

steobrien:
1. `# ClientError: Connection attempt timed out after 10 seconds`
2. Yep I think that's a good call. Or accounting for this in `zero.connection.state`, potentially?

aboodman:
Superceded by https://github.com/rocicorp/mono/pull/6299. Sorry for slowness I was on vacation and had this half in my head and didnt want to hand off.

PR #6235, opened 2026-07-14
bench(zero-cache): add initial sync regression fixtures

0xcadams:
### What

Adds deterministic, production-shaped benchmark fixtures for initial sync and Binary COPY parsing. The profiles cover mixed tables, wide text rows, large JSON payloads, narrow row-heavy workloads, and chunk-boundary-sensitive parser streams.

### Why

The existing benchmarks did not represent observed wide and TOAST-sized payload shapes or isolate parser behavior when large fields are split across small COPY chunks. These fixtures establish repeatable baselines for evaluating initial-sync performance changes.

### Changes

- Adds profile-driven initial-sync fixtures with payload accounting and replica validation.
- Reworks the parser benchmark around deterministic binary streams, correctness checks, warmups, and a minimum of 50 samples and two measured seconds.
- Adds focused unit and PostgreSQL tests for fixture generation, corruption detection, indexes, and TOAST-sized replicas.

### Current (on a M5 Pro)

**Binary COPY parser**

| Profile | Field bytes | COPY chunk bytes | Columns | Rows | Median MiB/s | Average MiB/s |
|---|---:|---:|---:|---:|---:|---:|
| `contained-4k-31k` | 4,096 | 31,744 | 7 | 4,096 | 6,211.28 | 5,817.41 |
| `large-payload-270k-5.5k` | 276,480 | 5,632 | 7 | 64 | 1,312.26 | 1,173.02 |
| `wide-text-683k-31k` | 699,392 | 31,744 | 25 | 32 | 3,380.78 | 3,146.15 |
| `wide-text-683k-5.5k` | 699,392 | 5,632 | 25 | 16 | 446.32 | 432.01 |

**Initial sync**

| Profile | Rows | Payload MB (decimal) | Median MiB/s | Average MiB/s | Median elapsed | Average elapsed |
|---|---:|---:|---:|---:|---:|---:|
| `mixed-regression` | 250,000 | 493.787499 | 281.20 | 181.80 | 1.67s | 2.59s |
| `wide-text-scaled` | 1,000 | 683 | 345.40 | 303.61 | 1.89s | 2.15s |
| `wide-text-full` | 10,000 | 6,830 | 108.69 | 101.11 | 59.93s | 64.42s |
| `wide-text-narrow` | 250,000 | 32 | 13.09 | 12.01 | 2.33s | 2.54s |
| `large-payload-scaled` | 2,000 | 550 | 230.54 | 192.17 | 2.28s | 2.73s |
[...]

PR #6236, opened 2026-07-14
chore(zero-cache): add per-table phase timing logs

0xcadams:
### What

Adds per-table phase timings to initial sync completion logs so source waiting can be distinguished from destination processing for both binary and text COPY.

### Why

A table's total COPY duration and SQLite flush time do not show whether the remaining time is spent waiting for PostgreSQL or processing data inside Zero. This distinction helps determine whether parser, decoding, SQLite, or source-side work is the useful optimization target.

### Changes

- Adds `sourceWaitMs` and `processingMs` to the existing per-table COPY completion record.
- Measures parser, decoding, metric bookkeeping, SQLite statement, and final-flush work inside destination callbacks.
- Derives source wait as COPY stream wall time outside synchronous destination processing.

### DIFF
--- a/packages/zero-cache/src/db/initial-sync.bench.pg.ts
+++ b/packages/zero-cache/src/db/initial-sync.bench.pg.ts
@@ -29,7 +29,7 @@ const PROFILES = {
   'mixed-regression': {
     fixture: {fixture: 'mixed', rows: 250_000},
     warmupReps: 1,
-    reps: 10,
+    reps: 5,
     tableCopyWorkers: 4,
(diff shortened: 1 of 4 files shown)


tantaman:
lgtm. I'd also profile locally with various DB sizes. When we swapped to binary encoding, it sped up gigabugs copy by 50% because PG spent less time doing string encoding.

0xcadams:
Yeah, definitely, added https://github.com/rocicorp/mono/pull/6235 for this, based on some customer workload shapes

PR #6237, opened 2026-07-14
perf(zero-cache): batch initial sync copy metrics

0xcadams:
### What

Batches initial-sync COPY byte and chunk metric updates into 8 MiB increments instead of recording both OpenTelemetry counters for every PostgreSQL COPY chunk.

### Why

Initial sync currently calls the OpenTelemetry metrics API twice per COPY chunk. Large initial syncs can process many small chunks, making metric recording part of the hot path even though only the eventual byte and chunk totals matter.

Batching preserves the exported totals and labels while reducing calls into the metrics SDK.

### Changes

- Adds a small accumulator that flushes COPY byte and chunk totals every 8 MiB.
- Uses the accumulator in both binary and text COPY paths.
- Flushes residual totals on normal completion and stream destruction.
- Makes flushing idempotent so finalization and automatic destruction cannot duplicate metrics.

**Active OpenTelemetry initial sync**

| Profile | Rows | Payload MB (decimal) | Median MiB/s | Average MiB/s | Median elapsed | Average elapsed | Cumulative speedup vs [#6235](https://github.com/rocicorp/mono/pull/6235) |
|---|---:|---:|---:|---:|---:|---:|---:|
| `mixed-regression` | 250,000 | 493.787499 | 336.76 | 311.69 | 1.40s | 1.51s | +13.11% |
| `wide-text-scaled` | 1,000 | 683 | 340.87 | 345.82 | 1.91s | 1.88s | +0.08% |
| `wide-text-full` | 10,000 | 6,830 | 132.57 | 139.96 | 49.13s | 46.54s | +14.86% |
| `large-payload-scaled` | 2,000 | 550 | 323.84 | 229.82 | 1.62s | 2.28s | +25.34% |
| `large-payload-full` | 10,000 | 2,750 | 195.16 | 203.26 | 13.44s | 12.90s | +34.77% |

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/initial-sync-metrics.test.ts
+++ b/packages/zero-cache/src/services/change-source/pg/initial-sync-metrics.test.ts
@@ -1,0 +1,124 @@
+import {metrics} from '@opentelemetry/api';
+import {
+  AggregationTemporality,
+  DataPointType,
+  InMemoryMetricExporter,
+  MeterProvider,
+  PeriodicExportingMetricReader,
+} from '@opentelemetry/sdk-metrics';
+import {afterEach, describe, expect, test, vi} from 'vitest';
+
+afterEach(() => {
+  metrics.disable();
+  vi.resetModules();
+});
+
+describe('createCopyMetricBatcher', () => {
+  test('flushes below, at, and above the byte threshold', async () => {
+    const {COPY_METRIC_BATCH_BYTES, createCopyMetricBatcher} =
+      await import('./initial-sync.ts');
+    const record = vi.fn();
+    const batcher = createCopyMetricBatcher(record);
+
+    batcher.add(COPY_METRIC_BATCH_BYTES - 1);
+    expect(record).not.toHaveBeenCalled();
+
+    batcher.add(1);
+    expect(record).toHaveBeenLastCalledWith(COPY_METRIC_BATCH_BYTES, 2);
+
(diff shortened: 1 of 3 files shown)


packages/zero-cache/src/services/change-source/pg/initial-sync-metrics.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+afterEach(() => {
+  metrics.disable();
+  vi.resetModules();
+});
+
+describe('createCopyMetricBatcher', () => {
+  test('flushes below, at, and above the byte threshold', async () => {
+    const {COPY_METRIC_BATCH_BYTES, createCopyMetricBatcher} =
+      await import('./initial-sync.ts');
```
arv, line 19: prefer static imports

packages/zero-cache/src/services/change-source/pg/initial-sync-metrics.test.ts
```diff
@@ -0,0 +1,124 @@
...
+  DataPointType,
+  InMemoryMetricExporter,
+  MeterProvider,
+  PeriodicExportingMetricReader,
+} from '@opentelemetry/sdk-metrics';
+import {afterEach, describe, expect, test, vi} from 'vitest';
+
+afterEach(() => {
+  metrics.disable();
+  vi.resetModules();
```
arv, line 13: not needed?

arv (commented):
LGTM

PR #6238, opened 2026-07-15
perf(zero-cache): batch CVR row deletes + configurable deferral threshold

arv:
Reduces CVR flush cost, measured co-dominant with IVM advance (~46 ms/flush) in the RM→ViewSyncer fan-out investigation (see PR #6233 / lever A2 in `design-10x.md`).

## Changes

**Batch the per-row CVR delete (N+1 → 1).** `executeRowUpdates` issued one `DELETE … WHERE rowKey = …` per removed row, in both the synchronous and async row-flush paths. Replace with a single set-based delete driven by `json_to_recordset` of the `(schema, table, rowKey)` keys, mirroring the existing batched put. Rows are matched on their JSONB primary key; `rowKey` equality is canonical JSONB equality — the same comparison the per-row delete relied on.

**Make the deferral threshold configurable.** `deferredRowFlushThreshold` (the row-update count above which a flush writes row records back asynchronously instead of synchronously) was a hardcoded constructor default of 100 that production could not set. Expose it as `cvr.deferredRowFlushThreshold` (`ZERO_CVR_DEFERRED_ROW_FLUSH_THRESHOLD`, default 100) and plumb it through to `CVRStore`, so the sync/async row-flush split can be tuned per deployment.

## Testing

All view-syncer pg suites pass against local PG 16 (12 files / 170 tests), including `cvr-store.pg.test.ts` (deferred-row + stress + ownership/`FOR UPDATE` + row-catchup), `cvr.pg.test.ts` (real `deleteUnreferencedRows` deletes), and `view-syncer.pg.test.ts` (end-to-end row removals). Lint/format/check-types clean on the changed files.

## Deliberately not included (follow-up)

Folding the ownership `SELECT … FOR UPDATE` into the pipelined write batch to save one round trip per flush — kept out to keep this change small and low-risk (small payoff, touches the delicate flush transaction).


tantaman:
Any benchmarks on the delta?

packages/zero-cache/src/config/zero-config.test.ts
```diff
@@ -213,6 +213,13 @@ test('zero-cache --help', () => {
                                                                         Note that this number must allow for at least one connection per                                                           
                                                                         sync worker, or zero-cache will fail to start. See num-sync-workers                                                        
                                                                                                                                                                                                    
+     --cvr-deferred-row-flush-threshold number                          default: 100
```
tantaman, line 216: Besides memory use, is there any downside to buffering? Like it doesn't matter for latency iirc because the client is still required to wait for `pokeEnd` which happens after all CVR flushes complete.

arv: There is really no downside.

This option was there before as an argument to CVRStore but it wasn't reflected in the flags.

We could remove the flag if you do not think it is worth it.

arv:
> Any benchmarks on the delta?

Ran the actual hot-model test (where flush showed up as the bottleneck) before/after, saturating at 8 wps / 50 users / 2 sync workers, 45s:

cvr.flush-time: before (per-row) p50 24.6ms / mean 37.9ms — after (batched) p50 23.8ms / mean 35.3ms. Identical within noise.

Reason: 0 physical row deletes across every flush in both runs (~1,500+ each). This workload's flushes are all inserts — rows leaving a result set get written as refCounts:null tombstones via the insert path, not physically deleted, so the batched-delete path never executes. The 1–12× from the isolated bench only kicks in when a flush deletes a lot of rows (bulk unsubscribe/GC), which doesn't happen here. Correct and never slower, but a no-op for this workload; the flush cost is the metadata tx + row upsert, not deletes.

---

Given that it is questionable if this PR is worth the noise.

arv:
I'm closing this. We can always revisit batching the deletes later.

PR #6240, opened 2026-07-15
feat(zero-cache): event-driven flow-control release

arv:
## What

Makes the change-streamer's early **consensus-timeout** flow-control release **event-driven** instead of gated by the 1-second progress-monitor tick. Off by default behind a config flag.

## Why

Flow control resumes replication once a majority of subscribers have acked and `flowControlConsensusPaddingSeconds` has elapsed. Today that release condition (`Broadcast.checkProgress()`) is evaluated **only** by the Forwarder's 1 s `setInterval` (`forwarder.ts`). So even when the condition is already satisfied, the stream resumes up to ~1 s late.

This is the non-monotonic slow-subscriber penalty measured in the RM→ViewSyncer fan-out investigation: a *mildly*-slow subscriber (2–5 ms/msg) stays in the pending set at every checkpoint, and the whole stream waits for the next tick each time — collapsing healthy subscribers from ~2000/s to ~48/s (40×). `padding=0` doesn't help, because the release is still tick-gated.

## How

Once a majority acks, `Broadcast` schedules its own release exactly `flowControlConsensusPaddingSeconds` after the **most recent** completion (re-armed on each later completion via a monotonic generation guard so stale timers are ignored). The majority-plus-padding **condition is identical** to today — only the timing becomes precise. The 1 s tick and `checkProgress()` remain in place as a fallback and for logging, and `'all-subscribers'` release still takes precedence.

Gated behind a new flag, default **false** (no behavior change unless enabled):

```
changeStreamer.flowControlEventDrivenRelease
ZERO_CHANGE_STREAMER_FLOW_CONTROL_EVENT_DRIVEN_RELEASE
```

A negative `flowControlConsensusPaddingSeconds` still disables early release entirely, mirroring the existing `checkProgress()` path.

## Testing

- New unit tests in `broadcast.test.ts` (PG-free, injected timer): timer arms at majority; re-arms on later completions with stale timers ignored; `all-subscribers` precedence over a pending timer.
- Full `change-streamer` unit suite (`broadcast` / `forwarder` / `subscriber`) green; `--help` snapshot regenerated; `oxfmt` / `oxlint` / `tsc` clean.
- No perf numbers yet — this is a latency-precision change; a before/after on the fan-out harness (slow-subscriber scenario) is the natural follow-up.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/forwarder.ts
+++ b/packages/zero-cache/src/services/change-streamer/forwarder.ts
@@ -16,14 +16,16 @@ export type ProgressMonitorOptions = {
   // as the majority-plus-padding condition is met, instead of waiting for the
   // 1s progress tick to call checkProgress().
   eventDrivenRelease?: boolean | undefined;
-  // Injectable timer, for deterministic testing.
+  // Injectable timers, for deterministic testing.
   setTimeoutFn?: typeof setTimeout | undefined;
+  clearTimeoutFn?: typeof clearTimeout | undefined;
 };
 
 export class Forwarder {
   readonly #lc: LogContext;
   readonly #progressMonitorOptions: ProgressMonitorOptions;
   readonly #setTimeout: typeof setTimeout;
+  readonly #clearTimeout: typeof clearTimeout;
   readonly #active = new Set<Subscriber>();
   readonly #queued = new Set<Subscriber>();
   readonly #flowControlWaits = getOrCreateCounter(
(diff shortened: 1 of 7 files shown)


packages/zero-cache/src/services/change-streamer/forwarder.ts
```diff
@@ -5,18 +5,25 @@ import {
...
+import {Broadcast, type EarlyReleaseOptions} from './broadcast.ts';
 import type {ChangeTag, WatermarkedChange} from './change-streamer-service.ts';
 import type {Status} from './change-streamer.ts';
 import type {Subscriber} from './subscriber.ts';
 
 export type ProgressMonitorOptions = {
   flowControlConsensusPaddingSeconds: number;
+  // When true, broadcasts schedule their own consensus-timeout release as soon
+  // as the majority-plus-padding condition is met, instead of waiting for the
+  // 1s progress tick to call checkProgress() (see B2).
```
arv: remove see b2... it makes no sense in isolation

PR #6247, opened 2026-07-16
chore(deps): upgrade typescript to 7.0.2

arv:
Bump typescript from ~6.0.2 to ~7.0.2 across all workspace packages, except replicache-doc.

replicache-doc stays on ~6.0.2 because its documentation toolchain (typedoc / docusaurus) does not support the TypeScript 7 native compiler — typedoc's peer range tops out at typescript 6.0.x. A syncpack versionGroup marks this as a deliberate, ignored exception so `syncpack lint` still passes.

TypeScript 7 enforces the declaration-emit portability check (TS2883) more strictly. Six internal packages that only type-check (noEmit) and never emit .d.ts — zql, zql-integration-tests, zero-cache, zero-permissions, zql-benchmarks, zero-react — trip this on cross-package relative source imports under pnpm's symlinked layout. They opt out via declaration/declarationMap: false, since the check is moot for packages that don't emit declarations. Packages that do emit declarations (replicache, zero, zero-events) are unaffected and keep the check.

Regenerate the tsnapi API snapshots to reflect TS7's declaration emit ordering (alphabetized union literals, required-before-optional properties); the public API surface is unchanged.

### DIFF
--- a/.syncpackrc
+++ b/.syncpackrc
@@ -1,4 +1,12 @@
 {
+  "versionGroups": [
+    {
+      "label": "replicache-doc pins TypeScript 6.x: its typedoc/docusaurus toolchain does not support the TypeScript 7 native compiler",
+      "packages": ["replicache-doc"],
+      "dependencies": ["typescript"],
+      "isIgnored": true
+    }
+  ],
   "semverGroups": [
     {
       "range": "~",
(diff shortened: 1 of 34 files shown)


arv:
Closing. Vercel is not ready yet.

PR #6249, opened 2026-07-16
chore: after shedding VS load, keep it shedded till VS catches up

tantaman:
CleanShot 2026-07-16 at 13 51 11@2x


tantaman:
will probably move batching upstream so it benefits all downstream components : https://rocicorp.slack.com/archives/C013XFG80JC/p1784224782492969

PR #6256, opened 2026-07-22
chore(rmv2): add SQLite change log schema [2/n]

tantaman:
Add the _zero.changeLogStream table and partial retention index to fresh replicas and the additive v14 migration. Seed an idempotent synthetic begin/commit transaction at the replica's current stateVersion and reuse its writeTimeMs as the initial retained boundary.

Production state after deployment:

- Fresh, restored, and upgraded replicas contain the new table, index, and exactly one synthetic two-row transaction at their migration watermark.
- Existing stream history is not backfilled. A future SQLite reader could only serve requests at or after the synthetic seed boundary.
- This slice does not append ongoing changes, read from the table, or purge it. After seeding, the table remains inert and has negligible disk cost.
- PostgreSQL remains the authoritative change log for catchup, purge, subscriber delivery, and replication-slot ACK durability.
- _zero.changeLog2 and replica data semantics are unchanged. The migration is additive, does not raise the minimum safe version or trigger an auto-reset, and older builds safely ignore the table.
- Rollback followed by roll-forward is safe because seeding is idempotent and cannot duplicate the synthetic rows.

```
 Column         Purpose
━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 watermark      The lexicographically sortable version identifying the transaction. Every
                row belonging to a transaction gets the same watermark. For the latest
                transaction, it matches _zero.replicationState.stateVersion. Catchup and
                purge compare watermarks to select whole transactions. It is TEXT because
                Zero versions are encoded strings and may have forms such as 123.01, rather
                than being plain integers.
─────────────  ─────────────────────────────────────────────────────────────────────────────
 pos            The message’s zero-based position inside the transaction. 0 is begin,
                intermediate positions are data/schema/truncate/backfill messages, and the
                final position is commit. It is a dedicated stream counter, separate from
                _zero.changeLog2’s row-operation counter.
─────────────  ─────────────────────────────────────────────────────────────────────────────
 change         The canonical JSON text for the inner change object, such as
                {"tag":"begin"} or {"tag":"insert", ...}. It does not store the outer
                ["begin", ...], ["data", ...], or ["commit", ...] envelope, nor duplicate
                the watermark metadata. The reader reconstructs those around the stored
                text.
─────────────  ─────────────────────────────────────────────────────────────────────────────
 precommit      Populated only on the commit row. It contains the transaction watermark
                remembered from the corresponding begin. The writer verifies that it equals
                the final commit watermark. In well-formed SQLite rows it is therefore
                normally identical to watermark. It preserves the old PG concept, provides
                integrity/debugging information, and acts as an explicit marker that a
                watermark is a complete catchup boundary. It does not mean “the previous
                committed transaction.”
─────────────  ─────────────────────────────────────────────────────────────────────────────
 writeTimeMs    Populated only on the commit row. It is the local Unix epoch time in
                milliseconds captured while committing the transaction—not the original
                PostgreSQL commit timestamp. The exact same value is written to
                _zero.replicationState.writeTimeMs. It supports time-based retention and
                backup/replication lag accounting.
```

### DIFF
--- a/packages/zero-cache/src/services/replicator/schema/change-log-stream.ts
+++ b/packages/zero-cache/src/services/replicator/schema/change-log-stream.ts
@@ -1,0 +1,61 @@
+import {assert} from '../../../../../shared/src/asserts.ts';
+import type {Database} from '../../../../../zqlite/src/db.ts';
+
+export const CHANGE_LOG_STREAM_TABLE = '_zero.changeLogStream';
+export const CHANGE_LOG_STREAM_WRITE_TIME_INDEX =
+  '_zero.changeLogStream_writeTimeMs';
+
+export const CREATE_CHANGE_LOG_STREAM_TABLE = /*sql*/ `
+  CREATE TABLE "${CHANGE_LOG_STREAM_TABLE}" (
+    "watermark"   TEXT NOT NULL,
+    "pos"         INTEGER NOT NULL,
+    "change"      TEXT NOT NULL,
+    "precommit"   TEXT,
+    "writeTimeMs" INTEGER,
+    PRIMARY KEY ("watermark", "pos")
+  );
+`;
+
+export const CREATE_CHANGE_LOG_STREAM_INDEX = /*sql*/ `
+  CREATE INDEX "${CHANGE_LOG_STREAM_WRITE_TIME_INDEX}"
+    ON "${CHANGE_LOG_STREAM_TABLE}" ("writeTimeMs", "watermark")
+    WHERE "writeTimeMs" IS NOT NULL;
+`;
+
+export const CREATE_CHANGE_LOG_STREAM_SCHEMA =
+  CREATE_CHANGE_LOG_STREAM_TABLE + CREATE_CHANGE_LOG_STREAM_INDEX;
+
+type ReplicationState = {
+  stateVersion: string;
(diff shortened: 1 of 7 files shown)


packages/zero-cache/src/services/replicator/schema/change-log-stream.ts
```diff
@@ -0,0 +1,61 @@
...
+export const CHANGE_LOG_STREAM_WRITE_TIME_INDEX =
+  '_zero.changeLogStream_writeTimeMs';
+
+export const CREATE_CHANGE_LOG_STREAM_TABLE = /*sql*/ `
+  CREATE TABLE "${CHANGE_LOG_STREAM_TABLE}" (
+    "watermark"   TEXT NOT NULL,
+    "pos"         INTEGER NOT NULL,
+    "change"      TEXT NOT NULL,
+    "precommit"   TEXT,
+    "writeTimeMs" INTEGER,
```
tantaman, line 14: this is used to compute a grace period against which to drop row. E.g., `watermark - 60_000` or `min(backup_floor, time_floor)`

PR #6258, opened 2026-07-22
chore(rmv2): add SQLite change log writer [5/n]

tantaman:
Writes to the SQLite change log when enabled.

Production state after deployment:

- With the default sqliteChangeLogMode=off, runtime behavior is unchanged and the existing SQLite change-log table remains inert.
- Any non-off mode currently enables only shadow writes on the canonical backup replica, or on the primary serving replica when no backup replica exists. serving-copy and non-canonical process trees do not write.
- Replica changes, stream rows, and stateVersion commit atomically with a shared writeTimeMs. Rollbacks, failures, aborts, and source disconnects leave no partial stream transaction.
- PostgreSQL remains authoritative for catchup, purge, subscriber delivery, and replication-slot ACK durability. This slice does not serve or compare SQLite reads.
- SQLite purge and retention scheduling are not implemented yet, so enabling writes in production would grow the local log without bound. Keep the mode off except for explicitly disk-bounded probes until the purge slice lands.

### DIFF
--- a/packages/zero-cache/src/server/replicator.ts
+++ b/packages/zero-cache/src/server/replicator.ts
@@ -40,6 +40,7 @@ import {
 import {getShardConfig} from '../types/shards.ts';
 import {
   getPragmaConfig,
+  replicaLogsChangeStream,
   replicaFileModeSchema,
   setUpMessageHandlers,
   setupReplica,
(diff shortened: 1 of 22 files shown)


packages/zero-cache/src/server/replicator.ts
```diff
@@ -46,6 +47,18 @@ export default async function runWorker(
 
   const config = getNormalizedZeroConfig({env, argv: args.slice(1)});
   const mode: ReplicatorMode = fileMode === 'backup' ? 'backup' : 'serving';
+  const runningLocalChangeStreamer =
+    config.changeStreamer.mode === 'dedicated' && !config.changeStreamer.uri;
+  const logsChangeStream = replicaLogsChangeStream(
```
tantaman, line 67: check if logging the change stream to sqlite has been enabled.

PR #6259, opened 2026-07-22
fix(zero-cache): avoid restore attempt and purge lock when litestream is not configured

darkgnotic:
Motivation: https://github.com/rocicorp/mono/pull/6250


darkgnotic:
@asterikx

0xcadams (approved):
LGTM

PR #6260, opened 2026-07-22
chore(litestream): default to restoring with litestream v5

darkgnotic:
Default to restoring with litestream v5. Backups are still performed using v3.

This provides a release for safely rolling back from a (future) release that backs up with v5.

This can be disabled with `ZERO_LITESTREAM_RESTORE_USING_V5=false`.

Also bump to the v5 litestream version to 0.5.15.


0xcadams (approved):
LGTM

PR #6263, opened 2026-07-23
chore(rmv2): observe SQLite change log shadow writes [6/n]

tantaman:
Logs metrics so we can determine if the SQLite change log is performing correctly and roughly matching our current PG change log.

I say "roughly matching" because we will not know if it exactly matches without reading the contents back. A follow up PR does sampled reads to check.

There is a corresponding commit in CloudZero (https://github.com/rocicorp/cloudzero/pull/1396) that will add a dashboard to surface this information. It will look roughly like:

```
### A. Everything is fine

  ┌──────────────────────────────────────────────────────────────────────────────┐
  │ Cloud Zero — SQLite Change Log Rollout                         Last 6 hours  │
  ├──────────────────────────────────────────────────────────────────────────────┤
  │ Writer       Head lag     Hash mismatch   Unpaired    Invariants   Rollbacks │
  │ present: 1   0             0               0           0            0         │
  │              occasional                                                      │
  │              brief spikes                                                    │
  └──────────────────────────────────────────────────────────────────────────────┘

  SQLite head lag
    2 │       ╭╮
    1 │  ╭╮   │╰╮       ╭╮
    0 └──╯╰───╯ └───────╯╰────────────
         Briefly rises during a commit, then returns to zero

  Retained payload / rows
      │      /|      /|      /|
      │     / |     / |     / |
      │____/  |____/  |____/  |____
             purge    purge    purge
      Expected sawtooth or bounded growth

  Message/commit latency           Shadow commit rate
      stable p50/p95/p99             follows transaction traffic
      no upward trend                no unexplained gaps

  Failures                         Production correlates
      rollbacks: 0                  normal PG replication lag
      invariants: 0                 DB/WAL usage remains bounded

  The most important healthy signature is:

  Hash Mismatches = 0
  Unpaired Hashes = 0
  Head Lag returns to 0

  Some short head-lag spikes are normal because the commit is received before SQLite finishes writing it.

  ### B. Things are wrong

  A divergence or writer failure would make the top row immediately noisy:

  ┌──────────────────────────────────────────────────────────────────────────────┐
  │ Writer       Head lag     Hash mismatch   Unpaired    Invariants   Rollbacks │
  │ present: 1   847           3  RED          12 RED      2 RED        18        │
  └──────────────────────────────────────────────────────────────────────────────┘

  SQLite head lag
  850 │                              ╭────────────
  600 │                         ╭────╯
  300 │                    ╭────╯
    0 └────────────────────╯
                           SQLite is no longer catching up

  Retained payload / rows
      │                         /
      │                     ___/
      │                 ___/
      │____________ ___/
                   No purge; storage grows continuously

  Commit latency
      │                      ╭╮
      │                 ╭╮  ╭╯╰╮
      │_________________╯╰──╯  ╰────
                       sustained latency increase

  Failures
      hash mismatch:    nonzero → source and SQLite transaction differed
      unpaired:         nonzero → commit failed or one hash was unavailable
      invariant errors: transaction boundaries/state are inconsistent
      rollbacks:        processing errors or repeated source interruptions
```

### DIFF
--- a/packages/zero-cache/src/observability/metrics.ts
+++ b/packages/zero-cache/src/observability/metrics.ts
@@ -119,6 +119,27 @@ const LATENCY_HISTOGRAM_BOUNDARIES_S = [
 
 const histograms = cache<Histogram>();
 
+/**
+ * Creates a histogram for dimensionless values without applying the
+ * millisecond-to-second conversion used by the latency helpers below.
+ */
+export function getOrCreateValueHistogram(
+  category: Category,
+  name: string,
+  opts: HistogramOptions,
+): Histogram {
+  const {bucketBoundaries, ...metricOptions} = opts;
+  return histograms(name, name =>
+    getMeter().createHistogram(`zero.${category}.${name}`, {
+      ...metricOptions,
+      advice: {
+        ...metricOptions.advice,
+        explicitBucketBoundaries: bucketBoundaries,
+      },
+    }),
+  );
+}
+
 export function getOrCreateHistogram(
   category: Category,
   name: string,
--- a/packages/zero-cache/src/server/replicator.ts
+++ b/packages/zero-cache/src/server/replicator.ts
@@ -34,6 +34,7 @@ import {
 } from '../services/replicator/replicator.ts';
 import {
   getSQLiteChangeLogInfo,
+  getSQLiteChangeLogStartupInfo,
   logSQLiteChangeLogStartup,
   SQLiteChangeLogObserver,
 } from '../services/replicator/sqlite-change-log-observability.ts';
@@ -50,6 +51,7 @@ import {
   replicaFileModeSchema,
   setUpMessageHandlers,
   setupReplica,
+  type ReplicaFileMode,
   type WalMode,
 } from '../workers/replicator.ts';
 import {createLogContext} from './logging.ts';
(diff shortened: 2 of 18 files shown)


packages/zero-cache/src/observability/metrics.ts
```diff
@@ -119,6 +119,27 @@ const LATENCY_HISTOGRAM_BOUNDARIES_S = [
 
 const histograms = cache<Histogram>();
 
+/**
+ * Creates a histogram for dimensionless values without applying the
```
tantaman, line 123: i.e., a value (count) histogram rather than latency (ms/seconds) histogram.

packages/zero-cache/src/services/replicator/sqlite-change-log-observability.ts
```diff
@@ -0,0 +1,349 @@
...
+    .prepare(/*sql*/ `
+      SELECT state."stateVersion", config."replicaVersion"
+        FROM "_zero.replicationState" AS state,
+             "_zero.replicationConfig" AS config
+    `)
+    .get<{stateVersion: string; replicaVersion: string} | undefined>();
+  const aggregate = db
+    .prepare(/*sql*/ `
+      SELECT max("watermark") AS "headWatermark",
+             count(*) AS "rows",
```
tantaman: make sure this doesn't run when off since this can add startup time for large logs.

packages/zero-cache/src/server/replicator.ts
```diff
@@ -99,6 +105,12 @@ export default async function runWorker(
     dbPath,
   });
 
+  const changeLogInfo = readSQLiteChangeLogInfo(lc, dbPath);
+  logSQLiteChangeLogStartup(lc, fileMode, logsChangeStream, changeLogInfo);
```
tantaman: what're we looking for here?

PR #6264, opened 2026-07-23
feat(rmv2): add SQLite change log reader [7/n]

tantaman:
Adds the SQLite ChangeLog reader. Inert in production but exercised in tests.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
+++ b/packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
@@ -7,7 +7,7 @@ import {
   reconstructWatermarkedChange,
   type ChangeLogEntry,
 } from './change-log-codec.ts';
-import type {ChangeTag, WatermarkedChange} from './change-streamer.ts';
+import type {WatermarkedChange} from './change-streamer.ts';
 
 export type CatchupPlan =
   | {
@@ -31,35 +31,7 @@ type PlanRow = {
 
 type ChangeLogRow = ChangeLogEntry & {
   readonly pos: number;
-  readonly tag: string | null;
 };
-
-type TransactionCursor = {
-  watermark: string;
-  nextPos: number;
-  complete: boolean;
-};
-
-const CHANGE_TAGS: ReadonlySet<string> = new Set<ChangeTag>([
-  'begin',
-  'insert',
-  'update',
-  'delete',
-  'truncate',
-  'backfill',
-  'create-table',
-  'rename-table',
-  'update-table-metadata',
-  'add-column',
-  'update-column',
-  'drop-column',
-  'drop-table',
-  'create-index',
-  'drop-index',
-  'backfill-completed',
-  'commit',
-  'rollback',
-]);
 
 /**
  * Seeks to the final position so validating a boundary does not scan every row
@@ -168,7 +140,7 @@ export class SQLiteChangeLogReader implements Disposable {
     // The first query must be strictly after the requested transaction. Every
     // real stream position is far below this sentinel.
     let lastPos = Number.MAX_SAFE_INTEGER;
-    let transaction: TransactionCursor | undefined;
+    let lastTag: string | undefined;
 
     while (true) {
       this.#throwIfAborted(signal);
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
```diff
@@ -0,0 +1,285 @@
...
+  'drop-column',
+  'drop-table',
+  'create-index',
+  'drop-index',
+  'backfill-completed',
+  'commit',
+  'rollback',
+]);
+
+const PLAN_SQL = /*sql*/ `
```
tantaman, line 51: can a given subscriber be caught up from our change log?

tantaman: also tells us the range for the subscriber to query over.

We need a range because we do not pin a snapshot for the duration of the catchup. The snapshot is released between each chunk.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
```diff
@@ -0,0 +1,285 @@
...
+  SELECT
+    state."stateVersion" AS "headWatermark",
+    bounds."minWatermark" AS "minWatermark",
+    @fromWatermark > state."stateVersion" AS "subscriberAhead",
+    EXISTS (
```
tantaman: this exists worries me for very large TX catchup. Let me see what I can do here.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
```diff
@@ -0,0 +1,285 @@
...
+    ) AS "boundaryExists"
+  FROM "_zero.replicationState" AS state
+  CROSS JOIN (
+    SELECT min("watermark") AS "minWatermark"
+      FROM "${CHANGE_LOG_STREAM_TABLE}"
+  ) AS bounds
+`;
+
+/** Exported so the focused query-plan test covers the production query. */
+export const SQLITE_CHANGE_LOG_READ_BATCH_SQL = /*sql*/ `
```
tantaman, line 65: reads a batch at a time so we can co-operatively schedule

packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
```diff
@@ -0,0 +1,285 @@
...
+    // The first query must be strictly after the requested transaction. Every
+    // real stream position is far below this sentinel.
+    let lastPos = Number.MAX_SAFE_INTEGER;
+    let transaction: TransactionCursor | undefined;
+
+    while (true) {
+      this.#throwIfAborted(signal);
+      const rows = this.#db.transaction(() =>
+        this.#db
+          .prepare(SQLITE_CHANGE_LOG_READ_BATCH_SQL)
```
tantaman: nit: we should not prepare each time

packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
```diff
@@ -0,0 +1,285 @@
...
+
+    let lastWatermark = fromWatermark;
+    // The first query must be strictly after the requested transaction. Every
+    // real stream position is far below this sentinel.
+    let lastPos = Number.MAX_SAFE_INTEGER;
+    let transaction: TransactionCursor | undefined;
+
+    while (true) {
+      this.#throwIfAborted(signal);
+      const rows = this.#db.transaction(() =>
```
tantaman: nit: not needed

packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
tantaman: @darkgnotic - for reading from the SQLite change log I'm assuming we do not want to keep a transaction open for the full duration of catchup?

tantaman: I'm going to go ahead and land since this isn't called anywhere yet. But what happens is I find the bounds of the log that the client needs then release the TX. This lets the client catch up without pinning the wal.

PR #6266, opened 2026-07-23
chore(zero-cache): decouple PurgeLock from restore

darkgnotic:
Remove the logic that releases the purge lock when a litestream restore fails.

Instead, the purge lock is released if needed when the change-streamer schema "ensured" (which happens after initial sync). This has the same result but decouples change-log-specific logic from the litestream restore, the latter of which is being moved into the change-source initialization logic in order to support replica-specific backup locations (for litestream v5 / RMv2).

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
+++ b/packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
@@ -41,7 +41,6 @@ import {
 } from './change-streamer.ts';
 import * as ErrorType from './error-type-enum.ts';
 import {Forwarder} from './forwarder.ts';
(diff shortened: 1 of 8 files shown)


packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
```diff
@@ -77,14 +76,13 @@ export async function initializeStreamer(
   opts: TuningOptions,
   setTimeoutFn = setTimeout,
 ): Promise<ChangeStreamerService> {
-  // Make sure the ChangeLog DB is set up.
-  await initChangeStreamerSchema(lc, changeDB, shard);
```
tantaman, line 81: where do you call this now? I only see it called from tests.

PR #6268, opened 2026-07-24
chore(zero-cache): refactor/separate restore logic

darkgnotic:
Split the logic for replication-manager vs view-syncer restores. The main motivation is to abstract view-syncer specific logic out of the core litestream library to facilitate adding flexibility to how replicas are restored in the replication-manager.

* The `view-syncer`'s snapshot reservation and continuous retry logic is moved into the `replicator.ts` worker.
* The `replication-manager`'s restore logic is moved into `change-source/replica-restore.ts`. This is mainly a stats-tracking wrapper around the shared `tryRestore()` method remaining in `services/litestream`.
* `main.ts` no longer has any litestream restore logic

This is mostly a behavior-preserving refactoring, with the exception of simplifying the litestream log format to always be 'json'.

### DIFF
--- a/packages/zero-cache/src/services/change-source/common/replica-restore.ts
+++ b/packages/zero-cache/src/services/change-source/common/replica-restore.ts
@@ -1,0 +1,39 @@
+import type {LogContext} from '@rocicorp/logger';
+import type {LitestreamConfig} from '../../../config/normalize.ts';
+import {
+  tryRestore,
+  type ReplicaConstraints,
+  type RestoreResult,
+} from '../../litestream/commands.ts';
+import {
+  litestreamRestoreDuration,
+  litestreamRestoreMetricAttrs,
+  litestreamRestoreRuns,
+} from '../../litestream/metrics.ts';
+
+/** @returns `true` if the replica was restored, `false` if not found */
+export async function restoreReplica(
+  lc: LogContext,
+  config: LitestreamConfig,
+  replicaFile: string,
+  replicaConstraints: ReplicaConstraints,
+): Promise<boolean> {
+  const start = performance.now();
+  let result: RestoreResult | undefined;
+  try {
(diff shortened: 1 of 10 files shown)


packages/zero-cache/src/services/change-source/common/replica-restore.ts
```diff
@@ -0,0 +1,39 @@
...
+  type RestoreResult,
+} from '../../litestream/commands.ts';
+import {
+  litestreamRestoreDuration,
+  litestreamRestoreMetricAttrs,
+  litestreamRestoreRuns,
+} from '../../litestream/metrics.ts';
+
+/** @returns `true` if the replica was restored, `false` if not found */
+export async function restoreReplica(
```
tantaman, line 15: a little confusing that this has the same name as the `reportReplica` in the file above.

packages/zero-cache/src/services/change-streamer/snapshot.ts
```diff
@@ -46,3 +53,74 @@ const statusMessageSchema = v.tuple([v.literal('status'), statusSchema]);
...
+  void (async function () {
+    const abort = new AbortController();
+    process.on('SIGINT', () => abort.abort());
+    process.on('SIGTERM', () => abort.abort());
+
+    for (let i = 0; ; i++) {
+      let err: unknown;
+      try {
+        let resolved = false;
+        const stream = await reserveSnapshot(lc, config);
```
tantaman, line 72: naming again 😅

`reserveAndGetSnapshotStatus` vs `reserveSnapshot` -- given the similar names I'd expect them to execute roughly identically with the exception of return type. This isn't true tough. `reserveAndGetSnapshotStatus` loops continuously until it reserves vs `reserveSnapshot` is a 1-shot attempt

PR #6272, opened 2026-07-24
chore(rmv2): add SQLite catchup handoff [8/n]

tantaman:
This does SQLite catchup against the SQLite change log.

**Note:** This'll be changing quite a bit as the current change log in this PR (and prior PRs) is stored in the replica file. It should be its own SQLite file. Follow up PRs will undertake this refactoring.

----

This is currently off in production.

## Algo

A subscriber connects. It checks SQLite to see what transaction it has durably recorded. After that, we wait till all in-flight transactions are persisted to SQLite. The reason here is because we stream transactions out from the RM before saving to SQLite so a subscriber can connect in a window where:

a. transactions have left
b. those transactions are not in the change log

**Example:**

Subscriber watermark:       50
Forwarder already through: 101
SQLite currently through:   98

The sequence is:

1. Register subscriber with Forwarder
2. Capture requiredHead = 101
3. plan() reports SQLite head = 98
4. Park subscriber; do not start reading yet
5. SQLite writes 99, 100, 101
6. plan() now reports head >= 101
7. Pin that new head and read SQLite from 50 through it
8. Drain the subscriber’s live backlog
9. Continue direct live forwarding

While steps 4–7 happen:

- Transactions 99–101 were forwarded before registration, so they are not in this subscriber’s backlog. They must come from SQLite.
- Transactions forwarded after registration—say 102+—are buffered in the subscriber’s live backlog.
- If the eventual pinned SQLite head includes some of 102+, those transactions exist in both catchup and the backlog. Subscriber watermark deduplication
  removes the overlap.

The backlog is bounded and participates in flow control, so an extremely slow catchup can eventually backpressure live replication rather than grow memory
without limit.

## Catch cooperatively schedules at async/batch boundaries.

- Each catchup runs as a detached async task.
- It yields while waiting for the required head, barrier polling, and subscriber consumption.
- Each SQLite all() query and row reconstruction is synchronous and blocks the change-streamer event loop for that batch.
- There is no explicit setImmediate()/macrotask yield between batches.

Batches are bounded and each read transaction closes promptly. It might need explicit yielding for
event-loop fairness if benchmarks show that many fast catchups or oversized batches hurt live-stream latency.

## Catchups are concurrent, not globally serialized.

Every subscriber gets its own #run() invocation in packages/zero-cache/src/services/change-streamer/sqlite-change-log-catchup.ts:156.

They are concurrent async tasks but not parallel SQLite executions:
- All use one shared SQLiteChangeLogReader.
- All run on the same JavaScript thread.
- Synchronous SQLite batch queries execute one at a time.
- Tasks interleave when they reach an await.

There is currently no catchup concurrency limiter.

## Catchup is served from the change-streamer’s main JavaScript thread.

packages/zero-cache/src/services/change-streamer/change-streamer-service.ts:539 lazily creates SQLiteChangeLogReader directly. It does not send reads to
ThreadWriteWorkerClient.

More precisely, this is the main thread of the change-streamer worker process—not the overall zero-cache parent process and not the canonical SQLite write
worker.

## It does not pin one long-lived SQLite transaction.

The sequence is:

plan()
   → read head/min/boundary in one short statement
   → remember headWatermark as a value

read(..., through=headWatermark)
   → SELECT next LIMIT batchSize
   → fully exhaust statement and close its implicit read transaction
   → yield batch
   → repeat with a new short read transaction

The keyset query is bounded by the remembered head:

```
WHERE ("watermark", "pos") > (?, ?)
 AND "watermark" <= ?
ORDER BY "watermark", "pos"
LIMIT ?
```

### DIFF
--- a/packages/zero-cache/src/server/change-streamer.ts
+++ b/packages/zero-cache/src/server/change-streamer.ts
@@ -55,6 +55,8 @@ export default async function runWorker(
       backPressureLimitHeapProportion,
       flowControlConsensusPaddingSeconds,
       flowControlEventDrivenRelease,
+      sqliteChangeLogReadBatchRows,
+      sqliteChangeLogBarrierTimeoutMs,
     },
     autoReset,
     replicationLag,
@@ -165,6 +167,11 @@ export default async function runWorker(
           flowControlEventDrivenRelease,
           statementTimeoutMs: change.statementTimeoutMs,
           changeLogBatchSize: change.logBatchSize,
+          sqliteCatchup: {
+            replicaFile: replica.file,
+            readBatchRows: sqliteChangeLogReadBatchRows,
+            barrierTimeoutMs: sqliteChangeLogBarrierTimeoutMs,
+          },
         },
         setTimeout,
       );
--- a/packages/zero-cache/src/config/normalize.test.ts
+++ b/packages/zero-cache/src/config/normalize.test.ts
@@ -15,7 +15,7 @@ function configWith(litestream: Partial<ZeroConfig['litestream']>): ZeroConfig {
       sqliteChangeLogRetentionMs: 60_000,
       sqliteChangeLogReadBatchRows: 1000,
       sqliteChangeLogPurgeBatchRows: 1000,
-      sqliteChangeLogBarrierTimeoutMs: 30_000,
+      sqliteChangeLogBarrierTimeoutMs: 300_000,
     },
     change: {db: 'postgres:///change'},
     cvr: {db: 'postgres:///cvr'},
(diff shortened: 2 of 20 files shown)


packages/zero-cache/src/server/change-streamer.ts
```diff
@@ -165,6 +167,11 @@ export default async function runWorker(
           flowControlEventDrivenRelease,
           statementTimeoutMs: change.statementTimeoutMs,
           changeLogBatchSize: change.logBatchSize,
+          sqliteCatchup: {
+            replicaFile: replica.file,
+            readBatchRows: sqliteChangeLogReadBatchRows,
+            barrierTimeoutMs: sqliteChangeLogBarrierTimeoutMs,
```
tantaman, line 173: the barrier exists because we stream changes from the replicator to clients before committing to the change log. Thus it is a possibility that a view-syncer is ahead of the replicator when catching up. Unlikely but in the realm of the possible.

PR #6278, opened 2026-07-28
chore(rmv2): retarget the change-log reader to its own database [10/n]

tantaman:
Point the sqlite change log reader at the sidecar sqlite change log file.

Stack:
- https://github.com/rocicorp/mono/pull/6277
- https://github.com/rocicorp/mono/pull/6276

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
+++ b/packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
@@ -20,17 +20,25 @@ export type CatchupPlan =
       readonly kind: 'too-old';
       readonly minWatermark: string;
       readonly headWatermark: string;
-    };
+    }
+  // The log exists as a file but has no content to serve: the replicator has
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.ts
```diff
@@ -20,19 +20,27 @@ export type CatchupPlan =
       readonly kind: 'too-old';
       readonly minWatermark: string;
       readonly headWatermark: string;
-    };
+    }
+  // The log exists as a file but has no content to serve: the replicator has
+  // not created or reconciled it yet. Slice 7E declines and falls back to PG
```
tantaman, line 25: falling back to PG will probably never happen in practice as none of this will be turned on until we're ready to deprecate the PG path.

We will:
1. write the change log before deprecating PG
2. compact it
3. sample it

but not actually serve catchup

arv (commented):
LGTM

PR #6280, opened 2026-07-28
fix(zero-server): avoid primary key updates in CRUD

shayonj:
The server CRUD executor currently includes primary-key columns in the `SET` list for `update` and for the conflict branch of `upsert`, even though those values identify the existing row and cannot change through these APIs.

PostgreSQL treats a syntactically updated key column as possibly changing and takes a row lock that conflicts with foreign-key checks on concurrent child inserts. This change leaves primary keys in the `WHERE`, insert values, and conflict target, but excludes them from update assignments. A key-only update becomes a no-op, and a key-only upsert uses `DO NOTHING` on conflict.

The tests pin the generated SQL and verify with PostgreSQL that an open CRUD update or upsert transaction does not block a concurrent foreign-key insert.

### DIFF
--- a/packages/zero-server/src/custom.pg.test.ts
+++ b/packages/zero-server/src/custom.pg.test.ts
@@ -192,6 +192,44 @@ describe('makeSchemaCRUD', () => {
     });
   });
 
+  test('update and upsert allow concurrent foreign key inserts', async () => {
+    await pg`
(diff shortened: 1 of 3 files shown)


cobrien-tines (approved):

aboodman:
This makes sense to me. Thanks for the contribution.

PR #6284, opened 2026-07-29
chore(rmv2): decline SQLite catchup when the change log is not ready [12/n]

tantaman:
The change-streamer and the replicator that writes the log start independently (two different processe) so the log may not exist when the first subscription arrives to read it.

Production effect: none — `shouldUse` is still unwired, so no production subscription reaches this code.


arv (commented):
LGTM

PR #6288, opened 2026-07-29
chore(rmv2): add the incremental change-log purger

tantaman:
Retention-floor SQL with bounded progress, testable without a scheduler, on a file that can later reclaim space. Nothing calls it: slice 9 lands the scheduler.

New `sqlite-change-log-purger.ts` promotes the prototype purge statements out of `sqlite-change-log-ceiling.bench.ts`, with three corrections:

- The head cap reads `max("watermark")` from the log's own database instead of taking a `stateVersionFloor` from the replica. Log-first commit ordering makes the log's head lead the replica's `stateVersion`, and the two no longer share a file, so the caller had no way to supply the right value.
- `"watermark" < @externalFloor` is now strict. At `<=` the transaction *at* the confirmed backup watermark is purged, and the follower that restored to it gets `too-old` from `plan()` because its `commit` row is gone. The change log is excluded from the litestream backup, so that transaction is the only copy a restoring follower can catch up from.
- The eligible-commits query is bounded by a `LIMIT`. The bench read the whole eligible set every batch and consumed a prefix of it, so a drain re-scanned the remaining history once per batch -- inside the `BEGIN IMMEDIATE`, where the scan holds the write lock alongside the delete.

`maxRows` also stops being a soft target. It let the oldest transaction through unbounded, so a ten-million-row upstream transaction became one ten-million-row DELETE holding the change-log write lock -- which, once purge runs outside the writer's process, outlasts the writer's 30s busy_timeout and stops replication. SQLite cannot yield mid-statement, so an oversized minimum transaction is instead torn across chunks: only `min("watermark")` is ever torn, rows go ascending by `pos` so the boundary check's commit row survives, and a prefix delete removes that row and advances the minimum together. A crash mid-tear needs no recovery state -- the torn transaction is still the minimum and still eligible. Row cost varies with payload size, so the real bound is `maxDurationMs`: the purger measures each delete and adapts its row budget.

Change-log schema goes to v2:

- `auto_vacuum = INCREMENTAL`, as `applyChangeLogPragmas`' *first* statement. SQLite honours a none -> incremental transition only while the file is empty, and `journal_mode = wal2` materializes page 1, so appending the pragma leaves a log that silently never reclaims. `openChangeLogDBForWriting` reads the mode back and rebuilds once through the file-delete path the corruption case already takes, since `DROP TABLE` cannot enable it. A mode still wrong after that rebuild is logged and tolerated rather than thrown: it can only mean the pragma order itself is wrong, and failing there would be a crash loop on every task while rebuilding again would be a reseed per start.
- `_zero.changeLogMeta` gains `seededAtMs`, which slice 11's warm-up precondition needs and which is free to add while the version is moving. `min("writeTimeMs")` cannot answer it: the seed row carries the replica's last upstream commit time, which on an idle shard reads as warm.

Bench: adds an `autoVacuum` axis and points both purge sites at the shipped purger, so what is measured is what ships. The oversized-purge case now asserts that no batch exceeds the row target and that at least one is a partial tear -- the inverse of what the soft limit required.

auto_vacuum write-path cost, wal2 + synchronous=NORMAL, small-high-frequency, 32MB payload, 5 reps:

  | | NONE | INCREMENTAL | delta |
  | changes/sec      | 28,612     | 27,854     | -2.6%   |
  | log commit p50   | 8.79 us    | 8.83 us    | +0.5%   |
  | log commit p95   | 17.2 us    | 19.0 us    | +10.2%  |
  | log commit p99   | 24.8 us    | 27.6 us    | +11.2%  |
  | change-log bytes | 45,518,848 | 45,563,904 | +0.099% |

  SQLITE_CHANGE_LOG_LOG_JOURNAL_MODE=wal2 \
  SQLITE_CHANGE_LOG_LOG_SYNCHRONOUS=NORMAL \
  SQLITE_CHANGE_LOG_TARGET_PAYLOAD_MB=32 SQLITE_CHANGE_LOG_REPS=5 \
  pnpm --filter zero-cache run bench sqlite-change-log-ceiling \
    -t 'log durability trade'

The p99 delta is 2.8us against the 7A bound of 3x the combined layout's commit, so the pragma stays and slice 9c is worth having.

This amends P§9 invariant 5: every transaction *above* `min("watermark")` is wholly retained; the minimum may be partial while it is being purged, and while torn it keeps its commit row so it remains a valid catchup boundary. At most one transaction is torn at any time. New invariant 14 -- no transaction at or above the confirmed backup watermark is ever purged -- is the strict `<` above.

Production effect: none, no caller. One extra reseed for any staging log at
schema v1, on next replicator start. Rollback: revert. Unblocks slice 9a.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.test.ts
+++ b/packages/zero-cache/src/services/change-streamer/sqlite-change-log-reader.test.ts
@@ -43,6 +43,7 @@ const ANCHOR: ReplicaAnchor = {
   replicaVersion: '01',
   stateVersion: '02',
   writeTimeMs: 1234567890,
+  nowMs: 1234567890,
 };
 
 /** The replica path whose `-change-log` sibling the fixtures build. */
--- a/packages/zero-cache/src/services/replicator/change-log-db.test.ts
+++ b/packages/zero-cache/src/services/replicator/change-log-db.test.ts
@@ -1,6 +1,10 @@
 import {existsSync, mkdirSync, rmSync, statSync, writeFileSync} from 'node:fs';
-import {afterEach, describe, expect, test} from 'vitest';
-import {createSilentLogContext} from '../../../../shared/src/logging-test-utils.ts';
+import {LogContext} from '@rocicorp/logger';
+import {afterEach, describe, expect, test, vi} from 'vitest';
+import {
+  createSilentLogContext,
+  TestLogSink,
+} from '../../../../shared/src/logging-test-utils.ts';
 import {Database} from '../../../../zqlite/src/db.ts';
 import {DbFile, expectTableExact} from '../../test/lite.ts';
 import {CREATE_V14_CHANGE_LOG_STREAM} from '../change-source/common/replica-schema.ts';
@@ -26,6 +30,7 @@ const ANCHOR: ReplicaAnchor = {
   replicaVersion: '01',
   stateVersion: '05',
   writeTimeMs: 12345,
+  nowMs: 1_700_000_000_000,
 };
 
 /** The main file and the sidecars `deleteLiteDB` removes. */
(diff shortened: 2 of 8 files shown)


tantaman:
re-homing change log writers to change-streamer.

PR #6291, opened 2026-07-30
perf(replicache): batch btree scan child node reads

santhoshJus:
scanForHash descends an internal node's children serially, awaiting each child before reading the next. Since the kv store coalesces concurrent gets into a single batched read (e.g. one SQL statement), the serial awaits defeat that batching and issue one store read per child.

Prefetch the children about to be visited in parallel before the serial DFS, so their reads coalesce. Traversal order and results are unchanged (the recursion below reads from the now-warm node cache). Prefetch errors are swallowed so a speculative read can't throw for — or ahead of — a node the serial DFS hasn't reached. The serial recursion stays the authoritative reader and still throws, at the same point in the traversal, for any node the scan actually visits.

On a React Native op-sqlite client this cut cold-connect DAG hydration from 558 store reads to 18 (avg batch 2.0 -> 61) and native SQL time from ~4,500ms to ~1,200ms. The win is store-dependent: large on stores where each get is a discrete round-trip, ~neutral where concurrent gets already share a transaction (e.g. IDBStore).


tantaman:
Thanks. We really need to remove the replicache b-tree from Zero given all of Zero's needs can be served by using the fork functionality of Zero's b-tree.

The algo in that world looks like:

```ts
server_diffs = receive(server_update)
new_sync = apply_diffs(sync_head, server_diffs)
diffs = diff(new_sync, main_head)

// we do this re-wind so IVM stays in sync
main_head = apply_diffs(main_head, diffs)

run_rebase(main_head, outstanding_mutations)

// we do this so sync and main always stay close and sharing max structure. This makes diff O(log(height))
sync_head = fork(main_head)
```

For storing the Zero b-tree, we would need to stick ID's on each page of the B-Tree. And update the diff algo to compare node ids rather than object equality.

aboodman:
It still seems worth reviewing the pr itmt given stated perf impact?


a (phone)

tantaman:
> It still seems worth reviewing the pr itmt given stated perf impact?

yeah, just leaving notes.

aboodman:
This is so small it seems fine to land. Somebody on the team needs to validate the perf improvement however.

aboodman:
I confirmed the perf benefit locally and workd through whether there are any downsides (there are not consider zero eagerly loads all data anyway).

I agree w/ the bigger improvement you wrote about @tantaman and I looked into that but I am not sure right off how it will interact with (a) the sync protocol impl -- do we re-implement it at a higher layer inside zero proper? right now it's in replicache (b) cross-tab -- right now this is done in replicache too and i think coupled to replicache's storage layout.

anyway i think you have more of this paged into your head so landing the immediate fix.

packages/replicache/src/btree/read.test.ts
```diff
@@ -0,0 +1,240 @@
...
+  });
+});
+
+test('limit 1 scan reads only the root-to-leaf path', async () => {
+  const dagStore = new TestStore();
+  const value = 'x'.repeat(256);
+  const entries = Array.from({length: 20_000}, (_, i) => [
+    String(i).padStart(8, '0'),
+    value,
+  ]) as [string, string][];
```
tantaman, line 199: 20k row tree.

Before my updates for "opt-in", we would read 1.31MB to fetch a single row and this test would fail.

tantaman:
I changed this to require opting in. The problem is that replicache callers would start scanning MB of data just to get a single row. Zero can opt into this given it eagerly loads everything.

tantaman:
Given that this can devolve into reading several orders of magnitude more than the caller requests, I'm going to wait for @arv to review.

It seems like it should be fine for Zero where we always read everything... but I'm not 100% sure.

packages/replicache/src/btree/read.ts
```diff
@@ -117,7 +127,10 @@ export class BTreeRead implements AsyncIterable<Entry<FrozenJSONValue>> {
   // determining from an entry.key alone whether it is a regular key or an
   // encoded IndexKey in an index map. Without encoding regular map keys the
   // caller has to deal with encoding and decoding the keys for the index map.
-  scan(fromKey: string): AsyncIterableIterator<Entry<FrozenJSONValue>> {
+  scan(
+    fromKey: string,
+    {prefetch = false}: BTreeScanOptions = {},
```
arv, line 132: this is an internal API so no need for a property bag. a required param seems better.

arv (commented):
I like the general approach.

Let me take over this PR.

arv:
Closed in favor of #6414

PR #6292, opened 2026-07-30
perf(zero-server): introspect the server schema from pg_catalog

diegopereira99:
Issues are disabled on this repo, so the problem statement is inline rather than in a linked issue.

## Problem

`getServerSchema` introspects the server schema through `information_schema.columns`. Server processes run it on the first transaction they serve, so on serverless deployments it runs once per worker per route, on every cold start.

`CRUDMutatorFactory` caches the result per factory instance, so a long-lived process pays once. On Vercel Fluid / Lambda style workers the process *is* the cache lifetime, so the query runs again on every push-handler cold start, every custom-query route cold start, and every worker recycle (scale to zero, deploy, memory pressure). The per-call cost gets multiplied by request fan-out and deploy frequency.

Measured on a production healthcare SaaS on Postgres (517 tables), API on serverless functions, connecting as a non-superuser application role:

- **10,500 executions/day** of this one statement against a single database
- **~430 ms** mean execution time
- **37-40% of total database time** — the largest single source of load, ahead of every application query
- a second deployment recorded **106K executions in a 3 hour window** during a period of frequent deploys

`information_schema.columns` is a view over `pg_attribute` that derives `data_type`, `character_maximum_length`, `numeric_precision`, `numeric_scale` and `udt_name` through the `information_schema._pg_*` SQL functions and applies `has_column_privilege` per column. `zero-cache` already moved publication introspection off it in #1484 for the same reason ("the raw `pg_catalog` tables […] the former is much faster"); this query is on the request path.

Separately, `JOIN pg_catalog.pg_type t ON c.udt_name = t.typname` joins by type *name*, so a column whose type name is not unique in the database yields one row per matching schema and `getServerSchema` keeps whichever row arrives last.

## What changed

- The query now reads `pg_catalog.pg_attribute` / `pg_class` / `pg_namespace` / `pg_type` and joins `pg_type` by OID. It reproduces the view's definition for the columns Zero selects: the `_pg_char_max_length` / `_pg_numeric_precision` / `_pg_numeric_scale` arithmetic, the single level domain unwrap that `data_type` and `udt_name` perform, and the view's own `relkind`, `attnum`, `attisdropped`, temp-schema and privilege filters.
- The SQL moved into an exported `serverSchemaQuery()` so it can be tested directly, the way `publishedSchemaQuery` is in `zero-cache`. `getServerSchema`'s behavior is unchanged.
- New `packages/zero-server/src/schema.pg.test.ts`.
- Three mocks in `custom.test.ts` recognized the introspection query with `sql.includes('information_schema')`; they now match on `pg_attribute`.

Results are unchanged, down to the wire types: `length`, `precision` and `scale` stay `int4` (`information_schema` returns them as the `cardinal_number` domain over `int4`, which drivers surface as numbers), everything else stays `::text`.

## Equivalence

`schema.pg.test.ts` runs both queries — the new one and a verbatim copy of the one it replaces — against the same database and asserts strict deep equality of all 11 columns over 49 columns of a catalog built to hit every branch: domains over `varchar(n)`, `numeric(p,s)`, an enum, an array and `bigint`; enum arrays, domain arrays, `numeric(p,s)` arrays; composite types; an extension type (`citext`); `bit(n)` / `bit varying(n)` (whose typmod is *not* offset by 4, unlike `char` / `varchar`); `real` / `double precision` (which report binary precision 24 / 53); `smallint` / `integer` / `bigint` (which report scale 0); unparameterized `numeric` (null precision and scale); a view; a materialized view; a generated column; a dropped column; and a table outside `public`. A second test pins the values the compatibility check actually reads, per column, and a third covers the ambiguous type name case.

Before this PR we validated the same query against real data: field-by-field over 1,306 columns / 75 tables of the production database above, then over its whole catalog (8,762 columns) with the table filter removed — zero differences either time. Repeated for this PR on a synthetic 500 table catalog (8,073 columns, 900 in scope): `EXCEPT ALL` empty in both directions.

## Performance

Stock `postgres:17` container, 2,000 tables / 26,073 columns, 75 tables requested, as a role that owns nothing, 6 runs each:

| query | execution time |
| --- | --- |
| `information_schema.columns` | 10.3 – 13.9 ms |
| `pg_catalog` | 1.9 – 2.1 ms |

On the production database the same comparison was ~430 ms → ~11 ms at its 75 table scope, and 1,871 ms → 109 ms introspecting the whole 517 table catalog.

## Testing

`pnpm --filter zero-server run test` — 425 tests pass across the `pg-15`, `pg-16`, `pg-17`, `pg-18` and `no-pg` projects (Docker testcontainers, local macOS). `check-types`, `lint`, `check-format` and `verify-deps` are clean. I have not run the full repo suite or the browser/perf projects.

## Notes

- This query is running in our production deployment, applied to the published `@rocicorp/zero` build with `patch-package` plus a build-time guard that fails if the patch does not apply, so the SQL here is what is in production rather than a fresh translation.
- Materialized views stay excluded, because `information_schema.columns` excludes `relkind = 'm'` and the replacement keeps that filter. Same for other sessions' temp tables.
- Longer term, caching the schema beyond the process lifetime would attack the same problem from the other side and compose with this. That needs an invalidation story; making the query cheap does not.

### DIFF
--- a/packages/zero-server/src/custom.test.ts
+++ b/packages/zero-server/src/custom.test.ts
@@ -385,7 +385,7 @@ describe('CRUDMutatorFactory', () => {
     const factory = new CRUDMutatorFactory(simpleSchema);
     const {mockTx} = createMockTx();
 
-    // Mock the information_schema query that getServerSchema uses
+    // Mock the introspection query that getServerSchema uses
     const mockSchemaRows = [
       {
         schema: 'public',
@@ -431,7 +431,7 @@ describe('CRUDMutatorFactory', () => {
     const originalQuery = mockTx.query;
     mockTx.query = vi.fn((...args: unknown[]) => {
       const sql = args[0] as string;
-      if (sql.includes('information_schema')) {
+      if (sql.includes('pg_attribute')) {
         return Promise.resolve(mockSchemaRows);
       }
       return originalQuery.apply(
@@ -450,7 +450,7 @@ describe('CRUDMutatorFactory', () => {
     // Verify mutate works
     void tx.mutate.basic.insert({id: '1', a: 1, b: 'test'});
 
-    // The information_schema query + the insert
+    // The introspection query + the insert
     expect(mockTx.query).toHaveBeenCalled();
   });
 
@@ -520,7 +520,7 @@ describe('CRUDMutatorFactory', () => {
         wrappedTransaction: null,
         query: (...args: unknown[]) => {
           const sql = args[0] as string;
-          if (sql.includes('information_schema')) {
+          if (sql.includes('pg_attribute')) {
             serverSchemaFetchCount++;
             return Promise.resolve(mockSchemaRows);
           }
@@ -614,7 +614,7 @@ describe('CRUDMutatorFactory', () => {
         wrappedTransaction: null,
         query: (...args: unknown[]) => {
           const sql = args[0] as string;
-          if (sql.includes('information_schema')) {
+          if (sql.includes('pg_attribute')) {
             serverSchemaFetchCount++;
             return Promise.resolve(mockSchemaRows);
           }
--- a/packages/zero-server/src/schema.pg.test.ts
+++ b/packages/zero-server/src/schema.pg.test.ts
@@ -1,0 +1,249 @@
+import {beforeEach, describe, expect, test} from 'vitest';
+import {testDBs} from '../../zero-cache/src/test/db.ts';
+import type {PostgresDB} from '../../zero-cache/src/types/pg.ts';
+import {serverSchemaQuery} from './schema.ts';
+
+type Row = Record<string, unknown>;
+
+// A catalog that exercises every branch of the expressions in
+// serverSchemaQuery: domains (including over enums, arrays and parameterized
+// types), enum arrays, composite types, extension types, the bit / varbit and
+// float / integer precision special cases, views, dropped and generated
+// columns, and a table outside the public schema.
+const catalogSql = /*sql*/ `
+CREATE EXTENSION IF NOT EXISTS citext;
+
+CREATE SCHEMA alternate_schema;
+
+CREATE TYPE mood AS ENUM ('happy', 'sad');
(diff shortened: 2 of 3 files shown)


aboodman:
Thank you for this PR. The hard-coded oid types make me nervous. I remember we were already grumpy about the structure of having to do this schema read so frequently in the first place. 

Rocicorp: when somebody reviews this, we should think about restructuing at a higher level to not need the repeated reads in the first place.

diegopereira99:
Fair concern — pushed a commit replacing the numeric OIDs with `regtype` casts (`'pg_catalog.float8'::regtype` etc.). The parser resolves them at plan time so there's no runtime cost, and the query now reads as type names. They're schema-qualified because regtype input goes through `search_path`: with `search_path = myschema, pg_catalog` and a type named `int4` in `myschema`, an unqualified `'int4'::regtype` resolves to the wrong type. For context, builtin type OIDs are pinned at initdb and haven't been renumbered across PG versions, but with regtype nothing depends on knowing that.

On the higher-level restructuring: agreed, and our production numbers are the argument for it. On serverless (Vercel), every worker cold start pays this read — we measured 10.5K executions/day (~40% of our total database time) before switching to the pg_catalog query. A cached or precomputed server-schema would eliminate even the cheap version. Happy to help build that if you sketch the direction you'd want.

grgbkr:
This is a great change, thank you for the contribution and thorough tests.

I don't currently favor the restructuring options,
1. caching needs an invalidation solution
2. [Zero schema ](https://zero.rocicorp.dev/docs/schema#multiple-schemas) could be changed to include the detailed Postgres types (something like `amount: { type: "number", pgType: "numeric", precision: 8, scale: 3 }`), but this adds more boilerplate, an opportunity for bugs due to drift/inconsistencies, and couples Zero's schema api to Postgres (Zero tries to not couple itself too closely to Postgres so as to allow for future extension to additional databases).  

This solution is great, as its just a drop in.

PR #6293, opened 2026-07-30
chore(rmv2): put the sqlite change-log in the change-streamer

tantaman:
Puts the writing of the sqlite change log in the change streamer rather than in the replicator worker.

Production effect: none; `sqliteChangeLogMode=off`. Rollback: revert.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
+++ b/packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
@@ -4,7 +4,10 @@ import {resolver, type Resolver} from '@rocicorp/resolver';
 import {assert, unreachable} from '../../../../shared/src/asserts.ts';
 import {promiseVoid} from '../../../../shared/src/resolved-promises.ts';
 import {publishCriticalEvent} from '../../observability/events.ts';
-import {getOrCreateCounter} from '../../observability/metrics.ts';
+import {
+  getOrCreateCounter,
+  getOrCreateLatencyHistogram,
+} from '../../observability/metrics.ts';
 import {
   min,
   type AtLeastOne,
@@ -53,6 +56,10 @@ import {
 } from './sqlite-change-log-catchup.ts';
 import {SQLiteChangeLogReader} from './sqlite-change-log-reader.ts';
 import {
+  SQLiteChangeLogWriter,
+  type SQLiteChangeLogWriterOptions,
+} from './sqlite-change-log-writer.ts';
+import {
   Storer,
   type PurgeLock,
   type TuningOptions as StorerOptions,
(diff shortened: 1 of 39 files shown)


packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
```diff
@@ -418,6 +454,12 @@ class ChangeStreamerImpl implements ChangeStreamerService {
         // from the durable PG head. Commits observed only since process startup
         // are insufficient after a change-streamer restart.
         this.#lastForwardedCommitWatermark = lastWatermark;
+        // Per stream connection, not once per process: the resume watermark is
+        // re-read on every reconnect, so the log can be holding transactions
+        // above it -- a restarted backfill's abandoned ones, in bulk -- and the
+        // writer's plain INSERT would collide with the first re-delivery.
+        // Ordered before startStream so no change can arrive mid-reconcile.
+        this.#changeLogWriter?.reconcile(lastWatermark);
```
darkgnotic, line 462: Catching up on all of the awesome work!

I think there may be a mismatch in the requirements of the changeLog here. Let me try and more completely describe the job of the change log so that you can determine if there's a gap that needs to be closed.

The change-log entity (whether it be globally persisted in PG, or ephemerally recreated from replica backups) needs to both:
* maintain a buffer of changes for catching up subscribers to the head of the replication stream
* track the progress of what needs to be backfilled, via the "tableMetadata" and "backfilling" tables: https://github.com/rocicorp/mono/blob/ac745bd9759629b2ad9fc2ad6a2bfe1a1c14309d/packages/zero-cache/src/services/change-streamer/schema/tables.ts#L104

The latter are sort of like "cookies" that the change-streamer tracks, and always passes to the change-source when starting a change stream, so that the change source can initiate the appropriate backfills. (This obviates the need for the change-source to persist any backfill state upstream).

The PG change-log (storer.ts) maintains those backfill tables in [#trackBackfillMetadata()](https://github.com/rocicorp/mono/blob/ac745bd9759629b2ad9fc2ad6a2bfe1a1c14309d/packages/zero-cache/src/services/change-streamer/storer.ts#L796)

To eventually remove the dependence on the PG change-log, the replicator (change-processor) also maintains the metadata in [processTableMetadata()](https://github.com/rocicorp/mono/blob/ac745bd9759629b2ad9fc2ad6a2bfe1a1c14309d/packages/zero-cache/src/services/replicator/change-processor.ts#L593).

The idea and eventual goal is for the sqlite change-log to be completely initializable from the restored replica; the change-log would start at the replica's last watermark, and the backfill/metadata tables would be copied over and tracked in the change log's own tables for the purpose of initializing (and re-initializing) the change-source stream.

So the idea of "reconciling" the sqlite change-log with the PG change-log is somewhat contrary to that goal; this is fine as an incremental step, but eventually the SQLite change-log needs to be derived entirely from the restored replica.

This also leads to an interesting intermediate state (e.g. RMv1.5) when we're maintaining both change-logs: it is unlikely that the two change-logs will be initialized at the same watermark; the replica-derived change-log will likely be behind since the PG-change-log generally advances while the replica is being restored from litestream.

The strawman I had in mind was that [getStartStreamInitializationParameters()](https://github.com/rocicorp/mono/blob/ac745bd9759629b2ad9fc2ad6a2bfe1a1c14309d/packages/zero-cache/src/services/change-streamer/change-streamer-service.ts#L526) would query both the PG change-log and the SQLite change-log and start from whichever is earlier. The corollary to this behavior is that both change-logs would need to know to ignore changes that it has already seen (which should be a fairly simple change).

Sorry for not having a more complete spec from the start ... but I think the progress you've made so far is amazing and I don't think requires a huge directional change. Let me know if you want to sync up in realtime too ... there may very well be better strategies to this fairly tricky migration.

PR #6295, opened 2026-07-30
chore(rmv2): schedule the change-log purge in-process 

tantaman:
Add the purger which deletes history when all of these are true:

  - The purge candidate is before the oldest backup watermark still needed.
  - Every connected subscriber has acknowledged at least that watermark.
  - It is older than the configured retention window—60 seconds by default.
  - It is not the newest transaction.
  - No backup/snapshot restore currently reserves that history.
  - The SQLite writer is between transactions.

  Cleanup happens in small batches so it doesn’t monopolize the change-streamer. It yields
  between batches and automatically checks again as retained rows grow old enough.

  Operationally:

  - Default mode is still off, so the branch changes nothing by default.
  - In write mode and above, SQLite logging is now disk-bounded (because we purge)
  - This branch still doesn’t serve normal catch-up traffic from SQLite; it prepares the infrastructure for slice 10’s comparison rollout.
  - PostgreSQL cleanup behavior is unchanged.
  - Rollback is simply setting sqliteChangeLogMode=off.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
+++ b/packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
@@ -412,81 +412,89 @@ export class SQLiteChangeLogPurgeScheduler {
     let stopped: PurgeStopReason | undefined;
     let db: Database | undefined;
 
-    for (;;) {
-      // The floor is re-read each batch: a coalesced dispatch may have raised
-      // it, and the gate below is evaluated against what is actually used.
-      floor = this.#pendingFloor ?? floor;
-      if (this.#stopped) {
-        stopped = 'stopped';
-        break;
-      }
-      if (this.#pausedBy.size > 0) {
-        stopped = 'paused';
-        break;
-      }
-      const connection = this.#connection();
-      if (connection === undefined) {
-        db = undefined;
-        this.#lc.debug?.(
-          'skipping SQLite change-log purge: the writer has not opened the log',
-        );
-        stopped = 'writer-unavailable';
-        break;
-      }
-      db = connection;
-      if (
-        batches >=
-        Math.max(
-          1,
-          this.#opts.maxBatchesPerCycle ?? DEFAULT_MAX_BATCHES_PER_CYCLE,
-        )
-      ) {
-        stopped = 'batch-limit';
-        break;
-      }
-      if (connection.inTransaction) {
-        // Purge windows exist only between the writer's commits (§3.3).
-        await this.#awaitWriterIdle();
-        continue;
-      }
+    try {
+      for (;;) {
+        // The floor is re-read each batch: a coalesced dispatch may have raised
+        // it, and the gate below is evaluated against what is actually used.
+        floor = this.#pendingFloor ?? floor;
+        if (this.#stopped) {
+          stopped = 'stopped';
+          break;
+        }
+        if (this.#pausedBy.size > 0) {
+          stopped = 'paused';
+          break;
+        }
+        const connection = this.#connection();
+        if (connection === undefined) {
+          db = undefined;
+          this.#lc.debug?.(
+            'skipping SQLite change-log purge: the writer has not opened the log',
+          );
+          stopped = 'writer-unavailable';
+          break;
+        }
+        db = connection;
+        if (
+          batches >=
+          Math.max(
+            1,
+            this.#opts.maxBatchesPerCycle ?? DEFAULT_MAX_BATCHES_PER_CYCLE,
+          )
+        ) {
+          stopped = 'batch-limit';
+          break;
+        }
+        if (connection.inTransaction) {
+          // Purge windows exist only between the writer's commits (§3.3).
+          await this.#awaitWriterIdle();
+          continue;
+        }
 
-      const batchFloor = floor;
-      const outcome = await this.#lock.withLock(() => {
-        // Everything the wait above established is re-checked under the lock:
-        // the writer may have begun another transaction or failed soft while
-        // this batch's acquisition was queued behind a registration.
-        if (this.#connection() !== connection || connection.inTransaction) {
-          return {kind: 'retry'} as const;
-        }
-        const declined = this.#gate(batchFloor);
-        if (declined !== undefined) {
-          return {kind: 'declined', reason: declined} as const;
-        }
-        const result = this.#cacheFor(connection).purger.purgeBatch({
-          externalFloor: batchFloor,
-          retentionCutoffMs: this.#now() - this.#opts.retentionMs,
-          maxRows: this.#opts.batchRows,
-          maxDurationMs:
-            this.#opts.statementTargetMs ?? DEFAULT_STATEMENT_TARGET_MS,
+        const batchFloor = floor;
+        const outcome = await this.#lock.withLock(() => {
+          // Everything the wait above established is re-checked under the lock:
+          // the writer may have begun another transaction or failed soft while
+          // this batch's acquisition was queued behind a registration.
+          if (this.#connection() !== connection || connection.inTransaction) {
+            return {kind: 'retry'} as const;
+          }
+          const declined = this.#gate(batchFloor);
+          if (declined !== undefined) {
+            return {kind: 'declined', reason: declined} as const;
+          }
+          const result = this.#cacheFor(connection).purger.purgeBatch({
+            externalFloor: batchFloor,
+            retentionCutoffMs: this.#now() - this.#opts.retentionMs,
+            maxRows: this.#opts.batchRows,
+            maxDurationMs:
+              this.#opts.statementTargetMs ?? DEFAULT_STATEMENT_TARGET_MS,
+          });
+          return {kind: 'purged', result} as const;
         });
-        return {kind: 'purged', result} as const;
-      });
 
-      if (outcome.kind === 'retry') {
-        continue;
+        if (outcome.kind === 'retry') {
+          continue;
+        }
+        if (outcome.kind === 'declined') {
+          stopped = outcome.reason;
+          break;
+        }
+        batches++;
+        deletedRows += outcome.result.deletedRows;
+        if (!outcome.result.moreEligible) {
+          break;
+        }
+        // Release the loop — and the guard, released above — so fan-out,
+        // catchup, and registrations make progress between batches.
+        await this.#yield();
       }
-      if (outcome.kind === 'declined') {
-        stopped = outcome.reason;
-        break;
-      }
-      batches++;
-      deletedRows += outcome.result.deletedRows;
-      if (!outcome.result.moreEligible) {
-        break;
-      }
-      // Release the loop — and the guard, released above — so fan-out,
-      // catchup, and registrations make progress between batches.
-      await this.#yield();
+    } catch (e) {
+      // Earlier batches committed independently and must remain visible in the
+      // result and metrics. Finalize the partial cycle, including its floor
+      // probe, rather than replacing it with an empty skipped result.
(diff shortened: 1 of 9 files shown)


packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
```diff
@@ -0,0 +1,690 @@
...
+ * file. This is what lets a quiet source still drain eligible history: the
+ * retention cutoff advances with the clock even when no new cleanup watermark
+ * arrives to dispatch a cycle.
+ */
+const DEFAULT_RECHECK_INTERVAL_MS = 30_000;
+
+/**
+ * The backstop poll while waiting for the writer's open transaction to
+ * commit. The commit notification normally arrives first; this covers an
+ * interrupted stream, whose rollback does not notify.
```
tantaman, line 38: why?

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
```diff
@@ -0,0 +1,690 @@
...
+ */
+const DEFAULT_MAX_BATCHES_PER_CYCLE = 100;
+
+/**
+ * How long to wait before re-checking a log that is drained down to history
+ * that is not old enough to purge yet, or whose writer has not opened the
+ * file. This is what lets a quiet source still drain eligible history: the
+ * retention cutoff advances with the clock even when no new cleanup watermark
+ * arrives to dispatch a cycle.
+ */
```
tantaman: if the source is not quiet, what is prompting the drains?

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
```diff
@@ -0,0 +1,690 @@
...
+ * connection, in the stream loop's process, so a statement's duration stalls
+ * the appending path — and the fan-out behind it — directly. The bench's
+ * 100 ms bound was derived for the replicator's write worker; this budget is
+ * measured against an event loop that also owes progress to fan-out, catchup,
+ * and websocket keepalives, so it is deliberately tighter.
+ */
+const DEFAULT_STATEMENT_TARGET_MS = 20;
+
+/**
+ * Batches per cycle. This bounds one cycle's bookkeeping, not the drain:
```
tantaman: metadata bookkeeping about floors? which bookkeeping?

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
```diff
@@ -0,0 +1,690 @@
...
+import {SQLITE_CHANGE_LOG_PLAN_SQL} from './sqlite-change-log-reader.ts';
+
+/**
+ * The duration target for any one purge statement, enforced by the purger's
+ * measured-budget feedback. Purge runs on the writer's own
+ * connection, in the stream loop's process, so a statement's duration stalls
+ * the appending path — and the fan-out behind it — directly. The bench's
+ * 100 ms bound was derived for the replicator's write worker; this budget is
+ * measured against an event loop that also owes progress to fan-out, catchup,
+ * and websocket keepalives, so it is deliberately tighter.
```
tantaman, line 25: we need to benchmark / measure to see how many rows we can purge in 20ms.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
```diff
@@ -0,0 +1,690 @@
...
+ * commit. The commit notification normally arrives first; this covers an
+ * interrupted stream, whose rollback does not notify.
+ */
+const DEFAULT_IN_TRANSACTION_POLL_MS = 1_000;
+
+/** Why a cycle stopped before draining the eligible set. */
+export type PurgeStopReason =
+  // The scheduler was stopped.
+  | 'stopped'
+  // A snapshot reservation is open.
```
tantaman, line 46: due to view syncer restart? Or view syncer connected and started catching up?

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
```diff
@@ -0,0 +1,690 @@
...
+  | 'paused'
+  // The writer has not opened the change log (the file does not exist yet),
+  // or has failed soft and deleted it.
+  | 'writer-unavailable'
```
tantaman, line 50: how can this happen? Is not the file created at startup?

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
```diff
@@ -0,0 +1,690 @@
...
+  // No subscriber is connected to confirm the floor. Matches the PG arm's
+  // "No subscribers to confirm cleanup" bail.
+  | 'no-subscribers'
```
tantaman: right, but we should still eventually purge rather than letting a floor be pinned indefinitely.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
```diff
@@ -0,0 +1,690 @@
...
+
+/**
+ * The backstop poll while waiting for the writer's open transaction to
+ * commit. The commit notification normally arrives first; this covers an
+ * interrupted stream, whose rollback does not notify.
+ */
+const DEFAULT_IN_TRANSACTION_POLL_MS = 1_000;
+
+/** Why a cycle stopped before draining the eligible set. */
+export type PurgeStopReason =
```
tantaman, line 43: state machine diagram for all this would be nice

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
tantaman: The file is created lazily, not in the service constructor. `SQLiteChangeLogWriter.reconcile(lastWatermark)` opens or creates it on each stream connection immediately before `startStream`, so a backup-watermark-driven cleanup can be scheduled before the first reconcile. The other case is fail-soft: any SQLite reconcile/write error closes and deletes the file and leaves the writer disabled for the rest of the process. In `off` mode there is no scheduler at all.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
tantaman: Agreed. The existing ceiling benchmark can vary `SQLITE_CHANGE_LOG_PURGE_BATCH_ROWS` and `SQLITE_CHANGE_LOG_PURGE_MAX_DURATION_MS`, but it defaults to 100 ms and this PR records no representative 20 ms result. The purger adapts future chunk sizes only after measuring a statement, so 1,000 rows is the first-statement starting size and can overrun. We should benchmark production-shaped row-size distributions and durability modes at 20 ms, then choose the starting row count from those results; per-statement latency and deleted-row telemetry would also let us validate it in production.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
tantaman: `paused` means a `/snapshot` reservation was opened by a view-syncer before it restores and prepares a replica. That normally happens during a new view-syncer start/restart, or a retry that needs a restore; it is not caused by an already-running view-syncer doing ordinary catchup. The pause ends when the same task starts its `/changes` subscription, at which point it has prepared the replica and is beginning catchup, or when the reservation is cancelled/replaced.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
tantaman: Yes. The current implementation does not provide eventual cleanup here: the timer repeats the same `no-subscribers` decline forever. Also, the outer `#purgeOldChanges()` returns before invoking the SQLite scheduler when both subscriber ACKs and confirmed reservation watermarks are absent, so changing only `#gate()` would not fix it.

This needs a service-level policy change: after the reconnect grace period, use the verified backup watermark as the floor when there are no subscribers and no open snapshot reservation, while retention still supplies the reconnect buffer. The PG and SQLite arms should apply that policy consistently. I would leave this thread open pending that change.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
tantaman: Agreed. The current state machine is:

```mermaid
stateDiagram-v2
  [*] --> Idle
  Idle --> Running: purge(floor)
  Running --> Running: higher floor coalesced
  Running --> Running: batch + yield
  Running --> WaitingForWriter: writer transaction open
  WaitingForWriter --> Running: commit, rollback, or 1 s poll
  Running --> Paused: snapshot reservation
  Paused --> Running: last reservation resumes
  Running --> RetryLater: lagging/no subscriber, no writer, error, or young rows remain
  RetryLater --> Running: 30 s timer
  Running --> Running: 100-batch limit; immediate continuation
  Running --> Idle: drained; nothing remains below floor
  Idle --> Stopped: stop
  Running --> Stopped: stop
  WaitingForWriter --> Stopped: stop
  Paused --> Stopped: stop
  RetryLater --> Stopped: stop
```

A pause waits for any batch already holding the mutex, so its bound is one statement. `pendingFloor` persists while idle and can only rise. This diagram should live with the class documentation; the no-subscriber transition also exposes the indefinite-pin gap noted in the other thread.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
tantaman: By bookkeeping I mean the lifetime and accounting of one `#running` promise: accumulated batch/deleted-row counts and duration, one cycle log/metric record, and one post-cycle floor probe. It is not a safety or database-work bound. After 100 batches, `batch-limit` finalizes that cycle and schedules an immediate continuation. The comment should say cycle-level observability/accounting.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
tantaman: Fresh verified backup watermarks are the primary driver: `trackBackupWatermark()` schedules `#purgeOldChanges()`, which dispatches this scheduler with the newly safe floor. Once running, `moreEligible` plus the yield drains further batches, and the 100-batch limit schedules an immediate continuation.

The 30 s recheck is for cases where the floor does not change: rows below the existing floor age past retention, a lagging ACK advances without a notification, or the writer/file becomes available again. Source commits only call `onWriterIdle()` to wake a cycle already parked on an open transaction; they do not initiate a purge.

packages/zero-cache/src/services/change-streamer/sqlite-change-log-purge-scheduler.ts
tantaman: The wording here is wrong. An interrupted-stream abort does call `onWriterIdle()` in `ChangeStreamerImpl`; the silent case is a normal `rollback` message, because `SQLiteChangeLogWriter` only invokes `onCommit`. The 1 s poll is only a liveness backstop in case a transaction-ending path misses the notification; it is not needed for safety. The comment and test should use the normal-rollback or missed-notification case instead.

PR #6296, opened 2026-07-30
chore(rmv2): compare SQLite catchup output against PG

tantaman:
Adds sqliteChangeLogMode=compare: a SQLiteChangeLogComparator in the change-streamer samples committed transactions (stable by shard and watermark), runs both catchup read paths over them, and compares per-transaction digests via ChangeLogTransactionHasher. Ranges are compared only through min(pgHead, sqliteHead); head skew is never reported. Suspect observations re-read the bounds and report `inconclusive` when a purge, truncation, or reseed raced the pin. Change text is normalized to one form before hashing, pinned by a round-trip test through both stores. Diagnostics carry digest prefixes and row counts only, never payloads.

Presence (missing-pg / missing-sqlite) is checked for every transaction in the range; digest comparison is sampled per
sqliteChangeLogComparePercent (new hidden option, default 1).

Production default unchanged (mode=off); compare activates only at mode=compare or serve, is dark, and changes nothing subscribers are served. Rollback: return the mode to `write`.

Note: the comparator's PG read deliberately shares the subscriber catchup statement shape, including its limitation that change->'tag' de-escapes every string value and therefore fails on payloads containing an escaped NUL character (a pre-existing catchup issue introduced with the stringify-before-fanout rewrite in #5900); such rows report as `reader-error`, matching what a subscriber would experience.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/sqlite-change-log-comparator.ts
+++ b/packages/zero-cache/src/services/change-streamer/sqlite-change-log-comparator.ts
@@ -489,6 +489,7 @@ export class SQLiteChangeLogComparator {
         return this.#skipped('nothing-to-compare');
       }
 
+      const divergent: TransactionCompareResult[] = [];
       reader = new SQLiteChangeLogReader(this.#lc, this.#changeLogFile);
       const plan = reader.plan(from);
       if (plan.kind === 'not-ready') {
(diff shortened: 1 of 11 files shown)


packages/zero-cache/src/services/change-streamer/sqlite-change-log-comparator.ts
```diff
@@ -0,0 +1,984 @@
...
+ * stream is compared in bounded increments rather than all at once.
+ */
+const DEFAULT_COMPARE_INTERVAL_MS = 30_000;
+
+/**
+ * The per-cycle bound on enumerated transactions. Bounds one cycle's work and
+ * memory (two watermark lists plus one transaction's digest state); it does not
+ * bound coverage, since the next cycle resumes from where this one stopped.
+ */
+const DEFAULT_MAX_TRANSACTIONS_PER_CYCLE = 256;
```
tantaman, line 37: but we still bound on rows, right? Like a million row tx won't be completely scanned?

PR #6299, opened 2026-07-31
fix(zero-client): Connection cleanup

aboodman:
This started with [#6232](https://github.com/rocicorp/mono/pull/6232). A user was seeing `Connection attempt timed out after 10 seconds` as an unhandled promise rejection. That PR proposed adding a catch to the rejected promise to reduce the exception noise.

Looking at it more closely, the timeout was not actually timing out the connection.

`#connect()` did some asynchronous setup before it started waiting on `#connectResolver`. The timer started during that setup. If setup took longer than 10 seconds:

1. The timer rejected `#connectResolver` before anything was waiting on it, causing the unhandled rejection.
2. `#disconnect()` created a new resolver, but the run loop was still blocked waiting for the original `#connect()`.
3. If setup eventually finished, the original "timed out" connection continued and could still connect normally.
4. If setup never finished, the run loop never retried and the client was stuck until refresh.

So adding a catch would have hidden the reported error, but it would not have fixed the connection behavior.

This PR makes setup, socket creation, and waiting for the server acknowledgement one bounded connection attempt. The run loop owns the 10-second timeout and races the whole connection against cancellation. On timeout we disconnect, abandon that connection, and retry after the normal backoff. If abandoned setup later creates a socket, we close it instead of installing it.

The timeout is still reported through the normal connection error path. We are fixing the unhandled rejection without hiding the timeout or other legitimate connection failures.

We still do not know why connection setup sometimes takes longer than 10 seconds. Timeout errors now include a timestamped list of connection events so the next report should tell us where the connection was stuck.

Tests: `pnpm --filter zero-client run test` (644 tests)

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -147,6 +147,7 @@ import {
   isAuthError,
   isClientError,
   isServerError,
+  isZeroError,
 } from './error.ts';
 import {
   type HTTPString,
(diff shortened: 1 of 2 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -2055,8 +2082,73 @@ export class Zero<
...
+                        kind: ClientErrorKind.Internal,
+                        message: getErrorMessage(ex),
+                      },
+                      {cause: ex},
+                    );
+                if (!attempt.controller.signal.aborted) {
+                  this.#disconnect(lc, error);
+                }
+                ready.reject(error);
+                if (this.#connectionReadyResolver === ready) {
```
grgbkr, line 2135: I dont see how this.#connectionReadyResolver could have been reassigned. It is only assigned in #runLoop, so nothing could have concurrently reassigned it.

packages/zero-client/src/client/zero.ts
```diff
@@ -2082,38 +2175,47 @@ export class Zero<
...
-                const hiddenError = new ClientError({
-                  kind: ClientErrorKind.Hidden,
-                  message: 'Connection closed because tab was hidden',
-                });
-                this.#disconnect(lc, hiddenError);
-                break;
+            } finally {
+              if (
+                !this.#connectionManager.is(ConnectionStatus.Connected) &&
+                this.#connectionReadyResolver === ready
```
grgbkr, line 2215: Same here, I dont see how this.#connectionReadyResolver could have been reassigned. It is only assigned in #runLoop, so nothing could have concurrently reassigned it.

grgbkr: I think we should document that `this.#connectionReadyResolver` should only be reassigned by the `#runLoop` and replace these conditions with `assert(this.#connectionReadyResolver === ready)`.

When I see an `if (x === y)` statement, it strongly implies that there is a legitimate, expected code path where `x !== y`, and it took me a while to convince myself that that is not the case.

grgbkr (commented):
LGTM, I think this is a good change.

Changes to this code are mighty hard to review.

PR #6305, opened 2026-08-03
perf(zero-react): memoize the useQuery render path

charpeni:
### Problem

Every render of every `useQuery`/`useSuspenseQuery` — including re-renders for unrelated reasons — redoes O(query-size) work:

1. `addContextToQuery` re-runs argument validation and the full query builder chain (a query like zbugs' `issueDetail` allocates ~20 intermediate `QueryImpl`s).
2. `ViewStore.getView` recomputes the view key: `qi.hash()` (normalize AST → `JSON.stringify` → xxHash-based `h64`) plus `JSON.stringify(format)`.

All the caches on that path (`QueryImpl#hash`, AST normalization, hash cache) are keyed on **object identity**, and the query is rebuilt inline on every render (`useQuery(queries.issue({id}))`, `useQuery(z.query.issue.where(...))`) — so every cache misses on every render. Micro-benchmarks: ~21μs/render for a mid-size AST, ~117μs/render for an `issueDetail`-shaped request with validation, vs ~0.04μs when identity is stable. Multiply by hooks-per-screen × render frequency (typing, pokes re-rendering parents) and this is real main-thread time. The Solid binding already memoizes this exact computation with `createMemo`; React had no equivalent.

### Changes

- **`useStableQuery`** (new, internal): a per-hook cache that keeps the resolved query identity-stable across renders.
  - A **query request** reuses the previous resolved query when the query function, Zero instance, context, and deep-equal args match — skipping validation and the builder chain entirely.
  - A **plain builder query** is reused when its AST + format + customQueryID are deep-equal — a ~4μs compare instead of ~17μs of normalize/stringify/hash.
  - Falsy renders keep the cache, so a toggled-off query is reused when it comes back. Changing the Zero instance invalidates.
  - The ref is written during render; this is safe because it's a pure memoization cache — entries are derived only from the inputs, so StrictMode double-renders and discarded concurrent renders always leave it valid.
- **View-store key cache**: `getView` caches the `hash + format` string per query object in a `WeakMap` (clientID appended per call, since one query object can serve multiple Zero instances). With a stable query identity the per-render key cost drops to a `WeakMap.get` + concat.
- **`ViewWrapper.updateTTL` guard**: early-return when the TTL value is unchanged, avoiding a per-render re-hash in the query manager. Value changes (including representation changes like `'60s'` → `60000`) still propagate as before.

### Deliberately unchanged

- The view store stays keyed by **AST hash**, not `name:args` — divergent ASTs for the same query name must not collapse (see `query-internals.ts`).
- `enabled: false` still builds the query (once, memoized) rather than never: the disabled snapshot's shape (`undefined` vs `[]`) depends on `format.singular`.
- View lifecycle (subscribe refcounts, 10ms deferred destroy) is untouched; `getView` still runs every render.

### A subtlety worth knowing about

Equality checks run `deepEqual` in **both directions**. `deepEqual` only value-compares keys of its first argument and then compares own-key counts, so an explicitly-undefined key on one side can absorb a differently-named defined key on the other: `deepEqual({q: undefined, page: 1}, {page: 1, sort: 'x'})` is `true`. One-directional comparison would serve a stale query when a call site's args change shape like that (e.g. `{q: searchText || undefined, page}`). The symmetric compare closes the hole; a regression test covers the exact scenario.

### Testing

- 7 new tests: builder fn runs once across equivalent re-renders (and re-runs on args change); a distinct-but-deep-equal query object is never re-hashed and reuses the existing view; changing the Zero instance re-resolves; falsy toggle reuses the cache; the undefined-key masking scenario re-runs the builder; `enabled: false` returns `[]`/`undefined` per plurality without materializing; unchanged TTL doesn't call `updateTTL` on the underlying view. Each fails against the pre-change code.
- Full `zero-react` suite: 81/81. `lint`, `format`, `check-types` clean (including the `zero` umbrella package).

### DIFF
--- a/packages/zero-react/src/use-query.tsx
+++ b/packages/zero-react/src/use-query.tsx
@@ -1,6 +1,10 @@
 import {resolver} from '@rocicorp/resolver';
-import React, {useSyncExternalStore} from 'react';
+import React, {useRef, useSyncExternalStore} from 'react';
+import {deepEqual} from '../../shared/src/json.ts';
 import {TESTING} from '../../shared/src/testing.ts';
+import type {AST} from '../../zero-protocol/src/ast.ts';
+import type {Format} from '../../zero-types/src/format.ts';
+import type {CustomQueryID} from '../../zql/src/query/named.ts';
 import {
   type Immutable,
   addContextToQuery,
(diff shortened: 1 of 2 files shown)


packages/zero-react/src/use-query.tsx
```diff
@@ -77,6 +81,153 @@ const suspend: (p: Promise<unknown>) => void = reactUse
...
+
+/**
+ * deepEqual only value-compares the keys of its first argument and then
+ * compares own-key counts, so an explicitly-undefined key on one side can
+ * absorb a differently-named defined key on the other:
+ * `deepEqual({q: undefined, page: 1}, {page: 1, sort: 'x'})` is true.
+ * Comparing in both directions closes that hole — a defined key present on
+ * only one side is always caught by the pass iterating that side.
+ */
+function symmetricDeepEqual(
```
arv, line 105: Is this really needed? Our json deep equal is lax in undefined properties.

packages/zero-react/src/use-query.tsx
```diff
@@ -77,6 +81,153 @@ const suspend: (p: Promise<unknown>) => void = reactUse
...
+        )
+      ) {
+        cache.rawQuery = query;
+        return cache.q as Query<TTable, TSchema, TReturn>;
+      }
+    }
+  }
+
+  const q = addContextToQuery(query, zero.context);
+  if ('query' in query) {
```
arv, line 202: this branching seems wrong

arv (changes requested):
I like where this is going. It is conceptually an improvement.

arv:
See https://github.com/rocicorp/mono/pull/6430 for deepEquals fix

PR #6306, opened 2026-08-04
fix(d2s): `scalar` incorrectly compiled when the user incorrectly used `scalar`

tantaman:
`scalar: true` is only allowed if the thing it is applied to constraints over a primary key.

E.g.,

```ts
issue.whereExists('owner', o => o.whereName('foo'))
```

`scalar: true` is applicable if there is a unique index on `name` and not applicable otherwise.

When `scalar:true` was used in the inapplicable case, d2s would compile incorrectly.

---

This also fixes the fuzzer so it would have found these bugs. The fuzzer only produced valid `scalar: true` queries now it can produce invalid ones like the user can.

Maybe there's a type level fix here so the user cannot even create invalid ones.... tbd

### DIFF
--- a/packages/z2s/src/compiler.ts
+++ b/packages/z2s/src/compiler.ts
@@ -352,9 +352,15 @@ function where(
         ' OR ',
       )})`;
     case 'correlatedSubquery':
-      if (condition.scalar) {
-        return scalarSubquery(spec, condition, table);
-      }
+      // `scalar` is deliberately ignored here. It is a planner hint for the
+      // IVM engines, which have no cost-based planner and benefit from
+      // pre-resolving an at-most-one-row subquery to a literal comparison
+      // (see zqlite's `resolveSimpleScalarSubqueries`). Postgres decorrelates
+      // EXISTS into a semi-join on its own, so the hint buys nothing here —
+      // and honoring it as `parentField = (SELECT childField … LIMIT 1)` is
+      // only sound when the subquery matches at most one row *globally*.
+      // Applying it unconditionally silently dropped rows whenever the
+      // subquery was not pinned to a unique key.
       return exists(spec, condition, table);
     case 'simple':
       return simple(spec, condition, table);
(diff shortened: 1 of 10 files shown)


packages/z2s/src/compiler.ts
```diff
@@ -352,9 +352,15 @@ function where(
         ' OR ',
       )})`;
     case 'correlatedSubquery':
-      if (condition.scalar) {
-        return scalarSubquery(spec, condition, table);
-      }
+      // `scalar` is deliberately ignored here. It is a planner hint for the
+      // IVM engines, which have no cost-based planner and benefit from
```
grgbkr, line 356: but the ivm engine does have a cost-based planner...

or i guess it's zero-cache that resolves this bit?

packages/zql-integration-tests/src/chinook/fuzz/driver.ts
```diff
@@ -737,57 +755,21 @@ export async function checkYieldTail(
...
- * *simple*) scalar subquery and check that the production split agrees with the oracle — the
- * original `scalar: true` AST through z2s (`parentField = (SELECT childField … LIMIT 1)`)
- * vs the IVM over the **pre-resolved** AST (`parentField = <literal>`, the transform
- * zero-cache's pipeline-driver applies via {@link resolveSimpleScalarSubqueries}).
- *
- * `rawRows` (client-named) backs the synchronous scalar executor; it must match the data
- * loaded into the oracle so both sides pick the same row. A candidate whose subquery fails
- * to resolve (would leave the base IVM treating `scalar` as a plain EXISTS — a generation
- * regression, not an engine bug) is reported, not silently passed.
+ * **Scalar-invariance sweep:** `scalar` is a plan hint every engine in the differential is
```
grgbkr, line 758: not sure what "in the differential" means here.

grgbkr:
LGTM

PR #6308, opened 2026-08-04
fix(zero-client): avoid slow query warnings on reconnect

0xcadams:
### What

Prevents long-lived Zero queries from reporting their age as end-to-end materialization latency when the client reconnects. Materialization timing now represents only the view's initial completion.

### Why

After reconnect, the first poke deliberately re-confirms existing queries with `got(true)`. The materialization callback treated every confirmation as a new metric but measured from the view's original start time. This produced false slow-query warnings and polluted materialization metrics whenever a connection was interrupted.

See https://github.com/tldraw/tldraw/issues/9808 for the original report.

### Changes

- Record the end-to-end materialization metric only once per materialized view.
- Preserve reconnect query-state reconciliation behavior.
- Add a regression test covering a repeated `got(true)` after 309.7 seconds.

### DIFF
--- a/packages/zero-client/src/client/query-materialization-metrics.test.ts
+++ b/packages/zero-client/src/client/query-materialization-metrics.test.ts
@@ -425,6 +425,35 @@ describe('query materialization metrics', () => {
       view2.destroy();
     });
 
+    test('records the metric only once when got is re-confirmed', () => {
+      mockPerformanceNow([
+        100,
+        125, // Initial client materialization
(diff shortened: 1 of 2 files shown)


grgbkr:
LGTM. 

There appear to still be edge cases where very long query-materialization-end-to-end times will be reported due to being disconnected for a long time (tab backgrounded, internet not available, etc).  I don't see any straightforward way to filter these out, without missing those which really do need to wait on a connection.

PR #6309, opened 2026-08-05
fix(zero-cache): avoid rebuilds for nullability changes

0xcadams:
### What

Treat PostgreSQL column nullability updates as logical metadata changes in zero-cache replicas instead of rebuilding the physical SQLite table and its indexes. Replica versioning still occurs so clients observe the schema change.

### Why

Zero records nullability with a `|NOT_NULL` marker in SQLite's declared type, even though SQLite does not enforce that constraint. The change processor interpreted changes to this marker as physical storage-type changes, causing unnecessary table and index rebuilds for `SET NOT NULL` and `DROP NOT NULL`.

These rebuilds create substantial write pressure on large replicas and can cause replication transactions to fail.

### Changes

- Compare column storage types without the logical `|NOT_NULL` marker before deciding to rebuild.
- Update column metadata and bump replica versions for nullability-only changes.
- Add regression coverage for both setting and dropping `NOT NULL`, including assertions that table and index storage remains unchanged.
- Normalize logical nullability markers in schema-change fuzz comparisons.

### DIFF
--- a/packages/zero-cache/src/services/replicator/change-processor.ts
+++ b/packages/zero-cache/src/services/replicator/change-processor.ts
@@ -19,10 +19,13 @@ import {
   type LiteTableSpecWithReplicationStatus,
 } from '../../db/lite-tables.ts';
 import {
+  isArrayColumn,
+  isEnumColumn,
   mapPostgresToLite,
   mapPostgresToLiteColumn,
   mapPostgresToLiteIndex,
 } from '../../db/pg-to-lite.ts';
+import type {ColumnSpec} from '../../db/specs.ts';
 import type {StatementRunner} from '../../db/statements.ts';
 import type {LexiVersion} from '../../types/lexi-version.ts';
 import {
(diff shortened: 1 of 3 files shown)


packages/zero-cache/src/services/replicator/change-processor.ts
```diff
@@ -651,23 +651,31 @@ class TransactionProcessor {
     const oldSpec = mapPostgresToLiteColumn(table, msg.old, 'ignore-default');
     const newSpec = mapPostgresToLiteColumn(table, msg.new, 'ignore-default');
 
+    // NOT_NULL is encoded in the SQLite type string for upstream metadata, but
+    // is not enforced by SQLite and does not change how values are stored.
+    const oldStorageType = oldSpec.dataType.replace('|NOT_NULL', '');
```
darkgnotic: Maybe instead of using the old encoded-type strings, we could make the comparison between the structured `msg.old` and `msg.new` fields? Maybe factor out a helper like `differentStorageTypes()` and explicitly enumerate the fields that we care about ... a bit more code but seems like a good thing.

darkgnotic (commented):
LGTM, but I would feel more comfortable with a look from @tantaman as well.  🙏

PR #6310, opened 2026-08-05
fix(zero): apply postgres patch in Docker image

0xcadams:
### What

Ensure the Zero Docker image applies the repository's `postgres@3.4.7` runtime patch when installing the server's external dependencies.

### Why

The Zero server bundle externalizes the `postgres` package, so the Docker image installs it into a fresh pnpm workspace. That workspace did not include the repository's `patchedDependencies` configuration or its patch file, causing the image to use an unpatched driver despite local workspace installations being patched.

This reintroduced the connection cleanup failure addressed by the existing patch.

### Changes

- Move the postgres patch under `packages/zero` so it is available in the Docker build context.
- Configure the generated runtime workspace to apply the patch.
- Copy the patch into the image before installing runtime dependencies.
- Update the root workspace configuration to reference the relocated patch.


arv (commented):
LGTM

PR #6311, opened 2026-08-05
fix(zero-cache): allow serving transactions to spill

0xcadams:
### What

Use `BEGIN IMMEDIATE` for the serving replica's durable writer while retaining `BEGIN CONCURRENT` for snapshotters that simulate changes against historical snapshots and always roll them back.

This allows SQLite to spill dirty pages into WAL during large serving-replica transactions instead of retaining the complete write set in native memory until commit.

### Why

A production PostgreSQL migration changed several UUID columns to `NOT NULL`. Zero encoded the nullability change in SQLite's declared type and interpreted it as a physical type change, causing each serving replica to copy the complete table and rebuild its indexes.

The largest affected table repeatedly OOM-killed the view-syncers. The serving writer used `BEGIN CONCURRENT`, for which our SQLite build explicitly prohibits dirty-page spilling. As the replacement table and indexes were built, every dirty page remained in SQLite's native page cache until commit.

The Node `--max-old-space-size` setting did not protect against this because it limits V8's managed heap, not SQLite's native allocations. Native SQLite memory eventually pushed the process beyond its Kubernetes memory limit.

Avoiding rebuilds for nullability-only changes addresses this incident's immediate trigger (see https://github.com/rocicorp/mono/pull/6309), but it does not address the systemic problem. A legitimate type conversion, index rebuild, backfill, or large DML transaction could produce the same unspillable write set.

The serving replicator is the only connection that commits to the replica. Snapshotters use `BEGIN CONCURRENT` only to stage private changes against historical snapshots before rolling them back. The durable writer can therefore use `BEGIN IMMEDIATE`, acquire the WAL writer lock, and spill pages without changing snapshot isolation.

### Changes

- Start serving-replica transactions with `BEGIN IMMEDIATE`.
- Keep snapshotter transactions on `BEGIN CONCURRENT` and clarify the mixed-mode transaction model.
- Add coverage for an immediate writer committing while a concurrent WAL2 snapshot holds private changes.
- Add a regression test proving that a large immediate transaction spills uncommitted dirty pages into WAL and remains fully rollback-safe.
- Add direct coverage that the serving `ChangeProcessor` selects `BEGIN IMMEDIATE`.

### DIFF
--- a/packages/zero-cache/src/db/begin-concurrent.test.ts
+++ b/packages/zero-cache/src/db/begin-concurrent.test.ts
@@ -19,6 +19,34 @@ describe('db/begin-concurrent', () => {
     dbFile.delete();
   });
 
+  const startSpillingWriter = () => {
+    const writer = dbFile.connect(lc);
+    writer.pragma('journal_mode = DELETE');
+    writer.pragma('journal_mode = WAL2');
+    writer.pragma('cache_size = 10');
+    writer.pragma('cache_spill = ON');
+    writer.exec('CREATE TABLE large(id INTEGER PRIMARY KEY, value BLOB)');
+
+    const journalBytes = () =>
+      [`${dbFile.path}-wal`, `${dbFile.path}-wal2`].reduce(
+        (total, path) => total + (existsSync(path) ? statSync(path).size : 0),
+        0,
+      );
+    const before = journalBytes();
+
+    writer.prepare('BEGIN IMMEDIATE').run();
+    writer.exec(`
+      WITH RECURSIVE rows(id) AS (
+        VALUES(1)
+        UNION ALL
+        SELECT id + 1 FROM rows WHERE id < 2000
+      )
+      INSERT INTO large SELECT id, randomblob(4000) FROM rows;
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/db/begin-concurrent.test.ts
```diff
@@ -130,4 +131,69 @@ describe('db/begin-concurrent', () => {
     conn1.close();
     conn2.close();
   });
+
+  test('immediate writer commits alongside a concurrent snapshot', () => {
```
darkgnotic, line 163: Can you add an additional test in which the `BEGIN IMMEDIATE` is first, and has some pending writes, and then validate that a `BEGIN CONCURRENT` can successfully proceed?

I verified this locally but I think it's work codifying in a test.

darkgnotic (dismissed):
TIL about spilling and BEGIN CONCURRENT vs BEGIN IMMEDIATE.

I verified locally that the former can concurrent transaction can make progress while an immediate transaction is in progress.

PR #6312, opened 2026-08-05
chore(zero-cache): fix e2e lag metric

tantaman:
We were double counting old changes and missing changes that were only briefly present in the backlog.


0xcadams (commented):
LGTM this is neat. The upstream skew is a little hard to follow but makes sense.

PR #6313, opened 2026-08-05
chore(zero-cache): delay readiness for backup, support vfs monitoring

darkgnotic:
### Refactoring
* Moved common logic from `litestream3-backup-monitor.ts` into the `backup-monitor.ts` class
* The remaining logic is converted to a poll+publish API; renamed the class to `litestream3-prometheus-poller.ts`
* Converted / created poll+publish implementations for (1) local testing, i.e. `replica-poller.ts` and (2) litestream v5, i.e. `vfs-watermark-poller.ts`

### LTX / VFS implementation
* Iterated on `vfs-watermark-reader.ts` to fit into the poll+publish API instead of request/response
* Added logic to watch the local replica and close the VFS connection (which otherwise makes s3 LIST calls at a regular interval) when the backup is in sync with the local replica, to avoid incurring the cost for inactive dbs
* Simplified-away the worker wiring classes by using the ProcessManager

### Backup readiness
* Delay server-readiness until the first backup has been made by litestream in the case of initial-sync. This reduces the period of unavailability after a resync completes.
* This delay will also be applied in RMv2 when each replica writes to its own backup (as required by LTX)


packages/zero-cache/src/services/change-streamer/backup-monitor.ts
```diff
@@ -1,6 +1,91 @@
...
+  }
+
+  async run() {
+    this.#lc.info?.('starting backup monitor');
+    this.#initBackupLagMetric();
+
+    for await (const backedUp of this.#watermarks) {
+      if (!this.#latestBackup) {
+        this.#lc.info?.(`received first backup up watermark`, {backedUp});
+        this.#latestBackup = backedUp;
```
0xcadams, line 49: Should `#latestBackup` be updated for every watermark? As written, the backup-lag gauge retains the first backup timestamp for the pod’s lifetime and eventually false-alerts. Seems like this should move outside the `if`.

0xcadams: This is causing https://app.incident.io/rocicorp/response/incidents/1121 which is a false alert

darkgnotic: Thank you for catching this, and sorry for the false alert. Fixing now!

PR #6323, opened 2026-08-06
chore(rmv2): track backfill records in the SQLite change log

tantaman:
Updates `change-log-stream-writer.ts` create `ChangeLogCookieWriter` and use it as schema changes come through.

 
- Schema changes write "cookies" to the change log so we can resume a backfill
- Those cookies are eventually persisted in the replica

---

If we lose the log before the cookie makes it to the replica, that's fine. We resume from PG (upstream DB, not pg change log) from the replica's watermark. PG will send us the DDL again.

### DIFF
--- a/packages/zero-cache/src/services/replicator/change-log-db.ts
+++ b/packages/zero-cache/src/services/replicator/change-log-db.ts
@@ -10,6 +10,12 @@
  * message. Its head is therefore always at or above the watermark the stream
  * would resume from, which is what {@link reconcileChangeLog} anchors on.
  *
+ * Beside the buffer it holds the log's *cookie jar* — the backfill progress
+ * state folded up to the same head. That is `change-log-cookies.ts`; this file
+ * owns its tables' lifecycle, because a cookie set is only meaningful paired
+ * with the watermark it was folded to, and so must be created, dropped, and
+ * invalidated with the buffer rather than beside it.
+ *
  * Because it is disposable, it has no migration framework. Any inconsistency —
  * a schema change, a leftover file from a different replica, a gap left by a
  * crash or by running with the writer disabled, a corrupt file — is resolved by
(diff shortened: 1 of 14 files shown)


packages/zero-cache/src/services/replicator/change-log-db.ts
```diff
@@ -408,10 +426,24 @@ export type ReseedReason =
...
+ * it is a field rather than a derivation because the reseed that replaces the
+ * clear reads as intent at the call site, and because it is what the
+ * `cookie_reseed` counter is keyed on.
+ */
 export type ReconcileResult =
-  | {action: 'none'; head: string}
-  | {action: 'truncated'; head: string; rows: number}
-  | {action: 'reseeded'; head: string; reason: ReseedReason};
+  | {action: 'none'; head: string; cookiesStale: false}
+  | {action: 'truncated'; head: string; rows: number; cookiesStale: true}
```
tantaman, line 440: truncated -> change log was ahead of the replica

packages/zero-cache/src/services/replicator/change-log-db.ts
```diff
@@ -408,10 +426,24 @@ export type ReseedReason =
...
+  | {
+      action: 'reseeded';
```
tantaman, line 442: reseeded -> change log had to be wiped and will resume from the replica's watermark

packages/zero-cache/src/services/replicator/change-log-db.ts
```diff
@@ -408,10 +426,24 @@ export type ReseedReason =
...
+ * clears them (see `change-log-cookies.ts`). It is `action !== 'none'` today;
+ * it is a field rather than a derivation because the reseed that replaces the
+ * clear reads as intent at the call site, and because it is what the
+ * `cookie_reseed` counter is keyed on.
+ */
 export type ReconcileResult =
-  | {action: 'none'; head: string}
-  | {action: 'truncated'; head: string; rows: number}
-  | {action: 'reseeded'; head: string; reason: ReseedReason};
+  | {action: 'none'; head: string; cookiesStale: false}
```
tantaman, line 439: this will be "steady state" after we remove the PG log. Today `none` happens when the sqlite log and pg log are at the same head.

PR #6327, opened 2026-08-07
chore(rmv2): initialize the backfill records from the replica

tantaman:
- Postgres has the official log saying, “We processed up to page 100, and these backfills remain.”
  - SQLite now keeps its own copy of that.
  - This branch reads both logs whenever the change stream connects.
  - Because SQLite may be behind, it replays the intervening schema changes before comparing them.
  - It records whether they agree.

Zero still uses Postgres’s answer. A mismatch or unreadable replica is logged and measured, but does not change or stop replication. This lets the new SQLite-derived path run in production before cutting over.

Logs five outcomes:

  - equal
  - equal-after-fold (SQLite was behind but agreed after log replay)
  - divergent
  - inconclusive (insufficient history or too much to scan safely)
  - error

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/change-log-initializer.ts
+++ b/packages/zero-cache/src/services/change-streamer/change-log-initializer.ts
@@ -1,0 +1,486 @@
+/**
+ * Where a stream connection's initialization parameters come from.
+ *
+ * A change-streamer starts a stream from three things read at one position: the
+ * watermark to resume at, the {@link BackfillRequest}s to hand the change
+ * source, and the cookie set those requests were derived from. Postgres has
+ * always supplied all three (`Storer.getStartStreamInitializationParameters()`).
+ * The point of this workstream is that a restored replica can supply them
+ * instead, which is what lets the `cdc` schema be decommissioned.
+ *
+ * This is the two-source stage. Both are computed on every stream connection,
+ * compared, and Postgres's is used — so the replica-derived path runs in
+ * production, at full rate, before anything depends on it. The flip is
+ * `initFromPgChangeLog: false`, not a rewrite.
+ *
+ * The shape is {@link UpstreamAcker}'s, deliberately: two stores, one output, a
+ * flag per store, and the Postgres flag carrying the marker that retires it.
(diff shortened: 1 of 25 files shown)


packages/zero-cache/src/services/change-streamer/change-log-initializer.ts
```diff
@@ -0,0 +1,486 @@
...
+ * - `inconclusive` — the log does not span the interval, so the fold cannot be
+ *   performed. Routine on a cold log, and not a finding.
+ * - `error` — the comparison itself failed: the replica could not be read, or
+ *   the fold threw. Not one of the four classifications, because it is a
+ *   statement about this process rather than about the two stores. Kept
+ *   separate rather than folded into `inconclusive` so that a change-streamer
+ *   that has silently stopped being able to read its replica is visible before
+ *   that read becomes the only source.
+ */
+export type InitComparisonResult =
```
tantaman, line 77: comparison goes away once PG goes away.

packages/zero-cache/src/services/change-streamer/change-log-initializer.ts
```diff
@@ -0,0 +1,486 @@
...
+ * The most change-log rows one comparison will scan.
+ *
+ * The interval is `(replicaWatermark, pgWatermark]`, so it is normally a
+ * handful of transactions — but a replicator that has been stalled can leave it
+ * arbitrarily large, and this runs on the stream loop between reconciliation and
+ * `startStream`, where latency delays the resumption of replication. Over the
+ * cap the comparison declines rather than pays: a chronically lagging replicator
+ * shows up as `inconclusive`, which is charted, and not as a stall.
+ */
+const MAX_FOLD_SCAN_ROWS = 100_000;
```
tantaman, line 122: changing to 10k, 100k seems a bit much

tantaman: this is how many rows we scan to compare change logs while we're in RMv1.5

tantaman: todo: we should compare change log hash and cookie result in 1 pass. For a follow up

PR #6328, opened 2026-08-07
fix(zero-cache): bound poke parts by payload size

cesara:
Zero flushes a chunk every 100 rows regardless of how large those rows are, so a single chunk can be arbitrarily big. Sizing by bytes keeps each message inside the client's timeout, and makes fat-row schemas degrade slowly instead of failing completely.

### DIFF
--- a/packages/zero-cache/src/integration/integration.pg.test.ts
+++ b/packages/zero-cache/src/integration/integration.pg.test.ts
@@ -605,7 +605,8 @@ describe('integration', {timeout: 30000}, () => {
 
     const downstream = new Queue<unknown>();
     const ws = new WebSocket(
-      `ws://localhost:${port}/zero/sync/v${PROTOCOL_VERSION}/connect` +
+      // This test asserts the legacy JSON pokePart representation.
+      `ws://localhost:${port}/zero/sync/v${PROTOCOL_VERSION - 1}/connect` +
         `?clientGroupID=abc&clientID=def&wsid=123&schemaVersion=1&baseCookie=&ts=123456789&lmid=1`,
       encodeURIComponent(btoa('{}')),
     );
@@ -806,7 +807,8 @@ describe('integration', {timeout: 30000}, () => {
 
       const downstream = new Queue<unknown>();
       const ws = new WebSocket(
-        `ws://localhost:${port}/zero/sync/v${PROTOCOL_VERSION}/connect` +
+        // This test asserts the legacy JSON pokePart representation.
+        `ws://localhost:${port}/zero/sync/v${PROTOCOL_VERSION - 1}/connect` +
           `?clientGroupID=abc&clientID=def&wsid=123&schemaVersion=1&baseCookie=&ts=123456789&lmid=1`,
         encodeURIComponent(btoa('{}')), // auth token
       );
--- a/packages/zero-cache/src/integration/integration.pg.test.ts
+++ b/packages/zero-cache/src/integration/integration.pg.test.ts
@@ -17,7 +17,12 @@ import {
 } from '../../../zero-permissions/src/permissions.ts';
 import type {AST} from '../../../zero-protocol/src/ast.ts';
 import type {InitConnectionMessage} from '../../../zero-protocol/src/connect.ts';
-import type {PokeStartMessage} from '../../../zero-protocol/src/poke.ts';
+import {
+  LAST_POKE_PART_PROTOCOL_VERSION,
+  POKE_CHUNK_MESSAGE_TYPE,
+  type PokePartBody,
+  type PokeStartMessage,
+} from '../../../zero-protocol/src/poke.ts';
 import {PROTOCOL_VERSION} from '../../../zero-protocol/src/protocol-version.ts';
 import {createSchema} from '../../../zero-schema/src/builder/schema-builder.ts';
 import {string, table} from '../../../zero-schema/src/builder/table-builder.ts';
@@ -606,7 +611,7 @@ describe('integration', {timeout: 30000}, () => {
     const downstream = new Queue<unknown>();
     const ws = new WebSocket(
       // This test asserts the legacy JSON pokePart representation.
-      `ws://localhost:${port}/zero/sync/v${PROTOCOL_VERSION - 1}/connect` +
+      `ws://localhost:${port}/zero/sync/v${LAST_POKE_PART_PROTOCOL_VERSION}/connect` +
         `?clientGroupID=abc&clientID=def&wsid=123&schemaVersion=1&baseCookie=&ts=123456789&lmid=1`,
       encodeURIComponent(btoa('{}')),
     );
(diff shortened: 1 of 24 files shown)


packages/zero-cache/src/services/view-syncer/client-handler.ts
```diff
@@ -104,9 +105,14 @@ export function startPoke(
...
-// Semi-arbitrary threshold at which poke body parts are flushed.
-// When row size is being computed, that should be used as a threshold instead.
-const PART_COUNT_FLUSH_THRESHOLD = 100;
+// Keep poke parts small enough that clients observe progress while syncing.
+// A single patch can exceed this limit because rows are atomic at the protocol
+// level, but it will be sent in a part by itself.
+export const POKE_PART_FLUSH_THRESHOLD_BYTES = 1024 * 1024;
+
+function jsonByteLength(value: unknown): number {
+  return Buffer.byteLength(JSON.stringify(value));
```
grgbkr: This is going to increase latency of pokes, as stringifying the value is expensive, and Buffer.byteLength is O(n).  

We need to structure this so that the JSON.stringify(value) work is not thrown away, but rather is used (as we do need to stringify the value to send it over the socket).  Also we probably want to use string length (O(1)) instead of  Buffer.byteLength (O(n)).

packages/zero-cache/src/types/downstream.ts
```diff
@@ -0,0 +1,19 @@
+import type {Downstream} from '../../../zero-protocol/src/down.ts';
+
+const serializedDownstreams = new WeakMap<object, string>();
+
+/**
+ * Associates a downstream message with its already-computed JSON encoding.
+ * The message remains structurally unchanged for in-process consumers.
+ */
+export function setSerializedDownstream<T extends Downstream>(
```
grgbkr: Using a weak map to pass the serialized message through layers is a smell.  We should just update the method/function apis and pass the serialized message as a parameter.

grgbkr:
So unfortunately changing poke's format is going to be more complicated than this, because of the challenge of compatibility with old clients.

Our old compatibility requirement was that server major-version `X` must still support client major version `X - 1 `.  Your PR doesn't do that, client's at major version `X - 1` will not understand the binary poke parts and won't work.  We could accomplish this old requirement by doing one major-version where the client is update to understand binary poke chunks, and then in the next major-version have the server start sending binary poke chunks.

However, with cloudzero we have even harder compatibility requirements.  We essentially need to never break compatibility with old clients.   For this change it means we need to detect clients connecting with the old protocol version, and continue sending the existing style pokeParts to them.  Only clients connecting with the new protocol version should get the binary poke chunks.

We need to update the documentation in this file https://github.com/rocicorp/mono/blob/main/packages/zero-protocol/src/protocol-version.ts#L72 to reflect that due to cloudzero we can never increase MIN_SERVER_SUPPORTED_SYNC_PROTOCOL

This probably needs a bit of a design doc as to how we will keep things maintainable, covering how do we handle forking logic around old sync protocol versions, how do we handle testing of them, etc.

Some types of protocol changes can be done in a backwards compatible manner (add fields or messages old clients just ignore), but its not practical to do that with this change, because backwards compat would mean sending both the old poke part and the new poke chunks, and clearly we dont want to do that, would defeat the entire point of the change.

packages/zero-cache/src/integration/integration.pg.test.ts
```diff
@@ -605,7 +605,8 @@ describe('integration', {timeout: 30000}, () => {
 
     const downstream = new Queue<unknown>();
     const ws = new WebSocket(
-      `ws://localhost:${port}/zero/sync/v${PROTOCOL_VERSION}/connect` +
+      // This test asserts the legacy JSON pokePart representation.
+      `ws://localhost:${port}/zero/sync/v${PROTOCOL_VERSION - 1}/connect` +
```
grgbkr: I think you want to fix this protocol version so that its the last one that uses pokePart.  As written, once PROTOCOL_VERSION gets bumped again, this will be the wrong value.

packages/zero-cache/src/integration/integration.pg.test.ts
```diff
@@ -806,7 +807,8 @@ describe('integration', {timeout: 30000}, () => {
 
       const downstream = new Queue<unknown>();
       const ws = new WebSocket(
-        `ws://localhost:${port}/zero/sync/v${PROTOCOL_VERSION}/connect` +
+        // This test asserts the legacy JSON pokePart representation.
```
grgbkr, line 892: Seems we should add an integration tests for the new chunk representation.

packages/zero-cache/src/services/view-syncer/client-handler.ts
```diff
@@ -104,8 +105,10 @@ export function startPoke(
   };
 }
 
-// Semi-arbitrary threshold at which poke body parts are flushed.
-// When row size is being computed, that should be used as a threshold instead.
+// Keep poke parts small enough that clients observe progress while syncing.
+// A single patch can exceed this limit because rows are atomic at the protocol
+// level, but it will be sent in a part by itself.
+export const POKE_PART_FLUSH_THRESHOLD_CHARS = 1024 * 1024;
```
grgbkr, line 112: POKE_CHUNK_FLUSH_THRESHOLD_CHARS

grgbkr: is it actually bytes now since we switched to binary?

packages/zero-protocol/src/poke.ts
```diff
@@ -16,12 +16,13 @@ import {nullableVersionSchema, versionSchema} from './version.ts';
  * A poke begins with a `poke-start` message which contains the `baseCookie`
  * the poke is updating from and the `cookie` the poke is updating to.
  *
- * The poke continues with zero to many `poke-part` messages, each of which
- * can contain patch parts.  These patch parts should be merged in the order
- * received.
+ * Through protocol version 51, the poke continues with zero to many
```
grgbkr, line 19: I think we should add a single byte at the beginning of binary messages as a message type tag.  pokeChunks can be 0x00.  This allows us to extend to support more binary messages.

packages/zero-cache/src/types/downstream.ts
```diff
@@ -0,0 +1,6 @@
+import type {Downstream} from '../../../zero-protocol/src/down.ts';
+
+export type ViewSyncerDownstream = {
+  readonly message: Downstream;
+  readonly serialized: string | undefined;
```
grgbkr, line 10: document this type, what is its motivation?

packages/zero-cache/src/workers/connection.ts
```diff
@@ -412,6 +450,169 @@ export function send(
...
+  return new Promise((resolve, reject) => {
+    if (ws.readyState !== WebSocket.OPEN) {
+      const error = webSocketClosedError();
+      lc.debug?.(
+        `Dropping outbound binary message on ws (state: ${ws.readyState})`,
+      );
+      reject(error);
+      return;
+    }
+    ws.send(chunk, error => (error ? reject(error) : resolve()));
```
grgbkr: Shouldn't this have the WEBSOCKET_SEND_TIMEOUT_MS  timeout like the one you added to send.

packages/zero-client/src/client/zero-poke-handler.ts
```diff
@@ -119,11 +151,40 @@ export class PokeHandler {
       this.#receivingPoke = undefined;
       return;
     }
-    this.#pokeBuffer.push({...this.#receivingPoke, pokeEnd});
+    const receivingPoke = this.#receivingPoke;
+    if (receivingPoke.chunks.length > 0) {
+      try {
+        receivingPoke.parts.push(
```
grgbkr: I think this can be simplified, by encoding `PokePartBody[]` instead of `PokePatch[]`

Right now, `ClientHandler` builds `PokePartBody`, `DownstreamSender` on server does string surgery to strip `pokeID` into `PokePatch`, and `PokeHandler` on client maps over every patch to add `pokeID` back. Encoding `PokePartBody[]` directly simplifies both layers:

1. Server (connection.ts): `DownstreamSender` can extract the serialized part with a simple `serialized.slice(12, -1)` (stripping the `['pokePart', ...]` wrapper) and feed it straight to `PokeChunkEncoder`. We can delete `serializedPokePatch()`.
2. View-Syncer (ClientHandler): Sizing logic can be simplified to a soft threshold check (`serializedRowsLength >= POKE_PART_FLUSH_THRESHOLD_CHARS`) because `PokeChunkEncoder` downstream already guarantees hard 1MB binary frame bounding.
3. Client (zero-poke-handler.ts): `JSON.parse` produces `PokePartBody[]` directly. We can assign the decoded array directly without `.map()` object copies, or redundant `pokeID` validation on chunks
4. Protocol Types: We can drop the separate `PokePatch` / `PokePatches` schemas from zero-protocol.

packages/zero-cache/src/services/view-syncer/client-handler.ts
```diff
@@ -208,46 +211,69 @@ export class ClientHandler {
...
+
     const ensureBody = async () => {
       if (!pokeStarted) {
         await this.#push(['pokeStart', pokeStart]);
         pokeStarted = true;
       }
       return (body ??= {pokeID});
     };
+
+    const serializedLength = (rowsLength = serializedRowsLength) =>
```
grgbkr: The exact character accounting here (serializedLength(), prospective lookahead math, serializedBody.slice(0, -2), and length assertions) adds quite a bit of complexity.

Since POKE_PART_FLUSH_THRESHOLD_CHARS is just a heuristic to keep memory bounded in view-syncer and keep legacy JSON pokePart messages reasonably sized (and PokeChunkEncoder downstream handles hard 1MB binary frame bounding for v52+), we can treat this as a soft limit:

typescript
```
if (
  serializedRowsLength >= POKE_PART_FLUSH_THRESHOLD_CHARS ||
  partCount >= PART_COUNT_FLUSH_THRESHOLD
) {
  await flushBody();
}
```

This would let us drop the exact string-length math and slice-stitching while achieving the same memory-bounding behavior for both legacy and new clients.

grgbkr (approved):

packages/zero-cache/src/types/poke-chunk.ts
```diff
@@ -0,0 +1,73 @@
...
+    await this.#write(this.#patchCount === 0 ? '[]' : ']', emit);
+    await this.#flush(emit);
+  }
+
+  cancel(): void {
+    this.#finished = true;
+    this.#bufferedBytes = 1;
+  }
+
+  async #write(value: string, emit: EmitPokeChunk): Promise<void> {
```
arv, line 51: is value a valid JSON string here?

packages/zero-cache/src/types/poke-chunk.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  test('splits a single oversized patch without corrupting UTF-8', async () => {
+    const encoder = new PokeChunkEncoder(7);
+    const chunks: Uint8Array[] = [];
+    const emit = (chunk: Uint8Array) => {
+      chunks.push(chunk.slice());
+      return Promise.resolve();
+    };
+    const patch = {value: '🙂漢'.repeat(100)};
```
arv, line 33: nice test

packages/zero-cache/src/types/poke-chunk.test.ts
```diff
@@ -0,0 +1,86 @@
...
+});
+
+function joinAndDecode(chunks: Uint8Array[]): string {
+  const decoder = new TextDecoder('utf-8', {fatal: true});
+  const decoded: string[] = [];
+  for (const chunk of chunks) {
+    expect(chunk[0]).toBe(POKE_CHUNK_MESSAGE_TYPE);
+    decoded.push(decoder.decode(chunk.subarray(1), {stream: true}));
+  }
+  decoded.push(decoder.decode());
```
arv, line 84: why is this needed?

packages/zero-client/src/client/zero-poke-handler.ts
```diff
@@ -211,6 +259,33 @@ export class PokeHandler {
   }
 }
 
+function decodePokeChunks(chunks: Uint8Array[]): PokeParts {
+  const decoder = new TextDecoder('utf-8', {fatal: true});
+  const decoded = new Array<string>(chunks.length + 1);
```
arv, line 264: this creates a holey array. Just do `decoded = []`

arv (commented):
LGTM

Nice work

PR #6329, opened 2026-08-07
fix(zero-cache): timeout stalled websocket sends

cesara:
Add a timeout on the send so a stalled client cannot hold a pipeline for 85+ seconds;


grgbkr:
LGTM

PR #6330, opened 2026-08-07
fix(zero-client): warn on synced schema mismatch

cesara:


grgbkr:
I don't think this is an effective way to surface this to customers.
1. This is a surface visible to end-users (i.e. customers' of zero's customers), when the target of the message is zero's customers.
2. By default the zero client logs at level ERROR, and so this won't even be visible.  So this is unlikely to be seen in the first place.


This is an example of a suspected misconfiguration by customers.  Another example of suspected misconfiguration is being unable to connect to the db (probably a bad connection string, or bad ip allow list).  We need a way of surfacing these expected misconfiguration to users.  Probably somewhere in the cloudzero console, and potentially via a push communication channel (email, slack, ?).

PR #6333, opened 2026-08-10
chore(zero-cache): avoid an sqlite_master lookup in the common case

darkgnotic:
Avoid a lookup on `sqlite_master` in the common case of `ColumnMetadata.getInstance()`.


darkgnotic:
Replacing this with https://github.com/rocicorp/mono/pull/6334.

PR #6334, opened 2026-08-10
chore(zero-cache): avoid an sqlite_master lookup in the common case

darkgnotic:
Avoid a lookup on sqlite_master in the common case of ColumnMetadata.getInstance().


tantaman (commented):
lgtm

PR #6337, opened 2026-08-11
chore(rmv2): compare sqlite and pg change logs at a sampled rate

tantaman:
## What this does

Runs the SQLite change log's catchup **read path** against Postgres's, in production, without serving anything from it. On a 30-second cycle the change-streamer takes the range of committed transactions both logs hold, checks that every transaction present in one is present in the other, and — for a sampled percentage — reads the full catchup output through both stores' real read paths and compares a digest of what each would have served a subscriber. Postgres remains authoritative; nothing a subscriber receives changes.

## Rollout

- Default is `sqliteChangeLogMode=off`; **this PR has no production effect until enabled.**
- Enable per replication-manager with `--change-streamer-sqlite-change-log-mode=compare`. The sampling dial is `--change-streamer-sqlite-change-log-compare-percent` (default 1; presence checks always run at 100%).
- Roll back by returning the mode to `write` — both comparisons stop, the log keeps accumulating.
- Cost is bounded by construction: at most 256 transactions and 1,000 catchup rows per store per cycle, yielding to the event loop between sampled reads, so a busy shard is compared incrementally rather than all at once.

## Checking on it in production

- `zero_replica_sqlite_change_log_compare_result_total{outcome}` — per-transaction outcomes. `mismatch` is the finding. `inconclusive` means a purge or reseed moved the bounds mid-cycle and is deliberately not counted as divergence.
- `zero_replica_sqlite_change_log_compare_cycles_total{result}` — cycles and skip reasons. `cold-log` after a reseed is routine; persistent `log-unavailable` is not.
- `zero_replica_sqlite_change_log_compare_cycle_duration_seconds` — cycle cost on the serving process.
- A mismatch also logs one error line carrying **only the watermark** — payloads are customer data and are never logged. The watermark is enough to pull both stores' rows and diff them by hand.

No paging on divergence (standing decision): compare mode has no production impact, so mismatches are investigated via the dashboard at leisure.

### DIFF
--- a/packages/zero-cache/src/config/normalize.ts
+++ b/packages/zero-cache/src/config/normalize.ts
@@ -58,8 +58,7 @@ export function assertNormalized(
       sqliteChangeLogReadPercent <= 100,
     '--change-streamer-sqlite-change-log-read-percent must be an integer between 0 and 100',
   );
-  // Deliberately no mode restriction: the default is nonzero, and the sampled
-  // comparison only runs in `compare` mode and above.
+  // This setting has no mode restriction. Comparison starts in `compare` mode.
   assert(
     Number.isSafeInteger(sqliteChangeLogComparePercent) &&
       sqliteChangeLogComparePercent >= 0 &&
--- a/packages/zero-cache/src/config/normalize.ts
+++ b/packages/zero-cache/src/config/normalize.ts
@@ -46,6 +46,7 @@ export function assertNormalized(
   const {
     sqliteChangeLogMode,
     sqliteChangeLogReadPercent,
+    sqliteChangeLogComparePercent,
     sqliteChangeLogRetentionMs,
     sqliteChangeLogReadBatchRows,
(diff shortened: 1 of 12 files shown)


packages/zero-cache/src/config/normalize.ts
```diff
@@ -57,6 +58,13 @@ export function assertNormalized(
       sqliteChangeLogReadPercent <= 100,
     '--change-streamer-sqlite-change-log-read-percent must be an integer between 0 and 100',
   );
+  // This setting has no mode restriction. Comparison starts in `compare` mode.
+  assert(
+    Number.isSafeInteger(sqliteChangeLogComparePercent) &&
+      sqliteChangeLogComparePercent >= 0 &&
```
tantaman, line 64: percentage of transactions we compare logs for. E.g., 1% of transactions are compared against pg and sqlite. We hard cap at 1k rows so we don't spend forever scanning. We should however cap at bytes... to follow up on

PR #6338, opened 2026-08-11
fix(zero-cache): exit on replication source termination

0xcadams:
### What

Prevent Zero replication from remaining permanently wedged when the PostgreSQL logical replication connection terminates while the change-streamer is blocked by flow control.

The PostgreSQL change source now reports termination independently of change consumption. If termination occurs while waiting on Storer or subscriber backpressure, the change-streamer exits so its worker can be replaced and replication can resume from durable state.

### Why

During the incident, the Storer began applying backpressure and the change-streamer stopped requesting messages from the logical replication stream. After two missed PostgreSQL keepalive windows, the existing watchdog correctly destroyed the unresponsive replication connection.

However, the change-streamer was blocked in `readyForMore()` and could not observe the stream failure through normal iteration. It therefore never entered its reconnect path. Replication remained wedged for roughly 12 hours despite the watchdog firing after two minutes.

Normal in-process recovery is not reliable in this state because stopping the Storer may itself be queued behind the operation that is no longer making progress. Exiting the worker allows process replacement to force-abort that state and resume from the durable watermark.

This patch addresses the missing cancellation edge. It does not change watchdog behavior or attempt to diagnose the underlying reason the Storer stopped making progress.

### Changes

- Report unexpected PostgreSQL replication readable errors and closures through `sourceTerminated`.
- Exclude intentional local stream cancellation from fatal termination handling.
- Race source termination against both Storer and subscriber flow-control waits.
- Exit the change-streamer worker when termination interrupts a blocked flow-control wait.
- Bound cleanup before process replacement to 100ms and tolerate cleanup rejections.
- Add coverage for real PostgreSQL source termination and the incident’s Storer-backpressure failure mode.

### DIFF
--- a/packages/zero-cache/src/services/change-source/pg/logical-replication/stream.ts
+++ b/packages/zero-cache/src/services/change-source/pg/logical-replication/stream.ts
@@ -4,6 +4,7 @@ import {
   PG_OBJECT_NOT_IN_PREREQUISITE_STATE,
 } from '@drdgvhbh/postgres-error-codes';
 import type {LogContext} from '@rocicorp/logger';
+import {resolver} from '@rocicorp/resolver';
 import {defu} from 'defu';
 import postgres, {type Options, type PostgresType} from 'postgres';
 import {sleep} from '../../../../../../shared/src/sleep.ts';
@@ -42,6 +43,7 @@ export async function subscribe(
 ): Promise<{
   messages: SourceWithPendingQueue<StreamMessage>;
   acks: Sink<bigint>;
+  sourceTerminated: Promise<Error>;
 }> {
   const session = postgres(
     defu(
@@ -135,6 +137,7 @@ export async function subscribe(
     : undefined;
 
   let destroyed = false;
+  const sourceTerminated = resolver<Error>();
   const typeParsers = await getTypeParsers(db, {returnJsonAsString: true});
   const parser = new PgoutputParser(typeParsers);
   const messages = Subscription.create<StreamMessage>({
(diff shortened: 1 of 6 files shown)


packages/zero-cache/src/services/change-source/pg/logical-replication/stream.ts
```diff
@@ -135,6 +137,7 @@ export async function subscribe(
     : undefined;
 
   let destroyed = false;
+  const sourceTerminated = resolver<Error>();
```
darkgnotic, line 140: IIUC what this implements is a way to be notified that the ChangeStream ended in the form of a Promise.

Perhaps it might be worth making this part of the Source / Subscription API. I will think about this ...

packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
```diff
@@ -596,6 +598,20 @@ class ChangeStreamerImpl implements ChangeStreamerService {
           lastWatermark,
           backfillRequests,
         );
+        const waitForFlowControl = async (wait: Promise<void>) => {
+          if (!stream.sourceTerminated) {
+            await wait;
+            return;
+          }
+          const sourceError = await Promise.race([
```
darkgnotic, line 606: In particular, I think this is a pattern in that will result in a memory leak due to a growing chain of "then()" callbacks on the long-lived `stream.sourceTerminated` Promise.

https://levelup.gitconnected.com/unfixable-memory-leaks-in-promise-race-28e5c5a6032c

darkgnotic (commented):
Thanks @0xcadams !

I agree with the motivation of the change but will think about an alternate API / implementation based on a couple of observations.

packages/zero-cache/src/services/change-streamer/change-streamer-service.ts
0xcadams: Agreed yeah I fixed this in https://github.com/rocicorp/mono/pull/6344

0xcadams: But I will wait for your fix!

PR #6340, opened 2026-08-12
fix(zero-client): fix out of order mutations

tantaman:
see user report: https://rocicorp.slack.com/archives/C0AQDEA9N10/p1786539945269819

tldr:

Tabs sending mutations for other tabs would end up skipping mutations, causing mutations to arrive out of order.

The fix is to pull all mutations out of the array that we have not sent rather than indexing into the array. Indexing into the array doesn't work because the tab always puts its own mutations last.

### DIFF
--- a/packages/zero-client/src/client/zero.test.ts
+++ b/packages/zero-client/src/client/zero.test.ts
@@ -1749,116 +1749,125 @@ test('pusher sends one mutation per push message', async () => {
   ]);
 });
 
-test('pusher does not skip mutations when the pending list is reordered', async () => {
-  // The pending list is the client group's local commit chain, so it
-  // interleaves mutations from every client (tab) in the group. `persist`
-  // rebases this client's unpersisted mutations onto the current perdag head
-  // and `refresh` then rebuilds memdag as `perdagChain ++ ownUnpersisted`, so
-  // another client's mutation can move *ahead* of a mutation we already sent.
-  // We must still send it, and we must never send a gap in a client's mutation
-  // ID sequence (the server rejects that with an out of order mutation error).
-  const z = zeroForTest();
-  await z.triggerConnected();
-
-  const mockSocket = await z.socket;
-  const clientGroupID = await z.clientGroupID;
-
-  const mut = (clientID: string, id: number): Mutation => ({
-    type: MutationType.Custom,
-    clientID,
-    id,
-    name: 'mut1',
-    args: [{}],
-    timestamp: id,
-  });
-
-  const push = async (mutations: Mutation[]) => {
-    mockSocket.messages.length = 0;
-    await z.pusher(
-      {
-        profileID: 'p1',
-        clientGroupID,
-        pushVersion: 1,
-        schemaVersion: '1',
-        mutations,
-      },
-      'test-request-id',
-    );
-    return mockSocket.messages.map(raw => {
-      const [, {mutations}] = valita.parse(JSON.parse(raw), pushMessageSchema);
-      return `${mutations[0].clientID}:${mutations[0].id}`;
-    });
-  };
-
-  expect(await push([mut('c2', 1), mut('c1', 1)])).toEqual(['c2:1', 'c1:1']);
-
-  // c2:2 lands *before* c1:1, the last mutation we sent.
-  expect(await push([mut('c2', 1), mut('c2', 2), mut('c1', 1)])).toEqual([
-    'c2:2',
-  ]);
-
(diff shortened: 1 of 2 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -383,8 +381,25 @@ export class Zero<
...
+   * and the *cross-client* order of that chain is not stable between pushes:
+   * `persist` rebases this client's unpersisted mutations onto the current
+   * perdag head, and `refresh` then rebuilds memdag as
+   * `perdagChain ++ ownUnpersistedMutations`. So another client's mutation can
+   * move ahead of a mutation we have already sent. Tracking what was sent per
+   * client (rather than a single position in the chain) keeps the decision
+   * independent of that ordering, so we never skip a mutation and never send a
+   * gap in a client's mutation ID sequence.
+   *
+   * Cleared whenever the connection is (re)established or torn down, since the
```
grgbkr: It's not that the server forgets what it has seen on a socket, its that the client doesn't know what was received by the server before the connection was lost.  Only those acknowledged via lastMutationIDs via poke are not resent (because they are rebased out of the commit chain by poke, since they are reflected in the base snapshot).

packages/zero-client/src/client/zero.test.ts
```diff
@@ -1749,6 +1749,118 @@ test('pusher sends one mutation per push message', async () => {
...
+      },
+      'test-request-id',
+    );
+    return mockSocket.messages.map(raw => {
+      const [, {mutations}] = valita.parse(JSON.parse(raw), pushMessageSchema);
+      return `${mutations[0].clientID}:${mutations[0].id}`;
+    });
+  };
+
+  expect(await push([mut('c2', 1), mut('c1', 1)])).toEqual(['c2:1', 'c1:1']);
```
grgbkr: should we use the pushing client's own client id rarther than c1 in this tests?

packages/zero-client/src/client/zero.test.ts
```diff
@@ -1749,6 +1749,118 @@ test('pusher sends one mutation per push message', async () => {
...
+
+  expect(await push([mut('a', 1), mut('b', 1)])).toEqual(['a:1', 'b:1']);
+
+  // Tab a closed with a:2 and a:3 persisted but unsent. We are the only client
+  // left that can push them.
+  expect(
+    await push([mut('a', 1), mut('a', 2), mut('a', 3), mut('b', 1)]),
+  ).toEqual(['a:2', 'a:3']);
+
+  // After a reconnect the server has forgotten what it saw on the old socket,
```
grgbkr: its not really about the server forgetting https://github.com/rocicorp/mono/pull/6340/changes#r3768207547

packages/zero-client/src/client/zero.test.ts
```diff
@@ -1749,6 +1749,118 @@ test('pusher sends one mutation per push message', async () => {
...
+    );
+    return mockSocket.messages.map(raw => {
+      const [, {mutations}] = valita.parse(JSON.parse(raw), pushMessageSchema);
+      return `${mutations[0].clientID}:${mutations[0].id}`;
+    });
+  };
+
+  expect(await push([mut('a', 1), mut('b', 1)])).toEqual(['a:1', 'b:1']);
+
+  // Tab a closed with a:2 and a:3 persisted but unsent. We are the only client
```
grgbkr: It doesnt really matter if tab a is closed or not, a client always pushes mutations in its own client group.

packages/zero-client/src/client/zero.test.ts
```diff
@@ -1749,6 +1749,118 @@ test('pusher sends one mutation per push message', async () => {
...
+  ]);
+
+  // c2:3 must not be sent without c2:2 having been sent first.
+  expect(
+    await push([mut('c2', 1), mut('c2', 2), mut('c1', 1), mut('c2', 3)]),
+  ).toEqual(['c2:3']);
+});
+
+test('pusher sends the pending mutations of a closed client', async () => {
+  // Mutation recovery skips our own client group, so when another tab in the
```
grgbkr: This comment is not really relevant to this test, the push being tested is not a mutationRecovery push (its not for another client group).  

We currently have mutationRecovery disabled, we probably should rip the code for it out of client/zero.ts since its been disabled for so long is probably just causing confusion.

I think the important bit this test is covering is that on connection closed #lastMutationIDsSent is reset.

grgbkr:
Wow thats an old bug.  I guess it would heal on reconnect when this.#lastMutationIDSent was reset.

PR #6343, opened 2026-08-12
fix(zero-cache): preserve litestream restore errors

0xcadams:
### What

Preserve Litestream’s captured stdout and stderr when a replica restore fails, so Zero logs the underlying restore error instead of only the subprocess exit code.

### Why

PR #6245 changed Litestream restores to pipe stdout and stderr rather than inherit the pod streams. This prevented raw Litestream errors from bypassing Zero’s structured logging and alert classification. It also re-surfaced the captured output after determining whether a failure was retriable.

PR #6267 fixed snapshot retention and removed the restore-retry workaround, including its captured-output logging. The subprocess output remained piped, however, so failed restores silently discarded it. PR #6268 later refactored restore logic on `main` while preserving that behavior.

This surfaced during a prod replication incident. A replica restore ran for approximately 13 minutes before failing, but the only retained diagnostic was `litestream exited with code 1`. The underlying S3, WAL, filesystem, or Litestream error was lost.

### Changes

- Include captured Litestream stdout and stderr in the error propagated from a failed restore.
- Keep subprocess output piped so raw Litestream errors cannot bypass structured logging.
- Log the replication-manager restore failure once without duplicating the exit message.
- Add regression coverage proving both stdout and stderr survive a failed restore.

### DIFF
--- a/packages/zero-cache/src/services/change-source/common/replica-restore.ts
+++ b/packages/zero-cache/src/services/change-source/common/replica-restore.ts
@@ -62,10 +62,7 @@ export async function restoreReplica(
     );
     result = attempt.result;
   } catch (e) {
-    lc.error?.(
-      `error restoring backup. resyncing the replica: ${String(e)}`,
-      e,
-    );
+    lc.error?.(`error restoring backup. resyncing the replica`, e);
   } finally {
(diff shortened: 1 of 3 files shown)


darkgnotic (commented):
Thank you @0xcadams !

PR #6346, opened 2026-08-13
chore(zero-cache): respond to upstream disconnects even if waiting for back-pressure

darkgnotic:
Avoid wedging the change-streamer if back-pressure never clears.

(Partial, alternate strategy to https://github.com/rocicorp/mono/pull/6338)

The implementation of the policy has two parts:

### Part 1
* Add a memory-safe way to race the short-lived flow-control promises within the change-streamer loop, with the (long) lifetime of the change stream (i.e. the change-source `Source`), leveraging an AbortController with short-lived promises to avoid the [problematic memory behavior of Promise.race()](https://github.com/nodejs/node/issues/17469).
* This ensures that we stop waiting on flow-control if the upstream connection has ended, as opposed to hanging indefinitely (observed during the incident that back-pressure never clears even after many hours).

### Part 2
* Add a timeout when waiting for the storer to drain its entries before the usual reconnect to the change-source. 
* If the timeout fires, the service aborts and the process exits.

### DIFF
--- a/packages/zero-cache/src/types/streams.ts
+++ b/packages/zero-cache/src/types/streams.ts
@@ -40,6 +40,18 @@ export type Source<T> = AsyncIterable<T> & {
   cancel: (err?: Error) => void;
 
   /**
+   * Resolves the result of the specified `other` Promise, or a
+   * resolved/rejected Promise when the Source is terminated or
+   * rejects.
+   *
+   * This is similar to using `Promise.race([other, sourceDone])`, but
+   * in a memory safe way, as repeatedly calling `Promise.race([ ... ])` with
+   * a long-lived Promise results in leaking memory via a growing list of
+   * `then()` callbacks (https://github.com/nodejs/node/issues/17469).
+   */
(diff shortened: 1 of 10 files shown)


packages/zero-cache/src/types/streams.ts
```diff
@@ -39,6 +39,18 @@ export type Source<T> = AsyncIterable<T> & {
...
+   * Resolves the result of the specified `other` Promise, or a
+   * resolved/rejected Promise when the Source is terminated or
+   * rejects.
+   *
+   * This is similar to using `Promise.race([other, sourceDone])`, but
+   * in a memory safe way, as repeatedly calling `Promise.race([ ... ])` with
+   * a long-lived Promise results in leaking memory via a growing list of
+   * `then()` callbacks (https://github.com/nodejs/node/issues/17469).
+   */
+  doneOr<R>(other: Promise<R>): Promise<void | R>;
```
0xcadams, line 52: Nice

0xcadams (commented):
LGTM

PR #6351, opened 2026-08-13
feat(zero-cache): allow overriding the replication-stream inbound timeout

gerardatkonvo:
### What

Adds `ZERO_UPSTREAM_PG_STREAM_INBOUND_TIMEOUT_MS` (`upstream.pgStreamInboundTimeoutMs`), an optional override for the replication-stream inbound-liveness threshold, which is otherwise hard-derived as `2x wal_sender_timeout`. Defaults are unchanged. Filing as a PR directly since issues are disabled on this repo — happy to adjust naming/plumbing.

### Why

The two timeouts protect opposite parties with asymmetric false-positive costs. `wal_sender_timeout` is the **server** reaping dead clients; operators tune it low for their own reasons (CloudNativePG hard-defaults it to `5s` cluster-wide for fast standby-failover detection), and a false positive there costs a cheap reconnect. The inbound watchdog is the **client** detecting a dead server, and its false positive is expensive: destroy stream → abort the partially-decoded transaction → replay from the confirmed watermark.

A busy wal sender can be legitimately silent for longer than `2x wal_sender_timeout`: decoding through long stretches of WAL that is filtered out (changes on unpublished tables — including zero-cache's own changeLog churn when the change-db is co-located with upstream), or assembling a large transaction that pgoutput v1 only sends at commit. #6348's back-pressure suppression doesn't cover this case — `queued === 0` because nothing is arriving at all — so the watchdog fires against a healthy, progressing connection.

**Production incident that motivated this** (zero-cache 1.8.0, PG 17 / CloudNativePG, `wal_sender_timeout = 5s` → 10s fuse): a ~900 MB backlog of large TOASTed-row transactions produced repeated decode silences of 10.1–10.5s. Each tripped the watchdog: destroy → `aborting interrupted transaction` → replay → stall, every 60–90s; net catch-up ≈ 0 for over an hour. Confirming the mechanism: `ALTER ROLE <zero-user> SET wal_sender_timeout = '60s'` (fuse → 120s), with nothing else changed, drained the same backlog in ~2 minutes. We had to change a server-side setting we didn't want to touch — per-role, to preserve the operator's 5s for physical standbys — purely to move a client-side threshold that has no knob. Logs available if useful.

### Changes

- `computeLivenessTimings(walSenderTimeoutMs, inboundTimeoutOverrideMs?)`: a positive override replaces `inboundTimeoutMs` only. Manual-keepalive timing stays derived from `wal_sender_timeout` (that half genuinely belongs to the server contract). `wal_sender_timeout = 0` still disables liveness entirely, override or not — without server keepalives, inbound silence is normal on an idle healthy connection, so any silence-based watchdog would false-positive (#6244 semantics preserved). Non-positive/NaN overrides fall back to the default.
- Plumbed `upstream.pgStreamInboundTimeoutMs` → `initializePostgresChangeSource` → `PostgresChangeSource` → `subscribe()`, following the `lagReportIntervalMs` pattern.
- Teardown warning now reports the actual threshold instead of hard-coding "2x wal_sender_timeout".
- Tests for override honored / fallback / disabled-stays-disabled; updated the config `--help` snapshot.

### Testing

- `stream.test.ts` and `zero-config.test.ts`: all pass.
- Full `src/services/change-source/pg` suite: 1037 passed, 1 pre-existing flake (`change-source.backfill.pg.test.ts` "column renamed while being backfilled") which passes in isolation and is untouched by this change.
- `tsc` clean, `oxlint` no errors, `oxfmt` applied.

Related: #6047, #6244, #6348.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/zero-cache/src/config/zero-config.test.ts
+++ b/packages/zero-cache/src/config/zero-config.test.ts
@@ -53,6 +53,25 @@ test('zero-cache --help', () => {
                                                                                                                                                                                                    
                                                                         (Note that this option has no effect for Postgres versions before 17.)                                                     
                                                                                                                                                                                                    
+     --upstream-pg-stream-inbound-timeout-ms number                     optional                                                                                                                   
+       ZERO_UPSTREAM_PG_STREAM_INBOUND_TIMEOUT_MS env                                                                                                                                              
+                                                                        The time (in milliseconds) without any inbound message from the upstream                                                   
+                                                                        wal sender after which the replication stream is considered unresponsive                                                   
+                                                                        and torn down to force a reconnect.                                                                                        
+                                                                                                                                                                                                   
+                                                                        Defaults to 2x the server's wal_sender_timeout. That suits idle                                                            
+                                                                        streams, but a busy wal sender can be legitimately silent for longer —                                                     
+                                                                        e.g. while decoding through WAL from unpublished tables, or assembling a                                                   
+                                                                        large transaction that is only sent at commit. If the server's                                                             
+                                                                        wal_sender_timeout is aggressive (some managed environments                                                                
+                                                                        default it as low as 5 seconds), the resulting teardown aborts and                                                         
+                                                                        replays the in-flight transaction, which can prevent replication from                                                      
+                                                                        ever catching up on a large backlog. Set this option to widen the                                                          
+                                                                        client-side threshold without changing the server setting.                                                                 
+                                                                                                                                                                                                   
+                                                                        (This option has no effect when wal_sender_timeout is 0, which                                                             
+                                                                        disables inbound liveness detection entirely.)                                                                             
+                                                                                                                                                                                                   
      --mutate-url string[]                                              optional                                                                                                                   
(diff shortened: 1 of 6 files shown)


darkgnotic (approved):
Thank you @gerardatkonvo.

This seems like a reasonable mitigation for your use case, at least until we support streaming of in-progress transactions.

Note that the other mitigation that we recommend in general is to split large transactions into smaller pieces, which may be still be necessary in order to avoid exceeding any wal-size limit.

Please update your branch with changes from `main` and then we'll get this in!

gerardatkonvo:
> Thank you @gerardatkonvo.
> 
> This seems like a reasonable mitigation for your use case, at least until we support streaming of in-progress transactions.
> 
> Note that the other mitigation that we recommend in general is to split large transactions into smaller pieces, which may be still be necessary in order to avoid exceeding any wal-size limit.
> 
> Please update your branch with changes from `main` and then we'll get this in!

Done! Thanks for the quick response.

PR #6355, opened 2026-08-13
fix(zero-cache): clean up Litestream restore temp files

0xcadams:
### What

Prevent failed or interrupted Litestream restores from leaving temporary databases and staged WAL files that consume the replica volume. The restore path also records bounded directory snapshots around cleanup to expose any other files accumulating beside the replica.

### Why

The bundled Litestream implementations use several temporary file shapes during restore:

- `rocicorp/litestream` `zero@v0.0.9` restores into `${replicaFile}.tmp` and stages parallel WAL downloads as `${replicaFile}.tmp-<8-hex-index>-wal`.
- Litestream `v0.5.16` restores into `${replicaFile}.tmp`; its v0.3 compatibility path may also create normal SQLite sidecars such as `-wal` and `-shm`.

Our fork does not clean all of these files after a failed restore. Repeated attempts can therefore amplify disk usage while logical replica DB/WAL metrics remain comparatively small.

### Changes

- Delete the temporary SQLite family before every restore and again after the attempt.
- Remove the exact indexed-WAL staging pattern emitted by `rocicorp/litestream` `zero@v0.0.9`.
- Extend `deleteLiteDB` to remove rollback-journal (`-journal`) files.
- Preserve the original restore result if final cleanup fails, while logging a warning.
- Log up to 100 directory entries and sizes before and after cleanup for additional disk-usage diagnostics.
- Add regression coverage for stale files, failed restores, indexed WAL staging, rollback journals, cleanup-error precedence, directory diagnostics, and preservation of the real replica family.

### DIFF
--- a/packages/zero-cache/src/db/delete-lite-db.ts
+++ b/packages/zero-cache/src/db/delete-lite-db.ts
@@ -1,7 +1,7 @@
 import {rmSync} from 'node:fs';
 
 export function deleteLiteDB(dbFile: string) {
-  for (const suffix of ['', '-wal', '-wal2', '-shm']) {
+  for (const suffix of ['', '-wal', '-wal2', '-shm', '-journal']) {
     rmSync(`${dbFile}${suffix}`, {force: true});
   }
 }
--- a/packages/zero-cache/src/services/litestream/commands.test.ts
+++ b/packages/zero-cache/src/services/litestream/commands.test.ts
@@ -347,13 +347,16 @@ describe('litestream/commands restoreReplica', () => {
     const source = join(dir, 'source.db');
     const replica = join(dir, 'replica.db');
     const temporaryReplica = `${replica}.tmp`;
+    const stagedWAL = `${temporaryReplica}-deadbeef-wal`;
     createRestorableReplica(source, '01');
     writeSQLiteFamily(temporaryReplica);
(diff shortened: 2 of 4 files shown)


darkgnotic (approved):
This looks good to me. 

One thing I'm surprised about ... does litestream v5 really not create any tmp files? I would have expected it to download ltx files somewhere and then apply them to the db.

(I don't know this for a fact, but rather inferring from how I think it would work)

Is this something you (or your agent) could double check?

0xcadams:
> Litestream v5 **does create a temp file**, but it is the reconstructed SQLite database:
> 
> ```text
> ${replicaFile}.tmp
> ```
> 
> For Cloud Zero’s S3 restore path, it does **not** download LTX objects into separate local files:
> 
> 1. `OpenLTXFile()` returns the S3 `GetObject` response body.
> 2. `ResumableReader` reads that stream directly, reopening byte-range requests if interrupted.
> 3. The LTX compactor merges those readers through an `io.Pipe`.
> 4. The decoder writes SQLite pages directly into `${replicaFile}.tmp`.
> 5. Litestream fsyncs and renames that file to `${replicaFile}`.
> 
> The pinned `superfly/ltx v0.5.2` compactor only buffers one SQLite page per input. It does not spill LTX inputs to disk.

So, it's in-memory.

PR #6357, opened 2026-08-14
chore(zero-cache): proportional consensus timeout

darkgnotic:
(Part 1 of 2 subscriber flow control improvements)

Changes the consensus flow control policy from a fixed 1-second timeout to a timeout proportional to the time in which the consensus responded. The proportion determines the maximum reduction of throughput when a subscriber is lagging, while allowing for variance in subscriber (i.e. machine) processing times.

For example, the default value of 2.0 allows the slowest subscriber to be 3x slower than the majority in the steady state, while similarly bounding degraded throughput to 1/3 the normal rate.

The implementation uses the event-driven release feature (https://github.com/rocicorp/mono/pull/6240) and computes the timeout based on when consensus is reached.

Upcoming Part 2: Pathological Laggard detection


arv:
LGTM

Nice.

PR #6361, opened 2026-08-17
feat(zql): warn when a manual {flip: true} is costlier than the semi-join

gyash1512:
## Problem

`{flip: true}` does two things: it makes the subquery the driving side of the join, and it sets `flippable = false`, which removes the join from the planner's search. When the flipped subquery has no selective filter, this drives a full scan of that table on every hydration pass, holds the read snapshot open for the duration, and blocks WAL checkpointing on the replica for as long as it runs — and nothing reports it. The alternative is never costed, so the planner cannot correct it.

## Change

After `PlannerGraph.plan()` selects the best plan, re-cost it with each manual flip undone, using the same reset → flip → FOFI → unlimit → propagate → cost sequence the search uses. If the semi-join is cheaper, `lc.warn` with both estimates. Manual flips are still honored; this only surfaces the planner's opinion.

Also passes `lc` through `buildPipeline` at hydration in `pipeline-driver.ts`. Without it, `planQuery` gets `lc = undefined` in production, so neither this warning nor the existing `MAX_FLIPPABLE_JOINS` warning can fire.

Going through the planner rather than a row-count heuristic matters: when a sibling join is flipped, its constraint propagates into the OR branches and can make a "bad-looking" flip genuinely cheap — the check correctly stays quiet in that case (covered by a test).

## Testing

- `pnpm run check-types` clean in `zql` and `zero-cache`
- `zql`: 1318/1318 tests; `pipeline-driver`: 45/45
- New `planner-manual-flip.test.ts` (5 tests): two manual flips inside `or(...)` warns for the costly one only; quiet for a good flip / no flip / no `lc`; chosen plan left intact.

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
+++ b/packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
@@ -696,6 +696,7 @@ export class PipelineDriver {
         },
         queryID,
         costModel,
+        this.#lc.withContext('queryID', queryID),
       );
       const schema = input.getSchema();
       input.setOutput({
--- a/packages/zql/src/planner/planner-graph.ts
+++ b/packages/zql/src/planner/planner-graph.ts
@@ -379,6 +379,60 @@ export class PlannerGraph {
         'no plan was found but flippable joins did exist!',
       );
     }
+
+    if (lc) {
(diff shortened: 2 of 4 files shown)


tantaman:
this is useful but I think it should only run/warn in dev. Mainly because planning isn't that cheap.

PR #6363, opened 2026-08-18
chore(litestream): use rocicorp litestream-v5 fork for the time being

darkgnotic:
Build zero with rocicorp's litestream v5 fork, which supports a configurable VFS poll interval.

This is a stop-gap measure until support (i.e. https://github.com/benbjohnson/litestream/pull/1157) has been released from the official repo.


0xcadams:
LGTM

PR #6372, opened 2026-08-19
fix(release): use immutable skopeo image

0xcadams:
### What

Make Zero release Docker publishing use upstream’s immutable Skopeo image rather than its daily-rebuilt version tag.

### Why

Quay moved the mutable `v1.22.2` tag and garbage-collected the workflow’s pinned manifest, causing the Release Zero Docker publish job to fail with `manifest unknown`. These move because they rebuild every day on a mutable base for OS security patches. Because the workflow already pins a digest, the daily image rebuilds were not providing automatic updates.

### Changes

- Switch all three Skopeo action references to `v1.22.2-immutable`.
- Pin the verified immutable multi-architecture manifest digest.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -318,7 +318,7 @@ jobs:
           path: dist/docker
 
       - name: Push GHCR Docker image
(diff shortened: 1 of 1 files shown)


darkgnotic (approved):
Thank you @0xcadams !

arv (commented):
LGTM

PR #6374, opened 2026-08-19
chore(rmv2): add the change-log compare digest primitives

tantaman:
We are building a comparator to compare SQLite and PG change logs during rmv1.5.

Adds base functions to do that compare

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/change-log-compare-digest.ts
+++ b/packages/zero-cache/src/services/change-streamer/change-log-compare-digest.ts
@@ -1,0 +1,100 @@
+/**
+ * Pure building blocks for comparing catchup output between the Postgres and
+ * SQLite change logs: which transactions to sample, and how to reduce a
+ * catchup range to a single comparable digest.
+ *
+ * Nothing here touches either store. The comparator service supplies the
+ * batches.
+ */
+
+import {createHash} from 'node:crypto';
+import type {ShardID} from '../../types/shards.ts';
+import {extractChangeSubstring} from './change-log-codec.ts';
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/services/change-streamer/change-log-compare-digest.ts
```diff
@@ -0,0 +1,112 @@
...
+import {extractChangeSubstring} from './change-log-codec.ts';
+import {ChangeLogTransactionHasher} from './change-log-transaction-hash.ts';
+import type {WatermarkedChange} from './change-streamer.ts';
+
+/** Selects a stable sample by shard and watermark. */
+export function isSampledForCompare(
+  shard: ShardID,
+  watermark: string,
+  percent: number,
+): boolean {
```
tantaman, line 21: this function answers: do we compare this TX across both change logs or not?

packages/zero-cache/src/services/change-streamer/change-log-compare-digest.ts
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Builds one ordered digest for a catchup range.
+ *
+ * The row position makes missing, extra, changed, and reordered rows change the digest.
+ * The digest omits `precommit` because a commit already includes its watermark.
+ *
+ * The read stops at `maxRows` or at `maxBytes`, whichever comes first.
+ */
+export async function digestCatchupRange(
+  batches: AsyncIterable<readonly WatermarkedChange[]>,
```
tantaman, line 58: batches can be supplied by SQLite or PG

PR #6378, opened 2026-08-19
chore(rmv2): add `serve` mode for serving from sqlite change log

tantaman:
also adds tests to test rollback

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/sqlite-change-log-catchup.metrics.test.ts
+++ b/packages/zero-cache/src/services/change-streamer/sqlite-change-log-catchup.metrics.test.ts
@@ -1,0 +1,122 @@
+/**
+ * `observability/metrics.ts` caches every instrument in module scope, so an
+ * instrument created against a previous test's meter provider would be handed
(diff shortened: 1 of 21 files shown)


packages/zero-cache/src/services/change-streamer/sqlite-change-log-catchup.metrics.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  read(): AsyncIterable<readonly WatermarkedChange[]> {
+    return (async function* () {})();
+  }
+
+  close(): void {}
+}
+
+test('log_warm is recorded per catchup, not per coordinator', async () => {
```
tantaman, line 78: what is all this warmup stuff? This literally had me confused and arguing with Claude for over an hour.

Once we're fully on SQLite the SQLite log and the S3 backup start at the same point.

But while PG exists... the SQLite log starts at HEAD not BACKUP! So the SQLite log will have holes in it when it tries to serve catchup in rmv1.5

Thus we refuse to serve catchup from SQLite until it's log extends back 1 retention period. This is really awkward and unfortunate.

PR #6380, opened 2026-08-19
fix(zql): forward 'yield' sentinels when UnionFanIn probes sibling branches

tantaman:
https://rocicorp.slack.com/archives/C0B2S4EF4BD/p1787068441144619

## The bug

`UnionFanIn.#pushInternalChange` decides whether to forward an add/remove by asking each sibling branch whether it already holds the row:

```ts
if (first(input.fetch({constraint})) !== undefined) return;
```

`fetch` streams are `Stream<Node | 'yield'>` — they interleave `'yield'` sentinels for cooperative multitasking. So `first()` answers *"was the stream non-empty?"*, not *"did it produce a row"*. The call is type-correct and semantically wrong: **an empty branch that happens to emit a sentinel is misread as a branch that holds the row**, and the add/remove is silently dropped.

A dropped `add` leaves a downstream `Take` with an empty window while its own fetch still reports the row. The next edit trips `assert(takeState.bound, 'Bound should be set')` — in the view-syncer that rethrows and disconnects every client on the group, re-arming on rehydrate. A dropped `remove` is quieter: the take keeps a row that no longer matches.

Needs an `OR` containing a flipped `whereExists` plus a `limit`, so it only appears once the planner's cost model chooses to flip. Present on `main` and in 1.9.0.

## The fix

The probe now iterates explicitly, **forwarding** each sentinel and treating only a real node as a match.

### DIFF
--- a/packages/zql/src/ivm/union-fan-in.ts
+++ b/packages/zql/src/ivm/union-fan-in.ts
@@ -19,7 +19,7 @@ import {
   pushAccumulatedChanges,
 } from './push-accumulated.ts';
 import type {SourceSchema} from './schema.ts';
-import {first, type Stream} from './stream.ts';
+import type {Stream} from './stream.ts';
 import type {UnionFanOut} from './union-fan-out.ts';
 
 export class UnionFanIn implements Operator {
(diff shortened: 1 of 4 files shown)


packages/zql/src/ivm/union-fan-in.ts
```diff
@@ -170,7 +171,13 @@ export class UnionFanIn implements Operator {
...
       });
 
-      if (first(fetchResult) !== undefined) {
+      // `skipYields` is required. `fetch` interleaves 'yield' sentinels for
+      // cooperative multitasking, and a bare `first()` reads one as if it were
+      // a row: an empty branch that happens to yield is misread as a branch
+      // that holds the row, silently dropping this add/remove. That desyncs
+      // the push and fetch paths of a downstream `Take` and trips
+      // `assert(takeState.bound, 'Bound should be set')` on the next edit.
+      if (first(skipYields(fetchResult)) !== undefined) {
```
tantaman: hmmm... should we really skipYields in `union-fan-in`? I think we want to forward the yield so we can yield.

grgbkr:
LGTM

arv (commented):
LGTM

PR #6381, opened 2026-08-19
fix(zql): use the connection comparator for the overlay's startAt pruning

tantaman:
> Stacked on #6380 for a clean diff, but **independent of it** — different file, different root cause, different test. Can be rebased onto `main` and merged in either order.

Follow-up to #4926, which fixed exactly this confusion one call away.

## The bug

#4926 found that `generateWithStart` was being passed `indexComparator` when `req.start.row` is a bound in the **connection's** sort order, and introduced `connectionComparator` for it. It deliberately left `generateWithOverlay` on `indexComparator`, with a comment explaining that `generateWithOverlayInner` depends on index order to flush a not-yet-yielded add overlay.

That reasoning is correct — but it is about the **splice**. The same `compare` argument is also handed to `computeOverlays`, whose *only* use of it is `overlaysForStartAt`:

```ts
const undefinedIfBeforeStartAt = (row: Row | undefined) =>
  row === undefined || compare(row, startAt) < 0 ? undefined : row;
```

…and that compares against `startAt` — the very connection-sort bound #4926 was about. So the two orders get conflated again, one call deeper.

They coincide only when the overlay row and `startAt` agree on the constraint keys. An ordinary partitioned fetch agrees. `#fetchMulti`'s per-value sub-fetches do not.

## The fix

Give `generateWithOverlay` a separate `startAtCompare` parameter. `compare` keeps ordering the splice (index order, per #4926); `startAtCompare` orders `startAt`.

- `MemorySource` passes `connectionComparator`.
- `TableSource` passes its single comparator twice, since SQL already returns rows in connection order — **which is why `TableSource` never had this bug.**

### DIFF
--- a/packages/zql/src/ivm/memory-source.test.ts
+++ b/packages/zql/src/ivm/memory-source.test.ts
@@ -971,6 +971,7 @@ describe('multiConstraints overlay handling — both helpers', () => {
             overlay,
             1,
             compare,
+            compare,
             undefined,
             multiConstraints,
           ),
--- a/packages/zql/src/ivm/memory-source.ts
+++ b/packages/zql/src/ivm/memory-source.ts
@@ -383,6 +383,15 @@ export class MemorySource implements Source {
       // not yet yielded add overlay will be yielded when the first row
       // not matching the constraint is reached.
       indexComparator,
+      // `startAt` is `req.start.row`, a bound in the connection's sort order,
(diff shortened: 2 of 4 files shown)


grgbkr:
LGTM.  I don't think fetchMulti existed when #4926 was written.

PR #6390, opened 2026-08-20
fix(zero-cache): stop logging app data

cesara:
Removes row values, JWT payloads, bound SQL parameters and app-supplied error details from four log sites, three of which run at the default `info` level. Each site now logs column names, counts or error kinds instead, and postgres notices are reduced to severity, SQLSTATE and schema location.

### DIFF
--- a/packages/zero-cache/src/auth/write-authorizer.ts
+++ b/packages/zero-cache/src/auth/write-authorizer.ts
@@ -455,8 +455,9 @@ export class WriteAuthorizerImpl implements WriteAuthorizer {
         rowQuery,
       ))
     ) {
-      // Never log `op` or `authData`: `op.value` is the row the client is
-      // writing and `authData` is the decoded JWT payload.
+      // `op.value` is the row being written and `authData` is the decoded
+      // JWT payload, so neither can go into the log. Column names and policy
+      // counts are enough to debug a failed permission check.
       this.#lc.warn?.('Permission check failed', {
         action,
         phase,
--- a/packages/zero-cache/src/auth/write-authorizer.ts
+++ b/packages/zero-cache/src/auth/write-authorizer.ts
@@ -455,15 +455,16 @@ export class WriteAuthorizerImpl implements WriteAuthorizer {
         rowQuery,
       ))
     ) {
-      this.#lc.warn?.(
-        `Permission check failed for ${JSON.stringify(
-          op,
-        )}, action ${action}, phase ${phase}, authData: ${JSON.stringify(
-          authData,
-        )}, rowPolicies: ${JSON.stringify(
-          applicableRowPolicy,
-        )}, cellPolicies: ${JSON.stringify(applicableCellPolicies)}`,
-      );
+      // Never log `op` or `authData`: `op.value` is the row the client is
+      // writing and `authData` is the decoded JWT payload.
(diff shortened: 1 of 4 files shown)


packages/zero-cache/src/auth/write-authorizer.ts
```diff
@@ -455,15 +455,16 @@ export class WriteAuthorizerImpl implements WriteAuthorizer {
...
-      this.#lc.warn?.(
-        `Permission check failed for ${JSON.stringify(
-          op,
-        )}, action ${action}, phase ${phase}, authData: ${JSON.stringify(
-          authData,
-        )}, rowPolicies: ${JSON.stringify(
-          applicableRowPolicy,
-        )}, cellPolicies: ${JSON.stringify(applicableCellPolicies)}`,
-      );
+      // Never log `op` or `authData`: `op.value` is the row the client is
```
grgbkr: this comment is hard to parse.

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -1802,8 +1802,16 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
 
     for (const q of transformedCustomQueries) {
       if ('error' in q) {
-        const errorMessage = `Error transforming custom query ${q.name}: ${q.error}${q.details ? ` ${JSON.stringify(q.details)}` : ''}`;
-        lc.warn?.(errorMessage, q);
+        // `q.details` and app-supplied `q.message` come from the customer's
+        // own query handler and can carry arbitrary data. Log the query
+        // identity and the error kind only.
```
grgbkr: do these get passed through to the client and logged there?  just wondering if q.details and q.message are used at all now.

packages/zero-cache/src/services/view-syncer/view-syncer.ts
```diff
@@ -1802,8 +1802,16 @@ export class ViewSyncerService implements ViewSyncer, ActivityBasedService {
...
+        lc.warn?.('Error transforming custom query', {
+          id: q.id,
+          name: q.name,
+          error: q.error,
+          message: q.error === 'parse' ? q.message : undefined,
```
grgbkr: the comment suggests we shouldn't be logging q.message

grgbkr (approved):

packages/zero-cache/src/auth/write-authorizer.ts
```diff
@@ -455,15 +455,16 @@ export class WriteAuthorizerImpl implements WriteAuthorizer {
         rowQuery,
       ))
     ) {
-      this.#lc.warn?.(
-        `Permission check failed for ${JSON.stringify(
-          op,
-        )}, action ${action}, phase ${phase}, authData: ${JSON.stringify(
-          authData,
```
tantaman, line 462: womp, good catch.

packages/zero-cache/src/services/view-syncer/view-syncer.ts
cesara: fixed!

packages/zero-cache/src/services/view-syncer/view-syncer.ts
cesara: yes they are sent to client

PR #6393, opened 2026-08-21
chore(zero-cache): fix vitest.config.ts to be agent-friendly

darkgnotic:


arv (commented):
LGTM

Interesting. How was this found?

PR #6394, opened 2026-08-21
chore(zero-cache): monitor v5 backup progress with vfs-query

darkgnotic:
Use the `vfs-query` helper binary (https://github.com/rocicorp/mono/pull/6392) to poll for the backup watermarks instead of directly loading the VFS extension into the Node/SQLite process.

The new `vfs-watermark-poller.ts` (now in change-streamer/ rather than litestream/) is no longer spawned as a worker, since it no longer needs to perform vfs-backed blocking SQLite queries. It is instead a normal Node object that manages the spawning of a `vfs-query` subprocess when the backup is behind, which polls s3 at the configured interval while alive, and shutting the process down if/when the backup catches up.

Docker/build changes:
* Use the official litestream v5 distribution; the fork is no longer necessary since we are linking in VFS statically
* Build the `vfs-query` binary on the alpine image


aboodman:
it seems like the vfs-query program gets killed 1 second after it catches up with local if nothing else has changed in the meantime. This feels too aggressive to me. I believe it is fairly expensive (in s3 queries) to startup the worker. Should the worker stick around for 15s just in case there's another local change?

Sorry if I'm misunderstanding.

arv:
LGTM

Same comments as @aboodman's. Consider keeping the remote poller alive longer than 1s

PR #6399, opened 2026-08-21
fix(zero-cache): reject legacy queries if the user is on custom queries

tantaman:
Legacy queries could be arbitrarily deep JSON and blow up parsing. Now rejects them server side unless the user explicitly enables legacy queries in their zero config.

Closes a DOS hole and supercedes PR #6000


arv (approved):
LGTM

PR #6406, opened 2026-08-24
fix(zero-cache): validate litestream vfs executable

0xcadams:
### What

Validate that Litestream V5 backups have the VFS query executable required by the backup cleanup monitor.

### Why

The startup assertion checked the polling interval instead of the executable. Because the polling interval defaults to 15 seconds, configurations enabling V5 backups without ZERO_LITESTREAM_VFS_QUERY_EXECUTABLE passed normalization and failed later during monitor construction.

### Changes

- Check vfsQueryExecutable when V5 backups are enabled.
- Include the default VFS poll interval in the normalization test fixture so the regression test exercises the intended field.


darkgnotic (commented):
LGTM

PR #6409, opened 2026-08-25
chore(litestream): output `litestream restore` logs as they arrive

darkgnotic:
Output the stdout/stderr from `litestream restore` as it happens, rather than when it completes, to facilitate debugging.


0xcadams (commented):
LGTM

PR #6411, opened 2026-08-25
feat(tool): log leak pre push hook

cesara:


tantaman:
This is cool but given we already express a policy (no ast, row, RowValue, etc. types) taint analysis may be a better and more deterministic fit?

Given taint analysis will track the full data flow across the program but the LLM will... 🤷‍♂️

Let me ask Codex to create an example static analysis ruleset based on your PR's stated policy

tantaman:
wdyt of expanding the taint analysis instead of relying on an LLM API call?

https://github.com/rocicorp/mono/pull/6413

We can have the LLM iterate and expand our static & deterministic rules as we add logging.

0xcadams:
Yeah, what Matt said, or this is a CI thing and we have this as part of a simple LLM-based security review. We already wanted to add that (and intended to?).

Pre-commit hook calling claude won't work for me, because I don't have claude code installed. And it feels a bit invasive.

cesara:
thanks for the feedback, i'll try to codeql path and make it into a github action

PR #6416, opened 2026-08-26
fix(zero-cache): ignore pushes for closed connections

jdnvn:
when Zero encounters a connection error (e.g. outdated client schema) it invalidates the WebSocket's authentication context and begins closing the connection. during this time, if a custom mutation is pushed, it will call `mustGetConnectionContext()` and since the shutdown removed the context, Zero will throw the following error:

```
Connection auth state was not available for this websocket.
```

this PR fixes this by checking whether the connection context exists before pushing the mutation. if the context exists, we proceed with the push. if it does not exist, that means the connection is closing, so we drop the invocation and leave it to be retried by the client upon reconnecting.

### DIFF
--- a/packages/zero-cache/src/services/mutagen/pusher.test.ts
+++ b/packages/zero-cache/src/services/mutagen/pusher.test.ts
@@ -1585,6 +1585,23 @@ describe('pusher streaming', () => {
     vi.resetAllMocks();
   });
 
+  test('ignores a push after its websocket connection closes', () => {
+    const pusher = newPusherService({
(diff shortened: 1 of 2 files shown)


tantaman:
thanks

tantaman (approved):

jdnvn:
Hey @tantaman , thanks for the review! Looks like CI needs approval to run

tantaman:
Merging under https://github.com/rocicorp/mono/pull/6416

PR #6419, opened 2026-08-27
fix(zero-cache): improve detection of orphaned change-db connections

darkgnotic:
Improves detection of half-closed pg connections which manifest as queries never completing, often resulting in the change-db's `Storer` loop hanging the server indefinitely.

This PR subsumes previous attempts to do this at different levels:
* https://github.com/rocicorp/mono/pull/5819
* https://github.com/rocicorp/mono/pull/6348

While those two mechanisms are fine in an of themselves, this new ProgressMonitor covers both cases as well as other cases that were previously not covered in the Storer:
* Hanging when failing to get a connection
* Hanging on `BEGIN` and `COMMIT` of the TransactionPool, before any back-pressure kicks in
* Hanging on catchup requests

The ProgressMonitor covers all cases in the main change-streamer and storer loops. 

Context:
* https://rocicorp.slack.com/archives/C0ATEPDS694/p1787617588732219?thread_ts=1787546057.483619&cid=C0ATEPDS694

Miscellaneous:
* To conservatively avoid false positives, increase the statement response timeout from 20s to 30s since it now technically cover BEGINs and COMMITs in addition to INSERTs. Also, INSERTs have become slightly heavier with batching optimizations.

TODO:
* Remove the statement response timeout and backpressure timeout logic to simplify the code. This will be done separately to make this PR easier to cherry-pick.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/storer.ts
+++ b/packages/zero-cache/src/services/change-streamer/storer.ts
@@ -4,6 +4,10 @@ import {resolver, type Resolver} from '@rocicorp/resolver';
 import {type PendingQuery, type Row} from 'postgres';
 import {AbortError} from '../../../../shared/src/abort-error.ts';
 import {assert, unreachable} from '../../../../shared/src/asserts.ts';
+import {
+  BigIntJSON,
+  type JSONObject,
+} from '../../../../shared/src/bigint-json.ts';
 import {Queue} from '../../../../shared/src/queue.ts';
 import {promiseVoid} from '../../../../shared/src/resolved-promises.ts';
 import * as v from '../../../../shared/src/valita.ts';
@@ -154,6 +158,7 @@ export class Storer implements Service {
   readonly #statementTimeoutMs: number;
   readonly #changeLogBatchSize: number;
   readonly #drainTimeoutMs: number;
+  readonly #progressMonitor: ProgressMonitor;
 
   #approximateQueuedBytes = 0;
   #running = false;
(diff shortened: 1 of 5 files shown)


packages/zero-cache/src/services/change-streamer/storer.ts
```diff
@@ -206,6 +216,31 @@ export class Storer implements Service {
...
+   *
+   * Does not cancel `promise` on timeout (there is no cancellation mechanism
+   * for a postgres.js query); it just stops waiting for it, and swallows a
+   * later rejection so it doesn't surface as an unhandled rejection.
+   */
+  async #withTimeout<T>(name: string, promise: Promise<T>): Promise<T> {
+    const result = await orTimeout(promise, this.#statementTimeoutMs);
+    if (result === 'timed-out') {
+      void promise.catch(() => {});
+      throw new AbortError(
```
arv, line 237: AbortError does not seem right.

https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static

The builtin APIs use DOMException TimeoutError

packages/zero-cache/src/services/change-streamer/storer.ts
```diff
@@ -1123,3 +1220,90 @@ export class PurgeLocker {
...
+  start() {
+    clearInterval(this.#timer);
+    this.#timer = setInterval(
+      () => this.checkTaskProgress(new Date()),
+      this.#failureThresholdMs,
+    ).unref();
+  }
+
+  // Called by an internal periodic timer, but exported for testing.
+  checkTaskProgress(now: Date) {
```
arv, line 1275: maybe use number instead of Date?

arv (commented):
LGTM with nits

PR #6420, opened 2026-08-27
chore(zero-cache): remove obsolete query watchdog logic

darkgnotic:
Remove obsolete query watchdog logic for handing half-closed connections:
* https://github.com/rocicorp/mono/pull/5819
* https://github.com/rocicorp/mono/pull/6348

which have both been subsumed by the higher-level ProgressMonitor introduced in: https://github.com/rocicorp/mono/pull/6419


arv (commented):
LGTM

PR #6425, opened 2026-08-27
fix(zero-cache): reject/ignore invalid watermarks from litestream v3

darkgnotic:
Mitigates a one-off case of an invalid watermark that was temporarily reported from litestream's prometheus metrics, with a valid one following soon after.

The root cause of the invalid watermark is still being investigated, but this PR makes the situation recoverable.

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/litestream3-prometheus-poller.test.ts
+++ b/packages/zero-cache/src/services/change-streamer/litestream3-prometheus-poller.test.ts
@@ -106,6 +106,30 @@ litestream_replica_validation_total{db="/tmp/zbugs-sync-replica.db",name="file",
     await vi.waitFor(() => expect(scheduled).toEqual(['618ocqq8', '618p0bw8']));
   });
 
+  test('ignores an invalid watermark reported by litestream metrics', async () => {
+    const time = Date.UTC(2025, 3, 24);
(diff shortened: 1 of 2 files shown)


packages/zero-cache/src/services/change-streamer/litestream3-prometheus-poller.test.ts
```diff
@@ -106,6 +106,30 @@ litestream_replica_validation_total{db="/tmp/zbugs-sync-replica.db",name="file",
...
   });
 
+  test('ignores an invalid watermark reported by litestream metrics', async () => {
+    const time = Date.UTC(2025, 3, 24);
+    vi.setSystemTime(time);
+
+    // Real-world garbage watermark observed in production: not a valid
+    // lexicographic version, so it must be dropped rather than crash the
+    // poller or get scheduled for publishing.
+    const t1 = (Date.now() / 1000).toPrecision(9);
```
0xcadams, line 116: Wow. Very odd.

0xcadams (commented):
LGTM

0xcadams:
Another unexpected value here: `428e018b-c417-47b5-bcf3-91c94558b962` seen at 2026-08-27T17:08:01.134819202Z. See https://rocicorp.slack.com/archives/C0BU2PLMVFA

PR #6426, opened 2026-08-27
chore(zero-cache): Run PRAGMA optimize instead of analyze

darkgnotic:
PRAGMA analyze blocks the process for a long time on large dbs, and which ends up disabling other background logic (liek the purge locker) during server startup.

Use PRAGMA optimize instead, as is recommended by the official SQLite docs.

Context:
* https://rocicorp.slack.com/archives/C0BSRF00ETZ/p1787851785251979

Miscellaneous:
* The replicator worker also runs `PRAGMA optimize = 0x10002` at startup, so that step is removed from the non-migration path.

### DIFF
--- a/packages/zero-cache/src/db/migration-lite.ts
+++ b/packages/zero-cache/src/db/migration-lite.ts
@@ -143,21 +143,26 @@ export async function runSchemaMigrations(
         }
       }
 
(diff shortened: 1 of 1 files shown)


0xcadams (commented):
LGTM!

PR #6441, opened 2026-08-29
package graph generator to understand dependencies

tantaman:
the tldr is: I have a tough time trusting LLM changes given we have no strict module boundaries. A LLM puts up a diff and it changes zero-cache where I only expected zero-client changes, for example.

Idk to what extent the lack of bounds cause problems for LLMs themselves but I do always have less success with an LLM in the Zero repo than I do in other repos.

Some red flags this shows:

CleanShot 2026-08-29 at 10 57 23@2x

Zero-server should not:
- depend on zero-cache
- depend on zqlite

CleanShot 2026-08-29 at 10 58 46@2x

Zero-client should not depend on:
- zero-cache
- datadog


0xcadams:
This is cool. Can we do a lint rule? Or just fix our package boundaries once and for all....

notes.md
arv, line 1: remove?

arv:
This is nice. I didn't look at the code in detail. The idea is great. Can we make it part of CI to ensure that we did not break the layering.

This is related to the validate-deps npm script which checks that imports matches package.json and that there are no cycles between packages.

Copilot seems to have some good feedback too.

PR #6445, opened 2026-08-30
fix(zero-cache): prevent pg client corruption/blackholing by isolating connections

darkgnotic:
Fixes the mysterious hung pg query issue that occasionally causes the change-streamer to stop making progress.

This was originally theorized to be caused by TCP/connection issues (e.g. half-closed connections), but incidences continued to happen even with socket-level watchdog logic that should detect it (https://github.com/rocicorp/mono/pull/6220 and https://github.com/rocicorp/mono/pull/6221).

A recent hang detected by the application-level watchdog (https://github.com/rocicorp/mono/pull/6419) was accompanied by corrupted change-db state that should not have been possible given transactional atomicity semantics. 

Application and db logs identified the cause to be a single pg "backend", in which a partial transaction was in progress, being reused for a _second_ application-level "connection", causing the transaction of the latter to commit its statements along with the partial transaction of the former. 

Specifically, in zero-cache semantics, an in-progress change-log multi-INSERT transaction was hijacked by a purge transaction, resulting in a Frankenstein commit of the purge and the unfinished INSERT.

The postgres driver issue has been filed here: https://github.com/porsager/postgres/issues/1204

The workaround in the zero-cache is to use isolated connection pools (i.e. pg clients) for concurrent storer logic to make it impossible for connections to be swapped between its transactions.

### Related

Dynamic connection pool creation also allows safer catchup of multiple subscribers in parallel (i.e. without exhausting a fixed connection pool), which is included in this change.

### DIFF
--- a/packages/zero-cache/src/config/zero-config.test.ts
+++ b/packages/zero-cache/src/config/zero-config.test.ts
@@ -295,12 +295,6 @@ test('zero-cache --help', () => {
                                                                                    to sync multiple view-syncers without requiring multiple replication slots on                                              
                                                                                    the upstream database. If unspecified, the upstream-db will be used.                                                       
                                                                                                                                                                                                               
-     --change-max-conns number                                                     default: 5                                                                                                                 
-       ZERO_CHANGE_MAX_CONNS env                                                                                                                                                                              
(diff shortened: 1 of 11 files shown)


arv:
LGTM

PR #6454, opened 2026-09-01
fix(backfill): reset pipelines when backfills begin

darkgnotic:
Remove an incorrect optimization in the original backfill implementation (https://github.com/rocicorp/mono/pull/5518) where pipeline resets were skipped when backfilling tables are created.

Pipelines should still be reset to refresh the schema; otherwise, subsequent DML statements for the newly created (but backfilling) tables will be unrecognized by the pipeline and throw an error.

Context:
* https://rocicorp.slack.com/archives/C0BU1HX8ELA/p1788236811691309


arv (commented):
LGTM

PR #6459, opened 2026-09-01
fix(backfill): prevent non-shard lag reports from interrupting backfills

darkgnotic:
Fix a situation in which a backfill gets unnecessarily interrupted by a lag report from a different shard (e.g. `ZERO_APP_ID`), causing it to idle until a new transaction arrives.

The error stemmed from the fact that the non-shard lag report, although ignored, was improperly classified as a "transactional" message, thereby initiating a change-stream reservation unnecessarily. With no other changes in the stream, that reservation was held until new change messages came in.

With the fix, the non-shard lag reports are properly recognized as non-transactional, avoiding the spurious and problematic change-stream reservation.

Context:
* https://rocicorp.slack.com/archives/C0B3A4P3W56/p1788294436747829?thread_ts=1788279391.627549&cid=C0B3A4P3W56


arv (commented):
LGTM

PR #6468, opened 2026-09-02
fix(zero-client): carry the connect abort reason on the attempt, not only signal.reason

pawarren:
## The bug

On runtimes whose `AbortController` predates the 2021 `signal.reason` spec addition, the connection run loop converts **every retryable connect failure into the permanent `error` pause**, so Zero's own retry/backoff never runs.

React Native is the big affected runtime: RN bundles the `abort-controller@3` polyfill, whose `abort(reason)` silently discards the reason and whose signal has no `reason` property at all.

The mechanism:

1. `#disconnect(lc, reason)` passes a typed, retryable `ZeroError` (e.g. `ClientError({kind: ConnectTimeout})`) to `attempt.controller.abort(reason)`.
2. The run loop reads it back exclusively off the signal — the `abortHandler` rejects with `attempt.controller.signal.reason`, and `#connectAttempt` has three `throw signal.reason` sites.
3. On a reason-less `AbortController`, all four sites read `undefined`. The run loop's catch wraps it — `isZeroError(undefined)` is false — as `ClientError({kind: Internal})`, which is non-retryable, and the run loop pauses in the `error` state forever ("Run loop paused in error state", with `undefined` as the logged reason).

We hit this in production with our React Native app (Margins): on a mid-range Android device, cold boots that hit a legitimate `ConnectTimeout` during a large initial hydration wedged permanently instead of retrying, and an airplane-mode boot produced several fatal pauses within seconds. With a spec-compliant AbortController polyfill patched in downstream, the wedge disappears and the run loop retries itself to `Connected` — device-verified against production, which is how we isolated the cause to exactly this read.

## The fix

Record the reason on the `ConnectAttemptControl` in `#disconnect`, and prefer it at the four read sites with `attempt.abortReason ?? signal.reason`. Behavior on spec-compliant platforms is unchanged (the two values are identical there); on legacy runtimes the typed error now survives the abort and the existing retry/backoff classification handles it as designed.

This also makes the run loop robust independent of platform `AbortController` fidelity, rather than relying on a polyfill detail for correctness of the retry path.

## Test

New test in `zero.test.ts`, modeled on the existing `Connect timeout` test: stubs `AbortController` with a subclass whose `abort()` drops the reason argument (the legacy polyfill's behavior), times out a connect attempt, and asserts the run loop schedules another attempt rather than pausing.

- Red on the previous code: `expected 'error' to be 'connecting'` — the permanent pause.
- Green with the fix, and the retried attempt proceeds to `Connected`.

`zero.test.ts`, `zero.connection.test.ts`, `connection-manager.test.ts`, `connection.test.ts`: 196/196 passing; `tsc` and `oxlint` clean.

https://claude.ai/code/session_018RpmRy5DquXDnZ325PEphZ

### DIFF
--- a/packages/zero-client/src/client/zero.test.ts
+++ b/packages/zero-client/src/client/zero.test.ts
@@ -2843,6 +2843,55 @@ test('Connect timeout', async () => {
   connectionStatusCleanup();
 });
 
+test('connect timeout retries when AbortController drops abort reasons', async () => {
+  // React Native's bundled `abort-controller@3` polyfill predates the 2021
+  // `signal.reason` spec addition: `abort(reason)` silently discards the
+  // reason and `signal.reason` reads `undefined`. The run loop used to read
+  // the abort reason exclusively off the signal, so on such runtimes every
+  // retryable connect failure surfaced as `undefined`, was wrapped as a
+  // non-retryable internal error, and permanently paused the run loop.
+  // The polyfill leaves `signal.reason` undefined; a spec `abort()` with no
+  // argument leaves a generic AbortError. Either way the typed retryable
+  // error passed to `abort(reason)` is lost, which is the mechanism under
+  // test. Subclassing keeps `signal` a real AbortSignal so DOM listeners
+  // that take `{signal}` still work.
+  class LegacyAbortController extends AbortController {
+    override abort(): void {
+      super.abort();
+    }
+  }
+  vi.stubGlobal('AbortController', LegacyAbortController);
+
+  try {
+    const z = zeroForTest({logLevel: 'debug'});
+    await z.waitForConnectionStatus(ConnectionStatus.Connecting);
+    const firstSocket = await z.socket;
+    for (let i = 0; i < 10; i++) {
+      await vi.advanceTimersByTimeAsync(0);
+    }
+
+    // Time out the first connect attempt.
+    await vi.advanceTimersByTimeAsync(CONNECT_TIMEOUT_MS);
+    expectLogMessages(z).contain(connectTimeoutMessage);
+
+    // The timeout is retryable: the run loop must schedule another attempt
+    // rather than pausing in the error state.
+    expect(z.connectionStatus).toBe(ConnectionStatus.Connecting);
+    const nextSocketPromise = z.socket;
+    await vi.advanceTimersByTimeAsync(RUN_LOOP_INTERVAL_MS);
+    for (let i = 0; i < 10; i++) {
+      await vi.advanceTimersByTimeAsync(0);
+    }
+    expect(await nextSocketPromise).not.equal(firstSocket);
+
+    // And the retried attempt can then succeed.
+    await z.triggerConnected();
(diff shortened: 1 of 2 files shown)


pawarren:
Some supporting context from the production fleet where we found this, plus two standalone repros.

## The defect, as we hit it

React Native polyfills `AbortController` with `abort-controller@3.0.0` (`Libraries/Core/setUpXHR.js`) — last released 2019, before `signal.reason` existed. Zero's connection loop does `controller.abort(reason)` and later `throw signal.reason` (zero.js ~892/896/913/922/1120 in 1.9.0). On that polyfill the reason is dropped, so **every aborted connect attempt throws `undefined`**, which the run loop wraps as `ClientError { kind: Internal }` and treats as fatal: `"Run loop paused in error state"` — no backoff, no retry, however retryable the original condition was. Plain airplane mode parks the client permanently in ~200ms.

Environment: Zero 1.9.0, release builds on a mid-range Android device (Samsung SM-A075M, Android 16), production backend, 14 days of fleet logs. Verified by intervention: patching the polyfill on-device made the failure disappear.

## Repro A — the polyfill gap (no Zero required)

```js
// npm i abort-controller@3.0.0    (the exact package RN bundles)
const { AbortController } = require("abort-controller");
const controller = new AbortController();
controller.abort(new Error("ConnectTimeout")); // reason silently dropped
console.log(controller.signal.aborted); // true
console.log(controller.signal.reason);  // undefined  <- what Zero throws
// Node's native AbortController prints `Error: ConnectTimeout` instead.
```

## Repro B — the run-loop consequence

```js
// npm i @rocicorp/zero abort-controller@3.0.0
const AC = require("abort-controller");
globalThis.AbortController = AC.AbortController; // what RN setUpXHR.js does
globalThis.AbortSignal = AC.AbortSignal;

const { Zero } = require("@rocicorp/zero");
const z = new Zero({
  server: "http://127.0.0.1:9", // guaranteed connect failure
  userID: "repro",
  schema: /* any minimal schema */,
  kvStore: "mem",
});
// Polyfilled: one failed attempt, then the permanent
//   "Run loop paused in error state".
// Native (delete the two globalThis lines): retries on backoff
//   indefinitely, as designed.
```

## Fleet impact and verification

Before the fix: 33/71 connects took >15s over 7 days across 4/6 users; `totalTimeToConnect` p50 11.3s, p90 247s, max 94min; one account hit 4 permanent pauses in 14 days. With a spec-compliant polyfill patched in: backoff retries run (~9s, reconnecting), airplane-mode boots stop wedging, zero fatal pauses across the session.

## Why fix it in Zero rather than the polyfill

RN's bundled polyfill has been unmaintained for years (flagged at facebook/react-native#50015), so carrying the abort reason on the attempt object — as this PR does — removes the dependency on host polyfill compliance entirely and works everywhere today. We're shipping a local patch making `abort-controller@3` spec-compliant as a bridge; the day Zero stops reading `signal.reason`, we delete it.

arv (approved):
LGTM

arv:
Closed in favor of #6485

PR #6469, opened 2026-09-03
perf(zql): intern Query objects with a V8-style transition tree

arv:
Building the same query twice allocated two structurally identical objects, so `#hash` -- and every other per-instance memo -- never hit. `useQuery` rebuilds its whole chain on every render, so it paid for a fresh AST and a fresh hash each time.

Each node now keys the queries derived from it by the operation that produced them, the way V8 keys hidden-class transitions, so applying the same operation twice returns the same object and no AST is built at all on the second pass. A transition is identified by a memoized key naming the operation, the varying argument as a `Map` key rather than concatenated into a string, and -- only where those two do not settle it -- a small delta compared with `deepEqual`. Never the whole AST: the parent is already canonical, so two of its children are equal exactly when the operations that produced them are.

Children hold their parent strongly and parents hold children weakly, so holding a query holds its ancestor spine and re-deriving it gives the same object back. The first transition out of a node is held strongly instead, in fields: `new WeakRef` costs more than everything else a transition does, and strong first-children form a path rather than a tree, so what is retained is bounded by the depth of a builder chain. Relationship bases get a separate strong store, bounded by the schema. Nothing keyed by data is ever held strongly. No `FinalizationRegistry`: cleared entries are dropped by the lookup that walks past them and by a sweep once a node's weak store has doubled.

Interning is by construction path, so `q.where(a).where(b)` and `q.where(b).where(a)` are still distinct instances here; the next PR in the stack folds those together.

Measured in plain node, per query. "Warm" is a rebuild of a chain that already exists, which is what every render does; "cold" is a new value under a warm root, the weak path.

| | before | after |
|---|---|---|
| warm chain, build + hash | 2564 ns | 446 ns |
| warm 200-row list, build + hash | 124,753 ns | 12,266 ns |
| warm chain, build only | 982 ns | 447 ns |
| cold row, build only | 124 ns | 350 ns |
| cold row, build + hash | 636 ns | 872 ns |
| fresh root chain, build + hash | 1509 ns | 1332 ns |

Cold numbers are with the event loop yielding and GC running; a synchronous loop overstates the weak path, because `new WeakRef` keeps its target alive for the rest of the job.

Stacked on #6464.

### DIFF
--- a/packages/zql/src/query/query-impl.ts
+++ b/packages/zql/src/query/query-impl.ts
@@ -1,6 +1,9 @@
 // oxlint-disable no-explicit-any
 import {assert} from '../../../shared/src/asserts.ts';
-import type {ReadonlyJSONValue} from '../../../shared/src/json.ts';
+import type {
+  ReadonlyJSONObject,
+  ReadonlyJSONValue,
+} from '../../../shared/src/json.ts';
 import {
   type AST,
   type CompoundKey,
@@ -634,16 +637,11 @@ export class QueryImpl<
     if (cond.type === 'simple' && cond.left.type === 'column') {
       key = whereKey(cond.left.name, cond.op);
       const right = cond.right;
-      if (right.type === 'literal' && isPrimitive(right.value)) {
-        tValue = right.value;
-        delta = undefined;
-      } else {
-        const tag = rightTag(right);
-        if (tag !== undefined) {
-          tValue = tag;
-          delta = undefined;
-        }
-      }
+      tValue =
+        right.type === 'literal' && isPrimitive(right.value)
+          ? right.value
+          : rightTag(right);
+      delta = undefined;
     } else {
       // A sub-query an `exists` correlates to is itself interned, so its
       // identity settles that part of the tree without walking it.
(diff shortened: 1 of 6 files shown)


packages/zql/src/query/query-impl.ts
```diff
@@ -148,6 +214,40 @@ export class QueryImpl<
...
+   */
+  #expressionBuilder: ExpressionBuilder<TTable, TSchema> | undefined;
+
+  /**
+   * The query this one was derived from, or `undefined` for a root query.
+   *
+   * @internal Exposed for tests and debugging. The field behind it exists to
+   * keep the ancestor spine alive, not to be navigated.
+   */
+  get derivedFrom(): QueryImpl<any, any, any> | undefined {
```
arv: remove

packages/zql/src/query/query-impl.ts
```diff
@@ -148,6 +214,40 @@ export class QueryImpl<
...
+    return this.#parent;
+  }
+
+  /** @internal Exposed so tests can assert what is retained strongly. */
+  get transitionsForTesting():
```
arv: remove

packages/zql/src/query/query-impl.ts
```diff
@@ -607,6 +883,173 @@ export function asQueryImpl<
...
+    }
+    case 'and':
+    case 'or': {
+      let out = `${cond.type === 'and' ? 'A' : 'O'}${cond.conditions.length}:`;
+      for (const c of cond.conditions) {
+        const k = condKey(c);
+        if (k === undefined) {
+          return undefined;
+        }
+        out += k;
```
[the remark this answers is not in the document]
arv: Fixed in a2edb5ea6: string values are now length-prefixed and numbers terminated, and arrays/parameter references are encoded recursively with the same self-delimiting tags instead of `JSON.stringify` (which also wrote `Infinity`, `NaN` and `null` alike). Added your two-tree collision as a regression test, plus the JSON-lossy cases. Merged up through #6470 and #6471.

packages/zql/src/query/query-impl.ts
```diff
@@ -221,36 +385,39 @@ export class QueryImpl<
...
-    this.#newQuery(
-      this.#tableName,
-      {...this.#ast, limit: 1},
-      {
-        ...this.format,
-        singular: true,
-      },
-      this.customQueryID,
-      this.#currentJunction,
+  one(): Query<TTable, TSchema, TReturn | undefined> {
```
arv, line 425: why was this changed to a prototype method. Or put differently, why was it an instance method before?

Is this change safe?

packages/zql/src/query/query-impl.ts
```diff
@@ -607,6 +881,209 @@ export function asQueryImpl<
...
+let nextASTID = 0;
+
+function astID(ast: AST): number {
+  let id = astIDs.get(ast);
+  if (id === undefined) {
+    id = ++nextASTID;
+    astIDs.set(ast, id);
+  }
+  return id;
+}
```
arv, line 947: would it make sense to store the AST hash here instead? Then we could maybe reuse that later when we compute the final hash?

packages/zql/src/query/query-impl.ts
arv: Removed in c0a96667a, along with the one test that read it. The `#parent` field stays, since it is what keeps a query's ancestor spine alive; only the getter is gone. That test asserted the shape of the parent chain rather than anything observable, and what the field is for is covered by the tests that rebuild a chain and expect the same instances back.

packages/zql/src/query/query-impl.ts
arv: `transitionsForTesting` is gone in 96fb970e3, together with `Transitions.first`/`rest`/`restSize` and (in #6470) `hashIndexForTesting` and `HashIndex.map`. Those tests now stub `WeakRef` and count constructions and derefs instead, which gets at the retention properties from outside rather than by reading fields.

packages/zql/src/query/query-impl.ts
arv: Nothing in the interning needs it, so I can revert it if you'd rather keep the diff minimal. Why I did it: each instance allocated one closure per builder method, and construction is on the cold path this PR is trying not to regress. Isolated, 8 arrow fields vs prototype methods measures 23ns vs 10ns per construction in plain node.

Why they were instance fields before, I don't know. Git blame puts them at #5315 (Remove AbstractQuery), so they may be a leftover from when the hierarchy needed bound methods. The one with a stated reason was `where`, which was `function(){}.bind(this)` because it reads `arguments.length` and an arrow has no `arguments`. As a normal method that still works, and `#exists` keeps its binding explicitly where it is handed to `ExpressionBuilder`.

On safety, what I checked: the `Query` interface declares all of these with method syntax, so a prototype method satisfies it and the variance is unchanged; `RunnableQueryImpl` overrides none of them; and nothing in the repo detaches one from its receiver.

The one real behavior change is for callers outside the repo: `const {where} = q` or `arr.map(q.where)` used to work and now loses `this`. That is a public interface, so it is your call whether that is worth 13ns a construction.

packages/zql/src/query/query-impl.ts
arv: Tempting, but the reuse half does not work with the hash as it stands, and without the reuse it is strictly more expensive.

The hash is a streaming fold: `visitAST` mixes words into two module-level lanes and `finalize` combines them at the end. There is no way to splice a precomputed sub-hash into that stream and land on the same value you would get by walking the sub-tree inline, so a parent could not reuse a child's digest without changing what the hash produces.

That matters because `hash()` is not the only walk. The server's query ID for a non-custom query is `hashOfAST(qi.ast)` (query-delegate-base.ts:354), which has to keep walking the whole thing. So a compositional client hash would not help that path, and we would be maintaining two different foldings of the same AST.

Meanwhile the id is doing less work than a hash would. `astID` is a WeakMap get plus, on a miss, a set; storing a hash means walking the entire sub-AST at build time, on the cold path, for a value only used to tell two sub-queries apart. It is only assigned at the three sub-query sites rather than for every query, and that was already one of the larger cold-build costs.

Where the idea would genuinely pay is making `hashOfAST` itself Merkle-style, folding each sub-query's digest rather than re-walking it. That turns hashing a chain from O(n²) to O(n) if you hash at every step, and the server could reuse it too. But it changes every query ID on the wire, so it is its own change with a migration, not something to fold in here. Happy to prototype it separately if you want to see the numbers.

PR #6470, opened 2026-09-03
perf(zql): fold queries built by different paths into one instance

arv:
The transition tree interns by construction path, so `q.where(a).where(b)` and `q.where(b).where(a)` were distinct instances even though `normalizeAST` sorts conditions and `related` entries and makes their ASTs byte-identical.

Each root now keeps a hash index of the queries hashed beneath it. When a query is hashed and an equal one is already indexed, the transition that produced it is re-pointed at the existing instance, so the next rebuild along that path yields the shared object. The query being hashed keeps its identity, since it has already been handed out.

This runs when a query is *hashed*, not when it is built, so a chain nobody hashes -- a mutator's, or a server-side query built once per request -- pays nothing, and a hashed one pays a map lookup on top of a hash it was computing anyway. Content is verified with `deepEqual` before a path is redirected, so a hash collision costs a missed unification and nothing else. The index lives on the root, which is what scopes it: two roots never share one, so the schema and the delegate, neither of which the hash covers, can never be conflated.

Measured in plain node with the event loop yielding and GC running, per query, against the tree alone:

| | tree | tree + index |
|---|---|---|
| cold row, build + hash | 872 ns | 1053 ns |
| cold row, build only | 350 ns | 354 ns |
| warm chain, build + hash | 42 ns | 44 ns |
| fresh root chain, build + hash | 1332 ns | 1527 ns |

Stacked on #6469.

### DIFF
--- a/packages/zql/src/query/query-impl.ts
+++ b/packages/zql/src/query/query-impl.ts
@@ -259,11 +259,6 @@ export class QueryImpl<
     return this.#transitions;
   }
 
-  /** @internal Exposed for tests. Only a root has one. */
-  get hashIndexForTesting(): HashIndex<QueryImpl<any, any, any>> | undefined {
-    return this.#byHash;
-  }
-
   constructor(
     schema: TSchema,
     tableName: TTable,
--- a/packages/zql/src/query/query-impl.ts
+++ b/packages/zql/src/query/query-impl.ts
@@ -1,6 +1,6 @@
 // oxlint-disable no-explicit-any
 import {assert} from '../../../shared/src/asserts.ts';
-import type {ReadonlyJSONValue} from '../../../shared/src/json.ts';
+import {deepEqual, type ReadonlyJSONValue} from '../../../shared/src/json.ts';
 import {
   type AST,
   type CompoundKey,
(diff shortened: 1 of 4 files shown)


packages/zql/src/query/query-impl.ts
```diff
@@ -251,6 +259,11 @@ export class QueryImpl<
     return this.#transitions;
   }
 
+  /** @internal Exposed for tests. Only a root has one. */
+  get hashIndexForTesting(): HashIndex<QueryImpl<any, any, any>> | undefined {
```
arv: remove things that are public and only for testing

arv: Removed: `hashIndexForTesting` and `HashIndex.map` are gone here, and `transitionsForTesting` plus `Transitions.first`/`rest`/`restSize` are gone in #6469. The tests now stub `WeakRef` instead, so they collect a value on demand and count constructions/derefs. That asserts the properties more directly than the fields did: held strongly means no WeakRef is made and it survives collection, and a lookup stopping at its match derefs fewer entries than one walking a whole bucket. Dropped the two assertions that only described internal shape.

PR #6471, opened 2026-09-03
perf(zql): key `start` transitions by the row

arv:
Every `start` off one node landed in a single bucket, keyed by nothing and compared by `deepEqual` on the row, so paging to a row never seen before scanned every live sibling page -- up to `MAX_SCAN` deep comparisons -- before building anything. On a warm root that made a cold page ten times slower than not interning at all.

The row is now folded into the transition value, so a new page is one map lookup. Every key and value in the encoding is self-delimiting -- strings are length-prefixed and numbers terminated -- so two rows can only encode alike if they are the same row. The value tags the `where` path already used are tightened the same way, since `condKey` concatenates them.

Measured in plain node with the event loop yielding and GC running, a new page under a warm chain, per query:

| | before | after | uninterned |
|---|---|---|---|
| build only | 2496 ns | 350 ns | 241 ns |
| build + hash | 3351 ns | 1103 ns | 872 ns |

Stacked on #6470.

### DIFF
--- a/packages/zql/src/query/query-impl.ts
+++ b/packages/zql/src/query/query-impl.ts
@@ -731,11 +731,16 @@ export class QueryImpl<
     row: Partial<Record<string, ReadonlyJSONValue | undefined>>,
     opts?: {inclusive: boolean},
   ): Query<TTable, TSchema, TReturn> {
-    // The row is folded into the transition value so that paging to a row
-    // never seen before is one lookup, not a scan of every sibling row's
-    // delta. Key order is part of the encoding: a row spelled in another order
-    // interns separately, and the hash index folds the two together once they
-    // are hashed.
+    // Interned like every other builder method (see `#derive`): if this query
+    // has already been paged from this row, the same query object comes back.
+    // The row is encoded into a string and looked up in a `Map`, so paging to
+    // a row never seen before costs one lookup. The alternative, keeping the
+    // row object and comparing it with `deepEqual` against every row already
+    // paged from here, made a new page slower than not interning at all.
+    //
+    // The encoding keeps property order, so `{a, b}` and `{b, a}` are looked
+    // up separately. They still end up as one object once either is hashed,
+    // since their ASTs are equal; see `#canonicalize`.
     return this.#derive(
       opts?.inclusive ? 'start:inclusive' : 'start:exclusive',
       valueTag(row as ReadonlyJSONValue),
--- a/packages/zql/src/query/query-impl.ts
+++ b/packages/zql/src/query/query-impl.ts
@@ -692,16 +692,11 @@ export class QueryImpl<
     if (cond.type === 'simple' && cond.left.type === 'column') {
       key = whereKey(cond.left.name, cond.op);
       const right = cond.right;
-      if (right.type === 'literal' && isPrimitive(right.value)) {
-        tValue = right.value;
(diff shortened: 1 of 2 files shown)


packages/zql/src/query/query-impl.ts
```diff
@@ -1006,31 +1006,56 @@ function isPrimitive(v: LiteralValue): v is string | number | boolean | null {
...
+function valueTag(v: ReadonlyJSONValue): string {
+  switch (typeof v) {
+    case 'string':
+      return `:s${enc(v)}`;
+    case 'number':
+      return `:n${v};`;
+    case 'boolean':
+      return v ? ':t' : ':f';
+    default:
+      return v === null ? ':z' : `:a${enc(JSON.stringify(v))}`;
```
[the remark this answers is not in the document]
arv: Fixed in cbde72403: arrays and objects are now encoded recursively with the same self-delimiting tags (count-prefixed, numbers written via `String`), so `Infinity`, `NaN` and `null` no longer collide; `JSON.stringify` is gone from the encoding. Added tests for nested Infinity/NaN/null in both `start` rows and `IN` lists, asserting the interned query's AST keeps the original value.

packages/zql/src/query/query-impl.ts
```diff
@@ -731,10 +731,15 @@ export class QueryImpl<
     row: Partial<Record<string, ReadonlyJSONValue | undefined>>,
     opts?: {inclusive: boolean},
   ): Query<TTable, TSchema, TReturn> {
+    // The row is folded into the transition value so that paging to a row
```
aboodman: I don't understand this comment. "transition value"? "delta"? i don't know these terms.

arv: Rewrote it in 80fbf3ca8 without the jargon: it now says the row is encoded to a string and looked up in a Map so a new page is one lookup, that the alternative (deepEqual against every row already paged from here) was slower than not interning, and that property order matters for the lookup but not for the final identity. The terms came from `query-transitions.ts` / `#derive`, which the comment now points at.

packages/zql/src/query/query-impl.ts
```diff
@@ -731,10 +731,20 @@ export class QueryImpl<
     row: Partial<Record<string, ReadonlyJSONValue | undefined>>,
     opts?: {inclusive: boolean},
   ): Query<TTable, TSchema, TReturn> {
+    // Interned like every other builder method (see `#derive`): if this query
```
aboodman: Why does this instance of the pattern get this long comment explaining it? Isn't it the same in every call site of #derive?

aboodman: Is there osmething different abt what's happening here than in other builder methods?

arv: No, nothing different: it goes through `#derive` like every other builder method. The only thing specific to `start` is that its argument is an object, so it has to be encoded to a string to be a Map key. Cut the comment down to that in c27e6c753; the general explanation lives on `#derive`.

arv: No, nothing different: it goes through `#derive` like every other builder method. The only thing specific to `start` is that its argument is an object, so it has to be encoded to a string to be a Map key. Cut the comment down to that in c27e6c753; the general explanation lives on `#derive`.

PR #6472, opened 2026-09-03
perf(zql): Defer hydration until source is ready.

arv:
A common case (especially on React Native) is that `useQuery` gets called
before `Replicache.init` has completed. Therefore the view gets materialized
and registered before the source has any data.

This leads to us pushing every row from the data source through every
already-connected pipeline instead of pulling only the desired rows once
the data is there. The push path is also superlinear: doubling the rows
cost 4.7x the time on Hermes.

`materialize()` stays synchronous. When the delegate reports
`pipelinesReady === false`, the view is created over a `DeferredInput`
placeholder and the server query is registered as before. `ZeroRep.init`
loads the replica into the sources with no pipelines connected, then calls
`ZeroContext.markPipelinesReady()`, which builds each pending pipeline in
materialization order and hydrates it with a single `fetch({})`, delivering
the fetched rows to the existing view as add changes. A view reports
`complete` only once it is both hydrated and the server has sent got.

21 queries materialized before the replica loads, Android emulator
(Hermes, Expo SDK 57) and Node 24 (V8):

| Rows | Engine | Before    | After    | Speedup |
| ---- | ------ | --------: | -------: | ------: |
| 40k  | Hermes | 7,275 ms  | 1,473 ms | 4.9x    |
| 80k  | Hermes | 33,981 ms | 3,075 ms | 11x     |
| 40k  | V8     | 549 ms    | 120 ms   | 4.6x    |
| 80k  | V8     | 1,140 ms  | 235 ms   | 4.9x    |

`zql-benchmarks` `cold-boot-hydration` (13 queries, 30k rows, V8):

| Scenario                           | Time    |
| ---------------------------------- | ------: |
| Live pipelines (before)            | 380 ms  |
| Deferred pipelines (after)         | 60.6 ms |
| Load then materialize (lower bound)| 60.7 ms |

### DIFF
--- a/packages/zql/src/ivm/deferred-input.ts
+++ b/packages/zql/src/ivm/deferred-input.ts
@@ -1,0 +1,78 @@
+import {assert} from '../../../shared/src/asserts.ts';
+import {makeAddChange} from './change.ts';
+import type {Node} from './data.ts';
+import type {FetchRequest, Input, Output} from './operator.ts';
+import type {SourceSchema} from './schema.ts';
+import {consume, type Stream} from './stream.ts';
+
+/**
+ * A placeholder `Input` for a view whose pipeline cannot be built yet.
+ *
+ * Until {@link attach} is called it behaves like an empty pipeline: `fetch`
+ * yields nothing and pushes never happen. When the real pipeline is attached
+ * the view is hydrated by pushing every row of the pipeline's initial fetch
+ * through the normal `Output.push` path, so the view sees the data exactly as
+ * it would see rows arriving incrementally.
+ *
+ * This lets `materialize()` stay synchronous and return a view immediately
+ * while the expensive pipeline construction and hydration are postponed, for
+ * example until the client's replica has been loaded into the IVM sources.
+ */
+export class DeferredInput implements Input {
+  #output: Output | undefined;
+  #input: Input | undefined;
+  #destroyed = false;
+
+  get attached(): boolean {
+    return this.#input !== undefined;
+  }
+
+  get destroyed(): boolean {
+    return this.#destroyed;
+  }
+
+  setOutput(output: Output): void {
+    this.#output = output;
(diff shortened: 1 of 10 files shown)


tantaman (commented):
I have memories of solving this problem on web once or twice. Do we know why this wasn't an issue on web? That's the only concern I have, otherwise it looks good.

tantaman:
Idk if this would ever happen in practice but you can technically get a failed pipeline by calling `getSchema` before `markReady`:


```ts
test('a schema-triggered pipeline build is still hydrated when pipelines become ready', () => {
  const {delegate, addIssue} = newDelegate();
  addIssue('i1');

  const view = delegate.materialize(
    newQuery(schema, 'issue'),
    (
      _q,
      input,
      format,
      onDestroy,
      onTransactionCommit,
      queryComplete,
      updateTTL,
    ) => {
      // ArrayView fetches before reading the schema. Because the input is still
      // deferred, that fetch is empty. Reading the schema afterwards builds the
      // real pipeline, but must not make the later hydration a no-op.
      const v = new ArrayView(input, format, queryComplete, updateTTL);
      expect(input.getSchema().tableName).toBe('issue');
      v.onDestroy = onDestroy;
      onTransactionCommit(() => v.flush());
      return v;
    },
  );

  expect(view.data).toEqual([]);
  delegate.markReady();
  // This throws / `view.data` is empty!
  expect(view.data).toMatchObject([{id: 'i1'}]);

  view.destroy();
});
```

packages/zql/src/ivm/deferred-input.ts
```diff
@@ -37,10 +47,18 @@ export class DeferredInput implements Input {
   }
 
   getSchema(): SourceSchema {
```
arv, line 52: How about passing in the schema to the constructor. We always have the schema.

PR #6475, opened 2026-09-03
fix(zero-client): measure connect setup and server acknowledgement separately

arv:
## The defect

`CONNECT_TIMEOUT_MS` covered an entire connection attempt: reading the cookie, the client group ID and the active clients from IDB, preparing and opening the socket, and only then waiting for the server to acknowledge.

Those fail for unrelated reasons. Sharing one budget means local setup spends the server's time — on a slow cold boot the socket may not exist until most of the 10s is gone, and a perfectly healthy server is then reported as unreachable. The failure is indistinguishable from a real network problem, and `Connection attempt timed out after 10 seconds` is what you get either way.

## The change

Each phase gets its own budget, armed and cancelled at the transition between them.

- **setup** keeps a deadline, so a stuck setup still times out and retries rather than hanging until the 60s disconnect watchdog.
- **ack** starts a fresh budget once setup completes, so what the server is given no longer depends on how slow the disk was.

`#armConnectTimeout(lc, attempt, phase)` builds either one; `ConnectAttemptControl.clearTimeout` holds whichever is currently running and is swapped at the handover, so the run loop's `finally` still has exactly one thing to cancel.

The messages name the phase that expired — `Connection setup timed out after 10 seconds` versus `Server did not acknowledge the connection within 10 seconds`. The kind stays `ConnectTimeout`; a distinct kind per phase would be a public taxonomy change, so it isn't in this PR.

**Trade-off:** an attempt that fails in both phases can now take two budgets rather than one. That is the cost of not having to choose which of the two failures the single budget is for.

## Why not simply move the deadline to socket creation

That was the first attempt, and it broke `connect timeout during setup retries without an unhandled rejection`, which stubs `ActiveClientsManager.create` to hang forever and asserts the attempt times out and retries. Moving the deadline leaves a stuck setup unbounded until the 60s watchdog, with no retry in between — a real regression that the existing test was there to catch. Splitting keeps that bound and still stops setup from spending the server's time.

## Setup timing

The two pieces of local setup most likely to be slow are timed, each warning above 50ms. Both are measured **where the work runs**, not where an attempt awaits it — both start earlier and in parallel, so an await-site number would usually read as zero.

### Opening the database

Instrumented in Replicache's `#open`, warning above 50ms.

The timer stops at the head write, immediately **before** `await this.#zero?.init(headHash, this.memdag)`, so it measures getting to a usable database and excludes loading the replica into Zero's IVM sources — which is separate work and usually the larger of the two.

This has to be instrumented inside `#open`. Measuring it from Zero cannot work: `resolveReady()` is called *after* `#zero.init`, so nothing Zero can await — the cookie included — resolves until hydration has finished, and any number taken from that side reports a slow hydration as a slow disk.

The window starts at the top of `#open`, which includes waiting out a closing instance of the same name. That is rare, but it does delay the database being usable, so it is counted rather than hidden.

### Creating the ActiveClientsManager

This runs at construction, so by the time `#connectAttempt` reaches `await this.#activeClientsManager` the wait is usually already over — timing that await would measure nothing. It is timed around `ActiveClientsManager.create` instead, starting after the client group ID resolves, since waiting for that is the database opening and is already timed on its own.

The log names the lock backend. `navigator.locks` does not exist on React Native, so `getClientLockManager` falls back to `MockClientLockManager`, an in-process stand-in that only ever sees the current client. That changes both what the timing means — no real lock acquisition or `locks.query()` — and how much the resulting client list can be trusted, so a number without the backend beside it would be misleading.

## Tests

`slow setup does not spend the server acknowledgement budget` covers the behavior change directly: setup finishing at 9s, then a full budget for the server, ending in a successful connect. Under the old single budget that connect failed. The existing stuck-setup retry test is unchanged and still passes, which is the evidence that the setup bound survived.

There is no test on the log line — making `#open` deterministically cross 50ms is not worth the fixture.

`zero-client` 650 pass, `replicache` 798 pass; `format`, `lint` and `check-types` clean on both. `zero-idb.test.ts > logged-out client uses a private storage sentinel for idb naming` fails on this branch and fails identically on clean `main`, so CI will be red for a reason unrelated to this change.

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -219,6 +219,11 @@ type ConnectAttemptControl = {
   readonly controller: AbortController;
   readonly connected: Resolver<void>;
   readonly events: [date: Date, event: string][];
+  /**
+   * Cancels whichever phase deadline is currently running. Swapped when the
+   * attempt stops setting up locally and starts waiting for the server.
+   */
+  clearTimeout?: (() => void) | undefined;
 };
 
 function connectionReadyResolver(): Resolver<void> {
@@ -267,6 +272,12 @@ export const DEFAULT_DISCONNECT_TIMEOUT_MS = 60 * 1_000;
  * consider it timed out.
  */
 export const CONNECT_TIMEOUT_MS = 10_000;
+
+/**
+ * Setting up active-client tracking blocks the first connect attempt. Above
+ * this, say so.
+ */
+const SLOW_ACTIVE_CLIENTS_THRESHOLD_MS = 50;
 
 const CHECK_CONNECTIVITY_ON_ERROR_FREQUENCY = 6;
 
@@ -719,6 +730,7 @@ export class Zero<
         this.#deleteClientsManager.onClientsDeleted([
           {clientGroupID, clientID},
         ]),
+      this.#lc,
     );
 
     const onUpdateNeededCallback = (reason: UpdateNeededReason) => {
@@ -1820,6 +1832,10 @@ export class Zero<
       'waiting for the server acknowledgement',
     );
     lc.debug?.('Waiting for connection to be acknowledged');
+    // Local setup is done; the rest of the wait is the server's, and gets its
+    // own budget rather than whatever the setup left over.
+    attempt.clearTimeout?.();
+    attempt.clearTimeout = this.#armConnectTimeout(lc, attempt, 'ack');
     await attempt.connected.promise;
     if (signal.aborted) {
       throw signal.reason;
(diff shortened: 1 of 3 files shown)


packages/zero-client/src/client/zero.ts
```diff
@@ -1820,6 +1832,10 @@ export class Zero<
       'waiting for the server acknowledgement',
     );
     lc.debug?.('Waiting for connection to be acknowledged');
+    // Local setup is done; the rest of the wait is the server's, and gets its
+    // own budget rather than whatever the setup left over.
+    attempt.clearTimeout?.();
+    attempt.clearTimeout = this.#armConnectTimeout(lc, attempt, 'ack');
     await attempt.connected.promise;
```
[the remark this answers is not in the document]
arv: Guarded, and made the deadline own its cancellation: `#armConnectTimeout` returns a no-op when the attempt is already aborted, and otherwise clears itself on the signal, so no deadline can outlive its attempt regardless of who holds the cancel function. Worth noting the specific sequence looks unreachable today — `#disconnect` aborts the attempt and `#connectAttempt` already checks `signal.aborted` on both sides of `createSocket`, leaving only a synchronous stretch before the handover — so I could not write a test that fails without the guard; it stands as a local invariant rather than a demonstrated fix. I did not race the acknowledgement wait against the signal: that changes attempt lifecycle beyond this PR and the stray-timer risk is closed by the signal binding alone.

packages/zero-client/src/client/event-loop-lag.ts
```diff
@@ -0,0 +1,197 @@
...
+  timeoutMs: number,
+  callback: (blockedMs: number) => void,
+): () => void {
+  const startedAt = Date.now();
+  const releaseLag = lag?.acquire();
+  let timeoutID: ReturnType<typeof setTimeout>;
+
+  const check = () => {
+    const blockedMs = lag?.blockedMsSince(startedAt) ?? 0;
+    const remainingMs = timeoutMs - (Date.now() - startedAt - blockedMs);
```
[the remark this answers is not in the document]
arv: This one is stale — it refers to `event-loop-lag.ts`, which no longer exists. The lag-adjusted timeout approach was abandoned in favour of not making the timers tolerant of a blocked loop at all: a long block is a real defect and should stay visible. The PR now splits the connect deadline into setup and acknowledgement phases instead, so local setup no longer spends the server's budget.

aboodman:
What motivated this change @arv ? I spent a lot of time on this area recently and it's super subtle.

Opening IDB and so-on should not be taking 10 seconds. I think something else must be going wrong which is why I added additional logging to the error message so that we could find out from error reports what phase was lagging.

arv:
The motivator is slow cold startups (React Native inparticular). With the old code we had a single budget for determining if the network connection was successful. In pathological cases that budget could be spent on SQLite on RN and IDB on web (and on web also the web lock). So a slow enough startup could use most of the connection budget, leaving too little time to connect to a healthy server, which showed up as spurious network connection failures.

This PR gives each phase its own budget instead of sharing one: setup (SQL/IDB, and the web lock on web) stays bounded and still retries, and the wait for the server starts a fresh timer once setup is done — so slow local startup no longer eats into the server's time. I've also added logging around the SQL/IDB and web lock phases so we get more insight when something is going wrong.

I'm not saying a 10s startup is expected, but we've seen RN be unreasonably slow, and that or some other unknown seems to be triggering this case.

aboodman:
Before, if the connection timed out (including local work) it would just retry again in 5s. So I think all this does is possibly reduce time to getting a good connection (and avoid a short 5s of saying user is offline). Do you agree that is the effect of the PR?

If that is all we're doing why not just change timeout to 20s. The code already reports the phase that failed and this logic was subtle and I'm afraid of regressions.

packages/zero-client/src/client/zero.ts
```diff
@@ -2960,12 +3022,32 @@ async function makeActiveClientsManager(
...
+  const groupID = await clientGroupID;
+  const start = performance.now();
+  const manager = await ActiveClientsManager.create(groupID, clientID, signal);
+  const elapsed = performance.now() - start;
+
+  // navigator.locks does not exist on React Native, where this falls back to
+  // an in-process stand-in that only ever sees this client. That changes both
+  // what the timing means and how much the client list can be trusted, so say
+  // which one produced the number.
+  const locks = getBrowserGlobal('navigator')?.locks ? 'native' : 'fallback';
```
aboodman, line 3040: I don't understand why we need this log. We can already tell what type of lock we have by OS log line came from. but if we do want it, the log line should be at the place the decision is made - in ActiveClientsManager ctor.

packages/zero-client/src/client/zero.ts
```diff
@@ -2960,12 +3022,32 @@ async function makeActiveClientsManager(
...
+  if (elapsed > SLOW_ACTIVE_CLIENTS_THRESHOLD_MS) {
```
aboodman, line 3041: This PR added timing code to help track down what was going on: https://github.com/rocicorp/mono/pull/6299. The timing info is reported to sentry so that it's easier to get from the field. Do we really additionally want this log-based timing code too?

aboodman: See `addConnectEvent`

PR #6482, opened 2026-09-04
feat(zero-client): 'cached' result type — persisted, fingerprint-verified complete results at boot

pawarren:
## Summary

This PR adds a fourth `ResultType`, `'cached'`: a materialized view whose complete result the local store already holds from a previous sync boots at `'cached'` instead of `'unknown'`, and graduates to `'complete'` on the first applied poke. We have been running this in production (React Native + expo-sqlite store, and web) as a patch on `@rocicorp/zero@1.9.0` across our whole app, and are upstreaming it now that it has survived a full consumer migration plus on-device verification.

## Motivation

With a persistent kvStore, the replica routinely holds the *complete* result of last session's queries at construction time — but a freshly materialized view can only say `'unknown'`, which is indistinguishable from "partially hydrated replica". Application code that needs trustworthy data (empty states, entitlement gates, client-side derivations) must therefore wait for the first poke on every single app open.

That wait is not just flicker. Measured on device (mid-range Android, production backend): Zero's socket does not survive app backgrounding (AbruptClose ~20s after home-press), so **every reopen pays a fresh connect** — 132/169ms time-to-first-poke on a good network, 1124/1129ms on post-background resume, worse on typical LTE. `'cached'` lets the app paint real content instantly on every warm or offline boot and treat it as settled, while still knowing the server has not yet spoken on this connection.

A second motivation is client-side derivations: ZQL has no aggregates, so count-like UI (badges, stats) is derived in app code from full result sets. That derivation is only sound when the set is known complete. `'unknown'` cannot make that claim; `'cached'` can ("the full set as of last sync"), which extends soundness to boot time.

## Semantics

| resultType | claim | may render as settled? | satisfies complete-waiters (`preload().complete`, `run({type:'complete'})`)? |
|---|---|---|---|
| `'unknown'` | No claim; the replica may be partially hydrated. | app's choice (loading) | no |
| `'cached'` | The store holds the server-confirmed **complete** result of this exact query body from a previous sync. | yes (app's choice) | **no** |
| `'complete'` | The server confirmed the result **on this connection**. | yes | yes |
| `'error'` | The query errored. | error state | rejects |

Transitions: `'cached'` is only ever entered from `'unknown'` (pre-authoritative), never downgrades `'complete'`/`'error'`, graduates to `'complete'` when `markGotQueriesAuthoritative()` fires after the first applied poke, and reverts to `'unknown'` if the got key is deleted (eviction) while still pre-authoritative. Freshness stays a promise only the connection can keep: nothing that waits on `'complete'` resolves at `'cached'`.

## Mechanism

- **The got-queries key's value carries the registered body's AST fingerprint** (`hashOfAST` of the server-mapped normalized AST; `null` when unknown, e.g. keys written by an older client). A `'cached'` claim requires the *current* registration's fingerprint to match the stored value, so a query-body edit under an unchanged name/args reads `'unknown'` (the safe direction) instead of vouching for rows a different body synced. The fingerprint is computed lazily per entry, so cold-start registration pays no extra AST walk unless asked.
- **Stale fingerprints converge via refresh puts appended to the first applied poke.** The server never re-sends a got put for an already-got hash, so without this a body-edit release would refuse the cached claim on every subsequent boot, forever, for a never-evicted query. The refresh entries are computed while merging (`mergePokes`) and ride the same transaction as the poke's own rows — they commit iff the catch-up they vouch for commits. There is deliberately no authoritative gate on the refresh: applying that very poke is what marks the session authoritative, and gating on the flag would make the refresh land one poke late.
- **Torn-state rule:** a refresh put is never emitted for a hash the same merged poke deletes or clears — the entries were computed from pre-poke state, and a put appended after the del would reinsert the bit while the rows delete.
- **`Zero.hasCachedResult(name, args)`** exposes the bare existence read of the persisted got key (usable at construction, before any connection) for boot probes and diagnostics. The view surface is strictly stronger — it additionally verifies the body fingerprint — so rendering should gate on the view's resultType.
- **View plumbing:** `ArrayView` gains `markCached()`/`unmarkCached()`; `materializeImpl` buffers a synchronous cached signal that fires during registration (before the view exists) and applies it right after construction. Custom view factories that do not implement the optional methods simply never surface the state — `zero-solid` is untouched and its suite passes as-is.

One operational observation worth recording: a short-TTL per-view query (our `workById`) can age out of the CVR while the app is closed. The store still holds its got key, so on reboot the view briefly boots `'cached'`, and the connection's own catch-up poke delivers the `del` that reverts it — the "desire aged out while offline" shape. That is exactly the merge where a got del and a refresh put can coexist, which is why the torn-state rule above exists and is exercised in our integration harness.

## Evidence

All suites in this repo, at `zero/v1.9.0`:

- `zql`: 1302 passed
- `zero-client`: 647 passed
- `zero-react`: 77 passed (74 existing + 3 new `'cached'` snapshot/waiter tests added here)
- `zero-solid`: 59 passed (no changes needed)
- `replicache`: 798 passed (typecheck included)

`check-types` clean on every touched package; `oxfmt --check` clean; `oxlint --type-aware` 0 errors.

Downstream, in our app: a six-scenario integration harness driving a real v51 fake sync server (born-cached boot, graduation on first poke, body-edit fingerprint mismatch, fingerprint refresh convergence with the masking heartbeat removed, eviction-while-offline staging the got del inside the catch-up poke, and complete-waiter pinning) passes against this exact logic, and all three device arms were verified on Android against production: fresh sign-in boots false/`'unknown'`, warm boot and airplane-mode boot both paint real content at `'cached'` with no loading state.

## Design notes for upstream (where you may want to go further than this PR does)

1. **Server-supplied body fingerprint.** This PR stamps the fingerprint client-side while applying the poke, from the current registration's AST. That is internally consistent, but the server is the natural authority: if the `gotQueriesPatch` put carried the query's fingerprint (e.g. the transformation hash for custom queries), the stored value would be authoritative rather than inferred, and server-side changes to a custom query's transformation would invalidate cached claims without any client involvement.
2. **Born-cached as a view-factory parameter.** The port reaches views through optional `markCached()`/`unmarkCached()` methods after construction, because `ViewFactory` only communicates `queryComplete: true | ErroredQuery | Promise<true>`. A first-class initial-result-state parameter (e.g. `queryComplete: true | 'cached' | ErroredQuery | Promise<true>`) would let every view implementation (solid included) surface the state natively instead of via the optional-method retrofit, at the cost of touching the factory signature.

Related footgun we hit while adopting this (not changed by this PR): `DEFAULT_PRELOAD_TTL` is `'none'` while materialize's default is `5m`. For an app leaning on preload precisely to have data cached for the *next* boot, a defaulted preload's desire expires as soon as it is released — deleting the got key and with it the cached claim — which is the opposite of what "preload" suggests. Worth considering alongside this feature.

Happy to adjust semantics, naming (`'cached'` vs alternatives), or split the `hasCachedResult` accessor out if you'd rather take the view-level state alone.

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -75,6 +75,7 @@ import type {
 } from '../../../zero-protocol/src/pull.ts';
 import type {PushMessage} from '../../../zero-protocol/src/push.ts';
 import type {UpQueriesPatchOp} from '../../../zero-protocol/src/queries-patch.ts';
+import {hashOfNameAndArgs} from '../../../zero-protocol/src/query-hash.ts';
 import type {Upstream} from '../../../zero-protocol/src/up.ts';
 import type {NullableVersion} from '../../../zero-protocol/src/version.ts';
 import {nullableVersionSchema} from '../../../zero-protocol/src/version.ts';
@@ -158,6 +159,7 @@ import {
 } from './http-string.ts';
 import {Inspector} from './inspector/inspector.ts';
 import {IVMSourceBranch} from './ivm-branch.ts';
+import {toGotQueriesKey} from './keys.ts';
 import {type LogOptions, createLogOptions} from './log-options.ts';
 import type {MakeMutatePropertyType} from './make-mutate-property.ts';
 import {addCustomMutatorsProperties} from './make-mutate-property.ts';
@@ -864,6 +866,8 @@ export class Zero<
       schema,
       this.#lc,
       this.#mutationTracker,
+      hash => this.#queryManager.astFingerprintForHash(hash),
+      () => this.#queryManager.gotFingerprintRefreshEntries(),
     );
 
     this.#visibilityWatcher = getDocumentVisibilityWatcher(
@@ -978,6 +982,32 @@ export class Zero<
       }
     }
     return createLogOptions(options);
+  }
+
+  /**
+   * Reports whether the store CURRENTLY holds the server-confirmed complete
+   * result of the named query (name + args) on this client group — i.e.
+   * whether the persisted got-queries key exists. Readable at construction,
+   * before any connection. Eviction (deregistration + TTL expiry),
+   * client-group reset, and resync delete the key, reverting the answer to
+   * false — the "never asked" reading, which is the safe direction. The key
+   * and its rows only change together inside applied-poke transactions, and
+   * eviction deletes them together, so a `true` never vouches for torn local
+   * data.
+   *
+   * Materialized views surface the same fact natively — and STRONGER: a view
+   * additionally verifies the stored body fingerprint, so after a query-body
+   * edit this existence read can be true while views correctly hold at
+   * 'unknown'. Gate rendering on the view's resultType; use this for boot
+   * probes and diagnostics.
+   */
+  hasCachedResult(
+    name: string,
+    args: readonly ReadonlyJSONValue[],
+  ): Promise<boolean> {
+    return this.#rep.query(tx =>
+      tx.has(toGotQueriesKey(hashOfNameAndArgs(name, args))),
+    );
   }
 
   /**
--- a/packages/zero-client/src/client/query-manager.test.ts
+++ b/packages/zero-client/src/client/query-manager.test.ts
@@ -1239,7 +1239,8 @@ test('handleClosed marks queries as errored exactly once', () => {
 
   const ast: AST = {table: 'issue'};
   const queryHash = hashOfAST(ast);
-  const gotCallback = vi.fn<(got: boolean, error?: ErroredQuery) => void>();
+  const gotCallback =
+    vi.fn<(got: boolean | 'cached', error?: ErroredQuery) => void>();
 
   queryManager.addLegacy(ast, 0, gotCallback);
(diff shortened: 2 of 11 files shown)


arv:
I'm not sure we want to make this change.

At least I do not think now is the right time to do it

arv:
> I'm not sure we want to make this change.
> 
> At least I do not think now is the right time to do it

We are doing this. I'll post a new PR and close this.

packages/zero-client/src/client/zero.ts
```diff
@@ -980,6 +984,32 @@ export class Zero<
...
+   * eviction deletes them together, so a `true` never vouches for torn local
+   * data.
+   *
+   * Materialized views surface the same fact natively — and STRONGER: a view
+   * additionally verifies the stored body fingerprint, so after a query-body
+   * edit this existence read can be true while views correctly hold at
+   * 'unknown'. Gate rendering on the view's resultType; use this for boot
+   * probes and diagnostics.
+   */
+  hasCachedResult(
```
arv, line 1004: Let's not expose this. Maybe we can do this later but step 1 would just be to extend resultType.

@aboodman

PR #6484, opened 2026-09-04
feat(zero-cache): Default CVR garbage collection threshold to 1 week

arv:
Raises `--cvr-garbage-collection-inactivity-threshold-hours` (`ZERO_CVR_GARBAGE_COLLECTION_INACTIVITY_THRESHOLD_HOURS`) from 48 to 168 (`24 * 7`).

## Why

Purging a CVR disables the client group, so the next connection from that client re-downloads everything (`ClientStateNotFound` → local store dropped → full cold sync). At 48h, any app whose users come back weekly rather than daily pays that cold start on most sessions. This came up with a customer whose per-user dataset is large enough that the cold path is very visible.

The short lifetime dates from a period when this machinery was less stable; `deleteClient` and the purger have been reliable for a while now. Queries are not run for disconnected clients, so a longer retention costs CVR database storage, not compute.

## Changes

- `packages/zero-cache/src/config/zero-config.ts`: default `48` → `24 * 7`, plus a note in the option description about what a purge costs the client.
- `packages/zero-cache/src/config/zero-config.test.ts`: `--help` snapshot updated.

Cloud Zero will additionally expose this per stack as an advanced setting (separate PR in `rocicorp/cloudzero`).

### DIFF
--- a/packages/zero-cache/src/config/zero-config.test.ts
+++ b/packages/zero-cache/src/config/zero-config.test.ts
@@ -238,9 +238,12 @@ test('zero-cache --help', () => {
                                                                                    Note that this number must allow for at least one connection per                                                           
                                                                                    sync worker, or zero-cache will fail to start. See num-sync-workers                                                        
                                                                                                                                                                                                               
-     --cvr-garbage-collection-inactivity-threshold-hours number                    default: 48                                                                                                                
+     --cvr-garbage-collection-inactivity-threshold-hours number                    default: 168                                                                                                               
        ZERO_CVR_GARBAGE_COLLECTION_INACTIVITY_THRESHOLD_HOURS env                                                                                                                                             
                                                                                    The duration after which an inactive CVR is eligible for garbage collection.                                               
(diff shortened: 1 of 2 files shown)


arv:
**Verify Signed Commit Authors** is failing on this PR and I can't fix it from here.

The commit is signed, but by the remote Claude Code session's own SSH key (`ED25519 SHA256:32dP45eSMmVSt/G/CGvcxl/P+MO3Nwj9xeTh/GSA2wc`). That key is not in `rocicorp/.github` → `.github/signing/allowed_signers`, which lists only the team's laptop keys, so the action reports `No principal matched` and rejects the commit. This is independent of the commit author (`arv@roci.dev`); no author identity makes an unlisted key acceptable.

Getting it green needs someone with an allowed key to re-sign:

```
git fetch origin arv/cvr-gc-default-one-week
git checkout arv/cvr-gc-default-one-week
git commit --amend --no-edit -S
git push --force-with-lease
```

I'm not pushing anything speculative in the meantime — the code itself is unchanged from what CI otherwise saw, and the rest of the checks are unaffected by this.

---

PR #6485, opened 2026-09-04
fix(zero-client): carry the connect abort reason on the attempt, not …

arv:
…only signal.reason

On runtimes whose AbortController predates the 2021 signal.reason spec addition — React Native's bundled abort-controller@3 polyfill being the big one — abort(reason) silently discards the reason and signal.reason reads undefined. The connection run loop reads the abort reason exclusively off the signal, so on those runtimes every retryable connect failure (ConnectTimeout included) surfaces as undefined, gets wrapped as a non-retryable internal ClientError, and permanently pauses the run loop in the error state. Zero's own retry/backoff never runs.

Record the reason on the ConnectAttemptControl in #disconnect and prefer it at the four read sites, falling back to signal.reason. Behavior on spec-compliant platforms is unchanged.

Regression test stubs an AbortController whose abort() drops the reason argument and asserts a connect timeout still retries instead of pausing: red on the previous code ('error' where 'connecting' was expected), green with the fix.

### DIFF
--- a/packages/zero-client/src/client/zero.ts
+++ b/packages/zero-client/src/client/zero.ts
@@ -224,6 +224,18 @@ type ConnectAttemptControl = {
    * attempt stops setting up locally and starts waiting for the server.
    */
   clearTimeout?: (() => void) | undefined;
+
+  /**
+   * The error this attempt was aborted with. Read this instead of
+   * `controller.signal.reason`: on runtimes whose `AbortController` predates
+   * the 2021 `signal.reason` spec addition (e.g. React Native's bundled
+   * `abort-controller@3` polyfill), `abort(reason)` silently discards the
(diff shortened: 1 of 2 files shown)


arv:
#6468

arv:
@copilot resolve the merge conflicts in this pull request

packages/zero-client/src/client/zero.ts
```diff
@@ -1780,7 +1793,7 @@ export class Zero<
     // The run loop has already stopped waiting for this attempt if it was
     // canceled. Do not let setup that completed late create a socket.
     if (signal.aborted) {
-      throw signal.reason;
+      throw attempt.abortReason ?? signal.reason;
```
aboodman, line 1796: Remove the duplicate path and let's just always use `attempt.abortReason`. Do we need `signal.reason` at all?

PR #6488, opened 2026-09-04
feat(zero-cache): config to turn on SQLite-only change log mode

tantaman:
- A new hidden setting, pgChangeLogEnabled. defaults to true.
  - If false, SQLite handles:
      - resuming streams
      - storing incoming changes
      - serving catch-up data
      - deciding when changes are safe to purge
      - tracking backup progress
  - PostgreSQL’s old change log is left alone and is no longer used as a fallback.
  - Subscribers asking for data older than SQLite has receive a “watermark too old” error,
    prompting a fresh sync.
  - Disabling PostgreSQL requires SQLite to be fully authoritative and requires a Litestream
    v5 backup, preventing unsafe configurations.


arv (commented):
LGTM

PR #6512, opened 2026-09-08
fix(zero-throughput): direct import Fastify in src/server.ts for Vercel detection

grgbkr:
Resolves the Vercel deployment error Error: No entrypoint found which imports fastify by initializing and exporting the Fastify application directly in src/server.ts.


arv:
I'm not sure I understand what Vercel is doing here but if it works it works.

LGTM

PR #6513, opened 2026-09-08
fix(zero-cache): unblock backfills when the change stream is canceled

arv:
## Problem

When the change stream is canceled (reconnect, backoff or shutdown) while a backfill is in progress, the backfill could be left suspended forever:

- If the backfill was awaiting `ChangeStreamMultiplexer.reserve('backfill')` while the replication loop held the reservation mid-transaction, the replication loop exits its `for await` on cancel without releasing, so the reservation promise never settles.
- If the backfill was awaiting `#changeStreamReached()` before sending `backfill-completed`, nothing resolves that waiter after cancel.

`BackfillManager.cancel()` only set `canceledReason`, which is checked *after* the reservation is granted. The stranded `#runBackfill` kept the `streamBackfill` async generator suspended at a `yield`, so its `finally` never ran: `tx.setDone()` and `db.end()` were never called. Each occurrence leaked an upstream Postgres connection holding a `REPEATABLE READ` snapshot transaction, kept alive indefinitely by the `TransactionPool` keepalives (pinning xmin on upstream). Each reconnect during a long backfill on a busy source could add another.

## Fix

- `ChangeStreamMultiplexer`: on cancel, reject all pending reservation requests (and any subsequent ones) with an `AbortError`, so waiting producers unwind.
- `BackfillManager.cancel()`: mark the manager canceled, wake any watermark waiters, and clear the retry timer. The running backfill exits at the next message (finalizing its stream via `for await` teardown), and no retries are scheduled after cancelation.

## Tests

- `change-stream-multiplexer.test.ts`: cancelation rejects pending and subsequent reservations.
- `backfill-manager.test.ts`: two new tests covering a backfill blocked on a reservation and one blocked on the stream watermark. Both verify the backfill stream is finalized after cancel and that no retry is started. Both fail without the fix (timeouts / stream never finalized).

Ran `lint`, `format`, `check-types`, the `no-pg` tests for `change-source/common`, and the `pg-16` `change-source` suite against a local Postgres 16 (3 pre-existing failures there are environmental: they depend on the DB user being named `test`, and reproduce identically on unmodified `main`).

---

### DIFF
--- a/packages/zero-cache/src/services/change-source/common/backfill-manager.test.ts
+++ b/packages/zero-cache/src/services/change-source/common/backfill-manager.test.ts
@@ -1,5 +1,5 @@
 import type {LogContext} from '@rocicorp/logger';
-import {beforeEach, describe, expect, test} from 'vitest';
+import {beforeEach, describe, expect, test, vi} from 'vitest';
 import {createSilentLogContext} from '../../../../../shared/src/logging-test-utils.ts';
 import {must} from '../../../../../shared/src/must.ts';
 import {Queue} from '../../../../../shared/src/queue.ts';
@@ -26,6 +26,7 @@ describe('backfill-manager', () => {
   let backfillRequests: BackfillRequest[];
   let testStreams: (TestStreamItem[] | Error)[];
   let changes: Queue<ChangeStreamMessage>;
+  let finalizedStreams: number;
   let lc: LogContext;
 
   function initBackfillManager(commitThresholdBytes?: number) {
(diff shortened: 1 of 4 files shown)


arv:
CI status: all lint, type, test (PG 15–18) and Docker build checks pass. Two failures remain that I cannot fix from this session:

- **Verify Signed Commit Authors**: the commit is SSH-signed with this session's key, which is not in `rocicorp/.github`'s `allowed_signers`. The commit needs to be re-signed by the author (e.g. `git commit --amend --no-edit -S` and force-push) or the key needs to be allow-listed.
- **Vercel – zero-throughput**: this PR only touches `packages/zero-cache`; `apps/zero-throughput`'s Vercel config was added in the base commit (#6511) and this is its first preview deploy on a branch, so the failure is independent of this change.

---

PR #6516, opened 2026-09-08
fix(zero-cache): drop sync connections whose socket closed during auth

arv:
## Problem

`Syncer.#createConnection` awaits `resolveAuth()` (JWT verification, possibly a remote JWKS fetch) before creating the `Connection`. `installWebSocketReceiver` checks `ws.readyState` before handing the socket over, but nothing re-checks it after the await. If the client disconnected in the meantime, the `'close'` event had already fired, so:

- the `Connection` attached its `'close'`/`'error'` listeners and `createWebSocketStream` to a dead socket, and `Connection.close()` never ran;
- the initial `connected` message is sent with `'ignore-backpressure'` and is silently dropped on a non-open socket, so nothing surfaced the failure.

Each occurrence leaked the `Connection` (with its 3 s pong `setInterval`, the `ConnectParams` including the auth token, and the socket), its entry in the `ConnectionContextManager`, its `ref()`s on the `Mutagen` and `Pusher` services (so they were never stopped by the `ServiceRunner`), and a `ViewSyncerService` for the client group, until the same `clientID` happened to reconnect to the same syncer worker. When `initConnection` is sent as a WebSocket frame rather than in the `sec-websocket-protocol` header (query-heavy apps), no downstream stream is ever created that could surface the dead socket, so the leak was permanent.

## Fix

After auth resolves, bail out if the socket is no longer `OPEN`, before creating any per-connection or per-client-group state. Records a `closed_during_auth` connection-failure metric.

## Tests

New `websocket closed while resolving auth` tests in `syncer.test.ts`: the socket closes from within the mocked `verifyToken`, and the test asserts that no mutagen/pusher is created, no connection is registered, and no success metric is recorded. The control test verifies a still-open socket is registered as before. Without the fix, the first test fails (a success is recorded and services are created for the dead socket).

Ran `lint`, `format`, `check-types`, and the `workers` `no-pg` tests.

---

### DIFF
--- a/packages/zero-cache/src/workers/syncer.test.ts
+++ b/packages/zero-cache/src/workers/syncer.test.ts
@@ -823,6 +823,94 @@ describe('jwt auth missing options and missing endpoints', () => {
   });
 });
 
+describe('websocket closed while resolving auth', () => {
+  let syncer: Syncer;
+  let mutagens: MutagenService[];
+  let pushers: PusherService[];
+  let contextManagers: Map<string, ConnectionContextManagerImpl>;
+  let verifySpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    verifySpy = vi.spyOn(jwt, 'verifyToken');
+    verifySpy.mockReset();
+    const env = setupSyncer(lc, {
+      auth: {
+        secret: 'test-secret',
+      },
+    } as ZeroConfig);
+    syncer = env.syncer;
+    mutagens = env.mutagens;
+    pushers = env.pushers;
+    contextManagers = env.contextManagers;
+  });
+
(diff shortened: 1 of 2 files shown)


arv:
CI status: two failures remain that I cannot fix from this session:

- **Verify Signed Commit Authors**: the commit is SSH-signed with this session's key, which is not in `rocicorp/.github`'s `allowed_signers`. It needs to be re-signed by the author (e.g. `git commit --amend --no-edit -S` and force-push) or the key needs to be allow-listed.
- **Vercel – zero-throughput**: this PR only touches `packages/zero-cache`; `apps/zero-throughput`'s Vercel config was added in the base commit (#6511) and this is its first preview deploy on a branch, so the failure is independent of this change.

---

PR #6523, opened 2026-09-09
perf(zql): pull-function protocol for the fetch path

arv:
Replaces the iterator protocol on the fetch path with a pull protocol:

```ts
interface PullStream<T> {
  next(): T | undefined;
  close(): void;
}
```

`next()` returns the value directly — `undefined` marks the end — so there is
no `{done, value}` object per row, and one object per stream rather than one
per stage.

It is deliberately **not** `Iterable`, and there is no adapter back to one. An
escape hatch would let any unconverted consumer keep paying the exact cost this
exists to remove, and would let the old protocol creep back later.

This now also contains what was #6509 (de-generatoring `MemorySource#fetch`),
as its first commit. That PR is closed; reviewing this one covers both.

### Why

Profiling the ZQL benchmarks on Hermes (harness in #6507) put **5.4%** of
`hydrate: issues with creator` in generator resume plumbing alone — `next` and
`generatorPrototypeResume` frames — before counting what hides inside the
generator bodies. A four-deep pipeline over 5000 rows, measured in the app's
own runtime:

| | Hermes | V8 |
|---|---|---|
| 4 chained generators | 31.9 ms | 4.0 ms |
| 4 chained manual iterators | 16.2 ms | 1.2 ms |
| 4 chained push callbacks | 12.5 ms | 0.8 ms |

Relative to a plain loop that is 31x on Hermes and 74x on V8 — V8 optimizes
loops and hand-written iterators hard, and generators are a barrier it will not
cross, so removing a generator layer pays *more* there. This is not a
Hermes-specific fix.

### Results

Android emulator, median of 3 paired interleaved rounds, `main` → this branch:

| benchmark | main | this | | V8 |
|---|---|---|---|---|
| hydrate: issues filtered open | 101.23 | 68.64 | **−33.3%** | −38.5% |
| hydrate: issues limit 50 | 13.12 | 11.27 | **−13.9%** | −23.7% |
| hydrate: issues with creator + comments | 998.82 | 873.14 | **−12.6%** | −17.3% |
| hydrate: issues with creator | 454.66 | 395.81 | **−12.5%** | −21.9% |
| hydrate: issues only | 86.74 | 76.30 | **−12.2%** | −21.1% |

Trust the device numbers over V8's: per-round spreads on iOS are tight
(−30.9, −33.5, −33.3), while V8 on a loaded host produced one +109.8% round on
`limit 50` that the median absorbs.

`filtered open` is the largest win and the one that required going all the way:
at an intermediate stage where `Filter` still reached the source through a
compatibility adapter it was −1.7%, because every row still allocated a result
object.

### The one regression

`push: add comment (child relation)` is **+2.0%** (median of 6 paired rounds,
range 0.0–4.3%, 5 of 6 positive). Push does not run the fetch path, so it gets
none of the benefit while still paying the new protocol's per-object cost. A
profile shows the cost is diffuse — push is ~43% BTree-and-view comparison,
which this does not touch — and no converted class appears in its hot frames.
The other five push benchmarks are flat.

### Two behaviours that are easy to lose

- **`close()` replaces what `for...of` did implicitly.** Breaking out of a loop
  called `.return()`, and that is what closed SQLite cursors. Every early exit
  now closes explicitly; the `finally` blocks in `zero-cache` are load-bearing,
  not defensive. Without them an abandoned hydration leaks a cursor and the next
  write fails with "database connection is busy executing a query" — caught by
  `abandoned hydration tears down its pipeline` and the view-syncer hydration
  timeout tests.
- **The two merges genuinely differ.** `mergeFetches` advances a branch before
  emitting the node it selected from it; `mergeSortedStreams` emits first and
  refills after. Swapping the order reorders 'yield' markers — `UnionFanIn
  propagates yield` catches it.

`Exists` is the one filter that suspends (counting a relationship pulls a child
stream that can emit 'yield'). It now holds that position in explicit state
rather than a parked generator, which let the generator-based
`FilterOutput.filter` API be deleted outright.

### One Hermes-only edit

`generateRows` picked its method with `data[reverse ? a : b]()`. The computed
member forces a dynamic lookup Hermes cannot inline-cache; the ternary-of-calls
form measured 18% faster there. I could not get a trustworthy V8 number for it
in isolation — the computed form timed 1.0 ms for one branch and 19.9 ms for
the other while the ternary sat at ~9.6 ms both ways, stable across two probe
designs. 1.0 ms for 2M allocating calls is 0.5 ns each, so V8's escape analysis
is scalar-replacing the returned object asymmetrically and the microbenchmark
is not measuring the property access. Treat that edit as Hermes-motivated and
V8-unknown; it is subsumed by the end-to-end numbers above.

### Commits

Reviewable in dependency order: `MemorySource` first, then the protocol, then
each consumer package, then a refactor pass that folds the repeated pull-stream
shapes into combinators (28 classes → 15).

### Verification

zql 1448 · shared 774 · zqlite 199 · zero-cache 5084 · zero-client 659 ·
zero-solid 60 · zql-integration-tests 1169. Repo-wide `check-types` (44/44),
`lint` and `check-format` clean, and `@rocicorp/zero` builds.

Two failures are not from this branch: zero-client's `logged-out client uses a
private storage sentinel for idb naming` asserts a hardcoded schema version 52
against the current 53 and fails on `main` too, and the `zql-integration-tests`
pg suite intermittently exhausts Postgres connections ("sorry, too many clients
already") — the affected files pass on their own.

The benchmark numbers above were taken before the final refactor, rename and
`zero-solid` commits. Those are behaviour-preserving and the suites agree, but
they have not been re-measured.

### DIFF
--- a/packages/zero-cache/src/auth/write-authorizer.ts
+++ b/packages/zero-cache/src/auth/write-authorizer.ts
@@ -569,10 +569,13 @@ export class WriteAuthorizerImpl implements WriteAuthorizer {
     const input = buildPipeline(rowQueryAst, this.#builderDelegate, 'query-id');
     try {
       const res = input.fetch({});
-      for (const _ of res) {
-        // if any row is returned at all, the
-        // rule passes.
-        return true;
+      try {
+        // if any row is returned at all, the rule passes.
+        if (res.next() !== undefined) {
+          return true;
+        }
+      } finally {
+        res.close();
       }
     } finally {
       input.destroy();
--- a/packages/zero-cache/src/services/run-ast.ts
+++ b/packages/zero-cache/src/services/run-ast.ts
@@ -102,8 +102,13 @@ export async function runAst(
     // triggering early return on Take's #initialFetch assertion.
     // The subquery AST already has limit: 1, so at most one row is produced.
     let node: Node | undefined;
-    for (const n of skipYields(input.fetch({}))) {
-      node ??= n;
+    const rows = skipYields(input.fetch({}));
+    try {
+      for (let n = rows.next(); n !== undefined; n = rows.next()) {
+        node ??= n;
+      }
+    } finally {
+      rows.close();
     }
     input.destroy();
     return node ? ((node.row[childField] as LiteralValue) ?? null) : undefined;
--- a/packages/zero-cache/src/services/view-syncer/flipped-exists-fetch-filter.bench.ts
+++ b/packages/zero-cache/src/services/view-syncer/flipped-exists-fetch-filter.bench.ts
@@ -21,7 +21,8 @@ type Fetch = {
 
 function fetchRowCount(input: Input, filter: NoSubqueryCondition | undefined) {
   let count = 0;
-  for (const node of input.fetch({filter})) {
+  const stream = input.fetch({filter});
+  for (let node = stream.next(); node !== undefined; node = stream.next()) {
     if (node !== 'yield') {
       count++;
     }
--- a/packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
+++ b/packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
@@ -15,7 +15,7 @@ import {
 import {ChangeIndex} from '../../../../zql/src/ivm/change-index.ts';
 import {ChangeType} from '../../../../zql/src/ivm/change-type.ts';
 import type {Change} from '../../../../zql/src/ivm/change.ts';
-import type {Node} from '../../../../zql/src/ivm/data.ts';
+import type {Node, RelationshipStream} from '../../../../zql/src/ivm/data.ts';
 import {
   skipYields,
   throwOutput,
@@ -33,6 +33,7 @@ import {
   makeSourceChangeEdit,
   makeSourceChangeRemove,
 } from '../../../../zql/src/ivm/source.ts';
+import {pullOf, type PullStream} from '../../../../zql/src/ivm/stream.ts';
 import type {ConnectionCostModel} from '../../../../zql/src/planner/planner-connection.ts';
 import {MeasurePushOperator} from '../../../../zql/src/query/measure-push-operator.ts';
 import type {ClientGroupStorage} from '../../../../zqlite/src/database-storage.ts';
@@ -560,8 +561,13 @@ export class PipelineDriver {
       // triggering early return on Take's #initialFetch assertion.
       // The subquery AST already has limit: 1, so at most one row is produced.
       let node: Node | undefined;
-      for (const n of skipYields(input.fetch({}))) {
-        node ??= n;
+      const rows = skipYields(input.fetch({}));
+      try {
+        for (let n = rows.next(); n !== undefined; n = rows.next()) {
+          node ??= n;
+        }
+      } finally {
+        rows.close();
       }
       if (!node) {
         return undefined;
@@ -1369,9 +1375,9 @@ class Streamer {
       switch (type) {
         case ChangeType.REMOVE:
         case ChangeType.ADD: {
-          yield* this.#streamNodes(queryID, schema, type, () => [
-            change[ChangeIndex.NODE],
-          ]);
+          yield* this.#streamNodes(queryID, schema, type, () =>
+            pullOf([change[ChangeIndex.NODE]]),
+          );
           break;
         }
 
@@ -1385,9 +1391,9 @@ class Streamer {
           break;
         }
         case ChangeType.EDIT:
-          yield* this.#streamNodes(queryID, schema, type, () => [
-            {row: change[ChangeIndex.NODE].row, relationships: {}},
-          ]);
+          yield* this.#streamNodes(queryID, schema, type, () =>
+            pullOf([{row: change[ChangeIndex.NODE].row, relationships: {}}]),
+          );
           break;
         default:
           unreachable(change[ChangeIndex.TYPE]);
@@ -1399,7 +1405,7 @@ class Streamer {
     queryID: string,
     schema: SourceSchema,
     op: ChangeType.ADD | ChangeType.REMOVE | ChangeType.EDIT,
-    nodes: () => Iterable<Node | 'yield'>,
+    nodes: () => RelationshipStream,
   ): Iterable<RowChange | 'yield'> {
     const {tableName: table, system} = schema;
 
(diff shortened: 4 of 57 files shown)


packages/zql/src/ivm/memory-source.ts
```diff
@@ -369,19 +405,12 @@ export class MemorySource implements Source {
...
-        yield {row, relationships: {}};
-      }
-      return;
+      return {rows: rowsIterable, constraint: req.constraint};
     }
 
     const withOverlay = generateWithOverlay(
       startAt,
-      pkConstraint ? once(rowsIterable) : rowsIterable,
+      new RowScan(rowsIterable),
```
tantaman, line 413: this used to do `once` rather than a full scan if a pkConstraint existed

tantaman:
DM to Arv reproduced below:

ok. I think we need to manually close all these streams if they return early (either throw or early loop termination).

At first I assumed it'd just get GC'ed and we're fine. Problem is that the SQLite backend holds a resource from SQLite (the prepared statement) that needs to be freed as soon as the stream is no longer used or else you can get this error on fetch:


--- write with an abandoned cursor open ---
write FAILED: This database connection is busy executing a query

This is the fix (by Claude): https://github.com/rocicorp/mono/commit/d3fbd4c2f397b4ac4c4a26d9d7fa44da34193fed

and some tests that failed prior to the change: https://github.com/rocicorp/mono/commit/a475ecc341ffb61d47ced851806b0ac0565484f8

I or you should have it add a third commit that reproduces the SQLite failure since that's really the thing we're trying to protect against.

packages/zero-cache/src/auth/write-authorizer.ts
```diff
@@ -569,10 +569,13 @@ export class WriteAuthorizerImpl implements WriteAuthorizer {
     const input = buildPipeline(rowQueryAst, this.#builderDelegate, 'query-id');
     try {
       const res = input.fetch({});
-      for (const _ of res) {
-        // if any row is returned at all, the
-        // rule passes.
-        return true;
+      try {
```
arv, line 572: this can use withPull

packages/zero-cache/src/services/view-syncer/flipped-exists-fetch-filter.bench.ts
```diff
@@ -21,7 +21,8 @@ type Fetch = {
 
 function fetchRowCount(input: Input, filter: NoSubqueryCondition | undefined) {
   let count = 0;
-  for (const node of input.fetch({filter})) {
+  const stream = input.fetch({filter});
+  for (let node = stream.next(); node !== undefined; node = stream.next()) {
```
arv, line 25: forEachSkipYield?

arv: missing close

packages/zero-cache/src/services/view-syncer/pipeline-driver.ts
```diff
@@ -560,8 +561,13 @@ export class PipelineDriver {
       // triggering early return on Take's #initialFetch assertion.
       // The subquery AST already has limit: 1, so at most one row is produced.
       let node: Node | undefined;
-      for (const n of skipYields(input.fetch({}))) {
-        node ??= n;
+      const rows = skipYields(input.fetch({}));
+      try {
+        for (let n = rows.next(); n !== undefined; n = rows.next()) {
```
arv, line 566: forEach here too

packages/zero-cache/src/services/run-ast.ts
```diff
@@ -102,8 +102,13 @@ export async function runAst(
     // triggering early return on Take's #initialFetch assertion.
     // The subquery AST already has limit: 1, so at most one row is produced.
     let node: Node | undefined;
-    for (const n of skipYields(input.fetch({}))) {
-      node ??= n;
+    const rows = skipYields(input.fetch({}));
```
arv, line 105: forEach here too

packages/zql-integration-tests/src/helpers/runner.ts
```diff
@@ -688,8 +688,17 @@ function gatherRows(
       _queryComplete,
     ) => {
       const schema = input.getSchema();
-      for (const node of skipYields(input.fetch({}))) {
-        processNode(schema, node);
+      const stream = skipYields(input.fetch({}));
```
arv, line 691: forEach here too

packages/zql-integration-tests/src/helpers/runner.ts
```diff
@@ -715,8 +724,17 @@ function gatherRows(
       node.relationships,
     )) {
       const childSchema = must(schema.relationships[relationship]);
-      for (const child of skipYields(getChildren())) {
-        processNode(childSchema, child);
+      const children = skipYields(getChildren());
+      try {
```
arv, line 728: forEach here too

packages/zql/src/ivm/array-view.ts
```diff
@@ -144,19 +144,33 @@ export class ArrayView<V extends View> implements Output, TypedView<V> {
...
-        this.#format,
-        false /* withIDs */,
-        true /* mutate: #root is freshly created and not yet observed by any
-                 consumer, so build it in place to avoid O(N^2) array copies.
-                 Every later push() is immutable, preserving reference
-                 stability for unchanged subtrees. */,
-      );
+    // Pull protocol: no result object per node. 'yield' is skipped inline
+    // rather than through skipYields, which would put an iterator back.
+    const stream = this.#input.fetch({});
```
arv, line 149: forEach here too

packages/zql/src/ivm/catch.ts
```diff
@@ -129,8 +130,18 @@ export function expandNode(node: Node | 'yield'): CaughtNode {
         row: node.row,
         relationships: mapValues(node.relationships, getChildren => {
           const children: CaughtNode[] = [];
-          for (const child of getChildren()) {
-            children.push(expandNode(child));
+
+          const childStream = getChildren();
```
arv, line 134: forEach here too

packages/zql/src/ivm/data.ts
```diff
@@ -135,8 +142,17 @@ export function drainStreams(node: Node | 'yield') {
     return;
   }
   for (const stream of Object.values(node.relationships)) {
-    for (const node of stream()) {
-      drainStreams(node);
+    const children = stream();
```
arv, line 145: forEach here too

arv (commented):
A lot of loops can be replaced with the new forEach helper. I stopped reviewing after 10

arv:
abandoning this

PR #6524, opened 2026-09-09
perf(zql): de-generator the pure-passthrough push delegations

arv:
Stacked on `mlaw/ivm-pull-fns` (was #6523 / `arv/ivm-pull-fns`).

> **Rebased onto the cleanup fixes.** `mlaw/ivm-pull-fns` carries a475ecc34
> (14 reproductions of pull-protocol cleanup regressions) and d3fbd4c2f (their
> fixes) on top of the fetch-path conversion. These three commits rebase onto
> it cleanly and the whole suite passes, including the 14 new tests.
>
> **The results below predate that base and need re-taking.** d3fbd4c2f
> restores the single-row cap on primary-key scans in `MemorySource#fetch`,
> which the conversion had dropped -- a PK lookup was walking to the end of the
> index and rejecting every remaining row through the filter predicate. The
> push benchmarks materialize a view before pushing, so they ran that path.

Two changes to what is left of the iterator protocol after the fetch path moved
to pull functions.

### 1. The filter chain's push path was pure `yield*`

`Output.push` returns `Stream<'yield'>` so that a slow push can *suspend*. But
`Filter`, `FilterStart`, `FilterEnd`, `filterPush` and
`maybeSplitAndPushEditChange` never yield on their own — every branch of all
five is a plain delegation. Each was a generator layer costing a resume per
yield and buying nothing, the same shape as the old `generateRows`.

They are now ordinary functions that return the downstream stream:

```ts
push(change: Change): Stream<'yield'> {
  return filterPush(change, this.#output, this, this.#predicate);
}
```

Downstream stays lazy — calling a generator function does not run its body —
and the empty cases share one frozen `EMPTY_YIELDS` instead of each allocating
a generator. `buildFilterPipeline` inserts a FilterStart/FilterEnd pair into
*every* pipeline, so this removes at least two generator layers from every
push, filtered or not.

The operators that genuinely suspend (`take`, `exists`, `cap`, `skip`,
`union-fan-in`, the sources) are untouched. Replacing `Stream<'yield'>` itself
wants a resumable-continuation type rather than `PullStream<T>` — push carries
no values — and is a separate design.

**One semantic change:** the predicate now runs at `push()` call time rather
than at first `next()`. No call site anywhere stores a push stream without
consuming it at the same site — all 67 consumers are inline `yield*`/`for...of`/
`consume` — and the predicate is required pure, so it is not observable.

### 2. BTreeSet is no longer iterable

The iterator protocol allocates a result object per value, which is the cost
the fetch path exists to avoid, and an iterable API is an easy way to
reintroduce it by accident.

Finding the real callers by making `next()` throw turned up one on a hot path:

```
BTreeForwardIterator.next → toSorted (iterables.ts) → MemorySource#getOrCreateIndex → fetch
```

`toSorted(tree, cmp)` is `[...tree].sort()`, so **every secondary-index build
allocated one `IteratorResult` per row**. That is now `BTreeSet#toArray()`,
which drains via `nextValue()`.

With that gone, `next()` had no production callers left, so `ValueIterator` is
pull-only, `[Symbol.iterator]` is gone from the iterators and from `BTreeSet`,
and `drainValues()` collects a scan where a test needs an array.

Two things fell out:

- **`Storage#scan` is deleted.** It had no production callers — only its own
  tests — and was the last thing in zql iterating a BTreeSet. zqlite's `del`
  test used it only to verify deletion and now asserts with `get`.
- **`RowScan#close()` was calling `this.#rows.return?.()`, which was always a
  no-op** — the BTree iterators never defined `return`. Removing iterability
  made that visible. An index scan holds no external resource, so `close()`
  now just sets `#done`. Worth noting given how load-bearing `close()` is for
  the *SQLite* cursors; this particular call never did anything.

### Results

Android emulator (Hermes), `--run push`, 4 paired interleaved rounds, medians.
ops/sec, so **higher is better**:

| benchmark | base | this | | in-run noise |
|---|---|---|---|---|
| Filter / push add open issue (passes filter) | 391.2 | 423.9 | **+8.7%** | ±0.4% |
| Filter / push add closed issue (filtered out) | 417.9 | 441.0 | **+5.7%** | ±0.3% |
| MemorySource push: add/remove 1000 rows, sort 4 keys | 416.8 | 438.8 | +5.2% | ±0.4% |
| MemorySource push: add/remove 1000 rows, sort 2 keys | 424.9 | 440.9 | +4.1% | ±1.2% |
| MemorySource push: add/remove 1000 rows, sort 1 key | 426.6 | 441.4 | +3.9% | ±0.3% |

The two `Filter / push` benchmarks are the direct target and every round was
positive. The `MemorySource push` gains follow from the same cause: even a
filterless push crosses the FilterStart/FilterEnd pair.

Everything dominated by join or relationship work is flat — `Join / push edit
issue title` −0.3%, `push: add issue (with creator join)` −0.2%, the
relationship-heavy view 2.2–2.5% against ±3% noise. Those layers still suspend
for real and were untouched.

**This does not fix #6523's push regression.** `push: add comment (child
relation)` is +0.6% against ±2.4% noise — flat, not recovered.

### On the measurement

V8 could not resolve this at all. Five paired rounds on the host produced
nothing usable: `push: add issue (no join)` read −12.4% with all four of the
first rounds negative, then round five came back **+13.0%**. One
`hydrate: issues with creator` round swung +40.4%. Had I stopped at four rounds
I would have reported a win that is not there. The emulator's in-run spreads
are ±0.3–0.4% on the benchmarks that matter here, so the Android numbers are
the ones to trust.

Caveat: one `add open issue` round came in at +16.3% against +6.7/+8.4/+9.0.
The median ignores it, but treat 8.7% as approximate.

### Verification

shared 772 · zql 1447 · zqlite 198 — each down by exactly the tests removed
(two iterator-protocol tests, the `MemoryStorage.scan` test, the `scan prefix`
test). Types and lint clean.

The hydration A/B did not produce a usable result: the Android emulator was
too degraded after hours of benchmarking, with in-run noise at a median of
±23.9% and a uniform -76% swing across all ten benchmarks in one round. It
needs a cold emulator and a `pm clear` between arms, and should be re-run
together with the push suite against the new base.

### DIFF
--- a/packages/shared/src/btree-set.bench.ts
+++ b/packages/shared/src/btree-set.bench.ts
@@ -19,7 +19,8 @@ const midKey = NUM_ENTRIES / 2;
 describe('BTreeSet iterators', () => {
   bench('values() full scan', () => {
     let sum = 0;
-    for (const v of tree.values()) {
+    const it = tree.values();
+    for (let v = it.nextValue(); v !== undefined; v = it.nextValue()) {
       sum += v;
     }
     use(sum);
@@ -27,7 +28,8 @@ describe('BTreeSet iterators', () => {
 
   bench('valuesFrom() from mid', () => {
     let sum = 0;
-    for (const v of tree.valuesFrom(midKey)) {
+    const it = tree.valuesFrom(midKey);
+    for (let v = it.nextValue(); v !== undefined; v = it.nextValue()) {
       sum += v;
     }
     use(sum);
@@ -35,7 +37,8 @@ describe('BTreeSet iterators', () => {
 
   bench('valuesReversed() full scan', () => {
     let sum = 0;
-    for (const v of tree.valuesReversed()) {
+    const it = tree.valuesReversed();
+    for (let v = it.nextValue(); v !== undefined; v = it.nextValue()) {
       sum += v;
     }
     use(sum);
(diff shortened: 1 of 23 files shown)


arv:
abandoning this

PR #6526, opened 2026-09-09
fix(zero-cache): remove per-request worker listeners added by /profz

arv:
## Problem

`handleProfzRequest()` subscribed to the worker's `profileResponse` messages with `Receiver.onMessageType()`, which is a bare `on('message')` with no way to unsubscribe. The worker it targets is long-lived and shared (the memoized syncer worker in the dispatcher, or the change-streamer's sub workers), so every `/profz` and `/profrmz` request left a permanent `'message'` listener behind, each retaining the request's response map, i.e. the captured CPU profiles.

The change-streamer additionally built a fresh `broadcastWorker()` facade per request. Each one adds permanent `'message'` forwarders to every sub worker.

## Fix

- Add `subscribeToMessageType()` to `types/processes.ts`, which returns an unsubscribe function (`onMessageType` is now implemented on top of it), and use it in `handleProfzRequest()` with a `finally`.
- Create the change-streamer's broadcast profile worker once at startup instead of per request.

## Tests

- `processes.test.ts`: `subscribeToMessageType` delivers matching messages and removes the listener on unsubscribe.
- `profz.test.ts`: two `/profz` requests against an in-proc worker leave the worker's `'message'` listener count unchanged. Fails without the fix (2 leaked listeners).

Ran `lint`, `format`, `check-types`, the `no-pg` tests for `services` and `types`, and `server/change-streamer.pg.test.ts` against a local Postgres 16.

Finding 6 from the zero-cache memory-leak audit.

---

### DIFF
--- a/packages/zero-cache/src/server/change-streamer.ts
+++ b/packages/zero-cache/src/server/change-streamer.ts
@@ -381,10 +381,16 @@ export default async function runWorker(
     });
   }
 
-  const getProfileWorker =
+  // Create the broadcast facade once: each broadcastWorker() adds permanent
+  // 'message' forwarders to every sub worker, so creating one per /profz
+  // request would leak a forwarder per request.
+  const profileWorker =
     profileSubWorkers.length > 0
-      ? () => Promise.resolve(broadcastWorker(profileSubWorkers))
+      ? broadcastWorker(profileSubWorkers)
       : undefined;
+  const getProfileWorker = profileWorker
+    ? () => Promise.resolve(profileWorker)
+    : undefined;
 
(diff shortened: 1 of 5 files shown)


arv:
**Verify Signed Commit Authors** fails on e010a58 because the commit is signed with the Claude Code session's SSH key, which is not in the org allow-list (`rocicorp/.github` `.github/signing/allowed_signers`). This is not a problem with the change itself; every other check on this head is expected to pass. It cannot be fixed from the session: the branch needs to be re-signed and force-pushed with an allow-listed key (or the session key allow-listed). Same situation as #6513 to #6516, which were re-signed locally.

---

grgbkr:
Thanks!

PR #6527, opened 2026-09-09
fix(zero-cache): do not untrack a replacement service when its predecessor exits

arv:
## Problem

`ServiceRunner` removed instances from its map by id in the old instance's `finally`. When an instance became invalid (a `Mutagen` or `Pusher` whose last ref was dropped, or a stopped `ViewSyncer`) and a new instance was created under the same id while the old one was still shutting down, the old instance's completion deleted the *replacement* from the map. The next `getService()` then created a duplicate, and the untracked instance was never stopped or drained by the runner (it is not reachable via `getServices()` either).

## Fix

Only delete the map entry if it still points at the instance that finished.

## Tests

`runner.test.ts`: `replacement created during teardown stays tracked`. Fails without the fix (a third instance is created).

Ran `lint`, `format`, `check-types`.

Finding 7 from the zero-cache memory-leak audit.

---

### DIFF
--- a/packages/zero-cache/src/services/runner.test.ts
+++ b/packages/zero-cache/src/services/runner.test.ts
@@ -66,6 +66,28 @@ describe('services/runner', () => {
     expect(s1).not.toBe(s2);
   });
 
+  test('replacement created during teardown stays tracked', async () => {
+    const s1 = runner.getService('foo');
+    // The instance becomes invalid (e.g. its last ref was dropped) but is
+    // still running its (async) shutdown.
+    s1.valid = false;
+    const s2 = runner.getService('foo');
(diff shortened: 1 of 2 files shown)


arv:
**Verify Signed Commit Authors** fails on d135eb7 because the commit is signed with the Claude Code session's SSH key, which is not in the org allow-list (`rocicorp/.github` `.github/signing/allowed_signers`). This is not a problem with the change itself; every other check on this head is expected to pass. It cannot be fixed from the session: the branch needs to be re-signed and force-pushed with an allow-listed key (or the session key allow-listed). Same situation as #6513 to #6516, which were re-signed locally.

---

PR #6529, opened 2026-09-09
fix(zero-cache): release the catchup pool when the snapshot read fails

arv:
## Problem

`Storer.#startCatchup()` creates a `READONLY` `TransactionPool` and a dedicated connection pool (sized to the number of subscribers) for each catchup, and relies on the catchup tasks' completion to call `setDone()` and `end()` the connections. If the initial snapshot read (`SELECT "lastWatermark" FROM replicationState`) failed, the `catch` block failed the subscribers and rethrew without tearing the pools down. `processReadTask()` errors do not fail the pool, so its workers stayed parked, each holding an open `READ ONLY` transaction, together with their connections.

## Fix

Abort the transaction pool and end the connection pool on that path.

## Tests

`storer.pg.test.ts`: `failed snapshot read tears down the catchup pool`. It points the `subscriber-catchup` pool at a database without the CDC tables, and asserts via `pg_stat_activity` that no connection to that database survives after the failure. Without the fix one connection (idle in transaction) stays open indefinitely.

Ran `lint`, `format`, `check-types`, the full `storer.pg.test.ts` (33 tests) against a local Postgres 16, and the `no-pg` `change-streamer` tests.

Finding 9 from the zero-cache memory-leak audit.

---

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/storer.pg.test.ts
+++ b/packages/zero-cache/src/services/change-streamer/storer.pg.test.ts
@@ -1978,6 +1978,71 @@ describe('change-streamer/storer', () => {
     });
   });
 
+  describe('catchup snapshot failure', () => {
+    let catchupDB: PostgresDB;
+
+    beforeEach<PgTest>(async ({testDBs}) => {
+      // A database without the cdc tables, so that the catchup pool's
+      // snapshot read fails.
+      catchupDB = await testDBs.create('change_streamer_storer_catchup');
+      const catchupProvider: PostgresDBProvider = (applicationName, maxConns) =>
+        pgClient(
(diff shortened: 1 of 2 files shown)


arv:
**Verify Signed Commit Authors** fails on 9ddf11f because the commit is signed with the Claude Code session's SSH key, which is not in the org allow-list (`rocicorp/.github` `.github/signing/allowed_signers`). This is not a problem with the change itself; every other check on this head is expected to pass. It cannot be fixed from the session: the branch needs to be re-signed and force-pushed with an allow-listed key (or the session key allow-listed). Same situation as #6513 to #6516, which were re-signed locally.

---

PR #6532, opened 2026-09-09
chore(zero-cache): chaos test  sqlite change log

tantaman:
**tldr:** I want more confidence in the sqlite change log stuff so this simulates various failure scenarios. This was authored by Claude and is purely test code. It found a few bugs which fixes for were landed in other PRs.

Given it is only test code, I wouldn't review too closely fwiw. The main review I'd do is to check if the LLM provided description of its own code is a reasonable test. That description is below.

# RMv2 local soak — how `mlaw/soak` works

A local, reproducible end-to-end exercise of the SQLite change log in `serve`
mode against a *real* litestream v5 backup (minio/S3), with real view-syncer
restores, disconnects and reconnects. 10 commits over `main`, ~9.6k lines,
entirely under `apps/zbugs/**`. The operator's view is
[`apps/zbugs/scripts/rmv2-soak/README.md`](./apps/zbugs/scripts/rmv2-soak/README.md).

It exists because `zero-cache-compare` cannot reach the interesting paths: with
no backup URL its purge floor is pinned to the local replica head, so the log
drains as fast as it fills, `minWatermark` never has to cover anything, and
neither the snapshot-reservation path nor the demotion path can fire.

## The topology it stands up

```text
docker:  postgres:16 (6434)              minio (9000/9001) + mc bucket-init
         |                                ^
         | logical replication            | s3
         v                                |
  replication-manager  ---- litestream v5 backup ----
  :4850  (change-streamer :4851, litestream metrics :4852)
  NUM_SYNC_WORKERS=0, ZERO_APP_ID=rmv2_soak, ZERO_LOG_LEVEL=debug
  SQLITE_CHANGE_LOG_MODE=serve / read 100 / cold 100 / compare 100
         |  ws://localhost:4851/
         +--------------+--------------+
         v              v              v
      vs-0           vs-1           vs-2     each: own replica dir,
      :4860          :4870          :4880    own TASK_ID, own restore
```

Three view-syncers, not one, so a demotion of one is visibly *not* a demotion of
the others and the purge floor has more than one ack holding it.

## The run

```text
main()                                          # scripts/rmv2-soak.ts
  assertPorts / assertPrerequisites             # binaries + nobody holding :4850..
  receiver.listen(otlpPort)                     # in-process OTLP/HTTP receiver
  startInfra()                                  # docker compose: postgres + minio
  phase 0 (--baseline)                          # A/B control: change log OFF
    run every workload phase, snapshot throughput, tear down, reset the bucket
  phase 1  start RM (initial sync + first backup), start 3 view-syncers
  phases 2-5   for each of sustained / burst / quiet / fat:
    run its stages
    quiesceAndCompare(`after phase <name>`)     # <- gate runs here, not only at the end
  phase 6  mixed: for each chaos action (C10-C12 forced last):
    start a load stage underneath it            # never hit an idle system
    runChaosAction(action)
    quiesceAndCompare(`after <id>`)
  sampler.sample(); buildReport(); writeReport(); exit(pass ? 0 : 1)
```

Workload shapes (every duration `* --scale`): `sustained` 25/s, `burst`
3x 500/s with idle gaps, `quiet` (longer than `retentionMs`, so the purger has
work), `fat` 25/s at 8000-byte payloads, `mixed` a compressed cycle of the three.

## The gate

```text
quiesce(replicas):                              # oracle.ts:218
  stop writes
  INSERT sentinel  -> wait until every replica sees it
  DELETE sentinel  -> wait until every replica sees it gone
  wait until every replica reports the same stateVersion
  # PG cannot be read at a historical LSN; this is what pins both sides
  # to the same transaction set with no LSN bookkeeping

compare(replica R, table T):                    # the only gate that survives cutover
  pi(R.T) == liteRow(PG.T)   as multisets keyed by a unique key
  # pi drops _0_version and _zero.*; liteRow is the replicator's own
  # PG->SQLite mapping, so both sides canonicalize identically
  # key: shortest unique index that is *observably* non-null and unique,
  #      else whole-row multiset with counts (replicas have no PRIMARY KEY)

pass =                                          # report.ts:325
      every oracle check ok
  and no named tripwires                        # note: 'error-log' is excluded
  and no chaos action reported a finding
  and every required route has count > 0        # coverage is positive, not absence
```

Comparing the RM's own replica alongside each view-syncer's bisects for free:
RM clean + followers dirty is the change-log/catchup path; both dirty points at
the replicator or initial sync.

Two stated blind spots: a bug *inside* `liteRow` is invisible, and equality at a
bound cannot see a transient wrong state that heals.

## Where the numbers come from

```text
OTel  ->  the census of record            JSON logs  ->  events with a time
  each task exports OTLP/HTTP JSON          reseed (and why), reservation
  to an in-process receiver on a path       open/confirm, demotion, purge
  naming the task; no collector             passes, tripwires
  catchup_routes{source,reason},            needs ZERO_LOG_LEVEL=debug on the RM:
  purge probes, compare outcomes,           `serving <id> from SQLite catchup`
  log bytes, backup lag                     is a debug line
```

View-syncers stay at `info` so the burst phase's log volume does not perturb the
timings it is measuring.

## The headline measurement (§1.4)

The residual wait is a **reseed** window, not a restore window — the purge floor
is capped at the backup watermark, so only a log that *just reseeded* can sit
above it. One number would lie, so the report gives two:

```text
RM restart ─► log reseeds at S ─────────────────────────────► backup covers S
              │                                               │
              ├──────── reseedToCoveringBackupMs ─────────────┤  the exposure itself
              │         floor = litestream monitor-interval       (measured 4.53-4.68s
              │                 + one vfs poll interval            at scale)
              │
              └─ follower arrives ─► reservation open ─► confirmed
                                     ├─ reservationHoldMs ─┤     what a follower paid
                                     (p50 2ms; ~0 whenever the
                                      follower arrived late)
```

`reseedToConfirmMs` is recorded for reference only — it spans the RM's own
restart, so it overstates the stall. `reservationHoldMsC3` is the control: C3
restores against a *live* log and reads ~1ms.

## Chaos matrix

`--chaos` takes ids, `none`, or `all`; the default is C1-C8, C13, C14.

| #   | Action                                                        | Expected route / assertion                          |
| --- | ------------------------------------------------------------- | --------------------------------------------------- |
| C1  | SIGTERM a view-syncer (graceful drain), restart               | `sqlite/selected`                                   |
| C2  | SIGQUIT a view-syncer (abrupt), restart                       | `sqlite/selected`                                   |
| C3  | Kill a view-syncer, delete its replica, restart               | restore then `sqlite`; **must not demote**          |
| C4  | Kill mid-burst; a short outage, then one past retention       | stale long-gap replica discarded and restored       |
| C5  | SIGTERM the replication-manager, restart                      | a valid log resumes from its own head               |
| C6  | Delete only the change log, restart, wipe a view-syncer replica| `sqlite/selected-cold`, or `pg/cold-log` when cold reads are off |
| C7  | SIGSTOP the RM 30s, then SIGCONT                              | disconnect/reconnect, no data gap                   |
| C8  | SIGKILL the RM mid-burst, restart                             | reconcile by *truncation*, not reseed               |
| C9  | Stop minio 5 min under sustained writes, then restart         | live pages and app-scoped slot WAL grow, then drain |
| C10 | `readPercent` 100 -> 0, restart                               | every route becomes `pg/percentage`                 |
| C11 | `serve` -> `compare` -> `write`, restart each time            | writer stays, reads stop, comparison stops          |
| C12 | `write` -> `off`, restart                                     | does turning it off actually free the disk          |
| C13 | C4 and C6 together                                            | a follower already behind, meeting a reseed         |
| C14 | Wipe the RM's whole volume (replica, litestream state, log)   | restore into a fresh generation; nobody demoted     |

C1 and C2 are two actions rather than the same test twice because
`GRACEFUL_SHUTDOWN` and `FORCEFUL_SHUTDOWN` are genuinely different paths in
`life-cycle.ts`. C10-C12 always run last: they leave the log rolled back to
`write` and then `off`, which the rest of the run cannot continue from.

## Layout

```text
apps/zbugs/scripts/
├── build-litestream.sh      L1  litestream-v3, litestream-v5, vfs-query (idempotent)
├── change-log-traffic.ts    L3  the stage driver, quiet phase, residue mix
├── rmv2-soak.ts             L4  the orchestrator (above)
└── rmv2-soak/
    ├── config.ts            L4  paths, ports, the environment matrices
    ├── node.ts, cluster.ts  L4  process lifecycle and topology
    ├── infra.ts             L2  docker compose, minio, the S3 helpers
    ├── workload.ts          L3  the phase shapes
    ├── oracle.ts            L5  quiescence, liteRow canonicalization, keyed diffs
    ├── chaos.ts             L6  C1-C14
    ├── logs.ts              L7  JSON log events and tripwires
    ├── otlp.ts              L7  the in-process OTLP receiver
    ├── resources.ts         L8  disk, slot and purge sampling
    ├── report.ts            L8  the census, the gate, the report
    └── results/             recorded reports from runs worth keeping
```

## Gotchas the harness had to absorb

- **Every zero-cache worker is its own process group** (`childWorker` forks
  `detached: true`). A process-*group* signal reaches only the dispatcher, and a
  dead dispatcher orphans its change-streamer / backup-replicator / litestream /
  vfs-query — which keep holding the replica, so the next start dies with
  `SQLITE_BUSY: journal_mode = delete`. `node.ts` tracks the whole process
  *tree* and reaps it; the run refuses to start if its ports are held.
- **The app identity is isolated** (`ZERO_APP_ID=rmv2_soak`). Sharing the default
  `zero` means sharing the CDC ownership row and the replication slot: an
  ordinary local zbugs RM can null the soak's owner and kill its change-streamer.
- **The backup path is not the configured path.**
  `initializePostgresChangeSource` appends a generation id, so
  `s3://zero-replica/zbugs` is a prefix of nothing and the harness owns and
  clears the whole bucket.
- **Invoke it as `node scripts/rmv2-soak.ts --flag`.** pnpm forwards the `--`
  itself into a strict `parseArgs`, which dies with
  `ERR_PARSE_ARGS_UNEXPECTED_POSITIONAL`.
- `--retention-ms 0` is asserted `> 0` — it is the purger's retention floor, not
  a routing knob. The dial that zeroes the warm-up wait is `--cold-read-percent`.

## Recorded runs

`results/` keeps four reports with the branch, command and verdict for each. The
sequence they record is the point: the full isolated run **failed** on C9, whose
assertion demanded that the SQLite file shrink — SQLite reuses freed pages
instead of truncating. C9 was corrected to measure *live* pages before/during/
after, and to scope the WAL measurement to `rmv2_soak_0_*` slots; the corrected
run shows live usage 13.43MB -> 1.49MB and app-slot WAL 1.05MB -> 56.29MB ->
0.86MB across a minio outage, which is the bound the check was always meant to
assert.

### DIFF
--- a/apps/zbugs/.gitignore
+++ b/apps/zbugs/.gitignore
@@ -30,3 +30,7 @@ zero-schema.json
 *.db-shm
 *.db-wal
 *.db-wal2
+
+# rmv2 local soak (scripts/build-litestream.sh, scripts/rmv2-soak.ts)
+.litestream/
+.soak/
--- a/apps/zbugs/docker/docker-compose.minio.yml
+++ b/apps/zbugs/docker/docker-compose.minio.yml
@@ -1,0 +1,51 @@
+# S3-compatible backup target for the RMv2 local soak
+# (scripts/rmv2-soak.ts). Overlays the postgres compose file in this
+# directory, so bring both up under one project:
+#
+#   docker compose -f docker-compose.yml -f docker-compose.minio.yml up -d
+#
+# The bucket is created by a one-shot `mc` container that exits when it is
+# done, which is why it has no healthcheck of its own: the orchestrator waits
+# for it to exit 0.
+services:
+  minio:
+    image: minio/minio:RELEASE.2025-04-22T22-12-26Z
+    command: server /data --address ":9000" --console-address ":9001"
+    restart: unless-stopped
+    ports:
+      - 9000:9000
+      - 9001:9001
+    environment:
+      MINIO_ROOT_USER: minioadmin
+      MINIO_ROOT_PASSWORD: minioadmin
+      # A soak restarts minio (chaos action C9); do not spend startup time
+      # scanning the object store for drives that cannot have changed.
+      MINIO_SCANNER_SPEED: fastest
+    healthcheck:
+      test: ['CMD-SHELL', 'mc ready local || exit 1']
+      interval: 5s
+      timeout: 5s
+      retries: 20
+      start_period: 5s
+    volumes:
+      - zbugs_minio_data:/data
+
+  minio_init:
+    image: minio/mc:RELEASE.2025-04-16T18-13-26Z
+    depends_on:
+      minio:
+        condition: service_healthy
+    entrypoint:
+      - /bin/sh
+      - -c
+      - |
+        set -e
+        mc alias set soak http://minio:9000 minioadmin minioadmin
+        mc mb --ignore-existing soak/zero-replica
+        mc anonymous set none soak/zero-replica
+        echo "bucket soak/zero-replica ready"
+    restart: 'no'
+
+volumes:
+  zbugs_minio_data:
+    driver: local
--- a/apps/zbugs/package.json
+++ b/apps/zbugs/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc && vite build",
-    "check-types": "tsc -p tsconfig.json && tsc -p tsconfig.node.json",
+    "check-types": "tsc -p tsconfig.json && tsc -p tsconfig.node.json && tsc -p tsconfig.scripts.json",
     "check-types:client:watch": "tsc -p tsconfig.json --watch",
     "check-types:server:watch": "tsc -p tsconfig.node.json --watch",
     "db-up": "cd docker && docker compose up",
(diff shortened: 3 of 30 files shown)


tantaman:
@arv - can you approve the package-lock changes?

PR #6536, opened 2026-09-09
feat(zero-cache): cap how long a snapshot reservation may hold the ch…

tantaman:
…ange log

A reservation keeps the purge floor -- and, for as long as it is open, the purge scheduler itself, which `startSnapshotReservation` pauses -- from moving past the `minWatermark` a restoring view-syncer was promised. That is what makes a slow restore safe. It lives exactly as long as its WebSocket, and until now nothing bounded that: a client that is alive but never finishes pinned the log for as long as it stayed connected, with disk as the only limit.

A client that has *died* is already cleaned up by the socket's liveness pings, so this is aimed at the live-but-stuck case, and at a restore that is simply slower than anyone planned for.

`SnapshotReservations` now arms a timer when a reservation opens -- when the purge pause begins, not when the bounds are confirmed -- and releases the reservation if it is still open when the timer fires. The follower is not told: its `/snapshot` stream ends, which is exactly what happens when a task subscribes, and its client proceeds with the bounds it already has. What it has lost is the guarantee behind them, so its subscription may be answered with `WatermarkTooOld` and it will restore again.

That is also why the default is an hour rather than something tighter: a cap below the time a restore actually takes turns every restore into a loop. `--change-streamer-snapshot-reservation-max-age-ms` raises it for a deployment whose replicas take longer, and the warning that fires names the flag.

The expiry is reported as `result=expired` on the existing reservation duration histogram, distinct from a follower that subscribed (`closed`) or went away (`cancelled`), so the rate is visible rather than inferred.

Two things it deliberately does not do. It does not use the service's injected `setTimeoutFn`: tests drive the cleanup pass through that mock by index, and an hour-scale safety valve does not belong in the same queue. And it unrefs its timer, so a pending expiry never keeps a stopping process alive.

The soak's C16 holds a reservation for five minutes, well inside the default; its comment now says so, since past the cap the change-streamer is supposed to purge and C16 would report that as a violation.

Claude-Session: https://claude.ai/code/session_013jdBVFyXyX2KixRSjJR7ai

### DIFF
--- a/apps/zbugs/scripts/rmv2-soak/chaos.ts
+++ b/apps/zbugs/scripts/rmv2-soak/chaos.ts
@@ -1565,6 +1565,12 @@ export const CHAOS_ACTIONS: readonly ChaosAction[] = [
       // This holds a reservation without restoring anything, which is the only
       // way to ask the question without a replica big enough to take minutes
       // to download.
+      //
+      // The hold must stay under
+      // `--change-streamer-snapshot-reservation-max-age-ms` (an hour by
+      // default), which is the cap that takes a reservation back. Past it the
+      // change-streamer is *supposed* to purge, and this action would report
+      // that as a violation.
       const holdSeconds = Math.max(60, Math.round(300 * ctx.config.scale));
       const load = ctx.traffic.runStage({
         rate: 25,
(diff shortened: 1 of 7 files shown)


aboodman:
What was the motivating event for this change? Did we see vs not proceeding
while holding this lock?

a (phone)

PR #6540, opened 2026-09-09
fix(zero-cache): record rehydrated unchanged queries in the inspector by query id

arv:
## Problem

When a view-syncer restarts, `#hydrateUnchangedQueries` rehydrates the gotten queries and recorded their hydration metric and AST in the `InspectorDelegate` under the `transformationHash`. The initial hydration path in `#syncQueryPipelineSet`, `removeQuery()` and the inspector's lookups (`getASTForQuery`, `getMetricsJSONForQuery`) all use the query id. Those entries were therefore unreadable by the inspector and never removed, accumulating in the delegate's maps for the lifetime of the syncer worker.

## Fix

Key them by query id like the other path.

## Tests

`view-syncer-inspect.pg.test.ts`: `unchanged queries rehydrated on restart are recorded by query id`. Hydrates a query, stops the view-syncer, restarts it with a fresh `InspectorDelegate` (`restartViewSyncer` now returns the delegate) and checks that the AST and hydration metric are found under the query id. Fails without the fix (the AST is `undefined`).

Ran the full inspect pg test file against Postgres 16, plus `lint`, `format`, `check-types`.

---

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/view-syncer-inspect.pg.test.ts
+++ b/packages/zero-cache/src/services/view-syncer/view-syncer-inspect.pg.test.ts
@@ -32,6 +32,7 @@ import {
   nextPoke,
   permissions,
   permissionsAll,
+  restartViewSyncer,
   serviceID,
   setup,
   TEST_ADMIN_PASSWORD,
@@ -89,6 +90,9 @@ describe('view-syncer/service', () => {
   let delegate: InspectorDelegate;
   let customQueryTransformer: CustomQueryTransformer | undefined;
   let clearMocks: () => void;
+  let databaseStorage: Awaited<ReturnType<typeof setup>>['databaseStorage'];
+  let config: Awaited<ReturnType<typeof setup>>['config'];
+  let setTimeoutFn: Awaited<ReturnType<typeof setup>>['setTimeoutFn'];
 
   beforeEach<PgTest>(async ({testDBs}) => {
     ({
@@ -103,6 +107,9 @@ describe('view-syncer/service', () => {
       inspectorDelegate: delegate,
       customQueryTransformer,
       clearMocks,
+      databaseStorage,
+      config,
+      setTimeoutFn,
     } = await setup(testDBs, 'view_syncer_inspect_test', permissionsAll, {
       queryFetchMode: 'empty-validation',
     }));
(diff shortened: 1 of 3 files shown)


arv:
**Verify Signed Commit Authors** fails on 722036a because the commit is signed with the Claude Code session's SSH key, which is not in the org allow-list (`rocicorp/.github` `.github/signing/allowed_signers`). This is not a problem with the change itself; every other check on this head is expected to pass. It cannot be fixed from the session: the branch needs to be re-signed and force-pushed with an allow-listed key, as was done for the earlier batches.

---

arv:
The merge-queue run for this PR ([JS run 34466676796](https://github.com/rocicorp/mono/actions/runs/34466676796)) was dequeued because **Check Types** failed. The failure is not this PR's; it is a build-order race between two apps this diff does not touch.

What happened, from the [job log](https://github.com/rocicorp/mono/actions/runs/34466676796/job/102836947605):

- `@rocicorp/zero:build` completed at 10:34:59.
- turbo then started `zero-throughput:build`, whose script is `pnpm --filter @rocicorp/zero run build` (added in #6511), so it ran the zero build a second time. That build starts with `rm -rf packages/zero/out`.
- `zbugs:build` was running in parallel and loading `apps/zbugs/vite.config.ts` at 10:35:08, which resolves `@rocicorp/zero`, and failed with `Failed to resolve entry for package "@rocicorp/zero"` because `out/` had just been deleted.

The same `Check Types` job passed on this head (73717a2) at 06:49Z, and `pnpm --filter zero-cache run check-types` passes locally on main + this PR. The race only bites when turbo's scheduler happens to overlap those two tasks.

No fix for the race exists yet. Proposed patch, out of scope for this PR: turbo already builds `@rocicorp/zero` before `zero-throughput` via `dependsOn: ["^build"]`, so the app's own `build` script only exists for Vercel. Renaming it to `vercel-build` (which Vercel runs in preference to `build`) removes the duplicate build from turbo's graph without changing the deployment:

```diff
--- apps/zero-throughput/package.json
-    "build": "pnpm --filter @rocicorp/zero run build",
+    "vercel-build": "pnpm --filter @rocicorp/zero run build",
```

This PR needs to be re-queued; I cannot re-queue it from this session.

---

PR #6542, opened 2026-09-09
fix(zero-cache): do not retain every view-syncer from the DrainCoordinator

arv:
## Problem

`ViewSyncerService.readyState()` raced the per-instance `#initialized` promise against `DrainCoordinator.draining`, a single promise per syncer worker that never settles unless the server drains. Each `Promise.race()` attaches a reaction to that promise, and reactions are only released when the promise settles, so the coordinator accumulated one reaction (and the race closure behind it) for every view-syncer ever created.

## Fix

Replace the promise with a subscription, `DrainCoordinator.onDraining(listener)`, which returns an unsubscribe function. `readyState()` subscribes and unsubscribes as soon as `#initialized` settles either way. A listener registered after draining has started is called immediately, preserving the old promise semantics for a drain that begins before a view-syncer initializes.

## Tests

- New `drain-coordinator.test.ts`: listener fires once draining starts and is then released; fires immediately when already draining; an unsubscribed listener is released and never called.
- `view-syncer.pg.test.ts`: `readyState() releases its drain listener once initialized` (listener count goes from 1 to 0 after the first hydration) and `a drain requested before initialization stops the view-syncer` (the pre-existing behavior).

Ran the whole `view-syncer.pg.test.ts` against Postgres 16, plus `lint`, `format`, `check-types`.

---

### DIFF
--- a/packages/zero-cache/src/services/view-syncer/drain-coordinator.test.ts
+++ b/packages/zero-cache/src/services/view-syncer/drain-coordinator.test.ts
@@ -1,0 +1,55 @@
+import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest';
+import {DrainCoordinator} from './drain-coordinator.ts';
+
+describe('view-syncer/drain-coordinator', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  test('onDraining fires once draining starts', () => {
+    const coordinator = new DrainCoordinator();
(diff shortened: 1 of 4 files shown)


arv:
**Verify Signed Commit Authors** fails on aa4b584 because the commit is signed with the Claude Code session's SSH key, which is not in the org allow-list (`rocicorp/.github` `.github/signing/allowed_signers`). This is not a problem with the change itself; every other check on this head is expected to pass. It cannot be fixed from the session: the branch needs to be re-signed and force-pushed with an allow-listed key, as was done for the earlier batches.

---

PR #6543, opened 2026-09-09
fix(zero-cache): ignore the exit of a replaced vfs-query child

arv:
## Problem

`VfsWatermarkPoller` kills its `vfs-query` child when the local watermark catches up with the backup, and spawns a replacement as soon as it falls behind again. The child exits asynchronously, and its `close` handler unconditionally cleared `#remotePoller` and, since polling was wanted again, scheduled a backoff respawn. That orphaned the replacement (it could no longer be stopped by `#disableRemotePoller()`, and every later disable/enable spawned another child next to it) and started a third child after the backoff.

## Fix

The `close` handler ignores a child that is no longer the tracked poller.

## Tests

`vfs-watermark-poller.test.ts`: `a replacement spawned before the old child exits is not orphaned`. Catches up and falls behind again before the first child has exited, then checks that no third child is spawned after the backoff and that the replacement is still killed when the poller catches up. Fails without the fix (three spawns).

Ran the whole poller test file, plus `lint`, `format`, `check-types`.

---

### DIFF
--- a/packages/zero-cache/src/services/change-streamer/vfs-watermark-poller.test.ts
+++ b/packages/zero-cache/src/services/change-streamer/vfs-watermark-poller.test.ts
@@ -217,6 +217,42 @@ describe('change-streamer/vfs-watermark-poller', () => {
     expect(spawnMock).toHaveBeenCalledTimes(1);
   });
 
+  test('a replacement spawned before the old child exits is not orphaned', async () => {
+    setUp();
+    const poller = makePoller(
+      fakeVfsQuery(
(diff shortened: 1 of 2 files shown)


arv:
**Verify Signed Commit Authors** fails on af70191 because the commit is signed with the Claude Code session's SSH key, which is not in the org allow-list (`rocicorp/.github` `.github/signing/allowed_signers`). This is not a problem with the change itself; every other check on this head is expected to pass. It cannot be fixed from the session: the branch needs to be re-signed and force-pushed with an allow-listed key, as was done for the earlier batches.

---

PR #6544, opened 2026-09-09
fix(zero-cache): close a pipelined stream on an unexpected ack instead of rejecting unhandled

arv:
## Problem

In `streamOut`'s pipelined path the ack for each message is awaited in a void-ed async closure. A mismatched ack threw inside that closure, which became an unhandled rejection (crashing the process under Node's default policy) while the stream itself stayed open. This is a crash path rather than a leak, but it came out of the same audit.

## Fix

Route the error to the `WebSocketCloser` like the synchronous path does, so the socket is closed and the source canceled.

## Tests

`streams.test.ts`: `pipelined: an unexpected ack closes the stream`. A raw client acks the first message with the wrong id; the socket must close, the message must be reported unconsumed, and no `unhandledRejection` may fire. Without the fix the socket never closes and the test times out.

Ran the whole streams test file, plus `lint`, `format`, `check-types`.

---

### DIFF
--- a/packages/zero-cache/src/types/streams.test.ts
+++ b/packages/zero-cache/src/types/streams.test.ts
@@ -388,6 +388,33 @@ describe('streams with internal acks', () => {
     ]);
   });
 
+  test('pipelined: an unexpected ack closes the stream', async () => {
+    // A bad ack used to throw inside a void-ed async closure, which surfaced
+    // as an unhandled rejection instead of closing the stream.
+    const unhandled: unknown[] = [];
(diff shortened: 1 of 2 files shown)


arv:
**Verify Signed Commit Authors** fails on 002334d because the commit is signed with the Claude Code session's SSH key, which is not in the org allow-list (`rocicorp/.github` `.github/signing/allowed_signers`). This is not a problem with the change itself; every other check on this head is expected to pass. It cannot be fixed from the session: the branch needs to be re-signed and force-pushed with an allow-listed key, as was done for the earlier batches.

---

PR #6546, opened 2026-09-10
fix(replicache): fall back to MemStore when IndexedDB fails to open

Karavil:
## Problem

On WebKit, `indexedDB.open` can reject with `UnknownError: Unable to open database file on disk` or `UnknownError: Error creating Records table (13) - database or disk is full` (SQLITE_FULL). Both come from `SQLiteIDBBackingStore::getOrEstablishDatabaseInfo` in WebKit ([SQLiteIDBBackingStore.cpp](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/indexeddb/server/SQLiteIDBBackingStore.cpp)). We see both in production on iOS WKWebView and macOS Safari.

`IDBStore` starts `openDatabase()` in its constructor and nothing awaits or catches that promise, so the page gets an unhandled rejection per store. Every later `read()` and `write()` awaits the same rejected promise and rejects again, so `ReplicacheImpl` never finishes `#open()` and the client never becomes ready.

`IDBStoreWithMemFallback` already exists for exactly this shape of failure, but `newIDBStoreWithMemFallback` only wrapped on Firefox and `isFirefoxPrivateBrowsingError` only matched the private-browsing `InvalidStateError`.

## Fix

- When the initial `indexedDB.open` fails, `IDBStore` rejects `read()` and `write()` with a new exported `IDBOpenError` whose `cause` is the browser's error. A failure after the database opened (a transaction error, a reopen) is not wrapped. `IDBStore` also attaches a `catch` to the initial open, which removes the unhandled rejection.
- `IDBStoreWithMemFallback` switches to a `MemStore` when a `read()` or `write()` fails with an `IDBOpenError`. It logs once at `warn` (`Switching to MemStore because IndexedDB failed to open`, with the error), and every concurrent caller whose call failed that way retries on the memory store. Any other error is rethrown, so a transaction error on a database that did open never replaces a populated store with an empty one.
- The Firefox private-browsing special case is removed. Firefox fixed that bug, and its error also comes from `open`, so it would now take the same fallback.
- `newIDBStoreWithMemFallback` wraps on every browser.
- `dropIDBStoreWithMemFallback` falls back to `dropMemStore` when `deleteDatabase` rejects with a `DOMException` (the previous `try`/`catch` around a promise-returning call could not catch anything).

## Exposing the fallback

Apps need to be able to tell when they are running in memory:

- The KV `Store` interface gets an optional `kind`. `IDBStore` reports `'idb'` and `MemStore` `'mem'`. `IDBStoreWithMemFallback` reports its current store, so it changes from `'idb'` to `'mem'` after the switch. `SQLiteStore` takes `kind` as an optional last constructor argument (default `'sqlite'`); the op-sqlite, expo-sqlite and zero-sqlite providers pass `'op-sqlite'`, `'expo-sqlite'` and `'zero-sqlite'`. Custom stores can leave it unset.
- `Replicache` and `Zero` expose `kvStore`, so `zero.kvStore.kind === 'mem'` means IndexedDB fell back to memory. The switch happens on the first read or write, so check after the instance is ready.
- `idbName` and `dropDatabase` are unchanged.

## Tests

`idb-store-with-mem-fallback.test.ts`:
- open rejection with each WebKit message: the store serves reads and writes from memory, `kind` goes from `'idb'` to `'mem'`, and the switch is logged once at `warn` with an `IDBOpenError` whose `cause` is the browser error;
- two concurrent first writes on a store whose open failed both succeed and log once;
- a transaction error after a successful open is rethrown, the data written before it is still there, and nothing is logged;
- the wrapper is installed on a non-Firefox user agent.

`idb-store.test.ts`: when the initial open fails, `read()` rejects with an `IDBOpenError` and `write()` with the browser error as `cause`.

`replicache.test.ts`: `kvStore.kind` is `'idb'` by default and `'mem'` with `kvStore: 'mem'`; a custom provider's store is returned as-is.

Ran `lint`, `format` and `check-types` for replicache and zero-client, and the `idb-store`, `idb-store-with-mem-fallback`, `replicache` and SQLite store test files.

### DIFF
--- a/packages/replicache/src/kv/idb-store-with-mem-fallback.ts
+++ b/packages/replicache/src/kv/idb-store-with-mem-fallback.ts
@@ -1,6 +1,5 @@
 import type {LogContext} from '@rocicorp/logger';
 import {navigator} from '../../../shared/src/navigator.ts';
-import {promiseVoid} from '../../../shared/src/resolved-promises.ts';
 import {IDBStore} from './idb-store.ts';
 import {MemStore, dropMemStore} from './mem-store.ts';
 import type {Read, Store, Write} from './store.ts';
@@ -8,8 +7,15 @@ import type {Read, Store, Write} from './store.ts';
 /**
  * This store uses an {@link IDBStore} by default. If the {@link IDBStore} fails
  * to open the DB with an exception that matches
- * {@link isFirefoxPrivateBrowsingError} we switch out the implementation to use
- * a {@link MemStore} instead.
+ * {@link isFirefoxPrivateBrowsingError}, or if `indexedDB.open` itself rejects
+ * (WebKit's "Unable to open database file on disk" and "Error creating Records
+ * table (13) - database or disk is full"), we switch out the implementation to
+ * use a {@link MemStore} instead. Only the open failure qualifies, matched by
+ * identity through {@link IDBStore.openError}: a transaction error on a
+ * database that did open is rethrown, so a store that has data is never
+ * swapped for an empty one behind the caller's back. Every caller whose call
+ * failed with that open error retries on the memory store, so concurrent
+ * first calls all succeed and the switch is logged once.
  *
  * The reason this is relatively complicated is that when {@link IDBStore} is
  * created, it calls `openDatabase` synchronously, but that returns a `Promise`
@@ -21,11 +27,13 @@ import type {Read, Store, Write} from './store.ts';
 export class IDBStoreWithMemFallback implements Store {
   readonly #lc: LogContext;
   readonly #name: string;
+  readonly #idbStore: IDBStore;
   #store: Store;
   constructor(lc: LogContext, name: string) {
     this.#lc = lc;
     this.#name = name;
-    this.#store = new IDBStore(name);
+    this.#idbStore = new IDBStore(name);
+    this.#store = this.#idbStore;
   }
 
   read(): Promise<Read> {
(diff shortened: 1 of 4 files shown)


packages/replicache/src/kv/idb-store-with-mem-fallback.ts
```diff
@@ -84,24 +103,16 @@ export function newIDBStoreWithMemFallback(
...
-    return dropIDBStore(name);
-  }
-  try {
-    return dropIDBStore(name);
-  } catch (e) {
-    if (isFirefoxPrivateBrowsingError(e)) {
+  return dropIDBStore(name).catch((e: unknown) => {
+    if (e instanceof DOMException) {
       return dropMemStore(name);
     }
```
arv, line 113: This seems a bit too lax. Are we sure we have a mem store in all the cases where we get a DOMException in dropIDBStore?

arv (dismissed):
I like it.

Let's get a second pair of eyes on this as well.

@aboodman @grgbkr

aboodman:
Agree, good idea.

packages/replicache/src/kv/idb-store-with-mem-fallback.ts
grgbkr: If its not there its a no-op, so seems fine.

arv:
One thing we could do is to try to create an IDB instance of a second name. We could either store that in localStore or always try to use `-fallback` name... I'll play with that

PR #6561, opened 2026-09-11
feat(zero-client): add an initializing connection state before connecting

arv:
## The defect

Zero's first connect attempt starts by loading the local store: the replica is opened and loaded into the IVM sources before the cookie can be read, and then the client group ID and active clients are read. On a slow Android device with a few thousand rows that takes 23–33 s. All of it happened in `connecting`:

- `ConnectionManager` started the 60 s connecting window in its constructor, so local loading used up the time the server was supposed to get before Zero reported `disconnected`.
- The setup deadline from #6475 expired during the load, so the first few attempts timed out with `ConnectTimeout` against a server that had not been asked anything yet.
- From outside, a loading store and an unreachable server looked the same. Host availability logic that starts a timer at the first `connecting` showed "service down" on every cold launch while production was healthy.

## The change

A new `initializing` state comes before `connecting`, on both `ConnectionManager` and the public `ConnectionState`.

- `ConnectionManager` starts in `initializing`, with no connecting window and no timeout interval.
- The run loop waits for the local store before it starts: the cookie (which Replicache only yields once the replica is loaded), the client group ID and the active clients. It then calls the new `initialized()`, which moves to exactly the state the constructor used to create: `connecting`, `attempt: 0`, window starting now.
- `initialized()` does nothing if something else, like `close()`, has already moved the state on.
- If loading fails, the error is not handled there. The first connect attempt awaits the same promises and reports the failure through the usual disconnect path, as before.
- `initializing` is entered once per instance. Reconnects never go back to it, because the store is already loaded.
- Refresh stays disabled in `initializing`, as it was while this window was `connecting`. Mutations are not rejected in it.

Hosts can now tell a loading replica from an unreachable server by starting their timers at `connecting` instead of at creation.

## Compatibility

`ConnectionState` gains a member, so user code with an exhaustive `switch` on `state.name` will fail to compile until it handles `'initializing'`. A non-breaking alternative would have been a `phase` field on `connecting`. I went with a separate state because the two phases fail for unrelated reasons and should read differently.

## Trade-off

A local store that never finishes loading now stays in `initializing` forever with no retries, where it used to retry every 10 s and eventually report `disconnected`. The retries could never succeed, since every attempt awaited the same promise, and `disconnected` blamed the network. Hosts that want to cap it can put their own timeout on `initializing`.

## Tests

- `ConnectionManager`: tests that the constructor starts in `initializing` with no timer and never times out there, that `initialized()` starts the window when it is called, respects a custom timeout, and does nothing once the state has moved on. Existing tests that assumed a `connecting` start now call `initialized()` first.
- New: slow initialization (longer than both the setup deadline and the connecting window) stays `initializing`, produces no connect timeouts, then connects.
- New: a hung initialization stays `initializing` and makes no connect attempt.
- The #6475 setup-deadline tests used to make `ActiveClientsManager.create` slow or hang, and that is now part of initialization. They now stall `getDeletedClients` inside socket preparation instead, which the setup deadline still covers.
- `slow setup does not spend the server acknowledgement budget` was in `zero.test.ts` twice, verbatim. The copy is removed.

### DIFF
--- a/packages/zero-client/src/client/connection-manager.test.ts
+++ b/packages/zero-client/src/client/connection-manager.test.ts
@@ -42,15 +42,19 @@ describe('ConnectionManager', () => {
 
   describe('constructor', () => {
     test('starts in initializing state without a timeout', () => {
-      using setIntervalSpy = vi.spyOn(globalThis, 'setInterval');
-      const manager = new ConnectionManager({
-        disconnectTimeout: DEFAULT_TIMEOUT_MS,
-      });
+      const setIntervalSpy = vi.spyOn(globalThis, 'setInterval');
+      try {
+        const manager = new ConnectionManager({
+          disconnectTimeout: DEFAULT_TIMEOUT_MS,
+        });
 
-      expect(manager.state).toEqual({name: ConnectionStatus.Initializing});
-      expect(manager.shouldContinueRunLoop()).toBe(true);
-      expect(manager.isInTerminalState()).toBe(false);
-      expect(setIntervalSpy).not.toHaveBeenCalled();
+        expect(manager.state).toEqual({name: ConnectionStatus.Initializing});
+        expect(manager.shouldContinueRunLoop()).toBe(true);
+        expect(manager.isInTerminalState()).toBe(false);
+        expect(setIntervalSpy).not.toHaveBeenCalled();
+      } finally {
+        setIntervalSpy.mockRestore();
+      }
     });
 
     test('never times out while initializing', () => {
--- a/packages/zero-client/src/client/connection-manager.test.ts
+++ b/packages/zero-client/src/client/connection-manager.test.ts
@@ -41,29 +41,93 @@ describe('ConnectionManager', () => {
   };
 
   describe('constructor', () => {
-    test('starts in connecting state with default timeout', () => {
+    test('starts in initializing state without a timeout', () => {
+      const setIntervalSpy = vi.spyOn(globalThis, 'setInterval');
+      try {
+        const manager = new ConnectionManager({
+          disconnectTimeout: DEFAULT_TIMEOUT_MS,
+        });
+
+        expect(manager.state).toEqual({name: ConnectionStatus.Initializing});
+        expect(manager.shouldContinueRunLoop()).toBe(true);
+        expect(manager.isInTerminalState()).toBe(false);
+        expect(setIntervalSpy).not.toHaveBeenCalled();
+      } finally {
+        setIntervalSpy.mockRestore();
+      }
+    });
+
+    test('never times out while initializing', () => {
+      const manager = new ConnectionManager({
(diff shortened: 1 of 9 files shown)


packages/zero-client/src/client/connection-manager.test.ts
```diff
@@ -41,29 +41,93 @@ describe('ConnectionManager', () => {
   };
 
   describe('constructor', () => {
-    test('starts in connecting state with default timeout', () => {
+    test('starts in initializing state without a timeout', () => {
+      const setIntervalSpy = vi.spyOn(globalThis, 'setInterval');
```
arv: you can do `using setIntervalSpy = ...` instead of try/catch

packages/zero-client/src/client/zero.ts
```diff
@@ -2196,6 +2198,24 @@ export class Zero<
...
+    // active clients. On a slow device with a large replica this takes tens of
+    // seconds, and none of it depends on the server, so it happens in
+    // `initializing` instead of spending the connecting window and the setup
+    // deadline of the first attempts.
+    try {
+      await Promise.all([
+        this.#rep.cookie,
+        this.clientGroupID,
+        this.#activeClientsManager,
+      ]);
```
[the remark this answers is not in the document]
arv: Fixed in 4f71151bb: the initialization wait now races the close signal and returns from the run loop if close() wins. Added a test that closes during a hung initialization.

packages/zero-client/src/client/connection-manager.ts
```diff
@@ -207,6 +204,38 @@ export class ConnectionManager extends Subscribable<ConnectionManagerState> {
...
+
+    const now = Date.now();
+    this.#connectingStartedAt = now;
+    this.#state = {
+      name: ConnectionStatus.Connecting,
+      attempt: 0,
+      disconnectAt: now + this.#disconnectTimeout,
+    };
+    const nextStatePromise = this.#publishStateAndGetPromise();
+    this.#maybeStartTimeoutInterval();
```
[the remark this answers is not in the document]
arv: Fixed in 4f71151bb: initialized() and both connecting() branches now start the timeout interval before publishing, so a subscriber that closes synchronously leaves no interval behind. Added tests for both paths.

packages/zero-client/src/client/connection-manager.test.ts
arv: Done in 1f281716e. zero-client has `types: []`, so this also adds `ESNext.Disposable` to its tsconfig `lib` to make `using` type-check.

arv: Reverted in 0c1558397: `using` on the vitest spy throws "Object is not disposable" in WebKit, which only runs in CI (Test Shard 2/3). Back to try/finally, and the ESNext.Disposable lib is dropped again.

PR #6569, opened 2026-09-12
ci(release): build dev images on native runners

cesara:


arv:
Are we now building both amd64 and arm64?

cesara:
> Are we now building both amd64 and arm64?

we were before but were using qemu for arm64 which was taking roughly 40-50 minutes. we now have it building in an arm64 container and runs in parallel

arv:
👍🏼 

I thought I made that change before 🤷🏼

PR #6576, opened 2026-09-14
perf(zql): return delegated iterables instead of tail yield* generators

arv:
## What

Generator functions whose only `yield` is a `yield*` in tail position are now plain functions that `return` the delegated iterable. Paths that previously yielded nothing return a shared empty sentinel (`emptyArray`, or the new `emptyIterator`).

Converted: `Filter`/`FilterStart`/`FilterEnd` push, fetch and filter, `FanIn.filter`/`fanOutDonePushingToAllBranches`, `UnionFanIn.push`/`fanOutDonePushing`, `Cap#pushEditChange`, `DeferredInput.fetch`, `FlippedJoin#fetchChunked`, `maybeSplitAndPushEditChange`, `MemorySource`/`TableSource.genPush`, `generateWithOverlay(Unordered)`, `generateRows`, Take's `replaceBoundAndForwardChange`, `getChildNodes`, `iterateMutators`, and `hydrate`/`hydrateInternal`.

`shared/src/iterables.ts` gains `emptyIterator` and `makeEmptyIteratorWithReturn(value)`: a stateless, shareable empty `Generator` with a return value. It's used where a `IterableIterator<'yield', boolean>` is required (`FilterEnd.filter` → `true`, `Filter.filter` short-circuit → `false`). Tests cover both the matching and the differing behaviour.

## Why

Every `yield*` layer costs a generator resume per item flowing through it. Many IVM operators were pure pass-through wrappers, so hot fetch/push paths stacked several of these layers.

## Reviewer notes

- **Evaluation timing changes.** A generator body runs on the first `next()`, whereas a plain function runs its prelude (asserts, `#setTakeState`, `#accumulatedPushes.push`, `computeOverlays`, `Streamer.accumulate`) at call time. Every caller I found consumes the result immediately (`yield*` / `for…of`), so this should be behaviour-neutral. It's still worth a look anywhere a stream is created and iterated later, like the `chunkStreams` array in `FlippedJoin#fetchChunked`.
- `getChildNodes` now returns `Iterable<ViewNode>` instead of `Generator<ViewNode>`.

## Benchmarks

A/B against `a8fda3a2b`, with before and after arms alternated to cancel drift.

| | iOS / Hermes (Release, simulator) | V8 / Node (`vitest.config.bench.ts`) |
|---|---|---|
| Arms | 2 before / 2 after | 4 before / 4 after (ABBA ×2) |
| Benchmarks | 47 | 66 (in-memory + SQLite flipped-join / zero-throughput) |
| Geomean | **−5.9%** | **−5.5%** |
| Faster in every arm | 33 | 36 |
| Slower in every arm | 2 (+0.7%, code paths not touched) | 2 |

Highlights (similar on both engines):
- Filter fetch open issues: **−40% (Hermes) / −48% (V8)**
- MemorySource fetch scans: −16 to −23%
- Join fetch and relationship hydration, including the heavy wide+deep case: −10 to −14%
- Hydration −5 to −15%, cold boot −3 to −5%, push −2 to −9%
- V8 only: zero-throughput memory push −9 to −10%, flipped-join fetches −2 to −5%

Regression to be aware of: `ivm-memory › push: add comment (child relation)` is **+13% on V8** (about 4.95 → 5.6 µs median) and −1.5% on Hermes. On V8 the minimum improved about 4.5% and p99 is unchanged, which looks like a JIT/inlining shift rather than added work, but it hasn't been profiled yet.

Unrelated, pre-existing on `main`: `ivm-memory › push: add issue (with creator join)` exceeds vitest's 60 s test timeout under the bench config.

## Testing

- `check-types` and `lint` pass for `zql`, `shared`, `zqlite` and `zero-cache`; `oxfmt --check` passes.
- `zql` IVM tests: 815 passed. `shared` iterables tests: 90 passed.


packages/zql/src/ivm/memory-source.ts
```diff
@@ -1086,14 +1086,14 @@ function compareBounds(a: Bound, b: Bound): number {
...
   data: BTreeSet<Row>,
   scanStart: RowBound | undefined,
   reverse: boolean | undefined,
 ) {
-  yield* data[reverse ? 'valuesFromReversed' : 'valuesFrom'](
-    scanStart as Row | undefined,
-  );
+  return reverse
+    ? data.valuesFromReversed(scanStart as Row | undefined)
+    : data.valuesFrom(scanStart as Row | undefined);
```
arv: The form `obj[expr]()` cannot be optimized in Hermes.

PR #6583, opened 2026-09-15
refactor(zero-cache): move change-source init logic into separate file

darkgnotic:
This separates the logic for restoring / initializing the replication slot and local replica (now in its own `change-source-init.ts` file) from the runtime logical replication ChangeSource logic in `change-source.ts`.

Pure refactoring, no code changes.


arv (commented):
LGTM

PR #6586, opened 2026-09-15
feat(zero-server): let adapters set the transaction isolation level

pawarren:
Mutator transactions are opened with a bare `BEGIN`, so they run at the database default — `read committed` on a stock Postgres. There is no way to ask for anything stricter through the adapter, and for some connection setups there is no way to ask for it at all: the `options` startup parameter (`-c default_transaction_isolation=…`) is rejected by PgBouncer, and a session-level `SET` is unreliable through a transaction pooler, which leaves a role default as the only lever — a database-wide change to get a per-adapter property.

This matters because `read committed` is too weak for a mutator that reads a row and then writes a value derived from that read. Two concurrent runs both read the old value and both write, neither raises an error, and one write is silently lost — so there is nothing for a retry to catch. Measured on Postgres 17.6, two runs of a read-then-write mutator body each adding 1 to `qty = 1`:

```
READ COMMITTED    A: UPDATE 1   B: UPDATE 1   final qty = 2   (correct answer is 3, no error from either)
REPEATABLE READ   A: ERROR: could not serialize access due to concurrent update
                  B: UPDATE 1   final qty = 2, A retryable
```

At `repeatable read` the conflict surfaces as a serialization failure, which is retryable. At `read committed` it is invisible.

## What this adds

An optional `TransactionOptions` argument carrying `isolationLevel`, threaded to however each adapter opens its transaction:

| adapter | how the level is applied |
| --- | --- |
| Kysely | `.transaction().setIsolationLevel(level)` |
| postgres.js | `sql.begin('isolation level <level>', fn)` |
| drizzle | transaction config `{isolationLevel}` |
| Prisma | `$transaction(fn, {isolationLevel})`, mapped to Prisma's PascalCase spelling |
| node-postgres | `BEGIN ISOLATION LEVEL <LEVEL>` |

```ts
const zql = zeroKysely(schema, db, {isolationLevel: 'repeatable read'});
```

`IsolationLevel` is a closed union (`'read committed' | 'repeatable read' | 'serializable'`) spelled as Postgres spells it, so it is never a source of injection and can be passed through verbatim where an adapter takes it that way. Both types are exported from `mod.ts`.

## Compatibility

Fully backwards compatible — with no option given, the transaction is still opened with a bare `BEGIN` and every existing call site is unchanged.

Two type declarations were widened to carry the config through: `DrizzleDatabase`'s `transaction` config moved from `never` to a structural mirror of drizzle's own `PgTransactionConfig` (so a real drizzle database stays assignable under `exactOptionalPropertyTypes`), and `PrismaClientLike.$transaction` gained its optional options parameter.

## Verification

- `tsc --noEmit` on `packages/zero-server`: 0 errors, against a 0-error baseline on `main`.
- `oxlint --quiet --type-aware`: clean. `oxfmt` applied.
- 12 new tests — one per adapter for the option set and unset, plus the `beginStatement` renderer. Existing `kysely.test.ts` / `drizzle.test.ts` still pass.

The DB-backed `adapters.pg.test.ts` was not run locally (it needs a live Postgres via `testDBs`); happy to add a round-trip assertion there against `current_setting('transaction_isolation')` if you'd like the coverage.

## Context

This came out of a discussion with @aboodman about making custom mutators converge under concurrent writes — the plan being to retry a conflicted mutator and let the re-run's read reach the right answer. That plan needs the mutator transaction to be at `repeatable read` or higher for read-derived writes, and right now there is no adapter-level way to ask for it.


pawarren:
Closing this — @aboodman is right that it can already be done on the connection, and I got a key claim in the description wrong.

The description said a session-level `SET` was unreliable through a transaction pooler. I carried that over from a different problem (the libpq `options` startup parameter, which PgBouncer genuinely does reject) without testing the `SET` path. It works. Measured against a Supabase Supavisor pooler on port 6543, five sequential pool checkouts, with `pool.on('connect', c => c.query('SET SESSION CHARACTERISTICS AS TRANSACTION ISOLATION LEVEL REPEATABLE READ'))`:

```
["repeatable read","repeatable read","repeatable read","repeatable read","repeatable read"]
```

Both routes work on a direct connection too:

```
baseline (no option)                        -> read committed
options: '-c default_transaction_isolation=repeatable\ read'  -> repeatable read   (needs the escaped space)
pool.on('connect') SET                      -> repeatable read
```

So the premise that there was no way to ask for a stricter level was simply wrong, and the right answer is documentation rather than API surface.

The one thing the adapter option did that a connection-level setting cannot: it emits `BEGIN ISOLATION LEVEL …` per transaction, so it can put mutators at `repeatable read` while other consumers sharing the same pool stay at `read committed`. In our case a single `pg.Pool` backs both the mutator provider and the server-side query providers, so the pool-level setting raises the level for reads as well. That is harmless for us — a plain `SELECT` does not take a serialization failure at `repeatable read` — and where it is not harmless, two pools solve it. Thin reason for an API.

Leaving the measurement here in case it is useful for the docs. The thing that would actually help, from the thread this came out of: a mutator that reads and then writes a derived value needs `repeatable read` or higher to be *correct*, not merely to be consistent, because at `read committed` two concurrent runs both read the old value, both write, neither errors, and one write is silently lost. That is worth a line in the custom-mutator docs next to the retry behaviour, since nothing in the API surface hints at it.

Sorry for the noise.

PR #6587, opened 2026-09-16
chore(zero-cache): finish renaming replicas.version to replicas.generation

darkgnotic:
With the `generation` column rolled out in v1.10, it can be made `NOT NULL` in v1.11. 
Additionally, stop reading the `version` column in v1.11, so that it can safely be dropped in v1.12.


arv:
This broke zmail https://rocicorp.slack.com/archives/C013XFG80JC/p1789557115239649

arv (commented):
LGTM

PR #6592, opened 2026-09-16
test(shared): make Queue drain perf test robust to slow runners

arv:
`Queue > large queue drains efficiently (O(n) not O(n^2))` asserted an absolute 200ms bound and flaked on CI (`expected 202.42 to be less than 200`, run 35106774868 / Test (Shard 2/3)).

Instead, compare drain times for the same total number of items: 10 queues of 10k vs one queue of 100k. A linear dequeue gives a ratio of ~1, and an O(n^2) `Array.shift()` regression gives ~10. The test fails when the ratio reaches 5.

- Timings are averaged over rounds totalling ≥50ms, so coarse browser timers (1ms in WebKit) can't yield 0/0.
- The large drain stops once it exceeds the ratio budget, so a regression fails in ~140ms instead of hitting the test timeout. I checked this by temporarily swapping in an `Array.shift()` buffer: it fails in node and chromium with a ratio of 25–32. WebKit's `shift()` is not quadratic, so that stand-in still passes there.

Verified: node, chromium, firefox, webkit pass; lint/format/check-types clean.


arv:
Addressed the suppressed Copilot comment about the O(n²) path hitting the test timeout: the workload is now 10×10k vs 1×100k, and the large drain stops as soon as it goes over the ratio budget. With an `Array.shift()` queue swapped in, the test now fails in ~140ms (ratio ~25–32 vs limit 5) instead of timing out. Timings are also averaged over rounds totalling ≥50ms, because WebKit's 1ms timer resolution produced 0/0. Passes in node, chromium, firefox, and webkit.

PR #6610, opened 2026-09-18
fix(zbugs): load the issue creator in the list query so the detail pa…

aboodman:
…ge renders locally

The issue page returns null until the issue's creator is present, and the list query did not load creators. Clicking an issue right after the list rendered therefore waited on the server's issue-detail response, which is queued behind the issuePreloadV2 hydration on the client group's lock, so the page stayed blank for over a second.

Loading the creator with the list rows lets the detail page render from local data immediately. Measured locally against a 250k-issue dataset with a scripted click as soon as the first row appears: click-to-render dropped from 1.1-1.3s to about 0.1s.

Claude-Session: https://claude.ai/code/session_01Q7YmJ3gTGoRSfw4P6rsQTa


arv (commented):
LGTM

PR #6613, opened 2026-09-18
chore(rmv2): high availability replication mode

darkgnotic:
High availability replication mode, allowing multiple concurrent replication-managers.

### Status

* 🚧  Experimental, not yet production ready
* Enabled via a hidden option `ZERO_UPSTREAM_PG_HIGH_AVAILABILITY_REPLICATION`

### Overview

Enables running concurrent replication-managers with dedicated replication slots.
* replication-managers converge on the latest **generation** (i.e. initial-sync version) within a given epoch, coordinating to ensure that only one initial-sync occurs at a time.
  * resyncs can be reliably forced by incrementing the `ZERO_UPSTREAM_PG_REPLICATION_EPOCH`
* typical RM startup involves **forking** a currently active replica by creating a new slot and waiting for an active replica's `confirmed_flush_lsn` (and thus litestream v5 backup) to exceed the new slot's LSN
  * LSN advancement is guaranteed by including the `replicas` table in the metadata publication
* when there are no active replicas to fork from (e.g. task crashed), **resumption** is performed as a fallback, claiming an inactive slot after a grace period
* old slots are cleaned up via https://github.com/rocicorp/mono/pull/6568


aboodman:
💪

PR #6620, opened 2026-09-21
perf(zql): eliminate cascading backfills on push

grgbkr:
## Context

Follows #6617 (`TakeGate`). While #6617 capped parent push iterations in `Join`/`FlippedJoin` to the query limit, `Take` still performed eager, synchronous forward refills upon receiving each `REMOVE`.

When multiple rows were invalidated within a single transaction or child change:
1. **$O(K)$ Interleaved Refills**: Removing $K$ rows caused $K$ independent single-row forward seeks rather than a single batch fetch.
2. **Re-entrant Fetches**: In correlated queries (e.g. `EXISTS`), backfill fetches re-entered `FlippedJoin` before pending removals had finished propagating, creating intermediate state hazards and redundant lookups.

This PR introduces two-phase push in ZQL IVM to defer refills until invalidations have propagated.

## Changes

- **Two-Phase Push Engine**:
  - **Phase 1 (Change Propagation)**: `Take.push(REMOVE)` no longer performs forward refills inline. It marks dirty partitions, updates bounds, and forwards `REMOVE` changes downstream without fetching replacements.
  - **Phase 2 (Topological Reconciliation)**: `MemorySource.genPush` triggers `reconcile()` while the connection's push overlay is active. `Take.reconcile()` sweeps dirty partitions, opens `TakeGate`, and refills the entire deficit (`limit - size`) in a single batch fetch starting from the current bound.
- **Topological Invariant**: Because ZQL pipelines are unidirectional DAGs from source to sinks, initiating `reconcile()` at the modifying source guarantees upstream `Take` nodes reconcile and push `ADD`s before downstream `Take` nodes resolve their deficits. Exactly one push phase and one reconcile phase reaches steady state.
- **Operator Pass-Through**: Added `reconcile?(pusher: InputBase): Stream<'yield'>` to the `Output` interface and implemented transparent forwarding across intermediate operators (`Join`, `FlippedJoin`, `Filter`, `Skip`, `Cap`, `Exists`, `UnionFanOut`, `UnionFanIn`, `TakeGate`).

## Verification

- **Multi-Phase Push Batching Test**: Added tests in `packages/zql/src/ivm/runaway-push.test.ts` verifying that multiple parent removals emit all `REMOVE` events in Phase 1 before deficit rows are refilled in a single batch in Phase 2 (`[remove, remove, remove, add, add]`), with zero re-entrant fetches into `FlippedJoin`.
- **Snapshot Transformation Verification**: Validated in `exists.flip.push.test.ts` that the intermediate re-entrant fetch log during `FlippedJoin` push is completely eliminated, and in `take.push.test.ts` that refill fetches cleanly move from being interleaved during push to executing within Phase 2.


tantaman:
Found a few test cases which this fails. Cleaning them up and putting them up.

tantaman:
n/m, your commit 0c2be2430 fixed it.

tantaman:
ok, @grgbkr check these tests out: https://github.com/rocicorp/mono/pull/6620/changes/ac617c69f7cd807b05b78712735e145a6144f561

tantaman:
@grgbkr - https://github.com/rocicorp/mono/pull/6620/changes/1d041c64c28e11bbb105654d3ef845e02af1e7a1 expands the fuzzer. It currently has 30 failures from other cases besides the one I flagged.

PR #6624, opened 2026-09-21
Add worker-wide snapshot row cache for IVM pipeline advancement

aboodman:
## Summary

Introduces a worker-wide cache for replica rows read during snapshot diff computation, allowing multiple client groups (view-syncers) on the same sync worker to share row lookups. This reduces redundant SQLite reads when advancing IVM pipelines.

## Key Changes

- **New `SnapshotRowCache` class** (`snapshot-row-cache.ts`): A bounded, FIFO-evicting cache that stores rows read from replica snapshots. Entries are keyed by SQL statement, arguments, and version tags to ensure correctness across different snapshot versions. Includes comprehensive test coverage.

- **Updated `Snapshotter` and `Diff` classes** (`snapshotter.ts`):
  - Accept an optional `SnapshotRowCache` instance in the constructor
  - Pass the cache to `getRow()` and `getRows()` methods with appropriate version tags
  - Disable caching for diffs that straddle table-wide operations (RESET/TRUNCATE) to maintain correctness
  - Added `hasTableWideOpSince()` method to detect such operations
  - Refactored row lookup SQL generation with `selectPrefix()` memoization to reduce string rebuilding

- **Integration in `syncer.ts`**: Creates a single `SnapshotRowCache` instance per worker and passes it to all view-syncer factories, enabling cross-client-group sharing.

- **Configuration** (`zero-config.ts`): Added `snapshotRowCacheSize` option (default 50,000 entries) to control cache capacity, with 0 disabling the cache entirely.

- **Comprehensive test coverage** (`snapshotter.test.ts`, `snapshot-row-cache.test.ts`): Tests verify cache hit/miss behavior, sharing across Snapshotters at different versions, correct handling of table-wide operations, and fallback behavior when cache is disabled.

## Implementation Details

- **Version tagging**: New values are tagged with their `stateVersion`, while previous values are tagged with `prev.version` (and `curr.version` when the result depends on intermediate changes). This ensures entries are never stale.

- **FIFO eviction**: Uses insertion order rather than LRU, as all client groups traverse changes in log order. The oldest entries are the ones no longer needed by any group.

- **Correctness guarantee**: Caching is disabled for diffs spanning table-wide operations, since such operations change table schemas/contents and would invalidate cached rows.

- **No mutation of cached values**: Cached SQLite rows are shared by reference, so callers must not modify them. The `fromSQLiteTypes()` conversion produces new row objects.

https://claude.ai/code/session_01PjMkRDohV2RauVjcopZziq


packages/zero-cache/src/services/view-syncer/snapshot-row-cache.ts
```diff
@@ -0,0 +1,166 @@
...
+ * JSON numbers whereas the values read from SQLite are bigints; both address
+ * the same row. Strings are prefixed to distinguish them from numbers.
+ */
+function serialize(v: unknown): string {
+  switch (typeof v) {
+    case 'string':
+      return 's' + v;
+    case 'number':
+    case 'bigint':
+      return String(v);
```
arv: Wouldn't this cause a conflict?

packages/zero-cache/src/services/view-syncer/snapshot-row-cache.ts
```diff
@@ -0,0 +1,166 @@
...
+      }
+    }
+    return value;
+  }
+
+  clear(): void {
+    this.#entries.clear();
+  }
+
+  #key(tag: string, sql: string, args: unknown[]): string {
```
arv: We could potentially do a something along the lines of hashAST. Creating these strings have some overhead.

https://github.com/rocicorp/mono/blob/e4a4127ca25a0c995e2c5b4e6724186b690c134d/packages/zero-protocol/src/query-hash-visitor.ts#L400

packages/zero-cache/src/services/view-syncer/snapshot-row-cache.ts
```diff
@@ -0,0 +1,166 @@
...
+    let sqlID = this.#sqlIDs.get(sql);
+    if (sqlID === undefined) {
+      sqlID = this.#sqlIDs.size;
+      this.#sqlIDs.set(sql, sqlID);
+    }
```
arv: use getOrInsert

packages/zero-cache/src/services/view-syncer/snapshotter.ts
```diff
@@ -284,6 +295,24 @@ export class ResetPipelinesSignal extends Error {
...
+function selectPrefix(table: LiteTableSpecWithKeysAndVersion): string {
+  let prefix = selectPrefixes.get(table);
+  if (prefix === undefined) {
+    const cols = Object.keys(table.columns);
+    prefix = `SELECT ${cols.map(c => id(c)).join(',')} FROM ${id(
+      table.name,
+    )} WHERE `;
+    selectPrefixes.set(table, prefix);
+  }
+  return prefix;
```
arv: getOrInsertComputed

packages/zero-cache/src/services/view-syncer/snapshot-row-cache.ts
arv: Oh, It needs to collapse bigint to same as number.

I'll add a comment

packages/zero-cache/src/services/view-syncer/snapshot-row-cache.ts
```diff
@@ -0,0 +1,188 @@
...
+  readonly #maxEntries: number;
+  readonly #entries = new Map<string, unknown>();
+  // The keys of #entries in insertion order, so that the oldest can be
+  // evicted in O(1). (Finding the oldest key with a fresh Map iterator
+  // instead costs O(evictions) per insert, as V8 leaves deleted entries in
+  // the table until it is compacted and every new iterator walks over them.)
+  readonly #keys = new RingBuffer<string>();
+  // Interns the SQL text of each distinct read so that entry keys
+  // do not repeat the (long) column list of every statement.
+  readonly #sqlIDs = new Map<string, number>();
```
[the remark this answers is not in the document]
arv: Fixed in 3c7e5c8c2: the cache is cleared (including the interned statements) once `MAX_INTERNED_STATEMENTS` (10k) distinct statements have been seen.

packages/zero-cache/src/services/view-syncer/snapshot-row-cache.ts
```diff
@@ -0,0 +1,188 @@
...
+    mode: ReadMode,
+    args: unknown[],
+    read: () => T,
+  ): T {
+    const key = this.#key(tag, sql, mode, args);
+    const cached = this.#entries.get(key);
+    if (cached !== undefined) {
+      this.#hits++;
+      this.#reads.add(1, HIT);
+      return cached as T;
```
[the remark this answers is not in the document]
arv: Fixed in 3c7e5c8c2: `resetToHead()` marks the old Snapshot, and Diff checks `prev.reset` before each change's reads, so a replayed diff throws `InvalidDiffError` regardless of cache hits. Added a regression test for this sequence.

packages/zero-cache/src/services/view-syncer/snapshotter.ts
```diff
@@ -363,56 +423,121 @@ class Snapshot {
...
+        return cached.statement.get<any>(args);
+      });
+    return cache
+      ? cache.getOrRead(tag ?? '', sql, 'get', args, () => {
+          const row = read();
+          if (row !== undefined) {
+            this.#assertNotAdvanced(row);
+          }
+          return row;
+        })
```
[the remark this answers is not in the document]
arv: Fixed in 3c7e5c8c2 (same change as the other thread): the freshness check no longer depends on the rows read.

packages/zero-cache/src/services/view-syncer/snapshotter.ts
arv: Done in 78923aea4.

arv (approved):
LGTM

PR #6629, opened 2026-09-21
feat(zql): push provably safe predicates into subqueries.

tantaman:
This branch adds **correlated predicate pushdown**: it copies eligible parent filters on relationship columns into child subqueries. That can reduce rows read when updates flow through a query. An optional planner-aware mode also lets the planner use those filters when choosing join plans.

```text
Parent filter + relationship
            │
            └── copy filter to child subquery
                    ├── default: after planning; reduces reads
                    └── optional: before planning; can change join choices
```

The planner-aware mode is on by default and has a separate zero-cache flag. The branch also adds tests, benchmarks, fuzz coverage, and design notes for the behavior.

` ZERO_ENABLE_PLANNER_AWARE_PUSHDOWN=true` 

----

The planner now tracks which child filters were copied from a parent. It uses that information in two ways:

  - It leaves copied filters out of selectivity estimates, so an ordinary semi-join doesn’t look more selective than it really is.
  - When estimating a connection, it leaves a copied filter out of the cost calculation if the join constraint already binds that filter’s
    column. If the column isn’t bound, the filter still counts.

  This lets a flipped join get credit for the copied filter reducing rows, without counting the same restriction twice.

---

**On planning:**

Suppose the parent filter is work.id = W, and pushdown copies it to the child as reading.work_id = W.

  If the planner counts both as independent restrictions, it can underestimate the rows: the child’s copied filter looks selective, then the
  join constraint on work_id looks like another selective filter. The planner may therefore estimate the join as much cheaper than it is and
  choose a poor plan.

  In a flipped join, the child is the outer loop, so the child’s copied filter really does reduce the rows scanned. The planner lets that
  filter count there. When it later estimates the parent lookup, the join constraint already binds the matching column, so the planner omits
  the corresponding parent restriction from that cost calculation. The restriction is counted once, on the side where it removes rows.


designs/001_correlated_predicate_pushdown.md
```diff
@@ -0,0 +1,559 @@
+# 001: Correlated predicate pushdown
+
+- **Status:** Implemented (core scope; extensions remain under Future work)
+- **Date:** 2026-09-21
+- **Packages:** `zql` (the rewrite), `zero-cache` (kill switch)
+
+## Summary
+
+When a parent query pins a correlation column, for example
```
aboodman, line 9: I think we should have a thing in agents.md that forbids mannered prose and strongly discourages metaphor. It just makes everything harder to read.

Consider instead:

> When a parent query contains a `where` clause on the correlation clause, e.g., `user.where('id', X).related("documents")` ...

So much easier to understand w/o the metaphor. With _pin_ I need to stop and consider what pin means here. Is there a code concept in zql that I've forgotten called "pinning". Is this the Rust concept of `Pin`? Wait, no we're in TS. Is there some new TS concept of pinning or a community concept?

No, actually it's none of those. It's just Claude being flowery with its writing and coining a metaphor for no damned reason that just makes everything harder to parse.

designs/001_correlated_predicate_pushdown.md
```diff
@@ -0,0 +1,559 @@
...
+`user.where('user_id', X)`, every child row that can join to it has the same
```
aboodman, line 10: It would be better to use a more conventional example like user and docs, or posts and comments.

designs/001_correlated_predicate_pushdown.md
```diff
@@ -0,0 +1,559 @@
...
+value in the matching child column. We copy that condition into the child
```
aboodman, line 11: "We copy that condition" - not immediately clear if this is saying that currently happens or this design changes it to happen.

designs/001_correlated_predicate_pushdown.md
```diff
@@ -0,0 +1,559 @@
...
+  affinity, and the JS join uses `===`. Requiring equal Zero types keeps
+  the pass away from column pairs where these could disagree. Equal Zero
+  types can still hide different Postgres types (for example `text` and
+  `uuid`). For `=`, `IS`, and `IN`, the pushed condition binds the same
+  value against the child column that the join constraint already binds,
+  so it cannot add a new mismatch.
+- **Range, `LIKE`, `!=`, `NOT IN`.** These are sound, but gain little. Not
+  in v1.
+
+### Kill switch
```
aboodman, line 433: I worry about bitrot for the section down to Future Work. I don't think these docs should carry details about the implementation. This all could change and is easily derivable from source.

aboodman (approved):
I like the idea of a designs dir but it should be for humans as much as for agents.

